PR #15, opened 2022-11-25
fix: properly handle CycloneDX xml files

OliverNocon:
When scanning a CycloneDX BOM in xml format the package enumeration did not happen since `BOMFormat` is only available in json BOMs.

In order to properly work with xml format another field needs to be considered.


another-rex (approved):
Thank you for the fix!

oliverchang:
@OliverNocon thank you very much for the contribution! Would you mind signing the CLA so we can merge this?

OliverNocon:
@oliverchang there was some issue with recognizing my company's CLA under which I am covered. Seems like I managed now ...

PR #16, opened 2022-11-26
fix: set error codes differently based on if there are vulnerabilities present

### LINKED ISSUE #7: enable easy usage in CI
I had a CI script that just broke because logs are not sent to stdout when they were not before, I think a change in #3 

```
osv-scanner -r . | tee /tmp/vulns
2022/11/23 18:56:22 Scanning dir .
2022/11/23 18:56:22 Scanning /app/ at commit f187e875250ec11be568d5f00974ebcddba51aa3
2022/11/23 18:56:22 Scanned /app/go.mod file and found 37 packages
! [ -s /tmp/vulns ]
```

This used to exit 0, now it exits 1 because those log lines actually get piped to `/tmp/vulns` so the file is not empty.

How should `osv-scanner` be properly used in CI?

Thanks!

G-Rath:
I've not yet added any test cases for the exit code when vulnerabilities are present because that requires having a lockfile with a vulnerability (which'll annoy dependabot) and interacting with the production API (which'll be relatively slow, the OSV could change, etc) - I plan to address that by adding a flag to set the API URL which'll allow leveraging [`httptest`](https://pkg.go.dev/net/http/httptest) like what I do in `osv-detector` [for testing the API-based databases](https://github.com/G-Rath/osv-detector/blob/main/pkg/database/api-check_test.go#L139-L153).

As for the error codes themselves, I've effectively split the 0-255 range in half (rounded down) with the upper half (starting at 127, as "general error") being reserved for "non-results related errors" such as files not found, errors with the API, etc; and the lower half being reserved for errors about vulnerabilities i.e. some tools set the exit code based on the severity of the vulnerabilities found (something that I'm not convinced is useful enough to land right away, but this way we can do so with minimal impact).

Currently the only other exit code the detector is using that I think will eventually make its way into the scanner is 128, meaning "lockfile not found", because that's the only error so far that I've identified as being useful to differentiate for automatic tools because that could be expected (and essentially enables differing figuring out if the detector/scanner can do anything with a file or directory to the tool itself, which is a very good thing)

Resolves #7

### DIFF
--- a/cmd/osv-scanner/main.go
+++ b/cmd/osv-scanner/main.go
@@ -357,13 +357,13 @@ func run(args []string, stdout, stderr io.Writer) int {
 
 	if err := app.Run(args); err != nil {
 		r.PrintError(fmt.Sprintf("%v", err))
-		return 1
+		return 127
 	}
 
 	resp, err := osv.MakeRequest(query)
 	if err != nil {
 		r.PrintError(fmt.Sprintf("Scan failed: %v", err))
-		return 1
+		return 127
 	}
 
 	filtered := filterResponse(r, query, resp, &configManager)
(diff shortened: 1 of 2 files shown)


another-rex (approved):

another-rex:
Thanks! The error code split and having a specific code for no packages/lockfiles found sounds like a good idea to me.

PR #31, opened 2022-12-09
fix: include `replace` directives in `go.mod`

### LINKED ISSUE #30: go.mod `replace` not being taken into account when reading
For example, Kubernetes go.mod has these lines:
```
require k8s.io/client-go v0.0.0
...
k8s.io/client-go => ./staging/src/k8s.io/client-go
```

and osv-scanner will incorrectly show all vulnerabilities for client-go that have introduced 0.

G-Rath:
Resolves #30

In theory this adds support for `replace`, though in practice it's only going to be useful when the replacement package is remote because there's no way to resolve the local package to the corresponding remote package name that'll be in OSVs - it's tempting to try and infer that from the target (e.g. if you're replacing `golang.org/x/net` with `./fork/net`, then it could be reasonable to assume its a fork of `golang.org/x/net`) but in practice there's no way to know when that shouldn't be applied either.

Likewise for the version, if there isn't one set as part of the replacement then there's not really anything that can be done because there's no requirement from Go that a version be specified for the replacement package anywhere on disk - the best that could be done would be to try and look for a git commit, but I think that is out of scope for this section of code + PR.

### DIFF
--- a/pkg/lockfile/parse-go-lock.go
+++ b/pkg/lockfile/parse-go-lock.go
@@ -44,11 +44,33 @@ func ParseGoLock(pathToLockfile string) ([]PackageDetails, error) {
 	}
 
 	for _, replace := range parsedLockfile.Replace {
-		packages[replace.Old.Path+"@"+replace.Old.Version] = PackageDetails{
-			Name:      replace.New.Path,
-			Version:   strings.TrimPrefix(replace.New.Version, "v"),
-			Ecosystem: GoEcosystem,
-			CompareAs: GoEcosystem,
+		var replacements []string
+
+		if replace.Old.Version == "" {
+			// If the left version is omitted, all versions of the module are replaced.
+			for k, pkg := range packages {
+				if pkg.Name == replace.Old.Path {
+					replacements = append(replacements, k)
+				}
+			}
+		} else {
+			// If a version is present on the left side of the arrow (=>),
+			// only that specific version of the module is replaced
+			s := replace.Old.Path + "@" + replace.Old.Version
+
+			// A `replace` directive has no effect if the module version on the left side is not required.
+			if _, ok := packages[s]; ok {
+				replacements = []string{s}
+			}
+		}
+
+		for _, replacement := range replacements {
+			packages[replacement] = PackageDetails{
+				Name:      replace.New.Path,
+				Version:   strings.TrimPrefix(replace.New.Version, "v"),
+				Ecosystem: GoEcosystem,
+				CompareAs: GoEcosystem,
+			}
 		}
 	}
 
--- a/pkg/lockfile/parse-go-lock.go
+++ b/pkg/lockfile/parse-go-lock.go
@@ -9,6 +9,16 @@ import (
 
 const GoEcosystem Ecosystem = "Go"
 
+func deduplicatePackages(packages map[string]PackageDetails) map[string]PackageDetails {
+	details := map[string]PackageDetails{}
+
+	for _, detail := range packages {
+		details[detail.Name+"@"+detail.Version] = detail
+	}
+
+	return details
+}
+
 func ParseGoLock(pathToLockfile string) ([]PackageDetails, error) {
 	lockfileContents, err := os.ReadFile(pathToLockfile)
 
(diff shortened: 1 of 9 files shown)


pkg/lockfile/parse-go-lock.go
```diff
@@ -22,16 +22,25 @@ func ParseGoLock(pathToLockfile string) ([]PackageDetails, error) {
...
+	for _, replace := range parsedLockfile.Replace {
+		packages[replace.New.Path+"@"+replace.New.Version] = PackageDetails{
+			Name:      replace.New.Path,
+			Version:   strings.TrimPrefix(replace.New.Version, "v"),
+			Ecosystem: GoEcosystem,
+			CompareAs: GoEcosystem,
+		}
+	}
+
+	return pkgDetailsMapToSlice(packages), nil
```
oliverchang, line 45: Should we move this out of https://github.com/google/osv-scanner/blob/4452f8cee25976ed656624dac8cb7c3cdc922436/pkg/lockfile/parse-npm-lock.go into a better filename? e.g. `utils.go` or something similar?

oliverchang: Alternatively, we can just use https://pkg.go.dev/golang.org/x/exp/maps#Values

pkg/lockfile/parse-go-lock.go
```diff
@@ -22,16 +22,25 @@ func ParseGoLock(pathToLockfile string) ([]PackageDetails, error) {
...
 			Version:   strings.TrimPrefix(require.Mod.Version, "v"),
 			Ecosystem: GoEcosystem,
 			CompareAs: GoEcosystem,
-		})
+		}
 	}
 
-	return packages, nil
+	for _, replace := range parsedLockfile.Replace {
+		packages[replace.New.Path+"@"+replace.New.Version] = PackageDetails{
```
oliverchang, line 37: Should this be referencing `replace.Old` instead in the key? 

The test files only contain `replace`, but if there was a `require` for the old module to be replaced, that'll still  get included right?

pkg/lockfile/parse-go-lock_test.go
```diff
@@ -121,3 +121,47 @@ func TestParseGoLock_IndirectPackages(t *testing.T) {
...
+
+	packages, err := lockfile.ParseGoLock("fixtures/go/replace-many.mod")
+
+	if err != nil {
+		t.Errorf("Got unexpected error: %v", err)
+	}
+
+	expectPackages(t, packages, []lockfile.PackageDetails{
+		{
+			Name:      "./fork/net",
```
oliverchang, line 155: What are the semantics for when there are multiple `replace` for the same `Old` module? 

It doesn't seem to make sense to replace both here, if in practice only one is going to be chosen to be used.

oliverchang (changes requested):
Thanks as always @G-Rath !

G-Rath:
@oliverchang my original work was just plain wrong - I've since re-implemented based off the spec.

pkg/lockfile/parse-go-lock.go
G-Rath: Correct, per the spec a `require` entry is required:

> Note that a replace directive alone does not add a module to the [module graph](https://go.dev/ref/mod#glos-module-graph). A [require directive](https://go.dev/ref/mod#go-mod-file-require) that refers to a replaced module version is also needed, either in the main module’s go.mod file or a dependency’s go.mod file. A replace directive has no effect if the module version on the left side is not required.

pkg/lockfile/parse-go-lock_test.go
G-Rath: From what I can tell the spec does not specify the behaviour in this case - my assumption is that it is order-dependent, but I'm also not sure if its an acceptable state beyond parsing.

pkg/lockfile/parse-go-lock.go
G-Rath: If this was a language that had file-based scopings & imports (like Node) I'd agree, but since everything in Go is visible to the surrounding files in a folder/package, I generally take a light "rule of three" based approach unless there's another really compelling reason (like a big refactor or the specific function(s) are large and/or complex), to avoid bloating PRs etc.

Likewise with applying generics: since Go compiles to a static binary, and this is a pretty straightforward utility function that most likely won't change anytime soon, there's no real advantage (from what I understand) to switching right now since the same/similar compiled code will end up in the binary (and currently it might actually be a bit slower/bigger since generics are still being optimized).

oliverchang:
Thanks @G-Rath !

@another-rex can you help review this?

another-rex (approved):
LGTM! Thanks

oliverchang (approved):

PR #37, opened 2022-12-13
go releaser slsa

another-rex:


.slsa-goreleaser.yml
```diff
@@ -0,0 +1,36 @@
...
+  - GO111MODULE=on
+  - CGO_ENABLED=0
+
+# (Optional) Flags for the compiler.
+# flags:
+  # - -trimpath
+  # - -tags=netgo
+
+# The OS to compile for. `GOOS` env variable will be set to this value.
+goos: linux
```
oliverchang, line 15: while we're at it, can we build mac and Windows binaries as well?

oliverchang (approved):

PR #46, opened 2022-12-14
Add gradle lockfile support

abhisek:
Add support for:

1. `gradle.lockfile`
2. `buildscript-gradle.lockfile`

Reference: https://docs.gradle.org/current/userguide/dependency_locking.html#lock_state_location_and_format


another-rex:
Thank you for the PR!

another-rex:
CC @G-Rath

pkg/lockfile/parse-gradle-lock_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	packages, err := lockfile.ParseGradleLock("fixtures/gradle/5-pkg")
+
+	if err != nil {
+		t.Errorf("Got unexpected error: %v", err)
+	}
+
+	if len(packages) != 5 {
+		t.Errorf("Expected %d packages got %d", 5, len(packages))
+	}
```
G-Rath, line 72: Can we use `expectPackages` here? I think it's worth asserting the actual packages to ensure they're all being parsed correctly

pkg/lockfile/ecosystems_test.go
```diff
@@ -35,7 +36,7 @@ func TestKnownEcosystems(t *testing.T) {
 
 	// npm, yarn, and pnpm, and pip and poetry, all use the same ecosystem
 	// so "ignore" those parsers in the count
```
G-Rath, line 38: This comment should be updated:

```suggestion
	// npm, yarn, and pnpm, and pip and poetry, and maven and gradle, all
	// use the same ecosystem so "ignore" those parsers in the count
```

G-Rath (changes requested):
Awesome stuff! You need to update the list of lockfiles in the readme, and it would also be good to have a few tests covering a lockfile with a bad line (or two).

abhisek:
@G-Rath Thanks for the suggestions. I have updated the PR.

pkg/lockfile/parse-gradle-lock_test.go
```diff
@@ -0,0 +1,128 @@
...
+	packages, err := lockfile.ParseGradleLock("fixtures/gradle/does-not-exist")
+
+	expectErrContaining(t, err, "could not open")
+	expectPackages(t, packages, []lockfile.PackageDetails{})
+}
+
+func TestParseGradleLock_OnlyComments(t *testing.T) {
+	t.Parallel()
+
+	packages, err := lockfile.ParseGradleLock("fixtures/gradle/only-comment")
```
G-Rath, line 21: nit: technically this file should be `only-comments`, which aligns with the test name

pkg/lockfile/parse_test.go
```diff
@@ -95,6 +96,7 @@ func TestParse_FindsExpectedParsers(t *testing.T) {
 		"poetry.lock",
 		"pubspec.lock",
 		"requirements.txt",
+		"gradle.lockfile",
```
G-Rath, line 99: `buildscript-gradle.lockfile` should be included on this list (effectively this should have all lockfiles that are listed in the readme as being supported)

G-Rath (changes requested):
Looks good - just need to update the list in `TestParse_FindsExpectedParsers`

pkg/lockfile/parse_test.go
abhisek: Updated to include `buildscript-gradle.lockfile`. But had to decrement the asserted count because both `gradle.lockfile` and `buildscript-gradle.lockfile` use the same parser.

pkg/lockfile/parse-gradle-lock_test.go
abhisek: Renamed :)

pkg/lockfile/parse_test.go
G-Rath: sweet as - that test is more meant to help catch people forgetting to update this section of tests if a new parser is added, which is always a bit fuzzy (arguably I should have used a `map` against the function name and then check that each was called at least once, which I think could be possible, but not worth the effort imo)

G-Rath (approved):

another-rex (approved):

PR #47, opened 2022-12-14
add signing to sign the binaries generated using cosign

cpanato:
Add a step in goreleaser to sign the binaries generated using sigstore cosign.

cc @another-rex

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -13,6 +13,7 @@ jobs:
     outputs:
       hashes: ${{ steps.hash.outputs.hashes }}
     permissions:
+      id-token: write  # needed for signing the binaries
       contents: write  # for goreleaser/goreleaser-action to create a GitHub release
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


laurentsimon:
The project already generates SLSA attestations, see https://github.com/google/osv-scanner/blob/main/.github/workflows/goreleaser.yml#L44-L53 and https://github.com/google/osv-scanner/releases/tag/v1.0.0

I don't think cosign signatures are needed, since attestation provide the same guarantees plus additional info (GH context, etc)

asraa:
> I don't think cosign signatures are needed, since attestation provide the same guarantees plus additional info (GH context, etc)

+1 although I think we could add a snippet into the README.md about how to verify the SLSA provenance in the release assets and what information is there with a quick jq command. WDYT? I can do as a follow-up if people agree

oliverchang:
> > I don't think cosign signatures are needed, since attestation provide the same guarantees plus additional info (GH context, etc)
> 
> +1 although I think we could add a snippet into the README.md about how to verify the SLSA provenance in the release assets and what information is there with a quick jq command. WDYT? I can do as a follow-up if people agree

Yes please!

cpanato:
closing

PR #48, opened 2022-12-14
Ci updates

cpanato:
- use git hashes instead of tags and update to latest
- format json and udpate .sh files


laurentsimon (approved):
Thanks!

another-rex (approved):
Thanks!

PR #54, opened 2022-12-15
Save json output for future use

helmutkemper:
Save json file for future use in reports

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,8 @@ reason = "No external http servers are written in Go lang."
 ## JSON output
 By default osv-scanner outputs a human readable table. To have osv-scanner output JSON instead, pass the `--json` flag when calling osv-scanner. 
 
+If you want to save the JSON file for future use, you can specify a path by passing `-output /path/to/file.json` or `-o /path/to/file.json`. 
+
 ### Output Format
 ```json5
 {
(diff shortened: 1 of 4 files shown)


helmutkemper:
I'm doing my chaos test for microservices and with this new feature, it's easy to save a test report

```go
package osvScanner

import (
  "github.com/helmutkemper/chaos/factory"
  "testing"
  "time"
)

func TestLinear(t *testing.T) {

  primordial := factory.NewPrimordial().
    NetworkCreate("test_network", "10.0.0.0/16", "10.0.0.1").
    Test(t)

  factory.NewContainerFromGit(
    "osv:latest",
    "https://github.com/helmutkemper/osv-scanner.git",
  ).
    ReplaceBeforeBuild("Dockerfile", "./Dockerfile").
    Volumes("/scan", "../../").
    Create("osv", 1).
    Start()

  if !primordial.Monitor(3 * time.Minute) {
    t.Fail()
  }
}

helmutkemper:
it is standard for some programs to use -o together with the --output option

oliverchang:
> it is standard for some programs to use -o together with the --output option

Sorry my question was unclear! 

Would `osv-scanner .... -json > output.json` work equally well for you here, as opposed to `osv-scanner ... -o output.json`  which this PR adds? Are there reasons why adding an explicit flag for this is helpful for users?

helmutkemper:
make the option clear and standardized

PR #59, opened 2022-12-15
- add conan.lock to the list

SSE4:
/cc @prince-chrismc

there are several formats of conan lockfile available in the wild (lockfiles are versioned separately from the conan client itself, as not every conan release changes the format of lockfile):
- initial, unversioned (`0.0`) - added in **conan 1.17**
- `0.1` added in **conan 1.18**, added versioning to the lockfile format
- `0.2` added in **conan 1.21**, added deterministic lockfiles
- `0.3` added in **conan 1.22**, changed how nodes reference each other
- `0.4` added in **conan 1.28**, the format mostly used currently in production and encountered most often 
- `0.5` is used by **conan 2.0+** (currently, as of December 2022, still in beta state, and not yet released)

formats 0.0  - 0.4 are mostly similar and use `graph_lock` with a list of nodes.
format 0.5 is a major overhaul, it's a completely new implementation developed from scratch.

some information about 0.0 - 0.4 lockfiles (current format used in production):
- [doc: lockfiles](https://docs.conan.io/en/latest/versioning/lockfiles.html)
- [doc: using recipe revisions and lockfiles](https://docs.conan.io/en/latest/howtos/using_rrev_and_lockfiles.html)
- [Conan 1.34 : New lockfile feature "bundles", Resource shift to Conan V2.0, Merged Tribe Proposals](https://blog.conan.io/2021/03/10/New-conan-release-1-34.html)
- [Conan 1.32: New validate() Method, First configurable Generator (MSBuildDeps), Renamed Multiple Toolchains & Generators, 2 New Meson Classes, Improve Lockfile Support for Private Requirements, Support for build_requires to Affect package_id (Workaround)](https://blog.conan.io/2020/12/14/New-conan-release-1-32.html)
- [Conan 1.29: Updates to multiple generators, new tools.rename function, experimental toolchain for gnu make, lockfile bug fixes](https://blog.conan.io/2020/09/07/New-conan-release-1-29.html)
- [Conan 1.28: Lockfile improvements, Four new conanfile.py attributes, Improved Clang support on Windows, Support components in pkg_config generator, Define generator filenames](https://blog.conan.io/2020/08/06/New-conan-release-1-28.html)
- [Conan 1.17: The Conan user survey, lockfiles, new graph commands, package ID modes for revisions, apple-clang 11.0 support and usability improvements](https://blog.conan.io/2019/07/15/New-conan-release-1-17.html)

more information about 0.5 lockfiles used in conan 2.0:
- [New 2.0 lockfiles](https://github.com/conan-io/tribe/pull/34)

the format itself is not documented, but can be learnt by studying the source code [graph_lock.py](https://github.com/conan-io/conan/blob/develop/conans/model/graph_lock.py). as it's currently not expected lockfiles to be manually or automatically edited by users or tools, they are only generated by the conan client itself during the graph resolution process, and afterwards, lockfiles are fully immutable.

for the version comparison, conan client itself uses [version ranges](https://docs.conan.io/en/latest/versioning/version_ranges.html) to compare versions, but it's only for packages that follow SemVer (a majority of C and C++ libraries don't follow SemVer, or even predate it). as different libraries use various versioning schemes, conan client cannot make any assumtions.
however, lockfile represents a resolved graph state, with all the version ranges already computed. therefore, versions in lockfiles are compared as regular strings.


prince-chrismc:
It needs to be implemented in https://github.com/google/osv-scanner/tree/main/pkg/lockfile 🙊

SSE4:
@prince-chrismc I've added some parser

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,177 @@
...
+	GraphLock    ConanGraphLock `json:"graph_lock,omitempty"`
+	ProfileHost  string         `json:"profile_host,omitempty"`
+	ProfileBuild string         `json:"profile_build,omitempty"`
+	// conan v0.5+ lockfiles use "requires", "build_requires" and "python_requires"
+	Requires       []string `json:"requires,omitempty"`
+	BuildRequires  []string `json:"build_requires,omitempty"`
+	PythonRequires []string `json:"python_requires,omitempty"`
+}
+
+const ConanEcosystem Ecosystem = "conan.io"
```
G-Rath, line 47: Just flagging this isn't a valid ecosystem currently per the [OSV spec](https://ossf.github.io/osv-schema/#affectedpackage-field) - I'll leave it to @another-rex & @oliverchang to decide what to do about this

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,177 @@
...
+			Version:   reference.Version,
+			Ecosystem: ConanEcosystem,
+			CompareAs: ConanEcosystem,
+		})
+	}
+}
+
+func parseConanV2Lock(lockfile ConanLockFile) []PackageDetails {
+
+	packages := make([]PackageDetails, 0, len(lockfile.Requires)+len(lockfile.BuildRequires)+len(lockfile.PythonRequires))
```
G-Rath, line 144: You'll need to wrap these in `uint64` to make CodeQL happy, otherwise it will flag this as a potential vulnerability since the return type of `len` technically allows negative numbers even though in practice that should never be possible.

See [here](https://github.com/google/osv-scanner/blob/main/pkg/lockfile/parse-composer-lock.go#L42-L43) for an example of how I've addressed that in the past

G-Rath:
@SSE4 I've done an initial review and nothing huge has jumped out at me as needing to be changed - I'm not familiar with this ecosystem/tool at all; would you happen to have any references to the lockfile (ideally a spec would be great, but not all ecosystems have those).

It would also be good to know how the version comparison is actually done if you know that too.

pkg/lockfile/parse-conan-lock.go
oliverchang: Indeed! 

We should at the very least add a TODO here to say that this is tentative and subject to change depending on the OSV schema. 

This is definitely a useful addition to have for the OSV-Scanner, but the database itself also currently does not have conan.io advisories. Is there also an existing Conan vuln DB that we could have publishing OSV entries?

prince-chrismc: > Is there also an existing Conan vuln DB that we could have publishing OSV entries?

Conan is the C++ package manager 😉 We have ConanCenter which offers a ton of C and C++ packages. https://cve.mitre.org/index.html or https://nvd.nist.gov/ are the types of places I'd expect find know vulnerabilities.

https://conan.io/center/openssl has numerous ones https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=openssl.
[openssl/3.0.5](https://conan.io/center/openssl?version=3.0.5) has [CVE-2022-3602](https://nvd.nist.gov/vuln/detail/CVE-2022-3602) for example.

I saw in the readme [these are listed as alias](https://github.com/google/osv-scanner#output-format) could they not be the IDs as well? Some like the Openssl Alpine that I see on https://osv.dev/list?ecosystem=Alpine&q=openssl was my initial reaction.

pkg/lockfile/parse-conan-lock.go
SSE4: thanks, I am not that familiar with Go or GoLang, so negative length or size type sounds a bit suspicious for me. fixed it.

pkg/lockfile/parse-conan-lock.go
SSE4: thanks, I've opened a PR to ovs-schema: https://github.com/ossf/osv-schema/pull/101
for the database, sorry, but I have no idea. maybe @prince-chrismc or @jcar87 know and can comment. perhaps, X-Ray might have some sort of database, but I am not familiar with it.

SSE4:
> @SSE4 I've done an initial review and nothing huge has jumped out at me as needing to be changed - I'm not familiar with this ecosystem/tool at all; would you happen to have any references to the lockfile (ideally a spec would be great, but not all ecosystems have those).
> 
> It would also be good to know how the version comparison is actually done if you know that too.

thanks, I've updated the PR description. lemme know if you have any further question.

SSE4:
@G-Rath how do we move this forward? please let me know if you need anything else from my side.

G-Rath:
@SSE4 the main thing that needs addressing is [this comment](https://github.com/google/osv-scanner/pull/59/files#r1051688292) - otherwise everything looks ok

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,181 @@
...
+	GraphLock    ConanGraphLock `json:"graph_lock,omitempty"`
+	ProfileHost  string         `json:"profile_host,omitempty"`
+	ProfileBuild string         `json:"profile_build,omitempty"`
+	// conan v0.5+ lockfiles use "requires", "build_requires" and "python_requires"
+	Requires       []string `json:"requires,omitempty"`
+	BuildRequires  []string `json:"build_requires,omitempty"`
+	PythonRequires []string `json:"python_requires,omitempty"`
+}
+
+const ConanEcosystem Ecosystem = "conan.io"
```
SSE4, line 47:
```suggestion
// TODO this is tentative and subject to change depending on the OSV schema
const ConanEcosystem Ecosystem = "conan"
```

SSE4:
> @SSE4 the main thing that needs addressing is [this comment](https://github.com/google/osv-scanner/pull/59/files#r1051688292) - otherwise everything looks ok

okay, I've changed ecosystem name `conan.io` to `conan` and added TODO you have suggested.

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,182 @@
...
+	ProfileHost  string         `json:"profile_host,omitempty"`
+	ProfileBuild string         `json:"profile_build,omitempty"`
+	// conan v0.5+ lockfiles use "requires", "build_requires" and "python_requires"
+	Requires       []string `json:"requires,omitempty"`
+	BuildRequires  []string `json:"build_requires,omitempty"`
+	PythonRequires []string `json:"python_requires,omitempty"`
+}
+
+// TODO this is tentative and subject to change depending on the OSV schema
+const ConanEcosystem Ecosystem = "conan"
```
SSE4, line 48:
```suggestion
const ConanEcosystem Ecosystem = "ConanCenter"
```

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return reference
+}
+
+func parseConanV1Lock(lockfile ConanLockFile) []PackageDetails {
+	var reference ConanReference
+	packages := make([]PackageDetails, 0, len(lockfile.GraphLock.Nodes))
+
+	for _, node := range lockfile.GraphLock.Nodes {
+
```
G-Rath, line 103: nit

```suggestion
	for _, node := range lockfile.GraphLock.Nodes {
```

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,182 @@
...
+		})
+	}
+
+	return packages
+}
+
+func parseConanRequires(packages *[]PackageDetails, requires []string) {
+
+	for _, ref := range requires {
+
```
G-Rath, line 131: nit

```suggestion
func parseConanRequires(packages *[]PackageDetails, requires []string) {
	for _, ref := range requires {
```

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,182 @@
...
+			Version:   reference.Version,
+			Ecosystem: ConanEcosystem,
+			CompareAs: ConanEcosystem,
+		})
+	}
+}
+
+func parseConanV2Lock(lockfile ConanLockFile) []PackageDetails {
+
+	packages := make(
```
G-Rath, line 145: nit

```suggestion
func parseConanV2Lock(lockfile ConanLockFile) []PackageDetails {
	packages := make(
```

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func parseConanLock(lockfile ConanLockFile) []PackageDetails {
+
+	if lockfile.GraphLock.Nodes != nil {
+
+		return parseConanV1Lock(lockfile)
+	} else {
+		return parseConanV2Lock(lockfile)
+	}
```
G-Rath, line 165: nit

```suggestion
	if lockfile.GraphLock.Nodes != nil {
		return parseConanV1Lock(lockfile)
	}

	return parseConanV2Lock(lockfile)
```

G-Rath (commented):
Got a couple of nitty comments about newlines

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+
+	parts = strings.SplitN(ref, "/", 2)
+	if len(parts) == 2 {
+		reference.Name = parts[0]
+		reference.Version = parts[1]
+	} else {
+		reference.Version = ref
+	}
```
G-Rath, line 94: I suspect I'm about to open a can of worms with this (which I missed in my first review, sorry about that), but this looks to me like it's possible for a package to not have a name - is that true? technically the `osv.dev` API allows that if you have a commit, but it's not really expected elsewhere.

it also looks like none of your tests are actually covering this line (unless I've missed something)

pkg/lockfile/parse-conan-lock.go
SSE4: I don't see a change here :thinking:

pkg/lockfile/parse-conan-lock.go
```diff
@@ -0,0 +1,177 @@
...
+	return reference
+}
+
+func parseConanV1Lock(lockfile ConanLockFile) []PackageDetails {
+	var reference ConanReference
+	packages := make([]PackageDetails, 0, len(lockfile.GraphLock.Nodes))
+
+	for _, node := range lockfile.GraphLock.Nodes {
+
+		if node.Path != "" {
```
G-Rath, line 104:
```suggestion
	for _, node := range lockfile.GraphLock.Nodes {
		if node.Path != "" {
```

pkg/lockfile/parse-conan-lock.go
G-Rath: hmm I think GitHub bugged out - I've remade the change [here](https://github.com/google/osv-scanner/pull/59/files#r1064243726) across more lines so hopefully it'll stick

pkg/lockfile/parse-conan-lock.go
prince-chrismc: All the official packages that are open-source that could possibly be in public databases with CVE will have a name.

It's technically possible for someone to have private packages but I would safe those are safe to skip.

Perhaps a warning or just skipping the package is an option?

SSE4: I've just got some responses:
```
It is possible to have self.version defined for a consumer, but not have self.name defined
In conan export or conan create it should never happen, I understand that it has reached the lockfile because it is a consumer
Well, not having to define a name, a conanfile.txt has neither a name nor a version, a conanfile.py also works fine without a name or version
But if you want to give it a version, you can.
```
so, I think they might theoretically appear in a lockfile, but they should be completely safe to skip.
I'll try to add some test case, if I can generate such a lockfile, otherwise, I'll just skip them in code (if Name is missing or empty).

SSE4: I've added code to skip packages with no name

G-Rath:
@SSE4 I might have missed it, but it looks like you don't have a test for when a package exists without a name? if so I think that is important to have, to close off the loop; otherwise LGTM 👍

SSE4:
> @SSE4 I might have missed it, but it looks like you don't have a test for when a package exists without a name? if so I think that is important to have, to close off the loop; otherwise LGTM +1

I'll add some tests with synthetic lockfiles, as I can't generate such with client.

SSE4:
> @SSE4 I might have missed it, but it looks like you don't have a test for when a package exists without a name? if so I think that is important to have, to close off the loop; otherwise LGTM +1

now all done

SSE4:
@G-Rath can we move this one forward?

G-Rath (approved):

pkg/lockfile/parse_test.go
```diff
@@ -90,6 +90,7 @@ func TestParse_FindsExpectedParsers(t *testing.T) {
 	t.Parallel()
 
 	lockfiles := []string{
+		"conan.lock",
```
another-rex, line 93: nit: Move this line down a few lines to maintain alphabetical order

pkg/lockfile/parse.go
```diff
@@ -18,6 +18,7 @@ func FindParser(pathToLockfile string, parseAs string) (PackageDetailsParser, st
 
 //nolint:gochecknoglobals // this is an optimisation and read-only
 var parsers = map[string]PackageDetailsParser{
+	"conan.lock":                  ParseConanLock,
```
another-rex, line 21: nit: Move this line down a few lines to maintain alphabetical order

README.md
```diff
@@ -126,6 +126,7 @@ osv-scanner --sbom=/path/to/your/sbom.json
 
 A wide range of lockfiles are supported by utilizing this [lockfile package](https://github.com/google/osv-scanner/tree/main/pkg/lockfile). This is the current list of supported lockfiles:
 
+- `conan.lock`
```
another-rex, line 129: nit: Move this line down a few lines to maintain alphabetical order

another-rex (approved):
Looks good to me, just waiting for (https://github.com/ossf/osv-schema/pull/101) to be accepted before merging this in.

SSE4:
> Looks good to me, just waiting for ([ossf/osv-schema#101](https://github.com/ossf/osv-schema/pull/101)) to be accepted before merging this in.

it's merged now

another-rex (approved):

Issue #55, opened 2022-12-15
Implement others ways of installation


jwillker:
The project has a high chance of gaining a lot of adoption. 
To help with adoption, the project could provide other ways to install besides go install or download binaries. This facilitates to use of the osv-scanner in different machines, servers, CI/CD, etc.

My proposal is that the project could have the following:
- doc with curl | bash 
- docker image
- brew package
- aurs package
- snap
- Linux packages(apk, deb, rpm)

All this can be done using the goreleaser. 
I can help with PRs if these features make sense!

andrewpollock:
My initial thoughts on `curl | bash` are coloured by my past history with things like:

- https://www.idontplaydarts.com/2016/04/detecting-curl-pipe-bash-server-side/
- https://www.onsecurity.io/blog/careless-with-curl-dont-be/

and as such, given the space we're operating in, I'd prefer for OSV Scanner to have a very solid supply chain security story.

I totally agree with the spirit of this issue though. As a Debian Developer, I'd like to see it packaged for Debian :-)

another-rex:
I am not very familiar with goreleaser so any help would be appreciated, +1 on being cautious with curl | bash though.

I think the first step is producing a docker image, which can also be used when making the CI/CD action.

koenhendriks:
I did try to build a docker image from the current Dockerfile in the repo, but I don't seem to be able to run the osv-scanner with arguments.

When trying to scan a directory:
`docker run --rm koenhendriks/osv-scanner -v ${PWD}:/app -r /app`

returns:
```
osv-scanner version: dev
commit: n/a
built at: n/a
```

And when trying to run with `--lockfile` I get the same:
`docker run --rm koenhendriks/osv-scanner -v ${PWD}:/app --lockfile=/app/composer.lock`

returns:
```
osv-scanner version: dev
commit: n/a
built at: n/a
```

Running it from go locally works fine.

jwillker:
> I am not very familiar with goreleaser so any help would be appreciated, +1 on being cautious with curl | bash though.
> 
> I think the first step is producing a docker image, which can also be used when making the CI/CD action.

@another-rex I just open a PR to implement a docker image publish feature using Goreleaser https://github.com/google/osv-scanner/pull/63

Can you review and comment the open points please?

jayvdb:
https://github.com/taiki-e/install-action now supports osv-scanner

G-Rath:
[...]

PR #63, opened 2022-12-15
feat: add docker publish method

jwillker:
As I mentioned in this issue: https://github.com/google/osv-scanner/issues/55, the project could have an official docker image.

This implementation will generate a docker image on the Github container registry for amd64 and arm64 architecture for each git tag and the latest version, using Goreleaser. https://goreleaser.com/customization/docker/, https://goreleaser.com/customization/docker_manifest/

The image labels follow the annotation recommendations of https://github.com/opencontainers/image-spec/blob/main/annotations.md.

The action.yaml was changed to work with the image instead of building from Dockerfile.

Some open points:
- Makes sense to generate an image on DockerHub too? Like: https://hub.docker.com/r/google/osv-scanner
- The build step was removed from Dockerfile as the Goreleaser recommends: https://goreleaser.com/errors/docker-build/#do, but this will be more difficult for users to build the image from the source. The alternative is to have two Dockerfiles.
- For ghcr login we can use the `GITHUB_TOKEN,` but it can be used with a PAT. ([Doc](https://github.com/docker/login-action#github-container-registry))

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,8 @@ reason = "No external http servers are written in Go lang."
 ## JSON output
 By default osv-scanner outputs a human readable table. To have osv-scanner output JSON instead, pass the `--json` flag when calling osv-scanner. 
 
+When using the --json flag, only the JSON output will be printed to stdout, with all other outputs being directed to stderr. So to save only the json output to file, you can redirect the output with `osv-scanner --json ... > /path/to/file.json`
+
 ### Output Format
 ```json5
 {


another-rex:
> Makes sense to generate an image on DockerHub too? Like: https://hub.docker.com/r/google/osv-scanner

> The build step was removed from Dockerfile as the Goreleaser recommends: https://goreleaser.com/errors/docker-build/#do, but this will be more difficult for users to build the image from the source. The alternative is to have two Dockerfiles.

Would prefer to have the goreleaser dockerfile named something different, 

> For ghcr login we can use the GITHUB_TOKEN, but it can be used with a PAT. ([Doc](https://github.com/docker/login-action#github-container-registry))

another-rex:
> Makes sense to generate an image on DockerHub too? Like: https://hub.docker.com/r/google/osv-scanner

For now I think just having github registry is fine. 

> The build step was removed from Dockerfile as the Goreleaser recommends: https://goreleaser.com/errors/docker-build/#do, but this will be more difficult for users to build the image from the source. The alternative is to have two Dockerfiles.

Agree with having two Dockerfiles, and would prefer for the goreleaser Dockerfile to be named something different. 
I think it's more intuitive for the default Dockerfile to be build-able immediately upon cloning the repository. 

> For ghcr login we can use the GITHUB_TOKEN, but it can be used with a PAT. ([Doc](https://github.com/docker/login-action#github-container-registry))

Happy to stick with GITHUB_TOKEN.

PR #68, opened 2022-12-16
Added installation instructions for Scoop

stevehipwell:
This PR adds installation instructions for using Scoop on Windows (https://github.com/ScoopInstaller/Main/pull/4231) which should make it nice and easy for anyone on a Windows machine to use this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,9 @@ The above all results in fewer, more actionable vulnerability notifications, whi
 ### Installation Process
 You may download the [SLSA3](https://slsa.dev) compliant binaries for Linux, macOS, and Windows from our [releases page](https://github.com/google/osv-scanner/releases).
 
-#### Windows
+#### Package Managers
 
-On Windows, you can install this using the [Scoop](https://scoop.sh/) package manager by running:
-
+If you're a [**Windows Scoop**](https://scoop.sh) user, then you can install osv-scanner from the [official bucket](https://github.com/ScoopInstaller/Main/blob/master/bucket/osv-scanner.json):
 ```shell
 scoop install osv-scanner
 ```
--- a/README.md
+++ b/README.md
@@ -35,9 +35,18 @@ The above all results in fewer, more actionable vulnerability notifications, whi
 ### Installation Process
 You may download the [SLSA3](https://slsa.dev) compliant binaries for Linux, macOS, and Windows from our [releases page](https://github.com/google/osv-scanner/releases).
 
+#### Windows
+
+On Windows, you can install this using the [Scoop](https://scoop.sh/) package manager by running:
+
+```shell
+scoop install osv-scanner
+```
+
 #### Install from source
 
 Alternatively, you can install this from source by running:
+
 ```bash
 $ go install github.com/google/osv-scanner/cmd/osv-scanner@v1
 ```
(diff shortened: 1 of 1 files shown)


jwillker:
@stevehipwell 

Relates https://github.com/google/osv-scanner/issues/55

I don't know how Scoop works but does this repo update itself for new osv-scanner versions? To be more clear, when a new version of osv-scanner is released, the Scoop will search for a new osv-scanner, and will it be available on their repository?

stevehipwell:
@jwillker Scoop keeps the package version up to date by watching the GH releases.

inferno-chromium:
> @jwillker Scoop keeps the package version up to date by watching the GH releases.

On Windows, you would do "go install" as well. I am not sure of community acceptance of Scoop and why we need to introduce this layer of complexity when "go install" is the standard for installing go packages.

stevehipwell:
@inferno-chromium that makes the assumption that all end users are expected to have the Go toolchain installed?

Scoop is pretty widely adopted for portable binaries; it adds zero additional complexity for projects and keeps versions up to date. I think it's worth adding to the README as osv-scanner is already available on Scoop.

I've generally found that the more ways there are to install binaries the better the adoption; and package managers which allow versions to be kept up to date are a good thing.

inferno-chromium:
> @inferno-chromium that makes the assumption that all end users are expected to have the Go toolchain installed?
> 
> Scoop is pretty widely adopted for portable binaries; it adds zero additional complexity for projects and keeps versions up to date. I think it's worth adding to the README as osv-scanner is already available on Scoop.
> 
> I've generally found that the more ways there are to install binaries the better the adoption; and package managers which allow versions to be kept up to date are a good thing.

Make sense. I would let @oliverchang @another-rex come back from vacation to comment here. This seems fine, i think you should move it after source section, in a dedicated "Install from package manager section" and underneath it, give bullet point on Scoop. We will be adding more package manager ways soon, so it makes sense there.

stevehipwell:
@inferno-chromium as the primary installation method in the current README is to downloading the binaries from the releases page wouldn't the package manager options logically follow next with installation from source remaining last?

README.md
```diff
@@ -35,16 +35,26 @@ The above all results in fewer, more actionable vulnerability notifications, whi
 ### Installation Process
 You may download the [SLSA3](https://slsa.dev) compliant binaries for Linux, macOS, and Windows from our [releases page](https://github.com/google/osv-scanner/releases).
 
+#### Windows
+
+On Windows, you can install this using the [Scoop](https://scoop.sh/) package manager by running:
+
```
another-rex, line 41:
```suggestion
#### Package Managers

If you're a [**Windows Scoop**](https://scoop.sh) user, then you can install osv-scanner from the [official bucket](https://github.com/ScoopInstaller/Main/blob/master/bucket/osv-scanner.json):
```

another-rex: Following how ripgrep shows installation instructions.

another-rex (commented):
Thanks! Happy to add instructions for installing from scoop.

Fully agree to have installing from source last, though I think it is a good idea to have the title be Package Manager rather than the OS, since each OS has multiple package managers.

another-rex (approved):

PR #69, opened 2022-12-16
Added descriptions for the commands in the Dockerfile & Changed the format of passing arguments to an Array in action.yml file

adityassharma-ss:
Pull Request Description:

Change 1: 
Changed the actions.yml file's arguments section, converted it to an array type:
`args: ['--skip-git', ${{ inputs.to-scan }}]`

Change 2:
Added descriptions to the Dockerfile commands to make it more understandable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,19 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


inferno-chromium:
These commands are self-explanatory, we should not add comments for obvious things as it makes the code less readable.

PR #70, opened 2022-12-16
feat: add docker publish method

jwillker:
> Reopening after closing by accident this [PR](https://github.com/google/osv-scanner/pull/63)

As I mentioned in this issue: https://github.com/google/osv-scanner/issues/55, the project could have an official docker image.

This implementation will generate a docker image on the Github container registry for amd64 and arm64 architecture for each git tag and the latest version, using Goreleaser. https://goreleaser.com/customization/docker/, https://goreleaser.com/customization/docker_manifest/

The image labels follow the annotation recommendations of https://github.com/opencontainers/image-spec/blob/main/annotations.md.

The action.yaml was changed to work with the image instead of building from Dockerfile.


jwillker:
@another-rex The other PR changes suggested were solved here. Can you review this PR?

JonZeolla:
@jwillker what about using https://github.com/ko-build/ko for this?

jwillker:
> @jwillker what about using https://github.com/ko-build/ko for this?

The osv-scanner already has a build process using goreleaser. The feature here is to extend the usage to release docker images. I don't know how the ko works in depth, and I don't know if it makes sense to add another tool to the project and change the build process.

What are the advantages of ko over Goreleaser?

inferno-chromium:
Rex is OOO next week, we will get to this early next year. I agree with @jwillker that we should just use existing tools.

oliverchang:
Thanks a ton for the contribution, and sorry for the delay on reviewing this! We haven't forgotten about this :)

@another-rex is still out until next week -- once he's back I'll let him review this.

.goreleaser.yml
```diff
@@ -28,7 +28,52 @@ builds:
       # - arm
       - arm64
     main: ./cmd/osv-scanner/main.go
-    binary: '{{ .ProjectName }}_v{{ .Version }}'
+
+dockers:
+  - image_templates:
```
another-rex, line 33:
```suggestion
  # Arch: amd64
  - image_templates:
```

.goreleaser.yml
```diff
@@ -28,7 +28,52 @@ builds:
...
+      - "--label=org.opencontainers.image.description=Vulnerability scanner written in Go which uses the data provided by https://osv.dev"
+      - "--label=org.opencontainers.image.licenses=Apache License 2.0"
+      - "--label=org.opencontainers.image.created={{.Date}}"
+      - "--label=org.opencontainers.image.name={{.ProjectName}}"
+      - "--label=org.opencontainers.image.revision={{.FullCommit}}"
+      - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--label=org.opencontainers.image.source={{.GitURL}}"
+      - "--label=org.opencontainers.image.url={{.GitURL}}"
+      - "--platform=linux/amd64"
+  - image_templates:
```
another-rex, line 49:
```suggestion
  # Arch: arm64
  - image_templates:
```
Just to make it clear the arch is what's different about these two builds, arm64 and amd64 looks very similar.

.github/workflows/goreleaser.yml
```diff
@@ -15,16 +15,26 @@ jobs:
     permissions:
       contents: write  # for goreleaser/goreleaser-action to create a GitHub release
     runs-on: ubuntu-latest
+    env:
+      DOCKER_CLI_EXPERIMENTAL: "enabled"
```
another-rex, line 19:
```suggestion
      # Required for buildx on docker 19.x
      DOCKER_CLI_EXPERIMENTAL: "enabled"
```

another-rex (approved):
Thanks for the changes! I just added some additional comments, but otherwise LGTM!

PR #75, opened 2022-12-17
Update README.md

Miaritiana:


inferno-chromium:
This does not impact markdown rendering, we don't really need this change.

PR #77, opened 2022-12-17
Update examples in README

hi-artem:
The way it is currently written may confuse people new to Golang.

You provide instructions on how to install binary from various sources, which will essentially have the same effect - the application binary is in OS path. However, the following usage examples assume that we are working with the source code.

This PR fixes the README examples, by making assumption `osv-scanner` is installed and can be directly called.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ as real git repositories.
 #### Example
 
 ```bash
-$ go run ./cmd/osv-scanner -r /path/to/your/dir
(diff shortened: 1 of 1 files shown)


inferno-chromium (approved):

inferno-chromium:
@oliverchang @another-rex - fyi. i think this is how it is used in other projects as well, so approving. I don't think we need to add "$(go env GOPATH)/bin/" to every command since we assume user know how to use go.

PR #88, opened 2022-12-20
Wrap source column when it exceeds 40 characters

another-rex:
Wrap source column when it exceeds 40 characters. An improvement to this is to calculate the max width of the table and only wrap to that.

### DIFF
--- a/internal/output/table.go
+++ b/internal/output/table.go
@@ -16,13 +16,12 @@ import (
 )
 
 const sourceLength = 40
-const headerRow = table.Row{"OSV URL (ID In Bold)", "Ecosystem", "Package", "Version", "Source"}
 
 // PrintTableResults prints the osv scan results into a human friendly table.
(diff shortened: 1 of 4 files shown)


internal/output/table.go
```diff
@@ -75,5 +79,14 @@ func PrintTableResults(vulnResult *models.VulnerabilityResults, outputWriter io.
...
 	}
+
 	outputTable.Render()
 }
+
+func wrapPath(path string) string {
+	slashPath := filepath.ToSlash(path)
+	moddedPath := strings.ReplaceAll(slashPath, "/", "/ ")
+	wrappedPath := wordwrap.WrapString(moddedPath, sourceLength)
+	unmoddedWrappedPath := strings.ReplaceAll(wrappedPath, "/ ", "/")
```
oliverchang, line 89: Could there be a source path with a space as the first character that would break this? 

This seems a bit too hacky to rely on. I'd rather we just write our own wrapping library for this case.

Issue #81, opened 2022-12-19
Support local DBs


oliverchang:
Currently the scanner works by using the OSV.dev API, which ensures the matching against latest live DB with little (targeted <15 minute latency from the upstream source) 

We should support a local mode, which supports taking in a local OSV DB.

One of the key prerequisites here is:
- Implementing version comparison rules for all our supported ecosystems. This is necessary for precise vulnerability matching based on the [OSV version matching algorithm](https://ossf.github.io/osv-schema/#evaluation) (@G-Rath has something for this that does this for the many of our ecosystems)

This will have some limitations:
- Commit based matching will not work -- the API indexes all commit hashes ingested, and it's not feasible to replicate this index locally.
- Potential performance issues?

G-Rath:
As you know this is (imo) pretty extensively supported by [`osv-detector`](https://github.com/G-Rath/osv-detector) - I'm happy for any code from that to be brought into the scanner, and think a good first step would be to look at that to see what people think is missing (or plain wrong).

> Implementing version comparison rules for all our supported ecosystems.

I'm hoping to get around to PRing that over to here this week 😅

> Potential performance issues?

Since the detector has been using a local database since day 1, I've spent a lot of time looking at the performance in a number of cases especially when:
  * I added support for deduplicating aliases (which added like 3 extra full loops), and extra loops that came from misc fixes like handling Python packages with unnormalized names
  * support for the osv.dev API was landed, to see if it was faster than using the local database
  * I overhauled `semantic` to use ecosystem-specific version comparisons, since that greatly changed the work being done by the detector

In all cases the detector performed extremely fast, with the slowest part always being downloading the database (which even then would typically only take <5 seconds, for all the dbs needed for a project) - in contrast with the API it would take 15+ seconds 😅

(not saying there couldn't be performances issues, just sharing my experience so far)

oliverchang:
@G-Rath any chance you'd have time to try tackle this one within the next few weeks? Otherwise we might be interested in taking this on as we've been seeing requests for this.

G-Rath:
@oliverchang yup probably - do you want to start with just an `--offline` flag for now and go from there?

oliverchang:
> @oliverchang yup probably - do you want to start with just an `--offline` flag for now and go from there?

SGTM! Learning from our mistake with --json / --format though, perhaps something like `--vulnerability-db` may be more extensible. e.g.

```
--db='https://api.osv.dev' # default
--db=offline 
````

[...]

PR #92, opened 2022-12-20
feat: add `semantic` package

G-Rath:
This brings over the `semantic` package from `osv-detector` along with the related generators that generate fixtures based on the versions listed in all current OSVs for the related ecosystem.

Relates to #81


pkg/semantic/parse_test.go
```diff
@@ -0,0 +1,22 @@
+package semantic_test
+
+import (
+	"errors"
+	"github.com/g-rath/osv-detector/pkg/lockfile"
```
oliverchang, line 5: Should these be pointing to osv-scanner instead?

oliverchang (commented):
thank you very much @G-Rath !!! Amazing work as always.

Still catching up on a mountain of things from the break, so I still need to review this in a bit more detail, but just one quick comment for now.

G-Rath:
Tests are now failing because of the new `nuget` and `Alpine` ecosystems - I'll have to look into their version comparison logics and ensure there's an appropriate comparator (`nuget` is because of #132)

oliverchang:
Thanks! I wonder if it's also worth making this an `internal` package for now, to give us more flexibility when it comes to making changes later (and not break semver rules)?

another-rex (approved):

generators/GenerateMavenVersions.java
```diff
@@ -0,0 +1,209 @@
+import org.apache.maven.artifact.versioning.ComparableVersion;
```
oliverchang, line 1: One more minor comment:

Should we put this in say `/test/semantic`? 

I chose `/test` based on https://github.com/golang-standards/project-layout#test, but maybe there's other locations for this too.

G-Rath: If you don't like `generators` (at least at the top level), then probably `scripts` or `tools` (though I didn't use that because I know it does have actual go-related uses) is most suitable.

I don't think `test` is appropriate because these are not tests.

oliverchang: Yeah my main nit was `generators` being in the top level.

`scripts` SGTM!

oliverchang (approved):

PR #100, opened 2022-12-26
Lint readme

wolf99:
Hiya, 
This PR lints the readme according to [markdownlint rules](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md).

If interested, I could add a further PR that adds a workflow that adds a markdown lint checker?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,22 +40,22 @@ You may download the [SLSA3](https://slsa.dev) compliant binaries for Linux, mac
 
 If you're a [**Windows Scoop**](https://scoop.sh) user, then you can install osv-scanner from the [official bucket](https://github.com/ScoopInstaller/Main/blob/master/bucket/osv-scanner.json):
 
-```shell
+```console
 scoop install osv-scanner
 ```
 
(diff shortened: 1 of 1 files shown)


wolf99:
Ive attempted to ad the GitHub email address that is used to sign these commits, to my Google account, so that I ca sign the CLA with that email address.
However the Google account settings requires alternative emails addresses to be verified. As the GitHub email address is not actually an email address as such, this is not possible.
I have signed the CLA using my personal email address but I will not be using that here due to wishing to retain my privay.

Maintainers can chose either:

- Fix the Google account requirements for verification of alternative email addresses for services like GitHub
- Ignore the exact details for the CLA requirement and take my word that I have signed it with my personal address and merge this PR
- Ignore this PR

As you like 🙂

G-Rath:
Thanks - personally I think it would be best to set CI up to check this, ideally with `prettier` with `proseWrap` set to `always` too; is that something you'd be ok to do?

I can't speak for the CLA requirement, though I suspect it probably can't be wavered but again I'm happy to put my name against it if you're ok with that; alternatively I'm happy to handle the whole thing :)

oliverchang:
Looks like the CLA is passing now! Thanks for contribution @wolf99.

oliverchang (approved):

PR #107, opened 2022-12-31
Add support for Alpine Linux /lib/apk/db/installed (Resolves #72)

### LINKED ISSUE #72: Add support for alpine's /lib/apk/db/installed
```
$ osv-scanner --lockfile /lib/apk/db/installed 
could not determine parser for /lib/apk/db/installed
```

It seems osv.dev has [alpine data available](https://osv.dev/list?ecosystem=Alpine) so it seems like a good match. :)

cmaritan:
Add support for Alpine APK Installed Database V2 ( resolves #72 ):

-  `/lib/apk/db/installed`

Reference: [https://wiki.alpinelinux.org/wiki/Apk_spec](https://wiki.alpinelinux.org/wiki/Apk_spec)

I tried to follow similar lockfile parsers (e.g. requirements.txt).
Added some tests, updated README.


pkg/lockfile/parse-apk-installed_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"testing"
+)
+
+func TestParseApkInstalled_Empty(t *testing.T) {
+	t.Parallel()
+
+	packages, err := lockfile.ParseApkInstalled("fixtures/apk/installed_empty")
```
G-Rath, line 11: nit: while `installed` isn't technically the file extension, I think it would be better to name the fixtures with `_installed` at the end, as that's more consistent with the others having their file extension at the end

pkg/lockfile/parse.go
```diff
@@ -32,6 +32,7 @@ var parsers = map[string]PackageDetailsParser{
 	"yarn.lock":                   ParseYarnLock,
 	"gradle.lockfile":             ParseGradleLock,
 	"buildscript-gradle.lockfile": ParseGradleLock,
+	"installed": 				   ParseApkInstalled,
```
G-Rath, line 35: It's probably not worth the extra complexity to try and solve right now, but I'm interested in how others feel about this - while I know `installed` is the name of the file, it feels a little too generic; I wouldn't be surprised if there's another tool out there that creates a file with the same name that we might one day want to support.

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,56 @@
...
+				continue
+			}
+			// Empty line follows a package info block. Append package before going to next one
+			packages = append(packages, curPkg)
+			curPkg = PackageDetails{}
+			continue
+		}
+		// File SPECS: https://wiki.alpinelinux.org/wiki/Apk_spec
+		if strings.HasPrefix(line, "P:") {
+			curPkg.Name = strings.Split(line, "P:")[1]
```
G-Rath, line 39: I think it would be better to use `TrimPrefix` here since that is exactly what we're doing:

```suggestion
			curPkg.Name = strings.TrimPrefix(line, "P:")
```

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,56 @@
...
+		} else if strings.HasPrefix(line, "V:") {
+			curPkg.Version = strings.Split(line, "V:")[1]
```
G-Rath, line 41: I think it would be better to use `TrimPrefix` here since that is exactly what we're doing:

```suggestion
			curPkg.Version = strings.TrimPrefix(line, "V:")
```

G-Rath (changes requested):
This is a great start, thanks!

In addition to my specific comments on the code, could you also:
  * run formatting
  * add a test for when the file doesn't exist
  * add a test that has the lines shuffled (with at least one `V:` before `P:) to confirm that we're not order-dependent
  * update `parse_test.go` to reflect the new parser

cmaritan:
Hello @G-Rath,

thank you for your precious feedback, I'm learning a lot in contributing to this project.
I updated the PR and think that all of your feedbacks have been added.
Let me know.

Thank you very much.
Regards.

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return packages, fmt.Errorf("could not open %s: %w", pathToLockfile, err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+
+	var curPkg PackageDetails = PackageDetails{}
+
+	for scanner.Scan() {
```
G-Rath, line 25: I'm wondering if we should take a similar approach to what I've done [in the `yarn.lock` parser](https://github.com/google/osv-scanner/blob/main/pkg/lockfile/parse-yarn-lock.go#L182-L196), which was to do the parsing in two phases: 1. group the lines for each package, 2. parse each collection of lines in `PackageDetails`.

While technically it's doing more loops, it should make it easier to reduce the amount of conditional branches (especially the duplicate ones outside the loop) which in my experience tends to be worth it - what do you think?

(if you do agree, then I think you should be able to use pretty much the exact same logic/structure as the `yarn.lock` parser, then add an extra check to handle if the name couldn't be found, which could go in a couple of different places)

cmaritan: No problems for me, I'll try and get back to you soon.
I think that additional loops are not an issue since lockfiles are generally small.
There will be more code, but aligned to other parsers.

cmaritan: Hello @G-Rath ,

I have a question:
 
> it should make it easier to reduce the amount of conditional branches
>(especially the duplicate ones outside the loop)

maybe I'm missing something but with your solution, aren't you performing the same pattern and number of conditional branches?
I'm referring to `parse-yarn-lock.go:` [here](https://github.com/google/osv-scanner/blob/main/pkg/lockfile/parse-yarn-lock.go#L32) and  [here](https://github.com/google/osv-scanner/blob/main/pkg/lockfile/parse-yarn-lock.go#L41)

cmaritan: Hello @G-Rath ,
committed changes. Tests ok.
Thank you again for your advice and support.

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if len(group) > 0 {
+		groups = append(groups, group)
+	}
+
+	return groups
+}
+
+func parseApkPackageGroup(group []string, pathToLockfile string) (PackageDetails, error) {
+	var pkg PackageDetails = PackageDetails{}
```
G-Rath, line 37: Let's set the `Ecosystem` and `CompareAs` here, since we know what those values should be and it reduces the potential for bugs by not having them set conditionally (even though we expect that condition should always be met).

Also, you can omit the type here

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,94 @@
...
+			pkg.Ecosystem = AlpineEcosystem
+			pkg.CompareAs = AlpineEcosystem
+		case strings.HasPrefix(line, "V:"):
+			pkg.Version = strings.TrimPrefix(line, "V:")
+		case strings.HasPrefix(line, "c:"):
+			pkg.Commit = strings.TrimPrefix(line, "c:")
+		}
+	}
+
+	if pkg.Name == "" {
```
G-Rath, line 53: We can simplify things a bit by moving this check out of this function - that way we can just always return `PackageDetails`, and the `ParseApkInstalled` function becomes responsible to deciding if we actually consider what we've parsed valid or not.

(You could also do that with `pkg.Version`, but I'd probably keep that in this function to be consistent with the similar function in the `yarn.lock` parser)

pkg/lockfile/parse-apk-installed.go
G-Rath: > maybe I'm missing something but with your solution, aren't you performing the same pattern and number of conditional branches?

Correct, but the footprint is smaller since it's just a length check + an append; the idea of that function is effectively to try and get us out of that raw unstructured state as fast as possible, so that we can do the more critical "business logic" stuff against a good structure which usually lets us avoid more complex versions of that pattern (which is where bugs can easily get introduced).

I view it as a lesser evil type situation - I won't say it's impossible to avoid having a duplicated check like that somewhere in this code, but I would expect it to have a significant tradeoff in at least one area (performance, readability, size, etc) that would make it worse than duplicating these three lines (I'm happy to be proven wrong though, if anyone has some ideas).

G-Rath (changes requested):
This is coming along really well, thanks for sticking with it! I think this should be the last main change we should make before this is good to go.

Also, I'm thinking we should rename the `groupPackageLines` and `parsePackageGroup` functions in the `yarn.lock` parser to `groupYarnPackageLines` and `parseYarnPackageGroup` respectively - are you ok doing that in this PR? otherwise doing it in another is fine.

pkg/lockfile/parse-apk-installed.go
cmaritan: Thank you @G-Rath , refactored as requested.
Also included commit to rename Yarn functions.

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,99 @@
...
+			pathToLockfile,
+		)
+	}
+
+	return pkg
+}
+
+func ParseApkInstalled(pathToLockfile string) ([]PackageDetails, error) {
+
+	file, err := os.Open(pathToLockfile)
```
G-Rath, line 67: nit: don't need this newline

```suggestion
func ParseApkInstalled(pathToLockfile string) ([]PackageDetails, error) {
	file, err := os.Open(pathToLockfile)
```

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,99 @@
...
+			pkg.Version = strings.TrimPrefix(line, "V:")
+		case strings.HasPrefix(line, "c:"):
+			pkg.Commit = strings.TrimPrefix(line, "c:")
+		}
+	}
+
+	if pkg.Version == "" {
+		_, _ = fmt.Fprintf(
+			os.Stderr,
+			"warning: malformed APK installed file. Found no version number in record. File: %s\n",
```
G-Rath, line 57: nit: I think we should try and include the `name` in here to make it easier to check what package is having problems - we can default so something like `"<unknown>"` if it's blank

pkg/lockfile/parse-apk-installed.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return groups
+}
+
+func parseApkPackageGroup(group []string, pathToLockfile string) PackageDetails {
+	var pkg = PackageDetails{}
+
+	pkg.Ecosystem = AlpineEcosystem
+	pkg.CompareAs = AlpineEcosystem
```
G-Rath, line 40: nit: lets set these in the assignment

```suggestion
	var pkg = PackageDetails{
		Ecosystem: AlpineEcosystem,
		CompareAs: AlpineEcosystem,
	}
```

G-Rath (approved):
I've left a few nit-y suggestions, but it's looking pretty solid overall, great stuff!

pkg/lockfile/parse-apk-installed.go
cmaritan: Hello,
we have two choices, I would go with first one, what you think?

**First one** is `parseApkPackageGroup` continuing to return empty string if package name is empty:
```go
	if pkg.Version == "" {
		var pkgPrintName string
		if pkg.Name == "" {
			pkgPrintName = "<unknown>"
		} else {
			pkgPrintName = pkg.Name
		}

		_, _ = fmt.Fprintf(
			os.Stderr,
			"warning: malformed APK installed file. Found no version number in record for package %s. File: %s\n",
			pkgPrintName,
			pathToLockfile,
		)
	}

	return pkg
}
```
**Second one**, `parseApkPackageGroup` directly returns `<unknown>` but I'm not seeing this string used in other parsers:
```go
	if pkg.Name == "" {
		pkg.Name="<unknown>"
	}
        if pkg.Version == "" {
			_, _ = fmt.Fprintf(
			os.Stderr,
			"warning: malformed APK installed file. Found no version number in record for package %s. File: %s\n",
			pkg.Name,
			pathToLockfile,
		)
	}

	return pkg
}
```

G-Rath: yup the first option is the way to go, at least in principle - alternatively you could avoid doing the name check/default by moving the version check to be after the name check [here](https://github.com/google/osv-scanner/pull/107/files#diff-743042e3836e671901997645039e8d1cc69fffb280b9d9377787c072d61c48d9R82).

btw If you decide to stick with your "first one", I would initialize `pkgPrintName` to the value of `pkg.Name` to avoid the `else`

cmaritan: Hello @G-Rath ,
followed your advice. Let me know if it's ready for merging.
Thank you again.

pkg/lockfile/parse.go
```diff
@@ -24,6 +24,7 @@ var parsers = map[string]PackageDetailsParser{
 	"Gemfile.lock":                ParseGemfileLock,
 	"go.mod":                      ParseGoLock,
 	"gradle.lockfile":             ParseGradleLock,
+  "installed":                   ParseApkInstalled,
```
another-rex, line 27:
```suggestion
  "installed":                   ParseApkInstalled,
```
```suggestion
	"gradle.lockfile":             ParseGradleLock,
        "installed":                   ParseApkInstalled,
```

G-Rath (approved):
LGTM

another-rex (approved):
Thanks for implementing this! LGTM.

pkg/lockfile/parse.go
another-rex: I agree it's too generic, it is not unlikely someone might just have a `installed` file in their repository completely unrelated to alpine, which will spit out potentially confusing warnings. 

A quick fix could be to check whether `installed` is in `/lib/apk/db/installed` and only scan that, though that is quite limited, since you might be copying it out, or mounting a docker image ...etc.

G-Rath: I wonder if this should be handled similar to sboms and csvs in `osv-detector` (which are a parser located in `lockfile`, but are never assumed when scanning directories - you have to explicitly say "this is a CSV I want you to check") - because `/lib/apk/db/installed` is very specific, so you're only ever going to get it checked if you run the scanner from anywhere along that path.

So what if instead it had a dedicated flag, like `--check-apk-installed`?

another-rex: That's a good idea, I made an issue here #124 to track it since as you said it's beyond the scope of this PR. We probably should solve this before the next release to avoid changing behavior.

PR #108, opened 2023-01-01
Update ldflags for reproducible builds

jfhovinne:
This should make it possible to reproduce the GoReleaser builds (tested locally with --snapshot and compared with GH release).

See https://goreleaser.com/customization/build/#reproducible-builds

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -13,7 +13,7 @@ builds:
       - -trimpath
     ldflags:
       # prettier-ignore
-      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.CommitDate}}'
+      - '-s -w'
     goos:
       # Further testing before supporting freebsd
       # - freebsd


G-Rath:
My understanding of the documentation you've linked to is that the current setup is what's required for a reproducible build - because:

> by default main.Date is set to the time GoReleaser is run ({{.Date}})

Could you provide some more details on what you're experiencing that makes you think it's not possible to reproduce the GoReleaser builds?

jfhovinne:
> My understanding of the documentation you've linked to is that the current setup is what's required for a reproducible build - because:
> 
> > by default main.Date is set to the time GoReleaser is run ({{.Date}})
> 
> So I would expect this change to result in future builds having the timestamp of when they were built, preventing them from being reproducible.

The documentation says that passing the main.Date variable is optional, but setting it to CommitDate is correct, indeed.

> Could you provide some more details on what you're experiencing that makes you think it's not possible to reproduce the GoReleaser builds?

Actually I see that I was using a wrong build parameter during my tests; I can now reproduce v1.0.2 binaries without any issue.

I would suggest to document the procedure in the readme though; I can open another PR for that.

PR #109, opened 2023-01-02
docs: describe build process

jfhovinne:
I would suggest to document the release process, and in particular how to reproduce the builds locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,13 @@ go install github.com/google/osv-scanner/cmd/osv-scanner@v1
 
 This requires Go 1.18+ to be installed.
 
+### Releasing
+
+Releases are created with [GoReleaser](https://goreleaser.com/).
+
+You can reproduce the builds locally by checking out a specific tag and calling `goreleaser build`, using the same
+Go version as the one used during the actual release (see goreleaser workflows).
+
 ### SemVer Adherence
 
 All releases on the same Major version will be guaranteed to have backward compatible JSON output and CLI arguments.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,6 +59,13 @@ $ go install github.com/google/osv-scanner/cmd/osv-scanner@v1
 
 This requires Go 1.18+ to be installed.
 
+### Releasing
```
G-Rath, line 62: I think it would be better to title this "Building", and frame it from that POV - so start with explaining that you can build from source using `goreleaser build --single-target`, and then mention that version builds can be reproduced by checking out the specific tag and doing a build.

Is probably also worth linking to the [`goreleaser build` docs](https://goreleaser.com/cmd/goreleaser_build/)

README.md
```diff
@@ -59,6 +59,19 @@ $ go install github.com/google/osv-scanner/cmd/osv-scanner@v1
...
+
+You can build from source using [GoReleaser](https://goreleaser.com/) and running:
+
+```bash
+$ goreleaser build --rm-dist --single-target --snapshot
+```
+
+See GoReleaser [documentation](https://goreleaser.com/cmd/goreleaser_build/) for build options.
+
+You can also reproduce the downloadable builds by checking out the specific tag and calling `goreleaser build`,
```
G-Rath, line 72: I think "running" makes more sense here, and is consistent with the first sentence in this section:

```suggestion
You can also reproduce the downloadable builds by checking out the specific tag and running `goreleaser build`,
```

G-Rath (approved):
just got one small comment about a word but after that's applied LGTM 👍

README.md
jfhovinne: Indeed, thanks.

oliverchang (approved):
Thanks! The README.md is getting big enough that I think we're gonna want to have a CONTRIBUTING.md doc soon where this can go into. 

We can move this there.

PR #127, opened 2023-01-09
Update README.md

ChronicLynx:
Correct toml config example to avoid error "Key 'IgnoredVulns.id' has already been defined."

The existing example will produce an error if you try to use it as is to build your config file. Ran into this issue today when implementing it into a project.


another-rex (approved):

another-rex:
Thanks for the fix!

PR #128, opened 2023-01-09
Alpine APK installed: support for flag check-apk-installed

### LINKED ISSUE #124: Alpine `installed` is too generic and might be conflict with other files with the same name

cmaritan:
Hello,
this (very) draft could resolve #124 by adding support to --check-apk-installed command line flag.
APK installed file still in lockfile like CSV but removed from parsers.
@another-rex and @G-Rath Let me know if it's somewhat aligned to your ideas.


cmd/osv-scanner/main.go
```diff
@@ -34,6 +34,12 @@ func run(args []string, stdout, stderr io.Writer) int {
 		Writer:    stdout,
 		ErrWriter: stderr,
 		Flags: []cli.Flag{
+			&cli.StringSliceFlag{
```
G-Rath, line 37: I think this should be a boolean flag rather than a slice, since my understanding is that there is only one expected `installed` file located at `/lib/apk/db/installed` - while some people might want to check arbitrary files as `installed`, this should be possible once #94 is landed and changing this flag afterwards would be a breaking change (+ worse-case if #94 stalls, we should easily add a `--apk-installed-files` flag alongside this one)

G-Rath (changes requested):

cmd/osv-scanner/main.go
```diff
@@ -34,6 +34,12 @@ func run(args []string, stdout, stderr io.Writer) int {
 		Writer:    stdout,
 		ErrWriter: stderr,
 		Flags: []cli.Flag{
+			&cli.BoolFlag{
+				Name:    "check-apk-installed",
```
another-rex, line 38: I'm thinking renaming this to be more generic like `--check-installed-packages`, and remove the short alias for now (we can add it in later if there's demand for it).

Then change the usage description to something like `"Scans installed packages for known vulnerabilities, supports: Alpine Linux"`. We can then expand this over time without adding additional flags, rather than having a very narrow flag specifically for Alpine.

What do you think @G-Rath?

pkg/osvscanner/osvscanner.go
```diff
@@ -82,6 +83,25 @@ func scanDir(r *output.Reporter, query *osv.BatchedQuery, dir string, skipGit bo
 	})
 }
 
+func scanApkInstalledfile(r *output.Reporter, query *osv.BatchedQuery, path string) error {
+	parsedLockfile, err := lockfile.ParseApkInstalled(path)
+
+	if err != nil {
+		return err
```
another-rex, line 90: How clear is the error output if osv-scanner can't find the "/lib/apk/db/installed" file?

another-rex (commented):
Thanks for the quick PR! Just some thoughts on making the name of the flag more generic.

cmd/osv-scanner/main.go
G-Rath: sounds fine to me

oliverchang:
Hmm, if we do implement good filesystem heuristics for scanning, it shouldn't introduce much (if any) noticeable overhead. 

I'd prefer that we keep things simpler with less settings to configure unless we really need it.

G-Rath:
@oliverchang I agree with @another-rex - as a user, I really want the scanner to have two modes:

1. check the system that I'm running in, go find stuff everywhere e.g. apk, apt, my home dir, `/etc`, wherever
2. check "this project" or "that docker image" aka "check what I tell you"

Because those two are not always the same - say for example I'm running the scanner in CI that is using an alpine image; I don't want it telling me about all the vulnerabilities that are in that CI image because the image isn't being used (at least by me) outside of CI.

I already feel similarly with the behaviour of git: right now the scanner is checking the git commit for all my projects whenever I use it, which will never trigger an alert because they're private projects.

So I think it makes sense to at least have a generic flag to switch the scanner between "check the system" and "check this project/dir".

oliverchang:
@oliverchang I agree with @another-rex - as a user, I really want the scanner to have two modes:

> 1. check the system that I'm running in, go find stuff everywhere e.g. apk, apt, my home dir, `/etc`, wherever

I don't think we support the "system I'm running on" part explicitly at all right now. It just happens to work in some cases. Our current two main targets are "project dir / filesystem in some target" and individual files. 

Given that it's harder to remove flags, I'd prefer we add this in the future if we need to. 

In the interests of releasing, I propose we:

1. Disable "installed" scanning for the next release. 
2. Merge in `--parse-as` support 
3. Re-enable "installed" scanning support, and support this as a `--parse-as` model
4. (Some time in the future) We add better system level scanning support, both for container images, and potentially also the system the tool is running on.

G-Rath:
> I don't think we support the "system I'm running on" part explicitly at all right now. It just happens to work in some cases. Our current two main targets are "project dir / filesystem in some target" and individual files.

Yeah I think this represents the first instance of supporting "system I'm running on" in a specific form (or, potential for that I guess)

> Given that it's harder to remove flags, I'd prefer we add this in the future if we need to.

Totally agree, so what you've proposed sounds like a good path forward 👍

another-rex:
It seems this will require more discussion and planning before continuing with implementation, so closing this PR for now until we set up a story for "system I'm running on" scanning.

G-Rath:
Just for future folks, the ability to scan APK has been supported by the scanner for a while now - see [here for more](https://google.github.io/osv-scanner/supported-languages-and-lockfiles/#alpine-package-keeper-and-debian-package-keeper).

PR #129, opened 2023-01-09
Workflow for linting markdown files

wolf99:
Following from #100 this PR adds a workflow to lint any markdown files within the repository.

On #100 it was suggested to use Prettier for this. Instead I have used [David Anson's markdownlint-cli2 action](https://github.com/marketplace/actions/markdownlint-cli2-action). There are a couple reasons not to use Prettier.

- Prettier does not support Go, so would be adding a very broad tool for one simple part of functionality
- Actions are simpler to use in GitHub workflows
- I'm not too familiar with Prettier 😄 

Developers still have the ability to run this check from a development environment using [markdownlint-cli2 itself](https://github.com/DavidAnson/markdownlint-cli2) or any intergration in their IDE, such as [this one for vscode](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint).

That said, if there is a strong preference for Prettier I would be happy to look into using that instead 🙂 .

### DIFF
--- a/.github/workflows/lint-code.yaml
+++ b/.github/workflows/lint-code.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: lint
+name: Lint Code
 
 on:
   push:
@@ -24,8 +24,8 @@ on:
 permissions: read-all
 
 jobs:
-  lint:
-    name: Lint and format
+  lint-go-code:
+    name: Lint and format Go code
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
--- a/.github/workflows/lint-docs.yaml
+++ b/.github/workflows/lint-docs.yaml
@@ -1,20 +1,24 @@
 name: Lint Docs
 
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      # Only run this workflow when these paths have changes
+      - '**.md'
+      - '.github/workflows/lint-docs.yaml'
 
 permissions: read-all
 
 jobs:
-  lint:
+  lint-markdown-docs:
     name: Lint Markdown
     runs-on: ubuntu-latest
     steps:
-      - name: Check out code
-        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
-      - name: Set up Go
-        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+      - name: Checkout code
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+
+      - name: Run markdownlint action
+        uses: DavidAnson/markdownlint-cli2-action@5b7c9f74fec47e6b15667b2cc23c63dff11e449e
         with:
-          go-version: '1.19'
-          check-latest: true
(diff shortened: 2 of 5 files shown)


G-Rath:
I'm not familiar with `markdownlint-cli2` so don't know how different it is by default vs `prettier`, but I'd say the two rank about the same in terms of boardness of tools since you can run it with `npx prettier` so long as Node is installed (see [here](https://github.com/G-Rath/osv-detector/blob/main/.github/workflows/checks.yml#L67-L75) for an example of how to set it up in GitHub Actions), and noting that `prettier` does support other files like YAML, JSON, etc which could be in this repo.

I think though there's no harm in trying to run both because `prettier` is focused on being a formatter, so won't catch stuff like broken links which I expect this linting tool too - also acknowledging I'm a guest on this project, so really it's @another-rex and @oliverchang's call on what tools they want to use so they might feel differently :)

(really, the big thing I'm a fan with `prettier` on markdown files, aside from how easy it is to run and the general consistency factor, is [`proseWrap`](https://prettier.io/docs/en/options.html#prose-wrap) set to `always` because I find it fits nicely with terminals and other file lengths - afaik other tools don't support that, but `markdownlint-cli2` might 🤷)

wolf99:
Hi @G-Rath 
markdownlint can be configured to fail if a given line lengths are exceeded.
Could you suggest what would be the acceptable line length? The old standard 80 char?

G-Rath:
@wolf99 how smart is markdownlint in applying that and does it support autofixing? most tools I've seen typically do line length checks by literally counting the number of chars, whereas prettier uses AST parsing meaning it understands situations where you can't actually split the content across multiple lines (i.e. such as with links and codeblocks).

Also does markdownlint support formatting codeblocks that have contents like JSON?

wolf99:
Re auto-fixing: yes this is supported. Although I am not sure if that is advisable in a PR context as peoples work can get mutated without them realizing what's happened. A failure in a workflow forces them to check the flow output.

Re the line lengths: In short, yes 🙂 
See: https://github.com/DavidAnson/markdownlint/blob/main/doc/md013.md

I haven't looked into how it does it, but it basically ignores certain cases such as tables, code blocks, etc.

Re formatting code within code blocks - no, this action formats markdown only. AFAIK it treat the contents of code blocks as code and ignores them (unless configured to fix things like line lengths specifically within them)

G-Rath:
> Re auto-fixing: yes this is supported. Although I am not sure if that is advisable in a PR context as peoples work can get mutated without them realizing what's happened. A failure in a workflow forces them to check the flow output.

Agreed, but we want to make it easy for people to resolve the problems being flagged (especially ones that are tedious and trivial!) - otherwise it can make it harder for people to try and contribute🙂

DavidAnson:
FYI, markdownlint and related tools do NOT try to fix line length issues. If that is important, you may want to look elsewhere.

wolf99:
Ok - so this is now working as expected, except for fixing inline.
I am not sure if that is agreed as wanted?
Happy to add if it is.

If you prefer Prettier given the above input from David (Thanks David!), then I can open another PR for that?

Precha24 (approved):
Good, thanks

Spiritaine (approved):

Spiritaine (approved):

PR #131, opened 2023-01-10
Do not highlight borders of result table

rhysd:
When I ran osv-scanner in my repo, I got the following output:

screenshot

This looked weird for me since the entire lower half of table is highlighted including borders. I reported this at https://github.com/jedib0t/go-pretty/issues/248 and the maintainer kindly added an option for this.

This PR avoids highlighting borders of the table using the option. Now the output looks like below:

スクリーンショット 2023-01-10 18 56 25

This looks better for me because it is clear that only the row is highlighted.

The `DoNotColorBordersAndSeparators` option requires the latest version of `go-pretty` so I updated the version in `go.mod` and `go.sum` (this was very small update from v6.4.3 to v6.4.4). I ran all tests with `run_tests.sh` and confirmed they passed successfully.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/BurntSushi/toml v1.2.1
 	github.com/CycloneDX/cyclonedx-go v0.7.0
 	github.com/google/go-cmp v0.5.9
-	github.com/jedib0t/go-pretty/v6 v6.4.3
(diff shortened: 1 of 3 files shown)


G-Rath (approved):

another-rex (approved):
Thanks for reporting this to go-pretty and the fix! Looks much better!

PR #141, opened 2023-01-12
Adding re-try for getting a Vuln for the given ID

davift:


G-Rath:
I wonder if it would be better to leverage an existing library like [`go-retryablehttp`](https://github.com/hashicorp/go-retryablehttp) for this - what do you think?

davift:
Dear G-Rath, thanks for your contribution. It is feasible to use the mentioned library. The solution was so simple leveraging the native features already available. Do you see any drawbacks?

G-Rath:
While generally I'm a fan of trying to leverage native features for simple things, in this case I think the library is a better option as it has some additional features (including exponential backoff) and can just be dropped-in which means we could look to use it for other HTTP requests in future without having to do as much work (by extension, this also means it's generally going to be a better tested experience since we're not the only ones using the library).

davift:
I absolutely understand your point but two things comes to my mind. First, I am dealing with a lot of SBOM at work and it is time consuming to initially inspect and later keep track of publicly known vulnerabilities such as supply chain attack. Second, give the simplicity of the feature it consumes less resources. Hope you are OK with my points.

G-Rath:
I'm afraid I still think that we should go with using `go-retryablehttp` since it is providing exactly what we want and more (that we also should want), and is well-maintained by a reputable company.

davift:
Don't understand me wrong, I am not doubting the reputation of the developers of the library proposed. I just have my own opinion based on my professional experience and I don't want to increase the web if I don't think it is required. I am currently overwhelmed with code that calls opensource libraries that call more, and more, and more... I am not a coder, far from this, but as a cyber-security researcher, this would be classified as extending the attack surface (the exact opposite of what I do every day).

G-Rath:
While you are technically correct, it feels you're like you're thinking about this very one dimensionally, since there are other important factors that should be considered such as the effort involved in implementing, maintaining, testing, etc.

Ultimately though I am guest on this project so really it's @another-rex call to make - since we've now both made our opinions clear, I think we should let him weight-in before we continue to avoid rabbitholing too much 🙂

another-rex:
While I agree in general importing in a specialize library has quite a few benefits, I think in this case because there is minimal complexity here I lean towards not importing in another external library. Especially since we currently only have two instances of http calls, both to endpoints we control.

If we need to add additional calls to other endpoints, we can bring in the go-retryablehttp library then.

internal/osv/osv.go
```diff
@@ -156,7 +157,16 @@ func MakeRequest(request BatchedQuery) (*BatchedResponse, error) {
 
 // Get a Vulnerabiltiy for the given ID.
 func Get(id string) (*models.Vulnerability, error) {
-	resp, err := http.Get(GetEndpoint + "/" + id)
+	var resp *http.Response
```
another-rex, line 160: Can you move this logic into a separate function, that way it can be reused for the initial query above as well?

davift: Done!

internal/osv/osv.go
```diff
@@ -119,6 +121,26 @@ func checkResponseError(resp *http.Response) error {
...
+			return nil, fmt.Errorf("Invalid HTTP method: %s", method)
+		}
+		if err == nil {
+			break
+		}
+		time.Sleep(time.Second)
+	}
+	return resp, err
+}
+
```
another-rex, line 143: Lets make this a bit more general and typo proof.
```suggestion
func makeRetryRequest(action func() (*http.Response, error)) (*http.Response, error) {
	var resp *http.Response
	var err error
	retries := 3
	for i := 0; i < retries; i++ {
		resp, err = action()
		if err == nil {
			break
		}
		time.Sleep(time.Second)
	}
	return resp, err
}
```

Now the calls can pass a function in to do whatever request is required.

internal/osv/osv.go
```diff
@@ -131,7 +153,7 @@ func MakeRequest(request BatchedQuery) (*BatchedResponse, error) {
 		}
 		requestBuf := bytes.NewBuffer(requestBytes)
 
-		resp, err := http.Post(QueryEndpoint, "application/json", requestBuf)
+		resp, err := makeRetryRequest("POST", QueryEndpoint, requestBuf)
```
another-rex, line 156:
```suggestion
		resp, err := makeRetryRequest(func() (*http.Response, error) {
			return http.Post(QueryEndpoint, "application/json", requestBuf)
		})
```

internal/osv/osv.go
```diff
@@ -156,7 +178,7 @@ func MakeRequest(request BatchedQuery) (*BatchedResponse, error) {
 
 // Get a Vulnerabiltiy for the given ID.
 func Get(id string) (*models.Vulnerability, error) {
-	resp, err := http.Get(GetEndpoint + "/" + id)
+	resp, err := makeRetryRequest("GET", GetEndpoint+"/"+id, nil)
```
another-rex, line 181:
```suggestion
	resp, err := makeRetryRequest(func() (*http.Response, error) {
		return http.Get(QueryEndpoint + "/" + id)
	})
```

another-rex (changes requested):
Thanks for the update! Just some final suggestions to make the function a bit more general.

davift:
Dear another-rex, 

Thanks for taking time to improve the code I proposed, it was a learning surprise. I can understand what you wrote but I could never do it myself.

Regards,

another-rex (approved):

internal/osv/osv.go
```diff
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
```
another-rex, line 7:
```suggestion
```

PR #144, opened 2023-01-13
Prettier lint md workflow

wolf99:
Alternative to #129 that uses markdownlint-cli action, this PR instead uses a Prettier action.

Initially implemented with Prettier argument `--check` to not change files in place, so as to experiment with the configuration.
Intention to later switch from `--check` to `--write`.

### DIFF
--- a/.github/workflows/lint-code.yaml
+++ b/.github/workflows/lint-code.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: lint
+name: Lint Code
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


wolf99:
OK, I'm out of ideas for how to use actions/checkout properly for this.
Will wait for a response on https://github.com/actions/checkout/issues/1108

G-Rath:
fwiw, I don't think it's worth the complexity of trying to have CI do commits on PRs anyway 🤷

wolf99:
Thanks @G-Rath, if no good answer in while from the issue I raised with actions/checkout, then I'll adjust the PR to not make the commit but instead fail if linting is needed.

Precha24 (approved):
Duplicate of #

Precha24 (approved):

PR #149, opened 2023-01-17
golangci checks in github action, and fixes initial linter issues

another-rex:
- Uses a modified .golangci config file from osv-detector. Some harder lints disabled for now.
- Fixes some of the linting errors.
- Add golangci to be an optional prereq in CONTRIBUTING.md
- Adds golangci github action


G-Rath (approved):
looks good to me! fwiw the only one I think we disagree on is `nlreturn` - I personally find it makes for more readable code, but I can live with it not being enforced if you feel differently 🤷

another-rex:
I don't feel too strongly about it either way. I added it back in with a block-size of 2 so it doesn't enforce cases where we have 
```
if err != nil {
    print_error(err)
    return err
}
```

internal/sbom/spdx.go
```diff
@@ -10,10 +12,10 @@ import (
 )
 
 type SPDX struct{}
-type spdxLoader func(io.Reader) (*spdx.Document2_2, error)
+type SPDXLoader func(io.Reader) (*spdx.Document2_2, error)
```
oliverchang, line 15: is this strictly needed? This makes the type public.

I think the rule is that the casing is consistent for acronyms (i.e. either always UPPERCASE or lowercase).

oliverchang (approved):
LGTM with minor comments. thanks!

run_lints.sh
```diff
@@ -2,4 +2,4 @@
 
 set -ex
 
-go vet ./...
+golangci-lint run
```
G-Rath, line 5: This is assuming that `golangci-lint` is installed and that is the correct version - my understanding is that you can use `go run` here instead to help handle both of those:

```suggestion
go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.50.1 run ./... --max-same-issues 0
```

(`--max-same-issues` is what I've got over in `osv-detector` as I feel it's better to have a lint tool show you all the results at once - feel free to remove if you disagree)

internal/sbom/spdx.go
another-rex: Reverted this change

run_lints.sh
another-rex: Thanks, that makes sense.

PR #152, opened 2023-01-17
Fix endpoint typo

another-rex:
Fix endpoint typo.  We should have some automated integration tests to catch this in the future.


oliverchang (approved):
LGTM. this didn't make its way to any release right?

oliverchang:
And +1 to more testing to catch this.

another-rex:
Nope, it was introduced in the PR that was merged on monday.

PR #157, opened 2023-01-21
chore: remove duplicated generators

G-Rath:
We all missed that I'd not committed the deleted generator files in #92 😂


oliverchang (approved):
thanks for catching!

PR #156, opened 2023-01-21
feat: Render output as a markdown table for use in github comments

deftdawg:
Add a `--markdown` command line switch that will render the output table as a markdown table for use when making comments to github and other places that use markdown.

Resolves #61

## Sample Output
Here's what the output looks like for `osv-scanner -D golang:1.18-rc-stretch --format markdown`:

### Raw
```
Scanned docker image with 189 packages
| OSV URL | Ecosystem | Package | Version | Source |
| --- | --- | --- | --- | --- |
| https://osv.dev/DLA-2948-1 | Debian | debian-archive-keyring | 2017.5+deb9u1 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-3022-1 | Debian | dpkg | 1.18.25 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-2952-1 | Debian | openssl | 1.1.0l-1~deb9u4 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-3008-1 | Debian | openssl | 1.1.0l-1~deb9u4 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-2963-1 | Debian | tzdata | 2021a-0+deb9u2 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-3051-1 | Debian | tzdata | 2021a-0+deb9u2 | golang:1.18-rc-stretch |
```
### Rendered
Scanned docker image with 189 packages
| OSV URL | Ecosystem | Package | Version | Source |
| --- | --- | --- | --- | --- |
| https://osv.dev/DLA-2948-1 | Debian | debian-archive-keyring | 2017.5+deb9u1 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-3022-1 | Debian | dpkg | 1.18.25 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-2952-1 | Debian | openssl | 1.1.0l-1~deb9u4 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-3008-1 | Debian | openssl | 1.1.0l-1~deb9u4 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-2963-1 | Debian | tzdata | 2021a-0+deb9u2 | golang:1.18-rc-stretch |
| https://osv.dev/DLA-3051-1 | Debian | tzdata | 2021a-0+deb9u2 | golang:1.18-rc-stretch |

Issue #85, opened 2022-12-19
Optimise human readable output for narrow terminals 


idkw:
When using the default textual output and the SOURCE path is long enough, the output is truncated.

All lines end with a ` ≈` and are truncated which make the output unusable because it misses the OSV URL entirely.

```
╭────────────────────────────────────────────────────┬───────────┬─────────────────────────┬───────────────────────────────────┬───────────────────────────── ≈
│ SOURCE                                             │ ECOSYSTEM │ AFFECTED PACKAGE        │ VERSION                           │ OSV URL (ID IN BOLD)         ≈
├────────────────────────────────────────────────────┼───────────┼─────────────────────────┼───────────────────────────────────┼───────────────────────────── ≈
│ xxxxx/xxxxx/node_modules/package-lock.json         │ npm       │ lodash                  │ 4.17.4                            │ https://osv.dev/vulnerabilit ≈
│ xxxxx/xxxxx/node_modules/package-lock.json         │ npm       │ lodash                  │ 4.17.4                            │ https://osv.dev/vulnerabilit ≈
│ xxxxx/xxxxx/node_modules/package-lock.json         │ npm       │ lodash                  │ 4.17.4                            │ https://osv.dev/vulnerabilit ≈
│ xxxxx/xxxxx/node_modules/package-lock.json         │ npm       │ lodash                  │ 4.17.4                            │ https://osv.dev/vulnerabilit ≈
│ xxxxx/xxxxx/node_modules/package-lock.json         │ npm       │ lodash                  │ 4.17.4                            │ https://osv.dev/vulnerabilit ≈
│ xxxxx/xxxxx/node_modules/package-lock.json         │ npm       │ lodash                  │ 4.17.4                            │ https://osv.dev/vulnerabilit ≈
│ xxxxx/xxxxx/node_modules/package-lock.json         │ npm       │ lodash                  │ 4.17.4                            │ https://osv.dev/vulnerabilit ≈
│
``` 

I know about the JSON output, but I'd like to have an untruncated textual output too because it is more readable.

oliverchang:
Re-opening as #87 was a partial fix.

oliverchang:
Thanks for trying osv-scanner and opening this issue! 

We've made some changes to mitigate this, and released https://github.com/google/osv-scanner/releases/tag/v1.0.2 which includes:

- Moving the source column to be the last column
- Better support for piping into `less -S`, which lets you horizontally scroll on narrow terminals.
[...]

PR #158, opened 2023-01-21
feat: create `--format` flag

G-Rath:
Helps with #156 and [related to](https://github.com/google/osv-scanner/issues/85#issuecomment-1361890538) #85.

I've gone with `--format` instead of `--output` because it seems more common (looking at CLI tools like ESLint, Rubocop, `bundler-audit`, rspec, Terminus) and also because `--output` tends to match the flag used by some CLI tools for "output to file" such as ESLint and cURL which the scanner might decide to support one day.

I've also kept it simple for now since there's only a couple of formats supported - in future it might make sense to employ enum / const types and look at some fancy map similar to what `lockfile` has for its parsers, but I don't think that's needed here yet.

(this'll need rebasing when #161 is landed)

### DIFF
--- a/cmd/osv-scanner/main.go
+++ b/cmd/osv-scanner/main.go
@@ -22,7 +22,7 @@ func run(args []string, stdout, stderr io.Writer) int {
 	var r *output.Reporter
 
 	cli.VersionPrinter = func(ctx *cli.Context) {
-		r = output.NewReporter(stdout, stderr, false)
+		r = output.NewReporter(stdout, stderr, "")
 		r.PrintText(fmt.Sprintf("osv-scanner version: %s\ncommit: %s\nbuilt at: %s\n", ctx.App.Version, commit, date))
 	}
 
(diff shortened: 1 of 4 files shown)


internal/output/reporter.go
```diff
@@ -9,24 +9,24 @@ import (
...
-	stdout       io.Writer
-	stderr       io.Writer
-	outputAsJSON bool
+	stdout io.Writer
+	stderr io.Writer
+	format string
 }
 
-func NewReporter(stdout io.Writer, stderr io.Writer, outputAsJSON bool) *Reporter {
+func NewReporter(stdout io.Writer, stderr io.Writer, format string) *Reporter {
```
another-rex, line 17: I generally prefer enums for these arguments, but I'm not familiar enough with Go to know if that's practical or not here, Go's enum seems very clunky.

another-rex (approved):
Looks good to me, just a question about whether an enum would be better here.

cmd/osv-scanner/main.go
```diff
@@ -57,6 +57,19 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+					if s != "table" && s != "json" {
+						return fmt.Errorf("unsupported output format \"%s\" - must be either \"table\" or \"json\"", s)
+					}
+
+					return nil
+				},
+			},
 			&cli.BoolFlag{
 				Name:  "json",
 				Usage: "sets output to json (WIP)",
```
another-rex, line 75:
```suggestion
				Usage: "deprecated, use --format json instead. Sets output to json (WIP)",
```

internal/output/reporter.go
G-Rath: Same - the main practical reason I didn't go with one here is that it'd require a cast since there's no way of having the cli library return an enum instead of a string; that's not the end of the world (and really what Go wants you to do in the first place), but combined with the fact that there are currently only a couple of supported formats I felt it would be best to keep it simple for now.

PR #161, opened 2023-01-22
fix: exit with generic non-zero code when there is a general error

G-Rath:
Addresses something I noticed in #94 - I was expecting it to require more hoop jumping but realised this should be a pretty robust/correct solution.

Right now there is only one primary cases where an error is printed but the actual error does not result in the function bailing out, though this will be expanded a little by the changes in #94.

(this'll need rebasing when #158 is landed)

### DIFF
--- a/cmd/osv-scanner/main.go
+++ b/cmd/osv-scanner/main.go
@@ -128,7 +128,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 		}
 
 		r.PrintError(fmt.Sprintf("%v\n", err))
+	}
 
+	// if we've been told to print an error, and not already exited with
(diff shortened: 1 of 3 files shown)


another-rex (approved):
Looks good, there are some printed errors in the parser code that does not go through the reporter, but I think the fix for that is to just have it use the reporter as well.

G-Rath:
Yeah those errors are really "soft panics" - I've resisted using the reporter because it would be a substantial API change (both in terms of the actual signatures + the fact that it means anyone using the parsers as a library would then-on have to pass this whole other structure, which feels very unwieldy).

If you feel really strongly about them, I'd recommend changing them to actual panics.

PR #167, opened 2023-01-27
Make OSV api public

### LINKED ISSUE #163: Make osv client public under pkgs
Mentioned here: https://github.com/guacsec/guac/issues/323

This will allow other go programs to use the osv client without copying the internal osv.go file.

another-rex:
Resolves #163 , once merged the library can be accessed here: 
> 	"github.com/google/osv-scanner/pkg/osv/v1"


pkg/osv/v1/osv.go
```diff
@@ -121,6 +121,7 @@ func checkResponseError(resp *http.Response) error {
 	return fmt.Errorf("server response error: %s", string(respBuf))
```
oliverchang, line 121: Can't make a comment on the actual lines, so I'll leave it here:

nit: Can we make `MaxQueriesPerRequest = 1000` private? That seems like an internal implementaton detail and it's unused elsewhere.

internal/output/table.go
```diff
@@ -6,8 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/google/osv-scanner/internal/osv"
 	"github.com/google/osv-scanner/pkg/models"
+	"github.com/google/osv-scanner/pkg/osv/v1"
```
oliverchang, line 10: I'm not sure including "v1" here is idiomatic. I believe that's always implied. 

See https://go.dev/doc/modules/release-workflow#breaking, which talks about how to publish breaking changes to APIs.

oliverchang (approved):

PR #168, opened 2023-01-27
Support for DPKG (Debian) parser

cmaritan:
As discussed in #164 [here](https://github.com/google/osv-scanner/pull/164#issuecomment-1404060289), this PR adds supports for DPKG parsing.
Structure is similar to APK parser.


KoenDG:
Just want to chime in this would be very useful

pkg/lockfile/dpkg-status.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+const DebianEcosystem Ecosystem = "Debian"
+
+func groupDpkgPackageLines(scanner *bufio.Scanner) [][]string {
```
G-Rath, line 12: noting that while this is exactly the same as what's in `parse-apk-installed` because of how Go compiles I don't think it's worth making into a utility function until we see it being repeated another time (the yarn parser has one that is very similar and you could deduplicate, but it'd require some extra complexity with passing in functions that I also don't think is worth it)

pkg/lockfile/parse.go
```diff
@@ -8,6 +8,10 @@ import (
 	"strings"
 )
 
+// To avoid linter error (goconst) on some parsers.
+// String used when pkgName is not present
```
G-Rath, line 12: nit: I think we should expand a bit more on this

```suggestion

// Represents when a package name could not be determined while parsing.
// Currently, parsers are expected to omit such packages from their results.
```

G-Rath (approved):
looks good! just got one comment about a doc comment :)

pkg/lockfile/parse.go
```diff
@@ -8,6 +8,10 @@ import (
 	"strings"
 )
 
+// Represents when a package name could not be determined while parsing.
+// Currently, parsers are expected to omit such packages from their results.
+const unknownPkgName = "<unknown>"
```
oliverchang, line 13: Can you explain a bit more about why this is required?

pkg/lockfile/dpkg-status.go
```diff
@@ -0,0 +1,105 @@
...
+func parseDpkgPackageGroup(group []string, pathToLockfile string) PackageDetails {
+	var pkg = PackageDetails{
+		Ecosystem: DebianEcosystem,
+		CompareAs: DebianEcosystem,
+	}
+
+	// TODO File SPECS:
+	for _, line := range group {
+		switch {
+		case strings.HasPrefix(line, "Package:"):
```
oliverchang, line 46: Note that according to the OSV-Schema (https://ossf.github.io/osv-schema/#affectedpackage-field), the package name for the "Debian" ecosystem should be the **source** package name. This is also what the existing Debian DB on OSV keys on.

oliverchang (commented):
Thank you for the contribution! Apologies for the delay in reviewing this.

pkg/lockfile/dpkg-status.go
```diff
@@ -0,0 +1,105 @@
...
+	return groups
+}
+
+func parseDpkgPackageGroup(group []string, pathToLockfile string) PackageDetails {
+	var pkg = PackageDetails{
+		Ecosystem: DebianEcosystem,
+		CompareAs: DebianEcosystem,
+	}
+
+	// TODO File SPECS:
```
oliverchang, line 43: Can you also expand on the implication of not supporting this in this TODO comment? e.g. are there going to be package names we miss from not supporting this?

cmaritan: My mistake, the comment is incomplete. I wanted to add link to `status` file format and it was difficult to find. I will update the comment with a useful link

pkg/lockfile/dpkg-status.go
cmaritan: Thanks for the info, I totally missed that!
So, I will refer to `Source: ` filed.
Analyzing it further, my understanding is that there are situations in which an override to the source package version is present in brackets when it differs from that of the binary package.
Example:
```
Package: bsdutils
Source: util-linux (2.36.1-8+deb11u1)
Version: 1:2.36.1-8+deb11u1
```
I will double check this and update PR,

pkg/lockfile/parse.go
cmaritan: Hello, if you refer to the string in itself , it is only used to specify that the value has been searched and not found.
From a technical point of view also empty string "" can be used if preferred.
For the `const`, it was necessary to use it solve a linter error (goconst). The string `<unknown>` is already used in apk parser code. Adding dpkg causes the linter to see multiple instances of the same constant string and thus it suggests using a const.
I can add my previously removed comment `// Avoid linter error (goconst) on some parsers` to make it clear
This const can be used also on apk parser to standardize the approach.

cmaritan:
I also realized that my current parser version is "too aggressive" and can lead to potential false positives.
 `Status:` field is ignored but it can also indicate (rarely I think) that a package is  not in installed state.
Will skip everything not in installed o "wanted installed" state in PR update.

cmaritan:
Hello @oliverchang , now my PR:

1. For package Name uses `Source` field  if present otherwise uses `Package` value (e.g. sudo package)
2. For package Version uses `Source` value between parenthesis if present otherwise uses `Version` field
3. Skip any package that is in `not-installed` or `config-files` states, when we are sure that package is not active in the system

pkg/lockfile/dpkg-status.go
```diff
@@ -0,0 +1,163 @@
...
+			sourcePresent = true
+			source := strings.TrimPrefix(line, "Source:")
+			name, version := parseSourceField(source)
+			pkg.Name = name // can be ""
+			if version != "" {
+				sourceHasVersion = true
+				pkg.Version = version
+			}
+
+		// If Source fieald has no version, use Version field
```
another-rex, line 91:
```suggestion
		// If Source field has no version, use Version field
```

pkg/lockfile/dpkg-status.go
```diff
@@ -0,0 +1,163 @@
...
+const DebianEcosystem Ecosystem = "Debian"
+
+func groupDpkgPackageLines(scanner *bufio.Scanner) [][]string {
+	var groups [][]string
+	var group []string
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if line == "" {
```
another-rex, line 20: nit: Invert this if statement to reduce indentation.

another-rex (approved):
LGTM, just some minor nits.

cmaritan:
Thank you @another-rex , committed suggested changes.

another-rex (approved):

PR #180, opened 2023-02-01
test(APKinstalled): correct wrong function call 

cmaritan:


G-Rath (approved):
ooo nice catch!

oliverchang (approved):
indeed, thanks!

PR #181, opened 2023-02-03
Add text output format

michaelkedar:
Adds `--format text` option for outputting scan results, per #85 

Outputs a numbered list of each vulnerability for each package, including all aliases.

Sample output format:
```
path/to/Cargo.lock:
  regex@1.3.1 is affected by the following vulnerabilities:
      1. https://osv.dev/GHSA-m5pq-gvj9-9vr8 - Rust's regex crate vulnerable to regular expression denial of service
         https://osv.dev/RUSTSEC-2022-0013   - Regexes with large repetitions on empty sub-expressions take a very lo...

path/to/go.mod:
  github.com/gogo/protobuf@1.3.1 is affected by the following vulnerabilities:
      1. https://osv.dev/GHSA-c3h9-896r-86jm - Improper Input Validation in GoGo Protobuf
         https://osv.dev/GO-2021-0053        - Due to improper bounds checking, maliciously crafted input to generate...

path/to/requirements.txt:
  jinja2@2.0 is affected by the following vulnerabilities:
      1. https://osv.dev/GHSA-462w-v97r-4m45 - High severity vulnerability that affects Jinja2
         https://osv.dev/PYSEC-2019-217      - In Pallets Jinja before 2.10.1, str.format_map allows a sandbox escape.
      2. https://osv.dev/GHSA-8r7q-cvjq-x353 - Incorrect Privilege Assignment in Jinja2
         https://osv.dev/PYSEC-2014-8        - The default configuration for bccache.FileSystemBytecodeCache in Jinja...
      3. https://osv.dev/GHSA-g3rq-g295-4j3m - Regular Expression Denial of Service (ReDoS) in Jinja2
         https://osv.dev/PYSEC-2021-66       - This affects the package jinja2 from 0.0.0 and before 2.11.3. The ReDo...
      4. https://osv.dev/GHSA-hj2j-77xm-mc5v - High severity vulnerability that affects Jinja2
         https://osv.dev/PYSEC-2019-220      - In Pallets Jinja before 2.8.1, str.format allows a sandbox escape.
      5. https://osv.dev/PYSEC-2014-82       - FileSystemBytecodeCache in Jinja2 2.7.2 does not properly create tempo...
```
(It's also in colour in terminals)

PR #183, opened 2023-02-03
feat: add experimental offline mode

### LINKED ISSUE #81: Support local DBs

G-Rath:
Resolves #81

~This is based off a lot of the core of the detector - it's not working yet because I need to figure how to handle passing in the queries to the local db given that the detector takes `PackageDetails`, but really the key thing there is how to handle PURL which comes from SBOMs that I don't really know how to use 😅 (idk if I'm just dumb or what, but for some reason I've still not been able to figure how to accurately generate one from a `Gemfile.lock`, `package-lock.json`, etc)~

~If someone could provide some sample SBOMs that would be very useful (I'll also do a PR adding tests using them as fixtures), and also happy to receive feedback on the general approach - there are some smaller bits to discuss, like if fields should be omitted from the JSON output vs an empty array, and the `Describe` related stuff too.~

This is now working, though personally it feels pretty awkward codewise - I know I'm bias but I feel like it would be better to trying to bring across the whole `database` package from the detector, as the API db is pretty much the same and then you'd have support for zips, directories, and the API with extra configs like working directories + an extensive test suite for all three (I don't think it would be as painful as one might first think, even with `osvscanner` having just been made public because that's relatively small).

Still, this does work as advertised - there's definitely a few things that could do with some cleaning up (including if fields should be omitted from the JSON output vs an empty array, and the `Describe` related stuff too) but am leaving them for now until I hear what folks think of the general implementation + my above comment.

I've also gone with two boolean flags rather than the url-based flag @oliverchang suggested because I didn't feel comfortable trying to shoehorn that into this PR as well, and now that we're using `--experimental` it should be fine to completely change these flags in future.


cmd/osv-scanner/main.go
```diff
@@ -34,6 +34,14 @@ func run(args []string, stdout, stderr io.Writer) int {
 		Writer:    stdout,
 		ErrWriter: stderr,
 		Flags: []cli.Flag{
+			&cli.BoolFlag{
+				Name:  "experimental-compare-locally",
```
oliverchang, line 38: nit: let's just name these `experimental-local-db` and `experimental-offline` instead. The offline flag can be more general as an explicit airgapped mode.  

side note: We may also want to add support for a `OSV_LOCAL_DB_PATH` env var or `--local-db-path` flags at some point to give users more control to support airgapped use cases.

oliverchang (commented):
nice, thanks! some initial comments.

G-Rath:
@oliverchang thanks for the initial review - a few of the things you've flagged I already know about but have deliberately left to show a possible future and their use-case is actually already being introduced in parallel PRs.

For now I'd like to focus on the overall direction, after which I'll adjust the whole PR to match and then you can review the finer points, but until then you don't need to spend time reviewing those :)

You've already started to touch on that by asking for a `Db` interface, but I don't think that's worth doing until there are multiple DBs since it's all internal - having said that, the detector already has three DBs including one for the API, so I'd be happy to port that across in which case you'd have your interface.

Let me know what you think

oliverchang:
> @oliverchang thanks for the initial review - a few of the things you've flagged I already know about but have deliberately left to show a possible future and their use-case is actually already being introduced in parallel PRs.
> 
> For now I'd like to focus on the overall direction, after which I'll adjust the whole PR to match and then you can review the finer points, but until then you don't need to spend time reviewing those :)

Sure! What kind of feedback re direction are you looking for in particular? The general structure? (It is a little hard to tease out all of these given the unrelated changes this seems to be bringing in :) ). That said I'll review this in a bit more detail later this week.

> 
> You've already started to touch on that by asking for a `Db` interface, but I don't think that's worth doing until there are multiple DBs since it's all internal - having said that, the detector already has three DBs including one for the API, so I'd be happy to port that across in which case you'd have your interface.

That seems like an easy/cheap thing to do today in a way that allows easy future extension, even if these interfaces are all private?

G-Rath:
> What kind of feedback re direction are you looking for in particular? The general structure?

Yeah something like that - overall this felt like a potentially large shift in structure with both expanding `models` to contain a fair amount of logic and adding the concept of a "db" that gets loaded; I am also probably overthinking this a little bit (especially with the private stuff, since that can be changeable at anytime in the future), but with the internal discussions you've been having I thought it would be best to try and get some initial high-level feedback to make sure it matched what you had in mind before continuing (of course if you're overall happy with it, then we can get into the nitty gritty)

(note I have now removed all of the unused functions since #181 has been closed, which would have been the main feature to use them at this point)

> That seems like an easy/cheap thing to do today in a way that allows easy future extension, even if these interfaces are all private?

Do you mean about just adding the `Db` interface, or about porting over the detectors dbs?

pkg/models/vulnerabilities_test.go
```diff
@@ -0,0 +1,183 @@
+package models_test
+
+import (
+	"encoding/json"
+	"github.com/google/osv-scanner/pkg/models"
```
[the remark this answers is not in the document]
G-Rath: fwiw it's probably best to try and configure the linter to enforce this, since it currently does not :)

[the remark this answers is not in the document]
G-Rath: I have opened #404 to ensure this is enforced via linting

G-Rath:
@oliverchang @another-rexI think you can probably start doing some reviewing if you have time during this week - I've still got to add some CLI tests, and do a maybe final review of the whole thing against our design doc but I don't think there's anything major missing.

If nothing else, I would like to know your thoughts on `ScannerActionsExperimental` - ideally I'd like to actually land that in its own PR pulling in the existing experimental flags, so if you're happy with that I'll get that done later this week too

G-Rath:
@oliverchang @another-rex I think you can probably start doing some reviewing if you have time during this week - I've still got to add some CLI tests, and do a maybe final review of the whole thing against our design doc but I don't think there's anything major missing.

If nothing else, I would like to know your thoughts on `ScannerActionsExperimental` - ideally I'd like to actually land that in its own PR pulling in the existing experimental flags, so if you're happy with that I'll get that done later this week too

internal/offline/zip.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/google/osv-scanner/pkg/models"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type ZipDB struct {
```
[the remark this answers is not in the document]
G-Rath: I'm not sure if it's worth it at this point given you've indicated you don't want to support additional databases and that this is a private package, but I'll re-review my feelings later this week when I do a general review of the codebase.

pkg/models/vulnerability.go
```diff
@@ -2,17 +2,24 @@ package models
...
 
 // Package identifies the affected code library or command provided by the
 // package.
 //
 // See: https://ossf.github.io/osv-schema/#affectedpackage-field
 type Package struct {
-	Ecosystem Ecosystem `json:"ecosystem" yaml:"ecosystem"`
-	Name      string    `json:"name" yaml:"name"`
-	Purl      string    `json:"purl,omitempty" yaml:"purl,omitempty"`
+	Ecosystem lockfile.Ecosystem `json:"ecosystem" yaml:"ecosystem"`
```
G-Rath, line 20: This is kind of a cop-out to get things working - I think technically this is a breaking change, but ultimately I think a decision needs to be made on how these constants are handled since now defined in at least two places and I'm pretty sure reconciling that would require a breaking change too technically?

(I'm pretty sure alternatively I can force Go to convert whenever these two types meet to avoid the breaking change, but have gone with the easier route for now until we've discussed since I could be wrong 🤷)

pkg/models/vulnerability.go
```diff
@@ -167,3 +174,155 @@ func (v Vulnerability) MarshalYAML() (interface{}, error) {
...
+		}
+	}
+
+	return false
+}
+
+func (v Vulnerability) IsAffected(pkg lockfile.PackageDetails) bool {
+	for _, affected := range v.Affected {
+		if affected.Package.Ecosystem == pkg.Ecosystem &&
+			affected.Package.NormalizedName() == pkg.Name {
```
another-rex, line 302: Does pkg.Name also need to be Normalized?

pkg/models/vulnerability.go
```diff
@@ -167,3 +174,155 @@ func (v Vulnerability) MarshalYAML() (interface{}, error) {
...
+	return affected
+}
+
+type Affects []Range
+
+// affectsVersion checks if the given version is within the range
+// specified by the events of any "Ecosystem" or "Semver" type ranges
+func (a Affects) affectsVersion(pkg lockfile.PackageDetails) bool {
+	for _, r := range a {
+		if r.Type != RangeEcosystem && r.Type != RangeSemVer {
```
another-rex, line 230: nit: This check seems to be duplicated in `containsVersion`. (This does allow a early exit here though, so remove it from containsVersion?)

pkg/models/vulnerability.go
```diff
@@ -167,3 +174,155 @@ func (v Vulnerability) MarshalYAML() (interface{}, error) {
...
+		if affected {
+			if e.Fixed != "" {
+				affected = vp.CompareStr(e.Fixed) < 0
+			} else if e.LastAffected != "" {
+				affected = e.LastAffected == pkg.Version || vp.CompareStr(e.LastAffected) <= 0
+			}
+		} else if e.Introduced != "" {
+			affected = e.Introduced == "0" || vp.CompareStr(e.Introduced) >= 0
+		}
+	}
```
another-rex, line 219: I think this assumes the events are sorted, and would not work correctly if it is not sorted?

The schema says that it's recommended to keep the events sorted, but explicitly says that it is not required: 
https://ossf.github.io/osv-schema/#requirements
> There must be at least one `"introduced"` object in the events array. While not required, it’s also recommended to keep the `events` array sorted according to the `affected[].ranges[].type` of the range.

internal/offline/zip.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, db.ArchiveURL, nil)
+
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve OSV database archive: %w", err)
+	}
+
+	req.Header.Set("User-Agent", osv.RequestUserAgent)
```
another-rex, line 97:
```suggestion
	if RequestUserAgent != "" {
		req.Header.Set("User-Agent", osv.RequestUserAgent)
	}
```

internal/offline/zip.go
```diff
@@ -0,0 +1,239 @@
...
+
+	resp, err := http.DefaultClient.Do(req)
```
another-rex, line 99: Provide an option for a custom HTTP client, similar to what was requested for the osv package: https://github.com/google/osv-scanner/pull/357.

another-rex (commented):
Thanks! Added some initial comments.

Also:
- What is the performance of the localdb when doing the local vuln matching like, is there any need for further optimisations, or is it already pretty fast as it is?
- We could add some logging or a loading indicator when downloading the local zip files.

pkg/models/vulnerability.go
G-Rath: This is an implementation of the algorithm provided [in the spec](https://ossf.github.io/osv-schema/#evaluation) which does assume events are sorted - while it feels like handling sorting could be messy, I'm not sure exactly how complex it would be to support but ultimately I think that would be better supported in a follow up PR (ideally with an update to the spec doc to reflect that part of the algorithm too)

pkg/models/vulnerability.go
G-Rath: tbh I might actually just axe this function entirely, and that'll let me get rid of `Affects` too

pkg/models/vulnerability.go
G-Rath: This can actually be removed since the osv.dev databases correctly conform to the spec - I originally had this because of https://github.com/github/advisory-database/issues/52

G-Rath:
> What is the performance of the localdb when doing the local vuln matching like, is there any need for further optimisations, or is it already pretty fast as it is?

This hasn't passed the @spencerschrock check yet, but generally downloading the databases is slower than the rest of the process so yes it's pretty fast.

I am interested in hearing how it performs against all the scorecard repos though because while I tried to benchmark I could have missed something 🤷 

> We could add some logging or a loading indicator when downloading the local zip files.

Whoops yeah adding some logs are another thing I still need to do 😅

pkg/models/vulnerability.go
G-Rath: actually I think probably implement `MarshalJSON` for this like I have for `Versions` in which case I'll need to keep the type still 🤔

internal/offline/zip.go
G-Rath: I don't think there's a gain with this being a private package?

another-rex: Oh good point!

spencerschrock:
> > What is the performance of the localdb when doing the local vuln matching like, is there any need for further optimisations, or is it already pretty fast as it is?
> 
> This hasn't passed the @spencerschrock check yet, but generally downloading the databases (which takes at least 5-10 seconds) is slower than the rest of the process so yes it's pretty fast.
> 
> I am interested in hearing how it performs against all the scorecard repos though because while I tried to benchmark I could have missed something 🤷



From the Scorecard weekly analysis side of things, I imagine a local/cached copy could certainly speed things up and would eliminate a lot of API traffic to osv (assuming we can configure it via our current entry point of `DoScan`). 

I'd be interested in seeing how much of a difference the lack of commit based matching hurts.

oliverchang:
> From the Scorecard weekly analysis side of things, I imagine a local/cached copy could certainly speed things up and would eliminate a lot of API traffic to osv (assuming we can configure it via our current entry point of `DoScan`).
> 
> I'd be interested in seeing how much of a difference the lack of commit based matching hurts.

Today, perhaps not so much. But in a few months we'll have a large chunk of the NVD imported that will be very useful for C/C++ in general where source/commit-based matching is the best we have.

pkg/models/vulnerability.go
oliverchang: Let's avoid breakingthe type here, as we have a few consumers of this package here. Converting between them is a little annoying, but still easily doable. 

It's also nice to reduce the dependencies between the `lockfile` and the `models` packages.

pkg/models/vulnerability.go
```diff
@@ -59,8 +66,8 @@ type Severity struct {
 type Affected struct {
 	Package           Package                `json:"package,omitempty" yaml:"package,omitempty"`
 	Severity          []Severity             `json:"severity,omitempty" yaml:"severity,omitempty"`
-	Ranges            []Range                `json:"ranges,omitempty" yaml:"ranges,omitempty"`
-	Versions          []string               `json:"versions,omitempty" yaml:"versions,omitempty"`
+	Ranges            Affects                `json:"ranges,omitempty" yaml:"ranges,omitempty"`
```
oliverchang, line 69: Are these also breaking changes?

cmd/osv-scanner/main_test.go
```diff
@@ -509,6 +509,209 @@ func TestRun_LockfileWithExplicitParseAs(t *testing.T) {
...
+				Scanned %%/fixtures/locks-many/composer.lock file and found 1 packages
+				No vulnerabilities found
+			`,
+			wantStderr: "",
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
another-rex, line 708: Running these in parallel causes the tests to fail on the initial run. I think it's because it tries to open the zip while it's being written to/updating.

We probably need some sort of locking mechanism to indicate that another process is currently updating the database. This can probably come in a followup PR though, and for now just removing the `t.Parallel()` should be good enough.

another-rex (commented):
Mostly looks good! Left some more comments.

cmd/osv-scanner/main_test.go
```diff
@@ -509,6 +509,209 @@ func TestRun_LockfileWithExplicitParseAs(t *testing.T) {
...
 }
 
+func TestRun_LocalDatabases(t *testing.T) {
+	t.Parallel()
+
+	tests := []cliTestCase{
+		// one specific supported lockfile
+		{
+			name:         "",
+			args:         []string{"", "-experimental-local-db", "./fixtures/locks-many/composer.lock"},
```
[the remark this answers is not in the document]
G-Rath: No that was a typo - Go arg parsing treats `-` and `--` the same

cmd/osv-scanner/main_test.go
G-Rath: I'm not sure if that is what's happening, because it shouldn't be creating the zip file in the first place if its 404ing.

However overall this does at least give a reason to support a flag for configuring the db location in addition to the env variable because then we'd be able to use a temp directory for each test (since you can't set env variables when running in parallel)

G-Rath: Yup, this was [a bug](https://github.com/google/osv-scanner/pull/183/commits/617e10d23ca653e9c8b11cf97c705bc007e8588a)

G-Rath:
Ok this should be good to go now I think, with the main discussion point I'm expecting to be about the introduction of setting the local db path which we'd previously decided would only be configurable by an env variable. I've since realised it makes it harder to test by not having it not just for the scanner as a CLI but anyone using the library that wants to write their own tests.

Also I've deliberately not :
  * added many CLI tests because the core behaviour is already covered by more package-specific tests and it's a lot more work to craft CLI tests to cover the same sorts of things
  * done anything fancy with messaging output or loading handling, like tracking if a db for an ecosystem failed to load as I think there's some pro/cons on the different ways to handle it and that the is good enough for a first implementation, so I'm leaving them as follow-up changes if/when needed (which can be based on user feedback)

Finally, I think this has discovered a bug with the sbom parsing as it looks reports the ecosystem for Alpine packages as APK instead of Alpine.

another-rex:
Fixed the alpine issue in #457

internal/offline/zip.go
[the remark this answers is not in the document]
oliverchang: Ack, this definitely can be left as a future refactoring improvement.

internal/offline/zip.go
```diff
@@ -0,0 +1,249 @@
...
+	"path"
+	"strings"
+
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/google/osv-scanner/pkg/models"
+	"github.com/google/osv-scanner/pkg/osv"
+)
+
+type ZipDB struct {
+	Name            string
```
oliverchang, line 25: Can you please add some comments for what all of these fields mean?

pkg/models/vulnerability.go
```diff
@@ -167,3 +174,130 @@ func (v Vulnerability) MarshalYAML() (interface{}, error) {
...
+
+			if slices.Contains(affected.Versions, pkg.Version) {
+				return true
+			}
+
+			if affectsVersion(affected.Ranges, pkg) {
+				return true
+			}
+
+			if pkg.Version == "" {
```
oliverchang, line 296: Can you add a comment for what this case is handling?

pkg/models/vulnerability.go
```diff
@@ -167,3 +174,130 @@ func (v Vulnerability) MarshalYAML() (interface{}, error) {
...
+		return semantic.MustParse(ar.Events[i].version(), pkg.CompareAs).CompareStr(ar.Events[j].version()) < 0
+	})
+
+	var affected bool
+	for _, e := range ar.Events {
+		if affected {
+			if e.Fixed != "" {
+				affected = vp.CompareStr(e.Fixed) < 0
+			} else if e.LastAffected != "" {
+				affected = e.LastAffected == pkg.Version || vp.CompareStr(e.LastAffected) <= 0
```
oliverchang, line 219: nit: is the `e.LastAffected == pkg.Version` bit needed here given that you have `<=` in the second part?

pkg/models/vulnerability.go
```diff
@@ -167,3 +174,130 @@ func (v Vulnerability) MarshalYAML() (interface{}, error) {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func (v Vulnerability) IsAffected(pkg lockfile.PackageDetails) bool {
+	for _, affected := range v.Affected {
+		if string(affected.Package.Ecosystem) == string(pkg.Ecosystem) &&
```
oliverchang, line 276: nit: this function is getting pretty indented. consider inverting this condition and continuing instead.

pkg/osvscanner/osvscanner.go
```diff
@@ -571,3 +580,36 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
 
 	return vulnerabilityResults, nil
 }
+
+func MakeRequest(
```
oliverchang, line 584: Does this need to be public?

internal/offline/zip.go
```diff
@@ -0,0 +1,249 @@
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve OSV database archive: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("db host returned %s", resp.Status)
+	}
+
+	defer resp.Body.Close()
```
oliverchang, line 114: Should this go before line 110? There can be a body opened even if the status vcode isn't 200 right?

internal/offline/check.go
```diff
@@ -0,0 +1,138 @@
...
+	return lockfile.PackageDetails{
+		Name:      query.Package.Name,
+		Version:   query.Version,
+		Commit:    query.Commit,
+		Ecosystem: query.Package.Ecosystem,
+		CompareAs: query.Package.Ecosystem,
+	}, nil
+}
+
+func setupLocalDBDirectory(localDBPath string) (string, error) {
```
oliverchang, line 46: Can you please add more comments inside this function to enumerate all the different cases that are being handled?

internal/offline/check.go
```diff
@@ -0,0 +1,138 @@
...
+
+		if err != nil {
+			return "", err
+		}
+	}
+
+	return path.Join(localDBPath, "osv-scanner"), nil
+}
+
+func Check(r reporter.Reporter, query osv.BatchedQuery, offline bool, localDBPath string) (*osv.HydratedBatchedResponse, error) {
```
oliverchang, line 85: nit: `Check` is pretty generic. Is there a more specific name we can give this function?

internal/offline/zip.go
```diff
@@ -0,0 +1,249 @@
...
+	for _, vulnerability := range db.Vulnerabilities(false) {
+		if vulnerability.IsAffected(pkg) && !vulnerabilities.Includes(vulnerability) {
+			vulnerabilities = append(vulnerabilities, vulnerability)
+		}
+	}
+
+	return vulnerabilities
+}
+
+func (db *ZipDB) Check(pkgs []lockfile.PackageDetails) (models.Vulnerabilities, error) {
```
oliverchang, line 241: nit: `Check` is pretty generic. Is there a more specific name we can give this function?

oliverchang (commented):
very nice!! Just mostly have some nits, otherwise this looks pretty good to me.

internal/offline/check.go
G-Rath: We could call it something like `CheckForVulnerabilities` but that feels pretty verbose vs names elsewhere in the scanner - while I agree `offline.Check` is still more generic than `osvscanner.DoScan`, I don't think it's as big of a problem given its a private package.

Maybe for symmetry we could rename it `MakeRequest` to match the `osv` package, so long as you don't think it'll be confusing do to `offline.MakeRequest`.

internal/offline/zip.go
G-Rath: see https://github.com/google/osv-scanner/pull/183#discussion_r1281130805 and I also disagree a bit more for this case as its on a struct.

internal/offline/check.go
G-Rath: I ended up heavily refactoring the internals of this function, which included catching a couple of bugs

pkg/models/vulnerability.go
G-Rath: I'm not sure it'd help matters much because the conditions would then be inverted which is typically considered hard to read

internal/offline/zip.go
oliverchang: How about `Scan` ?

pkg/models/vulnerability.go
oliverchang: Fair enough. I typically prefer early exiting to avoid having too indented code which ends up being difficult to read. Another alternative is to just factor our the indented bits into another function.

No strong opinions here so happy to leave as is.

internal/offline/check.go
oliverchang: `MakeRequest` does still seems better here to me.

internal/offline/check.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// if we're implicitly picking a path, try the temp directory before giving up
+	if implicitPath && localDBPath != os.TempDir() {
+		return setupLocalDBDirectory(os.TempDir())
+	}
+
+	return "", err
+}
+
+func Check(r reporter.Reporter, query osv.BatchedQuery, offline bool, localDBPath string) (*osv.HydratedBatchedResponse, error) {
```
oliverchang, line 90: I wonder if `offline` could be named a bit better to make it clearer. It's a bit confusing that an offline package accepts an `offline` argument. Maybe rename this to `updateDBs` or something similar?

internal/offline/zip.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/google/osv-scanner/pkg/osv"
+)
+
+type ZipDB struct {
+	// the name of the database
+	Name string
+	// the url that the zip archive was downloaded from
+	ArchiveURL string
+	// whether this database should make any network requests
+	Offline bool
```
oliverchang, line 30: Maybe we should name this `ShouldUpdate` or something similar ? That makes the intent a lot clearer, and currently as it's named it's a bit confusing in the context of this being inside an "offline" package.

oliverchang (approved):
thanks! LGTM with some final nits.

internal/offline/zip.go
G-Rath: I'm not sure that `Scan` is much better?

internal/offline/zip.go
G-Rath: what do you think about renaming the package itself to something like `local`? I'm not too fussed about renaming the property either, I just felt that `Offline` is pretty clear on what should (or should not) happen and that it is conceivable that other features get added that involve talking to a network.

oliverchang: `local` SGTM.

oliverchang (approved):

PR #185, opened 2023-02-04
fix(osv-query): omit SourceInfo from JSON marshaling

### LINKED ISSUE #184: Question: Is SourceInfo needed in Query to osv.dev?
When making a query to osv.dev,  `omit` field that contains `models.SourceInfo` data is included in JSON request body.
Removing the field from JSON marshaling seems not breaking osv-scanner output and avoid sending (possibly) unneeded data (in particular: local paths).

See:
https://github.com/google/osv-scanner/blob/f8eec77412e15fc00ace1e6891d38839c3c09f56/pkg/osv/osv.go#L129-L139
https://github.com/google/osv-scanner/blob/f8eec77412e15fc00ace1e6891d38839c3c09f56/pkg/osv/osv.go#L35-L40
https://github.com/google/osv-scanner/blob/f8eec77412e15fc00ace1e6891d38839c3c09f56/pkg/models/results.go#L70-L73

cmaritan:
Resolve #184 (if `models.SourceInfo` is confirmed to be unneeded!)

### DIFF
--- a/pkg/osv/osv.go
+++ b/pkg/osv/osv.go
@@ -36,7 +36,7 @@ type Query struct {
 	Commit  string            `json:"commit,omitempty"`
 	Package Package           `json:"package,omitempty"`
 	Version string            `json:"version,omitempty"`
-	Source  models.SourceInfo `json:"omit"`
+	Source  models.SourceInfo `json:"-"`
 }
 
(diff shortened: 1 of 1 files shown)


another-rex (approved):
Thanks! You're right it is not needed, did not know `-` is an option.

Issue #176, opened 2023-01-30
Add `io.Reader` variants to `lockfile` package


picatz:
At this time, the `lockfile` package's external functions only accept a path name, so it [expects a file to be available on disk](https://github.com/google/osv-scanner/issues/95#issuecomment-1364364700). For my use case, this isn't always true or ideal. Preferably, there would be an option to use a path name _and_ an `io.Reader`.

@cmaritan's work in https://github.com/google/osv-scanner/pull/164 with `ParseApkInstalledFromReader` serves as a good foundation to be expanded upon:

https://github.com/google/osv-scanner/blob/a4186252b74b5071dfb25f04a24a9fe92a1af436/pkg/lockfile/apk-installed.go#L74

https://github.com/google/osv-scanner/blob/a4186252b74b5071dfb25f04a24a9fe92a1af436/pkg/lockfile/apk-installed.go#L84

oliverchang:
Hey @picatz, thanks for the suggestion! 

Indeed this is something we'd like to have. We've also been discussing internally with some other potential users of this library around some other refactorings to make this package more usable generally (that includes this). 

We'll hopefully have something ready to share for input soon :)

G-Rath:
fwiw I've made a start on this - patch is attached; main thing I'm still doing is figuring out how to make the JSON/YAML/TOML/XML parsers play nice with an `io.Reader`, and then the test suites for all parsers should probably be turned into a couple of table-based tests so that they can easily run against both the file and `io.Reader` versions of each function.

Am pausing for now in favor of #81, but will hopefully be able to have it done in a couple of weeks.

```
Index: pkg/lockfile/parse-yarn-lock.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/pkg/lockfile/parse-yarn-lock.go b/pkg/lockfile/parse-yarn-lock.go
--- a/pkg/lockfile/parse-yarn-lock.go	(revision 5df9444f04bc86e3c64e68ef65bc8e2711edb32c)
+++ b/pkg/lockfile/parse-yarn-lock.go	(date 1675375026394)
@@ -3,6 +3,7 @@
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"net/url"
 	"os"
 	"regexp"
@@ -170,19 +171,33 @@
 	}
 }
 
-func ParseYarnLock(pathToLockfile string) ([]PackageDetails, error) {
[...]

Issue #102, opened 2022-12-27
Failed to determine version of "something" while parsing a yarn.lock - please report this!


AxetNord:
I don't know, where i may report about this. Report here.
[yarn.lock.txt](https://github.com/google/osv-scanner/files/10306499/yarn.lock.txt)

G-Rath:
Thanks, this is the right place to report this - I'll have a look tomorrow morning, but in the meantime it would be useful to know more about your project to better understand why your `yarn.lock` is structured as it is (assuming it's not just invalid).

Could you also confirm that running `yarn install` doesn't change the lock at all?

AxetNord:
before `yarn install`
[please report this.txt](https://github.com/google/osv-scanner/files/10306644/please.report.this.txt)
After `yarn install` no errors found. Sorry for that

G-Rath:
@AxetNord good to know, though it looks like this happened because your `yarn.lock` has properties quoted e.g.:

```
"word-wrap@^1.2.3":
  "integrity" "sha512-Hz/mrNwitNRh/HUAtM/VT/5VH+ygD6DV7mYKZAtHOrbs8U7lvPS6xf7EJKMF0uW1KJCl0H701g3ZGus+muE5vQ=="
  "resolved" "https://registry.npmjs.org/word-wrap/-/word-wrap-1.2.3.tgz"
  "version" "1.2.3"
```

Instead of:

```
"word-wrap@^1.2.3":
  integrity "sha512-Hz/mrNwitNRh/HUAtM/VT/5VH+ygD6DV7mYKZAtHOrbs8U7lvPS6xf7EJKMF0uW1KJCl0H701g3ZGus+muE5vQ=="
  resolved "https://registry.npmjs.org/word-wrap/-/word-wrap-1.2.3.tgz"
  version "1.2.3"
```

which is something that should be very easy to support but it would be useful to know how your `yarn.lock` got like that first - initially I wondered if it was because of the earlier versions of `yarn` but I've tried all the way back to v0.15.3 with no success.

PR #186, opened 2023-02-04
feat: support diagnostics in `lockfile` parsers

G-Rath:
This is something I probably should have done from the start, and is based on `Diagnostics` from [Terraform providers](https://developer.hashicorp.com/terraform/plugin/framework/diagnostics).

My main motivations are:
  1. there's a lot of output when running the tests (when a test fails)
  2. writing to stderr like this is a bit of a smell, which came about because I wanted to avoid requiring an arbitrary object (`Reporter`) for those using the `lockfile` as a library
  3. function signatures cannot be changed without a new major, but parsers might need to surface other auxiliary info in future, and with #176 every parser is going to have a pair function, so the work involved in this change can grow substantially

I'm opening this as a draft because frankly there's quite a lot of refactoring to do, turning all the existing tests into table-based tests (which should also help with landing #176 too), and some discussion points that we can have in the meantime (including "do other people think this is a good idea?").

For now I've mostly just focused on updating the individual parsers and converting their tests, but I'm aware some thinking will need to be done about what happens to `Parse` and co - there's a few different approaches which include "leave it for now", but I'm leaving that to last.

(also, I've not yet decided between `diag`, `diags`, or `diagnostics` - it should probably be the middle one tbh)


G-Rath:
@picatz interested in if you have any thoughts here, since you're the main person I know that is using the `lockfile` package

G-Rath:
I ended up [writing a script](https://gist.github.com/G-Rath/63d213b72f4efb3f8a967e93024d36e8) to easily convert the parser tests to be table-based - it should be useful even if this doesn't end up getting landed.

picatz:
👋 Hello @G-Rath, I think reporting diagnostics is a good idea conceptually, but I haven't dug through the implementation details yet! I'll find some time this week to do so and report back any other thoughts I have.

picatz:
Hey @G-Rath, just to follow up: I wasn't able to find time last week. But, this is something I'm definitely still interested in!

pkg/lockfile/apk-installed.go
```diff
@@ -58,21 +58,26 @@ func parseApkPackageGroup(group []string, pathToLockfile string) PackageDetails
...
 	}
 
 	return pkg
 }
 
 func ParseApkInstalled(pathToLockfile string) ([]PackageDetails, error) {
+	return parseFileAndPrintDiag(pathToLockfile, ParseApkInstalledWithDiagnostics)
+}
+
+func ParseApkInstalledWithDiagnostics(pathToLockfile string) ([]PackageDetails, Diagnostics, error) {
```
picatz, line 75: FWIW, thinking about a separate `Diagnostics` return value: in HCL/Terraform-land, the [diagnostics](https://github.com/hashicorp/terraform/blob/3d1a58d5b54bdcde038a6f06150e566f94e94357/internal/tfdiags/diagnostics.go#L15-L22) are returned _like an error_, not a separate thing. Diagnostics [carry a severity](https://github.com/hashicorp/terraform/blob/3d1a58d5b54bdcde038a6f06150e566f94e94357/internal/tfdiags/diagnostic.go#LL10) that determines if they're either a [warning](https://github.com/hashicorp/terraform/blob/3d1a58d5b54bdcde038a6f06150e566f94e94357/internal/tfdiags/diagnostic.go#L33) or an [error](https://github.com/hashicorp/terraform/blob/3d1a58d5b54bdcde038a6f06150e566f94e94357/internal/tfdiags/diagnostic.go#L32). 

The convention becomes to use the [`diags.HasErrors()`](https://github.com/search?q=repo%3Ahashicorp/terraform%20HasErrors&type=code) method to check if the returned diagnostics have any `error`s that need to be dealt with; otherwise you know the rest are just warnings that might be useful to log, or show the user, but not halt further execution.

While this is a _minor_ detail, I think the pattern is useful in these situations.

```go
pkgDetails, diags := ParseApkInstalledWithDiagnostics(pathToLockfile)
if diags.HasErrors() {
  // add info and return an `error` if appropriate
  return fmt.Errorf("failed to parse APK file %q in %q: %w", pathToLockfile, containerImage, diags.Err())
}
```

Either way, I want diagnostics! But, I figured I'd at least bring this up if you want to discuss it further 😄

pkg/lockfile/diagnostics.go
```diff
@@ -0,0 +1,24 @@
...
+	"os"
+)
+
+type Diagnostics struct {
+	Warnings []string
+}
+
+func (diag *Diagnostics) Warn(warn string) {
+	diag.Warnings = append(diag.Warnings, warn)
+}
```
picatz, line 14: Assuming there are ever cases where the warnings are duplicated, the consumer will need to consider de-duplication to present them to users. Not sure if it's worth it, but maybe a `map[string]struct{}` or something like that could provide that out the box? What do you think?

pkg/lockfile/dpkg-status.go
```diff
@@ -142,11 +146,10 @@ func ParseDpkgStatus(pathToLockfile string) ([]PackageDetails, error) {
 		}
 
 		if pkg.Name == "" {
-			_, _ = fmt.Fprintf(
-				os.Stderr,
-				"warning: malformed DPKG status file. Found no package name in record. File: %s\n",
+			diag.Warn(fmt.Sprintf(
+				"warning: malformed DPKG status file. Found no package name in record. File: %s",
```
picatz, line 150: 🤔 I don't think we _need_ to the `warning:` prefix. 

The type information alludes to that being `lockfile.Diagnostics.Warnings` at the moment. This seems like a display thing that an application would _add_ to an unstructured output like STDERR.

As a library consumer, I would know it is a diagnostic warning, and I can choose the display/styling in a CLI or whatever. I would much rather want to see a warning string like:

```console
found malformed DPKG status file %q
```

picatz (commented):
Spent some time this afternoon looking over the implementation. Left some thoughts I had based on my initial impressions. 

Overall, I think this definitely looks like the right direction!

pkg/lockfile/dpkg-status.go
G-Rath: Totally agree, I just didn't change it because this way the output with this change is 1:1 with what it is currently - I know this sort of CLI output shouldn't be considered breaking to change, but I think it's nice to start with being as close to 1:1 as possible as a way of proving its stability.

pkg/lockfile/diagnostics.go
G-Rath: I feel at this point it's best to leave that to the consumer to handle as there's no defined spec around these warnings and _hopefully_ we shouldn't need to expand this further, but also if we do it should be easy to add a `WarningsMap` or something.

pkg/lockfile/apk-installed.go
G-Rath: hmm that's interesting - I think it makes sense to consider because there's no real downside so long as `diags.Err()` exposes the underlying error? (actually, the main one I can think of is that IDEs like Goland would not tell you that you need to check if there was an error, which I'm pretty sure it does based on if the return is an `error`).

Though I'm not sure if there's an advantage to returning an object vs a slice?

G-Rath:
btw I hadn't realised this but go 1.20 added support for [multiple errors](https://tip.golang.org/doc/go1.20#errors), which I'm wondering might be a suitable alternative to this - while it's a little more cumbersome because it'd mean having to more deeply check the error value, it would avoid having to change the signature making it arguably less breaking given that there's a couple of ways diagnostics could be implemented (as showcased [in this comment][(https://tip.golang.org/doc/go1.20#errors](https://github.com/google/osv-scanner/pull/186#discussion_r1114915994))) 🤔

spencerschrock:
Just chiming in that being able to suppress the `os.Stderr` writes would be beneficial for [Scorecard](https://github.com/ossf/scorecard).

We currently call `DoScan` [here](https://github.com/ossf/scorecard/blob/35a7dd5b254443aa6a12cdc9679458be4b07d7de/clients/osv.go#L42) when checking for vulns. When we scan public repos, we sometimes encounter malformed `yarn.lock` files that write hundred of lines to `os.Stderr` due to #102.

I don't think this PR is enough to change the `DoScan` behavior (and that seems outside the scope of this PR), but I'm excited for the work being put in here.

G-Rath:
tbh I wish I thought of this sooner, but probably the easiest short-term solution could actually be to have a global var like this:

```
var HandleParserWarning = func (warning string) {
	_, _ = fmt.Fprintf(os.Stderr, warning)
}

// parse-maven-lock.go
HandleParserWarning(fmt.Sprintf(
	"Failed to resolve version of %s: property \"%s\" could not be found",
	mld.GroupID+":"+mld.ArtifactID,
	results[1],
))
```

(it could also have a "maybeFilePath" argument too 🤔 )

Then tools like scorecard could assign a function to do whatever they want, and it should be pretty easy to move away from later (it should be easy to continue to call the function along with whatever replacement solution comes along, and arguably it could just not be called anymore so long the function itself still exists).

@oliverchang @another-rex given it sounds like scorecard in particular are having a bit of a hard time with how uncontrollable these outputs are, does that sound like an acceptable middleground for now to make their lives a bit easier?

oliverchang:
> tbh I wish I thought of this sooner, but probably the easiest short-term solution could actually be to have a global var like this:
> 
> ```
> var HandleParserWarning = func (warning string) {
> 	_, _ = fmt.Fprintf(os.Stderr, warning)
> }
> 
> // parse-maven-lock.go
> HandleParserWarning(fmt.Sprintf(
> 	"Failed to resolve version of %s: property \"%s\" could not be found",
> 	mld.GroupID+":"+mld.ArtifactID,
> 	results[1],
> ))
> ```
> 
> (it could also have a "maybeFilePath" argument too 🤔 )
> 
> Then tools like scorecard could assign a function to do whatever they want, and it should be pretty easy to move away from later (it should be easy to continue to call the function along with whatever replacement solution comes along, and arguably it could just not be called anymore so long the function itself still exists).
> 
> @oliverchang @another-rex given it sounds like scorecard in particular are having a bit of a hard time with how uncontrollable these outputs are, does that sound like an acceptable middleground for now to make their lives a bit easier?

@spencerschrock how much of a problem is this for Scorecard right now? Is this something that needs a more immediate fix, or something that can wait a bit for a larger, more substantial refactoring? 

@G-Rath Something like `HandleParserWarning` does seem like a nice easy way to handle a lot of these problems, including the ones you brought up as the original motivation for this PR. Another nice way to generalize this might be to just expose a `WarningLogWriter` `io.Writer` instead. 

It's also a nice way of reducing the variations of each parser that we want to support. That said, I'm a little wary of adding new interfaces if we're going to refactor all of this again later. If we go with this I think we would have to commit to this approach for achieving the goals in the PR for the forseeable future.

G-Rath:
@oliverchang fwiw this is also impacting my ability to generally work on the scanner, because these warnings also get printed out by the test suite so I have to sift through a fair amount of output every time I make a change (which'll only get worse as more stuff is added) 🙃 

While I agree with generally trying to avoid a lot of big refactors, from what we've been talking about offline and what you've said over on issues like #176 it seems like it there's a fair bit to unpack before you'll feel confident doing such a refactor.

oliverchang:
> @oliverchang fwiw this is also impacting my ability to generally work on the scanner, because these warnings also get printed out by the test suite so I have to sift through a fair amount of output every time I make a change (which'll only get worse as more stuff is added) 🙃
> 
> While I agree with generally trying to avoid a lot of big refactors, from what we've been talking about offline and what you've said over on issues like #176 it seems like it there's a fair bit to unpack before you'll feel confident doing such a refactor.

Thinking more, either way the approach you suggested sounds good to me, with the possible exception of exposing an `io.Writer` to write logs to instead of a function. 

@another-rex do you have any thoughts?

another-rex:
Doing this quick compromise sounds good to me, and both having `io.Writer` or setting a full function seem fine, though having full function makes it slightly easier to do additional error handling if required.

spencerschrock:
> @spencerschrock how much of a problem is this for Scorecard right now? Is this something that needs a more immediate fix, or something that can wait a bit for a larger, more substantial refactoring?

Minimal. The vast majority of the output was taken care of by #250 (thanks @G-Rath).

oliverchang:
> Doing this quick compromise sounds good to me, and both having `io.Writer` or setting a full function seem fine, though having full function makes it slightly easier to do additional error handling if required.

I think an  `io.Writer` expresses the intention the best here -- as an abstraction dealing with a log output sink. We can still have custom `io.Writer` implementations if more involved error handling needs to be done.

G-Rath:
hmm I agree with Rex at this point, because I'd expect using a writer means `lockfile` then has to decide how to format the messages whereas with a function you could pass something like an error or an interface with extra detail, and then the consuming library can decide what to do with things.

But I think I'll probably have a better sense once I see an implementation of either solution.

G-Rath:
I'm going to close this for now since it requires rebasing + further discussion anyway, and the changes will still be retained by GitHub in case we want to dig up the implementation.

PR #189, opened 2023-02-05
feat: support passing `io.Reader` to `lockfile` parsers

### LINKED ISSUE #176: Add `io.Reader` variants to `lockfile` package

G-Rath:
Resolves #176, branched off #186


G-Rath:
Landing #260 has made this a bit more complex, because now the `requirements.txt` parser supports reading extra files based on the contents of the file currently being parsed which would involve another sizable change to the interface.

As such I'm going to put this PR on the backburner until the appropriate internal conversations have happened :)

G-Rath:
I'm going to close this for now since it requires rebasing + further discussion anyway, and the changes will still be retained by GitHub in case we want to dig up the implementation.

PR #190, opened 2023-02-05
fix: add missing toml tags to struct (and update linter)

G-Rath:
I was curious and ended up doing the work so might as well land it 🤷

This also removes one of the three "<linter> is disabled because generics" warnings.


another-rex:
Thanks!, can you also change the version in this file as well: .github/workflows/lint.yaml

another-rex (approved):

PR #191, opened 2023-02-07
Make scanner respect .gitignore files

### LINKED ISSUE #165: directory scanning should respect gitignore by default
OSV-scanner should respect the gitignore by default when walking directories and files.

A project might have a lot of deep directories (e.g. node_modules, build files...etc) that both slows down the scanner and produces a lot of noise. These directories are generally put in the .gitignore files, so ignoring them while scanning solves this by default.

A flag should be added to disable this .gitignore skipping.

michaelkedar:
Closes #165 

Also changes `GetCommitSHA` to no longer rely on the system's git executable.

By default, OSV-Scanner will skip files/directories that are ignored by a git project's `.gitignore` files. Added a flag `--no-ignore` to disable this behaviour.


cmd/osv-scanner/main.go
```diff
@@ -89,6 +89,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Usage:   "check subdirectories",
 				Value:   false,
 			},
+			&cli.BoolFlag{
+				Name:  "ignore-gitignore",
```
michaelkedar, line 93: Suggestions for an alternative name for this flag would be appreciated

michaelkedar:
Hmm, because the test files that should be ignored are actually in the repo, git doesn't actually recognize them as ignored files and the test doesn't work.

pkg/osvscanner/osvscanner.go
```diff
@@ -87,6 +101,33 @@ func scanDir(r *output.Reporter, query *osv.BatchedQuery, dir string, skipGit bo
 	})
 }
 
+// canCheckIgnore checks if the command `git check-ignore` will work on the system,
+// which is used to check .gitignore files
+func canCheckIgnore() bool {
+	cmd := exec.Command("git", "check-ignore")
```
oliverchang, line 107: Rather than expecting an external dependency, we should keep our tool as standalone as possible.

Can we instead use https://github.com/src-d/go-git/tree/master/plumbing/format/gitignore ? 

https://cs.github.com/kubernetes-sigs/bom/blob/4415dd6c071a181042f6ebcd197f37608c842f62/pkg/spdx/implementation.go#L499 has an example usage. 

Not sure if there's a higher level convenience parser function, but this appears to be how to read arbitrary .gitignore files which is still quite simple:

https://github.com/src-d/go-git/blob/8b0c2116cea2bbcc8d0075e762b887200a1898e1/plumbing/format/gitignore/dir.go#LL48C14-L48C14 may also provide a higher level/more convenient way to do what we want.

oliverchang (commented):
thanks for working on this!

cmd/osv-scanner/main.go
another-rex: `ripgrep` uses the flag `--no-ignore`, so that could be an alternative name.

pkg/osvscanner/osvscanner.go
michaelkedar: Will try using that package.

I did notice that the `git` command was used elsewhere already:
https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/osvscanner.go#L173-L174

maybe we could also replace with go-git there?

oliverchang: Good callout! Yes, we should as well.

cmd/osv-scanner/main.go
michaelkedar: Changed to `--no-ignore` now

cmd/osv-scanner/main_test.go
```diff
@@ -194,6 +194,36 @@ func TestRun(t *testing.T) {
...
+			name:         "",
+			args:         []string{"", "--recursive", "--no-ignore", "./fixtures/locks-gitignore"},
+			wantExitCode: 0,
+			wantStdout: `
+				Scanning dir ./fixtures/locks-gitignore
+				Scanned %%/fixtures/locks-gitignore/Gemfile.lock file and found 1 packages
+				Scanned %%/fixtures/locks-gitignore/composer.lock file and found 1 packages
+				Scanned %%/fixtures/locks-gitignore/ignored/Gemfile.lock file and found 1 packages
+				Scanned %%/fixtures/locks-gitignore/ignored/yarn.lock file and found 1 packages
+				Scanned %%/fixtures/locks-gitignore/subdir/Gemfile.lock file and found 1 packages
```
another-rex, line 220: Can you add a TODO here to state that this currently doesn't parse gitignore in subdirs?

pkg/osvscanner/osvscanner.go
```diff
@@ -87,6 +117,49 @@ func scanDir(r *output.Reporter, query *osv.BatchedQuery, dir string, skipGit bo
...
+type gitIgnoreMatcher struct {
+	matcher  gitignore.Matcher
+	repoPath string
+}
+
+func parseGitIgnores(dir string) (*gitIgnoreMatcher, error) {
+	// We need to parse .gitignore files from the root of the git repo to correctly identify ignored files
+	// Defaults to current directory if dir is not in a repo or some other error
+	// TODO: Won't parse ignores if dir is not in a git repo, and is not under the current directory (e.g ../path/to)
+	// TODO: What is the desired behaviour for non git-repos?
```
another-rex, line 129: I think the current behavior of defaulting to current dir as the root is good.

pkg/osvscanner/osvscanner.go
```diff
@@ -42,19 +46,45 @@ var VulnerabilitiesFoundErr = errors.New("vulnerabilities found")
...
 		if err != nil {
 			r.PrintError(fmt.Sprintf("Failed to walk path %s\n", err))
 			return err
 		}
 
+		if useGitIgnore {
+			match, err := ignoreMatcher.match(path, info.IsDir())
+			if err != nil {
+				r.PrintText(fmt.Sprintf("Failed to resolve gitignore for %s: %v", path, err))
+				// Don't skip if we can't parse know - potentially noisy for directories with lots of items
```
another-rex, line 78:
```suggestion
				// Don't skip if we can't parse now - potentially noisy for directories with lots of items
```

another-rex (commented):
Thanks for switching the git os calls over to the library as well! 

~~Do you think it's worth it to implement the subdir .gitignore scanning as well now? Or push it to a later PR.~~ I think it does do the subdirectory ignores now after reading through it again. Is that right?

another-rex (approved):

michaelkedar:
> I think it does do the subdirectory ignores now after reading through it again. Is that right?

Yep - `gitignore.ReadPatterns` evaluates subdirectories

pkg/osvscanner/osvscanner.go
michaelkedar: Cool - removed that TODO

pkg/osvscanner/osvscanner.go
```diff
@@ -87,6 +117,48 @@ func scanDir(r *output.Reporter, query *osv.BatchedQuery, dir string, skipGit bo
...
 
+type gitIgnoreMatcher struct {
+	matcher  gitignore.Matcher
+	repoPath string
+}
+
+func parseGitIgnores(dir string) (*gitIgnoreMatcher, error) {
+	// We need to parse .gitignore files from the root of the git repo to correctly identify ignored files
+	// Defaults to current directory if dir is not in a repo or some other error
+	// TODO: Won't parse ignores if dir is not in a git repo, and is not under the current directory (e.g ../path/to)
```
oliverchang, line 128: Can we create a bug to track this?   By "current directory", do you mean "current working directory"?

oliverchang (commented):
LGTM with one comment.

pkg/osvscanner/osvscanner.go
michaelkedar: Made #202

PR #193, opened 2023-02-07
Add --ok-empty CLI option

C-Pro:
Add `--ok-empty` flag that suppresses returning errors if no packages are found.
It is useful e.g. when `osv-scanner` is used in an organisation-wide [GitHub Required Workflow](https://docs.github.com//en/actions/using-workflows/required-workflows). Some projects/repositories may lack dependencies and checking for a particular return code in CI actions can be cumbersome, so I believe this flag can be useful.


G-Rath (changes requested):
Can you undo the unrelated formatting changes?

Also, the CLI already has a dedicated exit code for this situation so you can test for that in your action - I'm not saying this is a bad idea, but wondering if it is something that needs a new dedicated flag for?

(if people think a new flag is the answer, then I would suggest naming it something like `--allow-empty` or `--ignore-empty`)

C-Pro:
> Can you undo the unrelated formatting changes?

Done.

> Also, the CLI already has a dedicated exit code for this situation so you can test for that in your action - I'm not saying this is a bad idea, but wondering if it is something that needs a new dedicated flag for?

I am checking the code now, but one liner looks messy and magical.
`/osv-scanner -r . || if [ $? -eq 128 ]; then true; fi`

> (if people think a new flag is the answer, then I would suggest naming it something like `--allow-empty` or `--ignore-empty`)

Agree on this 👍 Renamed to `--allow-empty`.

oliverchang:
Thanks for this contribution!

We should document this somewhere in this repo, but our general philosophy to supporting flags is to be as minimal as possible, especially in cases when there's an easy alternative (as @G-Rath suggested). We've seen far too many overcomplicated kitchen sink CLI tools and would like to avoid building towards another one :)  

I'm not entirely convinced that `--allow-empty` is enough of a benefit over `/osv-scanner -r . || if [ $? -eq 128 ]; then true; else exit $?; fi`. In the organisation-wide required workflow case, this would only need to be done once, right?

C-Pro:
Org-wide workflow is just one example.
I think there'll be many use cases when people want to run scans on repositories without checking if they have dependencies or not.
I think this mode makes a lot of sense.
If you think otherwize, don't hesitate to close the PR. Workaround is not pretty, but it works 🤷

oliverchang:
Thanks for understanding @C-Pro. We'll close this for now, but will revisit if we see more users ask for this.

PR #198, opened 2023-02-08
govulncheck integration

another-rex:
An initial attempt at integrating govulncheck's library with OSV-Scanner.

TODOs before full PR:

~~- [ ] Add config options after design doc is finalized~~
~~- [ ] Allow user to configure tags and go versions~~
- [x] Update table output for inactive/not called vulnerabilities
- [x] Update exit code for not called vulnerabilities
- [x] Update README with feature and the go dependency if they want the vulncheck feature
- [x] Add go as dependency into osv-scanner docker containers.


cmd/osv-scanner/main.go
```diff
@@ -89,6 +89,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Usage:   "check subdirectories",
 				Value:   false,
 			},
+			&cli.BoolFlag{
+				Name:  "call-analysis",
```
oliverchang, line 93: Can we include `--experimental` in the flag name itself to make its experimental state more explicit?

internal/govulncheckshim/client.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, nil
+	}
+	response := convertToGvcOSV(*entry)
+
+	return &response, nil
+}
+
+func (ls *localSource) GetByAlias(ctx context.Context, alias string) ([]*gvcOSV.Entry, error) {
+	//nolint:prealloc // Need to be nil if none exists
+	var entries []*gvcOSV.Entry = nil
```
oliverchang, line 87: the default empty slice value is already nil right? we don't need to assign it here.

internal/govulncheckshim/client.go
```diff
@@ -0,0 +1,110 @@
...
+	vulnList         []models.Vulnerability
+	response         map[string]*models.Vulnerability
+	aliasToIDs       map[string][]*models.Vulnerability
+	moduleToIDs      map[string][]*models.Vulnerability
+	lastModifiedTime time.Time
+	client.Client
+}
+
+func newClient(vulns []models.Vulnerability) *localSource {
+	outLs := localSource{
```
oliverchang, line 23: nit: it wasn't obvious what `outLs` meant. Can we just call this variable `client` instead?

internal/govulncheckshim/client.go
```diff
@@ -0,0 +1,110 @@
...
+	"time"
+
+	"github.com/google/osv-scanner/pkg/models"
+	"golang.org/x/vuln/client"
+	gvcOSV "golang.org/x/vuln/osv"
+)
+
+type localSource struct {
+	vulnList         []models.Vulnerability
+	response         map[string]*models.Vulnerability
```
oliverchang, line 15: Is there a better name for this? Maybe `byID` or something similar?

internal/govulncheckshim/vulncheckshim.go
```diff
@@ -0,0 +1,77 @@
...
+	client := newClient(vulns)
+
+	cfg := &govulncheck.Config{Client: client}
+	var res *govulncheck.Result
+	var pkgs []*vulncheck.Package
+	pkgs, err := loadPackages([]string{scanPath}, filepath.Dir(scanPath))
+	if err != nil {
+		panic(err)
+	}
+	res, err = govulncheck.Source(context.Background(), cfg, pkgs)
```
oliverchang, line 41: nit: just return the result directly?

internal/govulncheckshim/client.go
another-rex: Good point, renamed all 3 to have better names:
vulnsByID      
vulnsByAlias   
vulnsByModule

internal/govulncheckshim/client_test.go
```diff
@@ -0,0 +1,244 @@
...
+				{
+					ID: "GO-2021-0098",
+				},
+			},
+		},
+		{
+			Alias: "GHSA-cx3w-xqmc-84g5",
+			ReturnValue: []struct{ ID string }{
+				// If the alias is the ID, do not return the entry
+				// TODO: Is this correct behavior?
```
another-rex, line 127: Is this correct behavior?

another-rex:
Still a draft at the moment, should be ready by end of today though.

internal/govulncheckshim/vulncheckshim.go
```diff
@@ -26,17 +27,18 @@ func (e *packageError) Error() string {
 }
 
 // RunVulnCheck runs govulncheck with a subset of vulnerabilities identified by osv-scanner
-func RunVulnCheck(path string, vulns []models.Vulnerability) (*vulncheck.Result, error) {
+func RunVulnCheck(path string, vulns []models.Vulnerability) (*govulncheck.Result, error) {
```
zpavlinovic, line 30: super nit: RunGovulncheck.

internal/govulncheckshim/client.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		if client.lastModifiedTime.Before(v.Modified) {
+			client.lastModifiedTime = v.Modified
+		}
+	}
+
+	return &client
+}
+
+func convertToGvcOSV(osv models.Vulnerability) gvcOSV.Entry {
```
zpavlinovic, line 48: I'd consider turning panics here into an error and propagate them through their callees. But this also depends on how you decide the scanner should communicate failures to users.

internal/govulncheckshim/client.go
```diff
@@ -0,0 +1,110 @@
...
+type localSource struct {
+	vulnList         []models.Vulnerability
+	vulnsByID        map[string]*models.Vulnerability
+	vulnsByAlias     map[string][]*models.Vulnerability
+	vulnsByModule    map[string][]*models.Vulnerability
+	lastModifiedTime time.Time
+	client.Client
+}
+
+func newClient(vulns []models.Vulnerability) *localSource {
```
zpavlinovic, line 22: I guess the precondition here is that all vulns apply to the program under consideration?

I mention this for the following reason.  A single `gvcOSV.Entry` can contain information on vulnerabilities in 1) [different modules/packages](https://github.com/golang/vulndb/blob/cf9ad2b3dc5b7ffca7e2fcff0de336bcd1e25e62/data/reports/GO-2020-0036.yaml) and/or 2) even different ranges within the same package. That is why each `gvcOSV.Affected` defines module, package, range, symbols, and platform info. govulncheck hence does a [fine grained filtering](https://github.com/golang/vuln/blob/8fb696828224a4da6b1492cd467a841f7230f0b5/vulncheck/vulncheck.go#L313) based on module info in the user program.

It could be that this is taken care of somewhere higher in the stack or that it does not matter for your current usage. I wanted to mention it as it is quite subtle.

another-rex: `vulns` should contain all the osv entries that affects any packages in the go.mod file of the program.

Since it's the OSV entry, it should contain the almost identical information to gvcOSV.Entry (e.g. GetByID/Alias/Module will return GO-2020-0036.yaml in it's entirety if it matches).

internal/govulncheckshim/client.go
another-rex: Updated to return an error instead of panicing, though I'm not sure what the callees or user can do if the two types differ, I believe this should be caught during development if the types do change to be incompatible. 

Still this  does make it clearer to the user and keeps the non analysis part running.

docs/working_docs/output.md
```diff
@@ -73,7 +73,15 @@ Sample output:
               "ids": [
                 "GHSA-c3h9-896r-86jm",
                 "GO-2021-0053"
-              ]
+              ],
+              // Call stack analysis is done using the `--experimental-call-analysis` flag
+              // and result is matched against data provided by the advisory to check if
+              // affected code is actually being executed.
+              "analysis": {
```
oliverchang, line 80: should we prefix this with "experimental-" for now?

docs/working_docs/usage.md
```diff
@@ -95,6 +95,17 @@ it should infer the parser based on the filename:
 osv-scanner --lockfile ':/path/to/my:projects/package-lock.json'
 ~~~~~~~~~~~~~~~~~~~
 
+### Scanning with call analysis (preview)
```
oliverchang, line 98: Should we add a line to our docs on SemVer adherence that we can break `--experimental` flags and `experimental-` fields in our JSON at any time?

internal/govulncheckshim/client_test.go
oliverchang: @zpavlinovic would you be able to comment on this?

pkg/osvscanner/vulnerability_result.go
```diff
@@ -47,8 +51,45 @@ func groupResponseBySource(r *output.Reporter, query osv.BatchedQuery, resp *osv
...
+				groupsByIDBySource[query.Source][id] = &groups[i]
+			}
+		}
+		pkg.Groups = groups
 		groupedBySource[query.Source] = append(groupedBySource[query.Source], pkg)
+		vulnBySource[query.Source] = append(vulnBySource[query.Source], pkg.Vulnerabilities...)
+	}
+
+	if callAnalysis {
+		for si, v := range vulnBySource {
```
oliverchang, line 71: Can we put this into its own function?

internal/govulncheckshim/vulncheckshim.go
```diff
@@ -0,0 +1,76 @@
...
+	fmt.Fprintln(&b, "Packages contain errors:")
+	for _, e := range e.Errors {
+		fmt.Fprintln(&b, e)
+	}
+
+	return b.String()
+}
+
+// RunGoVulnCheck runs govulncheck with a subset of vulnerabilities identified by osv-scanner
+func RunGoVulnCheck(path string, vulns []models.Vulnerability) (*govulncheck.Result, error) {
```
oliverchang, line 30: Are there any basic tests we can add for this?

pkg/osvscanner/vulnerability_result.go
```diff
@@ -47,8 +51,45 @@ func groupResponseBySource(r *output.Reporter, query osv.BatchedQuery, resp *osv
...
+			if si.Type == "lockfile" && filepath.Base(si.Path) == "go.mod" {
+				res, err := govulncheckshim.RunGoVulnCheck(filepath.Dir(si.Path), v)
+				if err != nil {
+					// TODO: Better error message to notify user that go is not installed
```
oliverchang, line 75: Can we at least just include "the Go toolchain is required to be installed" or something similar in the error message for now?

pkg/osvscanner/vulnerability_result.go
```diff
@@ -47,8 +51,45 @@ func groupResponseBySource(r *output.Reporter, query osv.BatchedQuery, resp *osv
...
+		groups := grouper.Group(grouper.ConvertVulnerabilityToIDAliases(pkg.Vulnerabilities))
+		// TODO(rex): Is there a better way to do this, this feels fragile
+		for i := range groups {
+			for _, id := range groups[i].IDs {
+				groupsByIDBySource[query.Source][id] = &groups[i]
+			}
+		}
+		pkg.Groups = groups
 		groupedBySource[query.Source] = append(groupedBySource[query.Source], pkg)
+		vulnBySource[query.Source] = append(vulnBySource[query.Source], pkg.Vulnerabilities...)
```
oliverchang, line 67: nit: s/vuln/vulns/ to make it clear there are multiple vulns in the value.

pkg/osvscanner/vulnerability_result.go
oliverchang: The code is also getting a little hard to follow here with many interdependencies. Can we maybe simplify this all by doing something like this:

(and reusing the existing loop for putting together all the final details): 
```go

// from lines 95
for source, packages := range groupedBySource {
  // Do final source analysis to annotate output with call graph analysis results. 
  doSourceAnalysis(source, packages)
  output.Results = ...
}

// Maybe these should go a separate file as well

type MatchedVulnerability struct {
  v *Vulnerability
  group *models.GroupInfo
}

func vulnsFromAllPkgs(pkgs) []MatchedVulnerability,  {
   // collect all vulnerabilities and the matching groups 
}

func doSourceAnalysis(source, pkgs) {
  allVulns := vulnsFromAllPkgs(pkgs)
  // Maybe change the type of the second arg for this as well to `MatchedVulnerability`, since this step always acts as a filter/post-processing step after we match vulns 
  res, err := govulncheckshim.RunGoVulnCheck(filepath.Dir(source.Path.Path), allVulns)
  ...
   
}

```

oliverchang (commented):
nice!! some comments

docs/working_docs/output.md
another-rex: yep, changed to experimentalAnalysis

docs/working_docs/usage.md
another-rex: Added a highlighted note here and in the installation.md part where we explain our semver adherence

internal/govulncheckshim/client_test.go
zpavlinovic: It would retrieve the OSV entry with ID GO-2021-0098.

However, govulncheck does not use that API of the client. It only uses GetByModule and LastModifiedTimestamp. This will be reflected in the simplified client we will roll out hopefully soon.

pkg/osvscanner/vulnerability_result.go
another-rex: Did a refactor of this and the logic is split up nicely now, though I don't quite get what MatchedVulnerabilities is for. PTAL

internal/sourceanalysis/sourceanalysis.go
```diff
@@ -0,0 +1,57 @@
...
+			idToGroupMap[vuln.ID] = &pv.Groups[groupIdx]
+			vulnList = append(vulnList, vuln)
+		}
+	}
+
+	return idToGroupMap, vulnList
+}
+
+// DoSourceAnalysis runs the language specific analyzers on the code given packages and source info
+func DoSourceAnalysis(r *output.Reporter, source models.SourceInfo, pkgs []models.PackageVulns) {
```
oliverchang, line 32: nit: rename this to just "Run" to avoid stutter since the package name is already "sourceanalysis".

internal/govulncheckshim/vulncheckshim.go
oliverchang: bumping this comment :)

pkg/osvscanner/vulnerability_result.go
oliverchang: Thanks! Looks a lot cleaner now :)

`MatchedVulnerability` was just an abstraction to avoid returning a separate map of vulns to groupinfos. It also seemed like it made logical sense to me because it represented this idea of a "matched" vulnerability with additional context -- context that we add during post-scanning analysis (i.e. grouping, source analysis)

There may also be an edge case that `MatchedVulnerability` would help with. A single vulnerability may affect multiple different packages. If there are multiple matches, then the current `map[string]*models.GroupInfo` keyed on vuln ID may not be enough to disambiguate those cases right?

oliverchang (commented):
looking good!! some final comments

internal/govulncheckshim/vulncheckshim.go
another-rex: Added a test for a small example project

pkg/osvscanner/vulnerability_result.go
another-rex: I'm not clear on how having `MatchedVulnerabiliity` will help with the edge case, as we still need to store/index by package information to identify which package the analysis result is for right?

oliverchang: The idea was that `MatchedVulnerability` can have duplicate vulns in there with different group mappings. 

The alternative here is for your map to have a slice of GroupInfos as values instead of a single one, which can also enable the fast lookup.

internal/sourceanalysis/sourceanalysis.go
```diff
@@ -0,0 +1,83 @@
...
+			r.PrintError(
+				fmt.Sprintf("Failed to run code analysis (govulncheck) on '%s' because %s\n"+
+					"(the Go toolchain is required)\n", source.Path, err.Error()))
+
+			return
+		}
+		// Add analysis information back into package list
+		for _, v := range res.Vulns {
+			for _, m := range v.Modules {
+				if idMatchedVulnMap[m.Path] == nil {
```
oliverchang, line 71: Looking into this a bit more. The `Vuln` struct is defined as:

```golang
type Vuln struct {
	// OSV contains all data from the OSV entry for this vulnerability.
	OSV *osv.Entry

	// Modules contains all of the modules in the OSV entry where a
	// vulnerable package is imported by the target source code or binary.
	//
	// For example, a module M with two packages M/p1 and M/p2, where only p1
	// is vulnerable, will appear in this list if and only if p1 is imported by
	// the target source code or binary.
	Modules []*Module
}
```

Assuming I'm understanding this correctly, let's take a project with lockfile that contains modules M1, M2, both of which are affected by OSV-1. Inside the project, let's say that we call vulnerable packages in M1, but only packages without vulnerabilities in M2. 

This means that govulncheck will give us:

```json
{
   "OSV": { OSV-1 },
   "Modules": [ M1 ],
}
```

Since M2 doesn't have any vulnerable packages imported. 

With the current code, this means that with our scan results, we find: OSV-1 affecting both M1 and M2, but we **don't** write `called: false` for M2 because that module is not in the `Modules` slice at all. This seems incorrect. 

I think instead we should be iterating over the modules, vulns that we have from our initial scan and cross referencing that with the results that govulncheck gives us to check the `IsCalled()`. We should also set `called` to `false` if a vulnerability was considered for source analysis but govulncheck did not return it or include the module (i.e. if it's not imported).

One approach that might be cleaner than all of this is something like:

```go
// Get rid of vulnsFromAllPkgs, MatchedVulnerability, and just iterate every single group we see in PackageVulns

// pre-process the govulncheck `res` to enable faster lookup. 

// slice naming confusing since a "package" in OSV-Schema actually refers to a Go module. 
for _, pkg :=  range pkgs {
  for _, groups := range pkg.Groups {
    for _, id := groups {
      // if the vulnerability for ID does not have source level info, continue
      // if the vulnerability for ID has source level info, check if it's in the listed vulns/modules returned by govulncheck. If not, set `called` to False. 
      // Set `called` to v.IsCalled()
     }
  }
}
```

another-rex: Rewrote source analysis according to the suggested approach, PTAL.

internal/sourceanalysis/sourceanalysis.go
```diff
@@ -0,0 +1,84 @@
...
+		for _, v := range res.Vulns {
+			gvcResByVulnID[v.OSV.ID] = v
+		}
+
+		for _, pv := range pkgs {
+			// Use index to keep reference to original element in slice
+			for groupIdx := range pv.Groups {
+				for _, vulnID := range pv.Groups[groupIdx].IDs {
+					gvcVuln, ok := gvcResByVulnID[vulnID]
+					if !ok { // If vulnerability not found, analysis has not been performed, so skip
```
oliverchang, line 57: If no modules have vulnerable pkgs imported, would this still work? I.e. if does govulncheck ever return OSV in cases where Modules is empty? 

If not, we'll have to check if the OSV has source level info.

internal/sourceanalysis/sourceanalysis.go
```diff
@@ -0,0 +1,84 @@
...
+						*analysis = make(map[string]models.AnalysisInfo)
+					}
+
+					if !containsModule {
+						// Code does not import module, so definitely not called
+						(*analysis)[vulnID] = models.AnalysisInfo{
+							Called: false,
+						}
+					} else {
+						// Codes does import module, check if it's called
```
oliverchang, line 75: Nit: singular code

oliverchang (commented):
Thanks!! Some hopefully final comments

internal/sourceanalysis/sourceanalysis.go
```diff
@@ -0,0 +1,84 @@
...
+	flatVulns := map[string]models.Vulnerability{}
+	for _, pv := range pkgs {
+		for _, vuln := range pv.Vulnerabilities {
+			flatVulns[vuln.ID] = vuln
+		}
+	}
+
+	vulnList := []models.Vulnerability{}
+	for _, v := range flatVulns {
+		vulnList = append(vulnList, v)
```
oliverchang, line 26: nit: just `vulns` as the variable name.

internal/sourceanalysis/sourceanalysis.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return vulnList
+}
+
+// Run runs the language specific analyzers on the code given packages and source info
+func Run(r *output.Reporter, source models.SourceInfo, pkgs []models.PackageVulns) {
+	vulnsSlice := vulnsFromAllPkgs(pkgs)
```
oliverchang, line 34: nit: let's just name this `vulns` (to be consistent with the above function too).

internal/sourceanalysis/sourceanalysis.go
another-rex: If it's not imported at all, it does not show up (it's treated as if no analysis is done at all, which is incorrect), however, this is also very niche, since if someone runs `go mod tidy` it removes those false positive entries from the `go.mod` file anyway.

oliverchang: Hmm, there's an important distinction here though right -- the project can still import from these modules (and hence not be removed by `go mod tidy`), just not any "vulnerable" ones.

Per 

```
	// For example, a module M with two packages M/p1 and M/p2, where only p1
	// is vulnerable, will appear in this list if and only if p1 is imported by
	// the target source code or binary.
```

It seems like this would be incorrect in such a case?

another-rex: Ah yes, in that case you are right, go mod tidy will keep it, and osv-scanner will show a false positive.

In this case we'll have to do some check of whether the vulnerability is a GO-20xx-xxxx based vuln, and set called to false if it never showed up in the list returned by govulncheck?

oliverchang: Yeah, but rather than keying just on the ID, a better approach would be checking if the ecosystem_specific contains Go source information.

another-rex: Ok, should be resolved, there is still some edge cases where multiple version ranges that's being affected are saved in different packages (possibly because the source level information is different between different versions).

This would require the semantic package and checking each affected entry is inside the version range or not. Leaving that as future work to implement when offline mode is implemented, as I believe there'll be a lot of overlap in the version range comparison code.

oliverchang (approved):
LGTM with one last comment!

oliverchang:
@julieqiu FYI

internal/testutility/utility.go
```diff
@@ -0,0 +1,63 @@
...
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/kr/pretty"
+)
+
+// LoadHelper loads a JSON fixture file and returns the decoded version.
+func LoadHelper[V any](t *testing.T, path string) V {
```
oliverchang, line 14: nit: `Helper` isn't too helpful of a name :) 

Maybe `LoadJSONFixture` instead ?

internal/testutility/utility.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		t.Fatalf("Failed to unmarshal val: %s", err)
+	}
+
+	if !reflect.DeepEqual(val, elem) {
+		t.Errorf("Not equal: \n%s", strings.Join(pretty.Diff(val, elem), "\n"))
+	}
+}
+
+// AssertMatchFixtureJSON creates a JSON file at path of the value val,
```
oliverchang, line 48: nit: fix the function name in the comment.

internal/testutility/utility.go
```diff
@@ -0,0 +1,63 @@
...
+// can be used with AssertMatchFixtureJSON to compare against future values.
+func CreateJSONSnapshot[V any](t *testing.T, path string, val V) {
```
oliverchang, line 50: Should we be consistent with naming and replace "Snapshot" with "Fixture"?

oliverchang (approved):
LGTM with some nits

Issue #199, opened 2023-02-08
Move output package from /internal to /pkg


dburriss:
**Please let me know if I am missing something here. I am very new to Go lang**.

I suggest changing the `output` package to a public package (move to /pkg folder).

The main reason for this change suggestion is that the public [func DoScan(actions ScannerActions, r *output.Reporter) (models.VulnerabilityResults, error)](https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/osvscanner.go#L363) function in the `osvscanner` package exposes an internal type in its public API. Namely, [output.Reporter](https://github.com/google/osv-scanner/blob/main/internal/output/reporter.go#L11) from the internal `output` package.

Another reason for this change is that the `ouput` package relies on public packages ie. the `models` package.

```mermaid
flowchart LR

    subgraph pkg
    config
    osvscanner-->models
    osv-->models
    osv-->lockfile
    end

    subgraph internal
    sbom
    semantic
    output
    end

    osvscanner-->output
    output-->models
    output-->osv
    config-->output
    semantic-->lockfile
    osvscanner-->sbom
```

Making the `output` package part of public packages by moving it to /pkg makes the public API usable by making it depend on only public types.

```mermaid
flowchart LR

    subgraph internal
    sbom
    semantic
    end

    subgraph pkg
    config
    osvscanner-->models
    osv-->models
    osv-->lockfile
    osvscanner-->output
    output
    output-->models
    output-->osv
    end


    osvscanner-->sbom
    config-->output
    semantic-->lockfile
    
```

Due to the reliance of `semantic.Parse` and `semantic.MustParse` on `filelock.Ecosystem` type, there still will exist a 2-way dependency between pkg and internal packages. This could be solved by moving the 2 above mentioned functions into the `lockfile` package. A possible destination is the parse.go file. Then parse.go could instead import the `semantic` package, with a few changes (making the semver parsing public internal).

```mermaid
flowchart TB

    subgraph pkg
    config
    osvscanner-->models
    osv-->models
    osv-->lockfile
    osvscanner-->output
    output
    output-->models
    output-->osv
    end

    subgraph internal
    sbom
    semantic
    end

    config-->output
    lockfile-->semantic
    osvscanner-->sbom
```

I will take a look at what this would look like in a Draft PR if anyone is interested in the changes? I made the 1st change already because I needed it but happy to PR back if something thinks this reasoning makes sense.

dburriss:
[...]

PR #200, opened 2023-02-08
Cleanup dependencies between pkg and internal packages

dburriss:
This addresses issue #199 .

The following tasks would need to be completed:
- [ ] Move output package from internal to pkg folder
- [ ] Invert dependency semantic -> lockfile
- [ ] Test


another-rex:
Closed by #341

PR #203, opened 2023-02-10
Moving Docs to a GitHub Page

hayleycd:
Based on conversation in issue #201 I have set up a GitHub page for OSV-Scanner's documentation. 

Notes:

- You can see the rendered documentation [here](hayleycd.github.io/osv-scanner). 
- Docs are divided into current and working. All docs are present on the website, but [working docs](https://hayleycd.github.io/osv-scanner/working/) do not have their urls in the nav. 
- This is largely a structural change to the docs, not a content change to the docs. Changes were made to READMEs and other files, but generally this change is structural.

Issue #201, opened 2023-02-08
Move Documentation to Github Page


hayleycd:
Hello!

This issue is to discuss moving the documentation out of the readme and into a github page. 

There are a few issues I want to iron out before opening a PR. 

1). What branch do we want to use to build the documentation? We can use "last-stable" or create a doc specific branch. Ideally, this branch will include up to date information, but not any information on features that are not yet available. We can also just use main, but I'm trying to avoid problems like the one that cropped up in #196. I'm not sure what the easiest thing to do from a maintenance standpoint and would like your thoughts. 

2). I am assuming the github page should be https://google.github.io/osv-scanner please let me know if a different url is preferred. 

3). Upon merge, I will need someone with access to the project settings to set up the github page. 

4). Are there any other questions or concerns that we can work out? 

For reference, [this](https://github.com/hayleycd/osv-scanner/tree/doc-structure/docs) is the doc folder of the branch I am working on, and you can see how the docs currently render [here](https://hayleycd.github.io/osv-scanner/)

another-rex:
> 1). What branch do we want to use to build the documentation? ...

I think instead of having a last-stable branch that is fixed to the release commit, we can update/cherry pick that branch with documentation updates as necessary until the next release. 

@oliverchang How is this done with oss-fuzz?

> 2). I am assuming the github page should be https://google.github.io/osv-scanner please let me know if a different url is preferred.

That url looks good to me.

oliverchang:
> > 1). What branch do we want to use to build the documentation? ...
> 
> I think instead of having a last-stable branch that is fixed to the release commit, we can update/cherry pick that branch with documentation updates as necessary until the next release.
> 
> @oliverchang How is this done with oss-fuzz?

OSS-Fuzz works by running at HEAD, so that wasn't a consideration there.

`last-stable` as a place that doesn't necessariy follow `main` but as somewhere that we can cherry-pick to SGTM. 

One thing I see other projects do is also provide an option to view the "main" documentation as well through a dropdown. I think this will be useful to both users and ourselves (to validate the documentation we're releasing looks good). 

> 
> > 2). I am assuming the github page should be https://google.github.io/osv-scanner please let me know if a different url is preferred.
> 
> That url looks good to me.

+1.

hayleycd:
> One thing I see other projects do is also provide an option to view the "main" documentation as well through a dropdown. I think this will be useful to both users and ourselves (to validate the documentation we're releasing looks good).

I'll look into that.

[...]

PR #206, opened 2023-02-12
Improve gitignore parsing

### LINKED ISSUE #202: Gitignore is not respected for files outside working directory that are not in a git repository
Running OSV-Scanner on a file/directory that is not in a git repository (or is not a git repository itself) currently parses gitignore files starting from the current working directory where OSV-Scanner was invoked.

This means that if a OSV-Scanner is run on a non-git file that is not under the current working directory (e.g. `osv-scanner -r ../foo/`), any gitignore files under the target will not be parsed.

This only affects non-git files/directories. If the target file/directory is part of a git repository then OSV-Scanner will parse gitignore files starting from the root of the file's containing git repository, as intended.

michaelkedar:
I had thought of some minor improvements for the gitignore parsing I had implemented:
- Parse gitignores for non-git files from their respective directories, instead of the current working directory.
  - Should fix #202, and I think also a potential issue on Windows scanning across drive letters.
- Fix some error messages not having newlines.
- Add a error message warning if the root file is ignored, instead of skipping silently.
- Add some notes in USAGE.md about gitignore behaviour.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -30,6 +30,13 @@ The recursive flag `-r` or `--recursive` will tell the scanner to search all sub
 
 Git directories are searched for the latest commit hash. Searching for git commit hash is intended to work with projects that use git submodules or a similar mechanism where dependencies are checked out as real git repositories. 
 
+### Ignored files
+
+By default, OSV-Scanner will not scan files that are ignored by `.gitignore` files. If the specified file is not part of a git repository, `.gitignore` files are parsed recursively starting from the target directory,
+otherwise they are parsed from the root of the git repository as typical.
+
+The `--no-ignore` flag can be used force the scanner to scan ignored files.
+
 ### Specify SBOM
 
 If you want to check for known vulnerabilities only in dependencies in your SBOM, you can use the following command:
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -16,6 +16,7 @@ import (
 	"github.com/google/osv-scanner/pkg/models"
 	"github.com/google/osv-scanner/pkg/osv"
 
+	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing/format/gitignore"
@@ -52,7 +53,7 @@ func scanDir(r *output.Reporter, query *osv.BatchedQuery, dir string, skipGit bo
 		var err error
 		ignoreMatcher, err = parseGitIgnores(dir)
 		if err != nil {
-			r.PrintError(fmt.Sprintf("Unable to parse git ignores: %v", err))
+			r.PrintError(fmt.Sprintf("Unable to parse git ignores: %v\n", err))
 			useGitIgnore = false
 		}
 	}
@@ -74,9 +75,12 @@ func scanDir(r *output.Reporter, query *osv.BatchedQuery, dir string, skipGit bo
 		if useGitIgnore {
 			match, err := ignoreMatcher.match(path, info.IsDir())
 			if err != nil {
-				r.PrintText(fmt.Sprintf("Failed to resolve gitignore for %s: %v", path, err))
+				r.PrintText(fmt.Sprintf("Failed to resolve gitignore for %s: %v\n", path, err))
 				// Don't skip if we can't parse now - potentially noisy for directories with lots of items
 			} else if match {
+				if root { // Don't silently skip if the argument file was ignored.
+					r.PrintError(fmt.Sprintf("%s was not scanned because it is excluded by a .gitignore file. Use --no-ignore to scan it.\n", path))
+				}
 				if info.IsDir() {
 					return filepath.SkipDir
 				}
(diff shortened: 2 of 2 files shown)


pkg/osvscanner/osvscanner.go
```diff
@@ -124,9 +128,23 @@ type gitIgnoreMatcher struct {
 
 func parseGitIgnores(dir string) (*gitIgnoreMatcher, error) {
 	// We need to parse .gitignore files from the root of the git repo to correctly identify ignored files
-	// Defaults to current directory if dir is not in a repo or some other error
-	// TODO: Won't parse ignores if dir is not in a git repo, and is not under the current directory (e.g ../path/to)
-	fs := osfs.New(".")
+	var fs billy.Filesystem
+
+	// Default to dir (or directory containing dir if it's a file) is not in a repo or some other error
+	f, err := os.Open(dir)
```
another-rex, line 134: I think you can combine the Open and Stat into `os.Stat`

another-rex (approved):
Thanks!, Looks good just one comment. Also the USAGE.md changes will clash with the documentation moving that is being done in #201, so if that get's merged in first the docs will have to be moved.

pkg/osvscanner/osvscanner.go
michaelkedar: Oh, I didn't see that function. Changed.

michaelkedar:
Re: #201 / #203
I don't mind waiting until that's merged and fixing it here. Or we can merge this now and fix the usage there.
Up to you.

another-rex:
That docs PR might take a bit longer, let's just merge this now.

USAGE.md
```diff
@@ -30,6 +30,13 @@ The recursive flag `-r` or `--recursive` will tell the scanner to search all sub
 
 Git directories are searched for the latest commit hash. Searching for git commit hash is intended to work with projects that use git submodules or a similar mechanism where dependencies are checked out as real git repositories. 
 
+### Ignored files
+
+By default, OSV-Scanner will not scan files that are ignored by `.gitignore` files. If the specified file is not part of a git repository, `.gitignore` files are parsed recursively starting from the target directory,
```
oliverchang, line 35: "recursively" seems a bit vague -- can you clarify what this means? 

That said -- does the behaviour actually make sense? If a file isn't part of a git repo, no gitignores should apply to it right?

USAGE.md
```diff
@@ -30,6 +30,13 @@ The recursive flag `-r` or `--recursive` will tell the scanner to search all sub
...
+otherwise they are parsed from the root of the git repository as typical.
+
+The `--no-ignore` flag can be used force the scanner to scan ignored files.
```
oliverchang, line 38: "can be used **to** force the scannner"

pkg/osvscanner/osvscanner.go
```diff
@@ -124,9 +128,19 @@ type gitIgnoreMatcher struct {
 
 func parseGitIgnores(dir string) (*gitIgnoreMatcher, error) {
 	// We need to parse .gitignore files from the root of the git repo to correctly identify ignored files
-	// Defaults to current directory if dir is not in a repo or some other error
-	// TODO: Won't parse ignores if dir is not in a git repo, and is not under the current directory (e.g ../path/to)
-	fs := osfs.New(".")
+	var fs billy.Filesystem
+
+	// Default to dir (or directory containing dir if it's a file) is not in a repo or some other error
+	finfo, err := os.Stat(dir)
```
oliverchang, line 134: Can we rename `dir` to `path` to avoid the confusing variable name?

USAGE.md
oliverchang: CC @another-rex

michaelkedar: "recursively" meaning it parses `foo/.gitignore`, `foo/bar/.gitignore`, `foo/bar/baz/.gitignore`, etc.
I guess it'd be more clear to say "`.gitignore` files are parsed in the target directory and all subdirectories"?

I think this behaviour is the probably the most useful:
- I don't think you'd typically have `.gitignore` files in directories that aren't git repositories.
   - If you do, I would assume that you'd want to use it.
- Say you have a folder `foo/`, that's not a repo, and inside have `foo/bar/` and `foo/baz/` that are git repos. Running `osv-scanner -r foo/` should pick up the gitignore files from `bar/` and `baz/`.

USAGE.md
michaelkedar: :upside_down_face: I guess I'll make another PR to fix this

USAGE.md
oliverchang: The way this is worded was rather unclear. It also doesn't clearly distinguish "specified file" from "target directory", and it wasn't clear what "not part of a git repository" means exactly. 

Would this alternative wording work:  

"All recursively scanned files are matched to a git repository (if it exists) and any corresponding .gitignore files are taken into account."

So just to make sure my own understanding is correct, using your example:  

Given `foo/` (not a git repo), `foo/bar` (git repo), and `foo/baz` (git repo):
- Scanning `foo/bar/blah` would use `foo/bar/.gitignore` if it exists
- Scanning `foo/baz/blah` would use `foo/baz/.gitignore` if it exists
- Scanning `foo/blah` would **also** use `foo/.gitignore` if it exists, even though it's not a git repo? 

There are variations of edge cases here. e.g. 
- If `foo/` is not a git repo AND has a gitignore,, and `foo/bar` is a git repo does not have a git ignore, does scannign something in `foo/bar` use the gitignore from `foo/` ?

michaelkedar: > So just to make sure my own understanding is correct, using your example:

These are all correct, as it currently works.

> Would this alternative wording work:
>
>"All recursively scanned files are matched to a git repository (if it exists) and any matching .gitignore files within that repository are taken into account."

I suppose, though I don't think it captures the (current) non-repository behaviour (e.g. dot point 3)

oliverchang: > > So just to make sure my own understanding is correct, using your example:
> 
> These are all correct, as it currently works.
> 
> > Would this alternative wording work:
> > "All recursively scanned files are matched to a git repository (if it exists) and any matching .gitignore files within that repository are taken into account."
> 
> I suppose, though I don't think it captures the (current) non-repository behaviour (e.g. dot point 3)

There's a bunch of edge cases (e.g. the last one I listed in my last comment) that would be resolved if we more closely followed git behaviour (which should take into account repo boundaries). Is that easy to support?

michaelkedar: Tracking repo boundaries correctly is a bit tricky. I've made the bug #209 for this.

USAGE.md
michaelkedar: Fixed in #208

pkg/osvscanner/osvscanner.go
michaelkedar: Fixed in #208

PR #211, opened 2023-02-15
Documentation Style Improvements

hayleycd:
Based on feed back since the launch of the new docs website, this PR includes the following changes:

1. Docs site now uses the OSV logo. 
2. Docs site now uses an OSV favicon. 
3. We had a number of ways we were representing console outputs. Some were displaying better than others. I have updated these instances to be more uniform, confirmed they display in a darker text, and utilize kramdown styling (the flavor of markdown the project uses).

I am up for further feedback though, and I encourage you to check out the current look [here](https://hayleycd.github.io/osv-scanner) and let me know what you think.


oliverchang (approved):
thanks!

another-rex:
Thanks! I disagree with the changes for number 3 though, as it removes a lot of highlighting that was actually desired.

The main problem if I remember is the console styling being inconsistent I believe? I think that can be fixed either by prefixing commands with a `$` sign:
~~~
```console
$ osv-scanner --sbom=/path/to/your/sbom.json
```
~~~
or by changing the language to bash
~~~
```bash
osv-scanner --sbom=/path/to/your/sbom.json
```
~~~
Both achieve the same darker text while retaining highlighting, though I slightly prefer the bash option since it makes lines easier to copy.

oliverchang:
> Thanks! I disagree with the changes for number 3 though, as it removes a lot of highlighting that was actually desired.
> 
> The main problem if I remember is the console styling being inconsistent I believe? I think that can be fixed either by prefixing commands with a `$` sign:
> 
> ```
> ```console
> $ osv-scanner --sbom=/path/to/your/sbom.json
> ```
> ```
> 
> or by changing the language to bash
> 
> ```
> ```bash
> osv-scanner --sbom=/path/to/your/sbom.json
> ```
> ```
> 
> Both achieve the same darker text while retaining highlighting, though I slightly prefer the bash option since it makes lines easier to copy.

@another-rex that's a good point. Can you put up a PR to change this?

hayleycd:
Thanks @another-rex. This is the kind of feedback I was soliciting. 

For the record, I did try the bash option. It added some interesting coloring to the text that I did not expect, and in my opinion the coloring did not add clarity. 

I am okay with it if the team likes it though. I would like to avoid the '$' option for copying purposes.

PR #212, opened 2023-02-16
Reinstated Bash Formatting

hayleycd:
Hi Team, 

Based on post-merge feedback on #211 , styling has been updated to use bash formatting in most places, and json and toml where appropriate. 

Some of the text colors do not have good contrast. I will continue working on that as time permits, with the hope to roll it out to other projects too but this can be merged now. 

You can see what it looks like [here](https://hayleycd.github.io/osv-scanner/)


another-rex (approved):
LGTM, thanks for the changes!

PR #221, opened 2023-02-17
feat: support `dependencyManagement` in Maven poms

### LINKED ISSUE #214: parse-maven-lock does not scan inside dependencyManagement tags
Dependencies can be also defined inside `dependencyManagement>dependencies` element when there is a set of projects that inherit from a common parent: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

The parser only scans inside `dependencies` tags and thus can miss packages in pom.xml files.

G-Rath:
Resolves #214


another-rex (approved):
Thanks! LGTM

oliverchang (approved):

Issue #218, opened 2023-02-17
Improve Report Output Documentation


Dentrax:
https://google.github.io/osv-scanner/output

* To make it more simple and human-readable ([like how trivy did](https://aquasecurity.github.io/trivy/v0.27.1/docs/vulnerability/examples/report/))
* Sync the doc with current [output reporters](https://github.com/google/osv-scanner/tree/main/internal/output)
* Use `<spoiler>` tag for outputs, and `3-backticks` for one-liner commands

PR #226, opened 2023-02-20
Update the output docs, make logo a lot bigger, make page slightly wider

another-rex:
Update the output docs, resolving #218. Added markdown output, put the output examples into collapsible details.

Also make logo a lot bigger, and main page slightly wider.

You can see the updated changes here: https://another-rex.github.io/osv-scanner/working/output


oliverchang (approved):

docs/working_docs/output.md
```diff
@@ -26,12 +31,52 @@ Sample output:
...
-Sample output:
+### JSON
+
+```bash
+osv-scanner --format json your/project/dir
+```
+
+Outputs the results as a JSON object to stdout, with all other output being directed to stderr - this makes it safe to redirect the output to a file with
+```bash
+osv-scanner --format json ... > /path/to/file.json
```
hayleycd, line 75: I don't like the three dots here, it isn't clear enough. 

I assume the full command is:

osv-scanner --format json your/project/dir >/path/to/file.json

But it feels like I am guessing.

docs/working_docs/output.md
```diff
@@ -10,13 +10,18 @@ permalink: /working/output
...
+### Table (Default)
 
 The default format, which outputs the results as a human-readable table.
 
-Sample output:
+```bash
+osv-scanner --format table your/project/dir
+```
+
+<details markdown="1">
```
hayleycd, line 23: I like using the dropdown! Thanks!

docs/working_docs/output.md
another-rex: Good point, updated to have a full command

hayleycd (approved):

PR #229, opened 2023-02-20
ParseAs for dpkg-status

cmaritan:
As previously done with apk `installed` lockfiles, I propose to add the parseAs option also for dpkg `status` files since they also have a very generic filename:

```bash
osv-scanner --lockfile 'dpkg-status:/var/lib/dpkg/status'
```


another-rex (approved):

G-Rath (changes requested):
You should add a CLI test like the one for `apk-installed`, to confirm it works

cmaritan:
> You should add a CLI test like the one for `apk-installed`, to confirm it works

Done!

G-Rath (approved):

PR #231, opened 2023-02-22
Upgrade to yaml v3

another-rex:
Update to yaml.v3, all tests still pass. Closes #118


oliverchang (approved):

G-Rath (approved):
fwiw I have not done this upgrade on the detector because it just increased the binary size without any immediate benefit - idk if you're worried about that, but maybe something to think about

another-rex:
Good point, for now it's not a metric we are measuring, but it is definately something to consider. Tested the two osv-scanner builds on linux and it seems to differ by about ~200kb, which I think is alright.

PR #250, opened 2023-02-28
fix: support yarn locks with quoted properties

### LINKED ISSUE #249: Errors parsing yarn.lock files with quoted properties
[Scorecard](https://github.com/ossf/scorecard) currently calls DoScan [here](https://github.com/ossf/scorecard/blob/35a7dd5b254443aa6a12cdc9679458be4b07d7de/clients/osv.go#L42) when checking for vulns. When we scan public repos, we sometimes encounter yarn.lock files that write hundred of lines to os.Stderr due to the issue discussed in #102 and #142.

Here are a few of the offending files that we've encountered:
https://github.com/Glissir/react-trello/blob/565dd41fd2263787412b6e131c9f21b26c7204cd/yarn.lock
https://github.com/nodeWechat/wechat4u/blob/93972863435ba0cfe98c316bbe6b739da390897d/yarn.lock

#102 mentioned being unable to reproduced with `yarn`, but a few search results have mentioned the issue is caused when `npm` manipulates the `yarn.lock` file:

https://github.com/npm/cli/issues/5126

https://stackoverflow.com/q/74272832
https://bstefanski.com/upgrading-one-dependency-causes-whole-yarnlock-change-fix

G-Rath:
Resolves #249

This files are actually valid, and generated by npm v7+


oliverchang (approved):
thanks!

another-rex (approved):

PR #251, opened 2023-03-01
Parse nested CycloneDX components correctly

### LINKED ISSUE #232: No output when scanning hierachical CycloneDX SBOM
When scanning a hierachical CycloneDX JSON file generated with `cyclonedx-cli merge --hierachical [...]` (https://github.com/CycloneDX/cyclonedx-cli) osv-scanner produces no output, although components are present (and reported when scanning the files before merge).

michaelkedar:
Should fix #232

### DIFF
--- a/internal/sbom/cyclonedx.go
+++ b/internal/sbom/cyclonedx.go
@@ -22,18 +22,30 @@ func (c *CycloneDX) Name() string {
 }
 
 func (c *CycloneDX) enumeratePackages(bom *cyclonedx.BOM, callback func(Identifier) error) error {
-	for _, component := range *bom.Components {
-		if component.PackageURL != "" {
-			err := callback(Identifier{
-				PURL: component.PackageURL,
-			})
-			if err != nil {
-				return err
+	// Components can have components, so enumerate them recursively.
+	var enumerateComponents func([]cyclonedx.Component) error
+	enumerateComponents = func(components []cyclonedx.Component) error {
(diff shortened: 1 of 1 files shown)


internal/sbom/cyclonedx.go
```diff
@@ -22,18 +22,30 @@ func (c *CycloneDX) Name() string {
...
-	for _, component := range *bom.Components {
-		if component.PackageURL != "" {
-			err := callback(Identifier{
-				PURL: component.PackageURL,
-			})
-			if err != nil {
-				return err
+	// Components can have components, so enumerate them recursively.
+	var enumerateComponents func([]cyclonedx.Component) error
+	enumerateComponents = func(components []cyclonedx.Component) error {
```
oliverchang, line 27: nit: can we just pull this out to it's own top level method (with `c *CycloneDX`)? 

(we probably could have also done `enumerateComponents := ` instead with the explicit `var`, but a top level function is still better)

oliverchang (commented):
thanks!!

internal/sbom/cyclonedx.go
michaelkedar: I've pulled this out, though now `enumeratePackages` looks a bit redundant.

(I couldn't do `enumerateComponents :=` because it needs to be defined for the recursive call)

oliverchang (approved):
thanks!

Issue #254, opened 2023-03-01
Add RPM/Red Hat ecosystem support


cmaritan:
As briefly discussed [here](https://github.com/google/osv-scanner/pull/164#issuecomment-1405729223), `osv.dev` still has not support for RPM ecosystem but maybe it's worth start discussing it.

Managing rpmdb is different from other lockfiles already supported until now by `osv-scanner` because they are not text files but instead:

- BerkeleyDB files for rpm before v4.16
- SQLite file for rpm v4.16 and later

Support using current osv-scanner API is straightforward and also the `io.Reader` support scenario should have no problems (see latest update of draft #164 ).

G-Rath:
Think this should be supported, just want to add a test case to the `scan` fixtures

krazylearner:
Hey please allow me to work on this issue.

PR #255, opened 2023-03-01
Rpmdb support

cmaritan:
Address #254


another-rex (commented):
Thanks! Looks like a reasonable implementation to me, my main concern is pulling in the relatively large number of dependencies dependencies to parse the sqlite database., especially before any support for redhat advisories are included in osv.dev .

pkg/lockfile/rpmdb.go
```diff
@@ -0,0 +1,49 @@
...
+			continue
+		}
+
+		var pkg = PackageDetails{
+			Ecosystem: RedHatEcosystem,
+			CompareAs: RedHatEcosystem,
+		}
+		pkg.Name = rpmPkg.Name
+		pkg.Version = rpmPkg.Version
+		packages = append(packages, pkg)
```
G-Rath, line 45: We might as well just do all together:

```suggestion
	packages = append(packages, PackageDetails{
		Name: rpmPkg.Name,
		Version: rpmPkg.Version,
		Ecosystem: RedHatEcosystem,
		CompareAs: RedHatEcosystem,
	})
```

pkg/lockfile/rpmdb_test.go
```diff
@@ -0,0 +1,64 @@
...
+		{
+			Name:      "htop",
+			Version:   "3.2.1",
+			Ecosystem: lockfile.RedHatEcosystem,
+			CompareAs: lockfile.RedHatEcosystem,
+		},
+	})
+}
+
+func TestParseRpmDB_SQLite_FileDoesNotExist(t *testing.T) {
```
G-Rath, line 48: nit: I typically had these kind of checks at the top as they're usually pretty short and test the first lines of the parser - while by itself not a big deal, I think we should still move these to be consistent with the existing tests

G-Rath (commented):
This looks cool! It would be good if you could have a look over the other parsers and consider if there's any situations they account for that should be handled for this parser/"lockfile" too, such as the same packages being listed multiple times

(I'm not super familiar with RedHat stuff, so it might all be being handled by the library)

cmaritan:
> Thanks! Looks like a reasonable implementation to me, my main concern is pulling in the relatively large number of dependencies dependencies to parse the sqlite database., especially before any support for redhat advisories are included in osv.dev .

Thanks for the feedback, I hadn't considered this point.

If it's a problem, we can surely keep aside this PR and merge it when osv.dev will have redhat advisories support.

cmaritan:
> This looks cool! It would be good if you could have a look over the other parsers and consider if there's any situations they account for that should be handled for this parser/"lockfile" too, such as the same packages being listed multiple times
> 
> (I'm not super familiar with RedHat stuff, so it might all be being handled by the library)

Usually rpmdb are written only by librpm, passing duplicates/invalid values should not be possible. 
Generating this test case would require direct insert of records inside sqlite db but maintainers keep its structure [undocumented](https://github.com/rpm-software-management/rpm/discussions/2211#discussioncomment-3793589) so it's possible to break other things.

Spiritaine (approved):

andrewpollock:
Rocky Linux (currently present in OSV.dev) uses RPM and I'm not sure about Mageia, but it's in progress... It might be worthwhile getting this PR merged?

picatz:
👋 Is there any status update on this PR? Would love to see this feature land.

another-rex:
@picatz Still in progress and is being actively worked on, rough estimate is around the end of October, there should be some significant progress by then.

picatz:
Awesome, I really appreciate the quick update @another-rex!

jessesmd:
Please also consider the epoch related to rpm packages as they have a specific format epoch:name-version-release.arch
Thank you!

PR #258, opened 2023-03-02
More specific cyclone dx parsing

### LINKED ISSUE #257: Performance issues from scanSBOMFile reading every file multiple times
Scorecard utilizes `osvscanner.DoScan` when performing its `Vulnerabilities` check. The time to complete the check is more than an order of magnitude higher than other checks. Running `pprof` shows a hot spot in `scanSBOMFile`.

It looks like when walking a directory, every file is potentially parsed as an SBOM:
https://github.com/google/osv-scanner/blob/6eab4504663395b4f5983ebe4570e8b816844ab5/pkg/osvscanner/osvscanner.go#L114-L117

There are currently two providers, SPDX and CycloneDX. While SPDX checks for a filename, there's no such check for CycloneDX:
https://github.com/google/osv-scanner/blob/6eab4504663395b4f5983ebe4570e8b816844ab5/pkg/osvscanner/osvscanner.go#L231-L238

I believe this means attempting to parse every file as a CycloneDX SBOM, twice. In large repositories, this adds up ( longest observed is 5 minutes) :
https://github.com/google/osv-scanner/blob/6eab4504663395b4f5983ebe4570e8b816844ab5/internal/sbom/cyclonedx.go#L50-L59

8sg3cyNvvMYG5HH

another-rex:
Make sure the file name follows the recognized file names here: https://cyclonedx.org/specification/overview/#recognized-file-patterns

Resolves #257 

Also 
- adds tests for sboms,
- makes SBOM scan logging output consistent with lockfile scan logging output
- Minor refactor of SBOMs

I believe we will also want something similar to parse-as in lockfiles for SBOMs as well in the future to allow file names that doesn't conform to the standard to be scanned.

### DIFF
--- a/cmd/osv-scanner/fixtures/locks-many/alpine.cdx.xml
+++ b/cmd/osv-scanner/fixtures/locks-many/alpine.cdx.xml
@@ -1,0 +1,604 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<bom xmlns="http://cyclonedx.org/schema/bom/1.4" serialNumber="urn:uuid:4243b783-229c-48e7-982b-febe2bb5bc2b" version="1">
+  <metadata>
+    <timestamp>2023-03-02T12:04:22+11:00</timestamp>
+    <tools>
+      <tool>
+        <vendor>anchore</vendor>
+        <name>syft</name>
+        <version>0.73.0</version>
+      </tool>
+    </tools>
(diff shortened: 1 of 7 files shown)


oliverchang (approved):
thanks!

G-Rath (approved):

PR #260, opened 2023-03-02
feat: support `-r` flag in `requirements.txt` files

### LINKED ISSUE #252: Python requirements : manage  nested requirements
Hello,

It could be nice to manage nested requirements, for example : 
base.txt
```
Django==4.1.7
```
production.txt:
```
-r ./base.txt
sentry-sdk==1.15.0
```
Then I would call 
```
osv-scanner --config=/Users/pmind/Projects/osv-scanner.toml --lockfile=requirements.txt:requirements/production.txt
```
And osv-scanner would read production + base as it would be only one file.

G-Rath:
Resolves #252


pkg/lockfile/fixtures/pip/with-multiple-r-options.txt
```diff
@@ -0,0 +1,5 @@
+-r ./one-package-constrained.txt
+-r ./multiple-packages-mixed.txt
+
+# this should take precedents as it comes after
```
another-rex, line 4: I don't think this comment is correct, even if it goes before pip will still install 1.2.3.
Pip doesn't allow conflicting requirements, and if not given exact versions, will try to find the version that satisfies  all requirements if it can, order does not seem to matter.

another-rex (dismissed):
Thanks!, just one comment.

pkg/lockfile/fixtures/pip/with-multiple-r-options.txt
G-Rath: hmm ok so then in this case Pip would error? I guess the easiest path forward would be to just treat them as separate packages for now, and then this can be improved in future if needed (especially since this is the existing behaviour regardless of if `-r` is present)

another-rex: In this case pip will install `1.2.3` because `requests` in `./multiple-packages-mixed.txt` doesn't have a version constraint.

Yeah I think treating them as separate packages is probably fine for now.

G-Rath: yeah sorry you're right, I was thinking of the original test case I wrote which was using `django` (which is contrained in `one-package-constrained.txt`)

another-rex (approved):

Issue #273, opened 2023-03-06
Deprecate "working" docs concept


hayleycd:
The current system of "working" and "current" docs is not useful in practice. 

Revert back to one version and build docs using a stable branch.

hayleycd:
Completed

PR #275, opened 2023-03-08
Remove "working doc" concept

hayleycd:
Please preview changes here: https://hayleycd.github.io/osv-scanner/

Based on discussion with @oliverchang I have removed the working doc concept. We will be building docs from a stable branch. 

There wasn't a ton of difference between the working docs and current docs at this point, and I tried to combine them and use the preview label. Mostly govulncheck related.

Addresses #273


oliverchang:
Thanks! Before this goes out, we also need to update the stable branch as well right?

oliverchang:
> @oliverchang
> 
> I'm not sure what you are getting at. I believe we are currently building the docs from the main branch, and we will want to specify another one to build from.

Right, so before this one gets merged, we need to update the stable branch and point docs to it right? Otherwise the docs will be built from this one which is no accurate for the last stable release.

hayleycd:
> > @oliverchang
> > I'm not sure what you are getting at. I believe we are currently building the docs from the main branch, and we will want to specify another one to build from.
> 
> Right, so before this one gets merged, we need to update the stable branch and point docs to it right? Otherwise the docs will be built from this one which is no accurate for the last stable release.

Yes! I think we are on the same page. Add the current state of docs to the stable branch and make sure the github page is building from it instead of main.

another-rex (approved):
LGTM!

PR #279, opened 2023-03-09
fix: trim leading zeros off when comparing numerical components in Maven versions

G-Rath:
This was found in the latest version of the generated maven versions fixture - I'm pretty sure this also is present in the API version too as it's also missing kind of function, but I've not tested that for sure.

I've not updated the generated fixture because I'm going to be following this up with a new generator workflow that will involve that, but this needs to be landed first otherwise _that_ PR will fail CI 😅

### DIFF
--- a/internal/semantic/fixtures/maven-versions.txt
+++ b/internal/semantic/fixtures/maven-versions.txt
@@ -33,6 +33,9 @@
 1 = 1-0
 1 = 1.0-0
 1.0 = 1.0-0
+
+1.1 = 1.01
+
 // no separator between number and character
 1a = 1-a
 1a = 1.0-a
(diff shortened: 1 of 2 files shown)


another-rex (approved):

oliverchang:
@michaelkedar can you please check if fixes are needed in OSV.dev for this?

michaelkedar:
> @michaelkedar can you please check if fixes are needed in OSV.dev for this?

OSV.dev does trim leading zeros for these, and there's already a test case for it :+1:

G-Rath:
oohh ok it's because this is actually the wrong fix 😅

G-Rath:
@michaelkedar I would recommend adding the test case I added in this PR to the OSV.dev test suite too, just to be on the safe side since this & #285 showcase that it is possible to have a buggy implementation pass every other test case successfully.

PR #285, opened 2023-03-10
fix: trim leading zeros off when comparing numerical components in Maven versions (better)

G-Rath:
While the fix I landed in https://github.com/google/osv-scanner/pull/279 was a fix, it wasn't the _best_ fix - really the bug was I'd used the wrong variable - @michaelkedar comment about [these tests](https://github.com/google/osv-scanner/pull/279#pullrequestreview-1334066731) catching this is the key (they pass because the whole string can be parsed as a number, meaning the incorrect condition is passing)


another-rex (approved):
Huh, good catch!

PR #292, opened 2023-03-13
Prevent `.gitignore` files from interfering with tests

### LINKED ISSUE #286: Alpine packaging: Tests break if other .gitignore files are present in parent directories
The tests for the [1.2.0 upgrade](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/45010) are currently failing in Alpine: https://gitlab.alpinelinux.org/ptrcnull/aports/-/jobs/989373

The errors look like this:

```
--- FAIL: TestRun (0.00s)
    --- FAIL: TestRun/#03 (1.78s)
        main_test.go:273: stderr
             got:
            /builds/ptrcnull/aports/community/osv-scanner/src/osv-scanner-1.2.0/cmd/osv-scanner/fixtures/locks-many/not-a-lockfile.toml was not scanned because it is excluded by a .gitignore file. Use --no-ignore to scan it.
            
             want:
    --- FAIL: TestRun/#10 (1.78s)
        main_test.go:273: cli exited with code 128, not 0
        main_test.go:273: stdout
             got:
            {
              "results": null
            }
            
             want:
            {
              "results": []
            }
        main_test.go:273: stderr
             got:
            Scanning dir ./fixtures/locks-many/composer.lock
            /builds/ptrcnull/aports/community/osv-scanner/src/osv-scanner-1.2.0/cmd/osv-scanner/fixtures/locks-many/composer.lock was not scanned because it is excluded by a .gitignore file. Use --no-ignore to scan it.
            
             want:
            Scanning dir ./fixtures/locks-many/composer.lock
            Scanned %%/fixtures/locks-many/composer.lock file and found 1 packages
```

I think this is because osv-scanner recognizes that the build path `/builds/ptrcnull/aports/community/osv-scanner/src/osv-scanner-1.2.0/` is inside of a git repository `/builds/ptrcnull/aports/` and then picks up a [`.gitignore`](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/.gitignore) file from outside the buildpath at `/builds/ptrcnull/aports/.gitignore`. This file excludes `/builds/ptrcnull/aports/community/osv-scanner/src/`, which causes osv-scanner tests to dismiss all test files, causing the tests to fail.

Not sure what's the best way to solve this, but figured I might share my analysis.

michaelkedar:
Should fix #286.
The issue was that the tests were expecting osv-scanner to be a git repository, and the downloading the GitHub release source files does not include the `.git` directory.

Fixed by initializing the `fixtures` directory as a git repository when running the tests, which both prevents git from walking past the folder, and parses the `.gitignore` files for the related test.

### DIFF
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -4,10 +4,13 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
+
+	"github.com/go-git/go-git/v5"
 )
 
 func dedent(t *testing.T, str string) string {
@@ -488,4 +491,15 @@ func TestRun_LockfileWithExplicitParseAs(t *testing.T) {
 			testCli(t, tt)
 		})
 	}
+}
+
+func TestMain(m *testing.M) {
+	// Temporarily make the fixtures folder a git repository to prevent gitignore files messing with tests.
+	_, err := git.PlainInit("./fixtures", false)
+	if err != nil {
+		panic(err)
+	}
+	code := m.Run()
+	os.RemoveAll("./fixtures/.git")
+	os.Exit(code)
 }


another-rex (approved):
Thanks!

G-Rath (commented):
huh TIL `TestMain` is a thing - would it make sense to have it at the top, as it's effectively the first thing that is run right and probably doesn't make sense to have intermingled with other tests

michaelkedar:
> would it make sense to have it at the top, as it's effectively the first thing that is run right and probably doesn't make sense to have intermingled with other tests

I personally like having it always at the end of the file, like how I'd put a `main` function. Some examples I've see online tend to also have it at the end.

Agree that it shouldn't be intermingled with other tests.

oliverchang (approved):

PR #294, opened 2023-03-14
fix: avoid infinite loops parsing Maven poms with syntax errors

### LINKED ISSUE #293: ParseMavenLock hangs when parsing some Maven archetypes files
Both of these small `pom.xml` files seem to cause `osv-scanner` to hang. One invocation ran for 5+ hours before the program was terminated for other reasons.

Reproducible with:
```
osv-scanner -L ./pom.xml
```

https://github.com/Open-Wide/owsi-nuxeo-boilerplate/blob/96a65826d8f610d17b269d0da1d577bbb5d46cb1/archetype/src/main/resources/archetype-resources/pom.xml

https://github.com/BeOne-PL/promena/blob/adbc25d2cd3acf1990f7188938fee25d834aa0db/sdk/promena-transformer/promena-transformer-archetype/src/main/resources/archetype-resources/configuration/pom.xml

spencerschrock:
Resolves #293

### DIFF
--- a/pkg/lockfile/fixtures/maven/invalid-syntax.xml
+++ b/pkg/lockfile/fixtures/maven/invalid-syntax.xml
@@ -1,0 +1,13 @@
+<project>
+  <properties>
+    <${Id}.version>${project.version}</${Id}.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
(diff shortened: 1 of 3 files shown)


G-Rath (approved):
That's what I get for not checking error values 🤦

oliverchang:
Thanks @spencerschrock!!

another-rex (approved):

PR #295, opened 2023-03-14
Set version in source code

another-rex:
Set the release version in the source code. 

An alternative to setting the version directly in source code is to get package repositories to build their osv-scanner with goreleaser instead of go directly. While goreleaser is often already in most repositories, it is not as straightforward:

- Package repositories sometimes have their own guidelines for how to build go, e.g. Arch Linux go binaries must be built with hardened options which we currently do not do with goreleaser.
- goreleaser also expects to be in a git repository (to get version tags), but most build systems prefers downloading the source code in a tarball instead of checking out the repository.

### DIFF
--- a/cmd/osv-scanner/main.go
+++ b/cmd/osv-scanner/main.go
@@ -13,7 +13,8 @@ import (
 )
 
 var (
-	version = "dev"
+	// Update this variable when doing a release
+	version = "1.2.0"
 	commit  = "n/a"
 	date    = "n/a"
 )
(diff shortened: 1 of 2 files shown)


cmd/osv-scanner/main.go
```diff
@@ -13,7 +13,8 @@ import (
 )
 
 var (
-	version = "dev"
+	// Update this variable when doing a release
+	version = "1.2.0"
```
oliverchang, line 17: Should this have a "-dev" suffix by default in non-release build contexts?

another-rex: Thought about that a bit, I feel like it probably should not, since it seems like all linux package repos will build their own version, the version of osv-scanner people get from them will then also have the -dev suffix, which I don't think is ideal.

The user can already tell if it's built by us directly or not by looking at whether "built on" and "commit" is filled out or not.

oliverchang: SGTM, that makes sense.

oliverchang (approved):

Issue #281, opened 2023-03-09
Improve `IgnoredVulns` ergonomics


ericcornelissen:
This report concerns the `IgnoredVulns` [configuration](https://google.github.io/osv-scanner/configuration/) option, in particular the way it handles vulnerabilities known by multiple identifiers. At least for Rust(/Cargo), `osv-scanner` reports both `RUSTSEC` and `GHSA` advisories but ignoring by either one of these for a given vulnerabilities does not ignore the other. 

## Example

I have a Rust project that `osv-scanner` currently finds two vulnerabilities for:

```sh
$ osv-scanner .
Scanning dir .
Scanning .../rust-project/ at commit 1234567890abcdef
Scanned .../rust-project/Cargo.lock file and found 159 packages
╭─────────────────────────────────────┬───────────┬────────────────┬─────────┬────────────╮
│ OSV URL (ID IN BOLD)                │ ECOSYSTEM │ PACKAGE        │ VERSION │ SOURCE     │
├─────────────────────────────────────┼───────────┼────────────────┼─────────┼────────────┤
│ https://osv.dev/GHSA-mc8h-8q98-g5hr │ crates.io │ remove_dir_all │ 0.5.3   │ Cargo.lock │
│ https://osv.dev/RUSTSEC-2023-0018   │           │                │         │            │
│ https://osv.dev/GHSA-wcg3-cvx6-7396 │ crates.io │ time           │ 0.1.45  │ Cargo.lock │
│ https://osv.dev/RUSTSEC-2020-0071   │           │                │         │            │
╰─────────────────────────────────────┴───────────┴────────────────┴─────────┴────────────╯
```

Where the `GHSA` and `RUSTSEC` URLs/IDs for the same package are in fact for the same vulnerability (the CLI output suggests this, and the [osv.dev website](https://osv.dev/list) refers to them as _Aliases_). However, if I want to ignore these by, say, the `RUSTSEC` identifiers using:

```toml
# osv-scanner.toml

[[IgnoredVulns]]
id = "RUSTSEC-2023-0018"
reason = "TODO"

[[IgnoredVulns]]
id = "RUSTSEC-2020-0071"
reason = "TODO"
```

then running `osv-scanner` still outputs the `GHSA` URLs/IDs (and exits with a non-zero exit code):

```sh
$ osv-scanner .
Scanning dir .
Scanning .../rust-project/ at commit 1234567890abcdef
Scanned .../rust-project/Cargo.lock file and found 159 packages
[...]

PR #300, opened 2023-03-15
Make `IgnoredVulns` also ignore aliases

michaelkedar:
#281 - Adds `includeAliases` flag to configuration toml to set wheter also ignore vulnerability aliases, which defaults to true.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -29,9 +29,14 @@ type Config struct {
 }
 
 type IgnoreEntry struct {
-	ID          string    `toml:"id"`
-	IgnoreUntil time.Time `toml:"ignoreUntil"`
-	Reason      string    `toml:"reason"`
+	ID             string    `toml:"id"`
+	IgnoreUntil    time.Time `toml:"ignoreUntil"`
+	Reason         string    `toml:"reason"`
+	IncludeAliases *bool     `toml:"includeAliases"` // Use pointer to tell apart unset (treated as true) and false
+}
+
+func (e *IgnoreEntry) ShouldIncludeAliases() bool {
+	return e.IncludeAliases == nil || *(e.IncludeAliases)
 }
 
 func (c *Config) ShouldIgnore(vulnID string) (bool, IgnoreEntry) {
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -346,29 +346,52 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
 	return nil
 }
 
-// Filters response according to config, returns number of responses removed
-func filterResponse(r *output.Reporter, query osv.BatchedQuery, resp *osv.BatchedResponse, configManager *config.ConfigManager) int {
+// filters vulnerabilities and groups according to config, returns filtered vulnerabilities and groups
+func filterVulns(r *output.Reporter, configToUse config.Config, vulns []models.Vulnerability, groups []models.GroupInfo) ([]models.Vulnerability, []models.GroupInfo) {
 	hiddenVulns := map[string]config.IgnoreEntry{}
+	outGroups := []models.GroupInfo{}
+	outVulns := []models.Vulnerability{}
 
-	for i, result := range resp.Results {
-		var filteredVulns []osv.MinimalVulnerability
-		configToUse := configManager.Get(r, query.Queries[i].Source.Path)
-		for _, vuln := range result.Vulns {
-			ignore, ignoreLine := configToUse.ShouldIgnore(vuln.ID)
+	for _, group := range groups {
+		keepIDs := []string{}
+		for i, id := range group.IDs {
+			ignore, ignoreLine := configToUse.ShouldIgnore(id)
 			if ignore {
-				hiddenVulns[vuln.ID] = ignoreLine
+				hiddenVulns[id] = ignoreLine
+				if ignoreLine.ShouldIncludeAliases() {
+					for _, id := range keepIDs {
+						hiddenVulns[id] = ignoreLine
+					}
+					for _, id := range group.IDs[i+1:] {
+						hiddenVulns[id] = ignoreLine
+					}
+					keepIDs = []string{}
+					break
+				}
 			} else {
-				filteredVulns = append(filteredVulns, vuln)
+				keepIDs = append(keepIDs, id)
 			}
 		}
-		resp.Results[i].Vulns = filteredVulns
+		if len(keepIDs) > 0 {
+			group.IDs = keepIDs
+			outGroups = append(outGroups, group)
+		}
 	}
 
-	for id, ignoreLine := range hiddenVulns {
-		r.PrintText(fmt.Sprintf("%s has been filtered out because: %s\n", id, ignoreLine.Reason))
+	for _, vuln := range vulns {
+		ignoreLine, ok := hiddenVulns[vuln.ID]
+		if ok {
+			if ignoreLine.ID == vuln.ID {
+				r.PrintText(fmt.Sprintf("%s has been filtered out because: %s\n", vuln.ID, ignoreLine.Reason))
+			} else {
+				r.PrintText(fmt.Sprintf("%s (alias of %s) has been filtered out because: %s\n", vuln.ID, ignoreLine.ID, ignoreLine.Reason))
+			}
+		} else {
+			outVulns = append(outVulns, vuln)
+		}
 	}
 
-	return len(hiddenVulns)
+	return outVulns, outGroups
 }
 
 func parseLockfilePath(lockfileElem string) (string, string) {
@@ -456,16 +479,16 @@ func DoScan(actions ScannerActions, r *output.Reporter) (models.VulnerabilityRes
 		return models.VulnerabilityResults{}, fmt.Errorf("scan failed %w", err)
 	}
 
-	filtered := filterResponse(r, query, resp, &configManager)
-	if filtered > 0 {
-		r.PrintText(fmt.Sprintf("Filtered %d vulnerabilities from output\n", filtered))
-	}
 	hydratedResp, err := osv.Hydrate(resp)
 	if err != nil {
 		return models.VulnerabilityResults{}, fmt.Errorf("failed to hydrate OSV response: %w", err)
 	}
 
-	vulnerabilityResults := groupResponseBySource(r, query, hydratedResp, actions.ExperimentalCallAnalysis)
+	filter := func(sourcePath string, vulns []models.Vulnerability, groups []models.GroupInfo) ([]models.Vulnerability, []models.GroupInfo) {
+		return filterVulns(r, configManager.Get(r, sourcePath), vulns, groups)
+	}
+
+	vulnerabilityResults := groupResponseBySource(r, query, hydratedResp, actions.ExperimentalCallAnalysis, filter)
 	// if vulnerability exists it should return error
 	if len(vulnerabilityResults.Results) > 0 {
 		// If any vulnerabilities are called, then we return VulnerabilitiesFoundErr
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -346,7 +346,7 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
 	return nil
 }
 
-// filters vulnerabilities and groups according to config, returns filtered vulnerabilities and groups
+// Filters vulnerabilities and groups according to config, returns filtered vulnerabilities and groups
 func filterVulns(r *output.Reporter, configToUse config.Config, vulns []models.Vulnerability, groups []models.GroupInfo) ([]models.Vulnerability, []models.GroupInfo) {
 	hiddenVulns := map[string]config.IgnoreEntry{}
 	outGroups := []models.GroupInfo{}
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -366,6 +366,7 @@ func filterVulns(r *output.Reporter, configToUse config.Config, vulns []models.V
 						hiddenVulns[id] = ignoreLine
 					}
 					keepIDs = []string{}
+
 					break
 				}
 			} else {
--- a/pkg/osvscanner/vulnerability_result.go
+++ b/pkg/osvscanner/vulnerability_result.go
@@ -11,17 +11,26 @@ import (
 	"github.com/google/osv-scanner/pkg/osv"
 )
 
+// Function used to filter vulnerabilities and groups.
+type vulnFilterFn func(sourcePath string, vulns []models.Vulnerability, groups []models.GroupInfo) ([]models.Vulnerability, []models.GroupInfo)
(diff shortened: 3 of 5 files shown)


michaelkedar:
I'm a bit iffy on the style & organisation of the code. Would appreciate feedback/suggestions.

pkg/osvscanner/vulnerability_result_internal_test.go
michaelkedar, line 1: There aren't any test cases actually defined in this file.

pkg/config/config.go
```diff
@@ -29,9 +29,14 @@ type Config struct {
 }
 
 type IgnoreEntry struct {
-	ID          string    `toml:"id"`
-	IgnoreUntil time.Time `toml:"ignoreUntil"`
-	Reason      string    `toml:"reason"`
+	ID             string    `toml:"id"`
+	IgnoreUntil    time.Time `toml:"ignoreUntil"`
+	Reason         string    `toml:"reason"`
+	IncludeAliases *bool     `toml:"includeAliases"` // Use pointer to tell apart unset (treated as true) and false
```
oliverchang, line 35: If we default this to true, we should just start with not including this flag at all. It's easier to add a flag later than to remove it. This will likely also simplify the implementation a lot. 

It's hard to come up with a reasonable and unconfusing use case where users will want to set this to false. 

@another-rex WDYT?

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,53 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+				hiddenVulns[id] = ignoreLine
+				if ignoreLine.ShouldIncludeAliases() {
+					for _, id := range keepIDs {
+						hiddenVulns[id] = ignoreLine
+					}
+					for _, id := range group.IDs[i+1:] {
+						hiddenVulns[id] = ignoreLine
+					}
+					keepIDs = []string{}
+
```
another-rex, line 380: I think this can just be simplified to
```suggestion
					for _, id := range group.IDs {
						hiddenVulns[id] = ignoreLine
					}
					keepIDs = []string{}
```

pkg/osvscanner/vulnerability_result_internal_test.go
another-rex: I think I gave up trying to figure out how to put these huge structures into a test at some point, but didn't remove the file. 

I'm still not sure how to ergonomically add tests that are actually useful here, maybe we need to break down the functions more and test smaller parts.

pkg/osvscanner/vulnerability_result.go
```diff
@@ -11,17 +11,26 @@ import (
...
 
+	totalFiltered := 0
+
 	for i, query := range query.Queries {
-		response := resp.Results[i]
-		if len(response.Vulns) == 0 {
+		vulns := resp.Results[i].Vulns
+		vulnCount := len(vulns)
+		groups := grouper.Group(grouper.ConvertVulnerabilityToIDAliases(vulns))
+		vulns, groups = filter(query.Source.Path, vulns, groups)
```
oliverchang, line 31: Instead of passing a filter here, which can make the logic a bit hard to follow, can we just call a separate `filterIgnores` that takes in the entire `models.VulnerabilityResults` and modifies it as required?

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,53 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+	outGroups := []models.GroupInfo{}
+	outVulns := []models.Vulnerability{}
 
-	for i, result := range resp.Results {
-		var filteredVulns []osv.MinimalVulnerability
-		configToUse := configManager.Get(r, query.Queries[i].Source.Path)
-		for _, vuln := range result.Vulns {
-			ignore, ignoreLine := configToUse.ShouldIgnore(vuln.ID)
+	for _, group := range groups {
+		keepIDs := []string{}
```
oliverchang, line 367: We should document the algorithm in comments here a bit more. IT can be a little hard to follow otherwise.

That said,  we can get rid of this complicated `keepIDs` logic if we just default to ignoring all aliases (without supporting the flag to flip it) right?

pkg/config/config.go
another-rex, line 35: Maybe change the name to ExcludeAliases, so defaulting to false is automatic, and makes it clearer that default is to include. (Not 100% sure about this, IncludeAliases also makes sense to me)

another-rex (commented):
Nice, mostly looks good to me, added some initial comments.

pkg/config/config.go
another-rex: Removing it for now sounds good to me.

michaelkedar: Happy to remove this.

pkg/osvscanner/vulnerability_result.go
michaelkedar: I've done a first pass at filtering the entire `models.VulnerabilityResults` - it's pretty deeply nested so I'll probably want to extract a few functions from it.

pkg/osvscanner/osvscanner.go
michaelkedar: > That said, we can get rid of this complicated keepIDs logic if we just default to ignoring all aliases (without supporting the flag to flip it) right?

Yeah, the logic is simpler without it

pkg/osvscanner/osvscanner.go
michaelkedar: Simplified a bit when taking out `includeAliases`

pkg/config/config.go
michaelkedar: Removed the flag entirely

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,68 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
-				filteredVulns = append(filteredVulns, vuln)
+// Filters results according to config, preserving order. Returns total number of vulnerabilities removed.
+func filterResults(r *output.Reporter, results *models.VulnerabilityResults, configManager *config.ConfigManager) int {
+	removedCount := 0
+	newResults := []models.PackageSource{}
+	for _, pkgSrc := range results.Results {
+		configToUse := configManager.Get(r, pkgSrc.Source.Path)
+		newPackages := []models.PackageVulns{}
+		for _, pkgVuln := range pkgSrc.Packages {
+			newVuln := filterPackageVulns(r, pkgVuln, configToUse)
```
another-rex, line 368: nit: Use plural `vulns` to make it clear this is multiple vulns.

pkg/osvscanner/vulnerability_result_internal_test.go
michaelkedar: It's not in scope for this PR, but I think we should either add at least one test case, or remove this file so it's clear that this isn't actually being tested.

I feel like a test would be useful to make sure the output structure/order doesn't change, but that'd still require defining the whole `models.VulnerabilityResults` structure somewhere.

pkg/osvscanner/vulnerability_result.go
michaelkedar: Inner function has been extracted.

pkg/osvscanner/osvscanner.go
michaelkedar: > We should document the algorithm in comments here a bit more. IT can be a little hard to follow otherwise.

Added some comments into the changed filter function.

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,68 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+			removedCount += len(pkgVuln.Vulnerabilities) - len(newVuln.Vulnerabilities)
+			if len(newVuln.Vulnerabilities) > 0 {
+				newPackages = append(newPackages, pkgVuln)
```
another-rex, line 371: Should this be newVuln instead of pkgVuln?

michaelkedar: Yes - good catch!

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,68 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+	for _, group := range pkgVulns.Groups {
+		ignore := false
+		for _, id := range group.IDs {
+			var ignoreLine config.IgnoreEntry
+			if ignore, ignoreLine = configToUse.ShouldIgnore(id); ignore {
+				for _, id := range group.IDs {
+					hiddenVulns[id] = true
+				}
+				// NB: This only prints the first reason encountered in all the aliases.
+				r.PrintText(fmt.Sprintf("%s has been filtered out because: %s\n", ignoreLine.ID, ignoreLine.Reason))
```
michaelkedar, line 398: Wondering if this should be something like
```
VULN-XXX and N aliases have been filtered out because: [reason]
```

another-rex: Looks like a good format to me.

oliverchang (commented):
nice! Can we also write some tests for this?

pkg/osvscanner/osvscanner_internal_test.go
```diff
@@ -0,0 +1,73 @@
...
+			r := output.NewVoidReporter()
+			// ConfigManager looks for osv-scanner.toml in the source path.
+			// Sources in the test input should point to files/folders in the text fixture folder for this to work correctly.
+			configManager := config.ConfigManager{
+				DefaultConfig: config.Config{},
+				ConfigMap:     make(map[string]config.Config),
+			}
+			got := tt.testCase.input
+			filtered := filterResults(r, &got, &configManager)
+			if !reflect.DeepEqual(got, tt.testCase.want) {
```
oliverchang, line 63: nit: can we use https://pkg.go.dev/github.com/google/go-cmp@v0.5.9/cmp#Diff instead? 

see example usage:

https://github.com/google/osv-scanner/blob/2cf2b20c3b25613e4ca0dbd4d803748ec7887f4a/pkg/grouper/grouper_test.go#L137

pkg/osvscanner/osvscanner_internal_test.go
```diff
@@ -0,0 +1,73 @@
...
+		testCase.path = path
+
+		return testCase
+	}
+	tests := []struct {
+		name     string
+		testCase testCase
+	}{
+		{
+			name:     "",
```
oliverchang, line 38: Can you add some detail to the names here to describe the cases you're testing?

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,75 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+		}
+	}
+	results.Results = newResults
+
+	return removedCount
+}
+
+// Filters package-grouped vulnerabilities according to config, preserving ordering. Returns filtered package vulnerabilities.
+func filterPackageVulns(r *output.Reporter, pkgVulns models.PackageVulns, configToUse config.Config) models.PackageVulns {
+	hiddenVulns := map[string]bool{}
```
oliverchang, line 386: nit: for consistency, can we name this `ignoredVulns` ?

oliverchang, line 386: nit: we don't actually need this to have a bool value. We can just use `map[string]struct{}` which is slightly more efficient (0 byte storage) and a common pattern.

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,75 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+			}
+		}
+		if !ignore {
+			newGroups = append(newGroups, group)
+		}
 	}
 
-	for id, ignoreLine := range hiddenVulns {
-		r.PrintText(fmt.Sprintf("%s has been filtered out because: %s\n", id, ignoreLine.Reason))
+	newVulns := []models.Vulnerability{}
```
oliverchang, line 415: nit: prefer

```golang
var newVulns []models.Vulnerability
```

Per https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,75 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+		newPackages := []models.PackageVulns{}
+		for _, pkgVulns := range pkgSrc.Packages {
+			newVulns := filterPackageVulns(r, pkgVulns, configToUse)
+			removedCount += len(pkgVulns.Vulnerabilities) - len(newVulns.Vulnerabilities)
+			if len(newVulns.Vulnerabilities) > 0 {
+				newPackages = append(newPackages, newVulns)
 			}
 		}
-		resp.Results[i].Vulns = filteredVulns
+		if len(newPackages) > 0 {
```
oliverchang, line 374: Add a comment here explaining why: the default behaviour is we don't want to include the pkgSrc at all if there are no vulns.

pkg/osvscanner/osvscanner.go
```diff
@@ -357,29 +357,75 @@ func scanDebianDocker(r *output.Reporter, query *osv.BatchedQuery, dockerImageNa
...
+func filterResults(r *output.Reporter, results *models.VulnerabilityResults, configManager *config.ConfigManager) int {
+	removedCount := 0
+	newResults := []models.PackageSource{}
+	for _, pkgSrc := range results.Results {
+		configToUse := configManager.Get(r, pkgSrc.Source.Path)
+		newPackages := []models.PackageVulns{}
+		for _, pkgVulns := range pkgSrc.Packages {
+			newVulns := filterPackageVulns(r, pkgVulns, configToUse)
+			removedCount += len(pkgVulns.Vulnerabilities) - len(newVulns.Vulnerabilities)
+			if len(newVulns.Vulnerabilities) > 0 {
```
oliverchang, line 370: nit: add a comment explaining that the  default behaviour is we don't include the package at all if there are no vulns.

oliverchang (commented):
nice!!! some minor comments.

pkg/osvscanner/fixtures/filter/all/configs/a/osv-scanner.toml
```diff
@@ -0,0 +1,15 @@
+[[IgnoredVulns]]
+id = "GHSA-mc8h-8q98-g5hr"
+reason = "Ignore 1"
+
+[[IgnoredVulns]]
```
oliverchang, line 5: nit: to make these tests easier to follow, can you please annotate these with comments to explain which ones are aliases of each other? 

Same comment for all other test osv-scanner.toml files.

oliverchang (approved):

pkg/osvscanner/fixtures/filter/all/configs/a/osv-scanner.toml
michaelkedar: Added some annotations to all toml files

oliverchang (approved):

Issue #303, opened 2023-03-16
Add a "verbose dry-run" mode to display what data is exchanged with OSV.dev


andrewpollock:
https://github.com/google/osv.dev/discussions/1131 raised some questions rooted in confidentiality/privacy concerns around what information is shared with (and subsequently logged by) OSV.dev infrastructure.

I thought it would be helpful for privacy-conscious users of OSV Scanner to be able to reason about what data is being exchanged with OSV.dev if they could run OSV Scanner in a "verbose dry-run" where the API calls were displayed without sending anything to OSV.dev.

junnhwan:
Hi! I'd like to work on this issue.

I'll implement a verbose dry-run mode that shows what data would be exchanged with OSV.dev (request payloads / package queries) without performing the actual network calls, building on the existing offline/dry-run related paths.

Please assign this to me if that works. Thanks!

junnhwan:
**Proposed approach** (building on discussion from #698 / #746):

Add a `--dry-run` flag that:

1. Still extracts packages from lockfiles/SBOMs/dirs as usual
2. Builds the deduplicated `POST /v1/querybatch` payload that would be sent to OSV.dev
3. Prints the request method, URL, and JSON body (via cmdlogger / stderr)
4. Does **not** perform any OSV.dev network calls (no querybatch, no `/v1/vulns/{id}`, no determineversion)
5. Returns empty vulnerability results (exit 0) so users can inspect payloads safely

Notes:
- License checks and deps.dev transitive resolution are out of scope for this flag (OSV.dev only, per the issue); `--licenses` will be skipped with a warning in dry-run
- Incompatible with `--offline-vulnerabilities` (offline already avoids the API)
- CLA signed

Happy to adjust based on maintainer feedback before or during review.

Rajkaran-122:
Hi @andrewpollock, I’d like to work on this issue and wanted to confirm the proposed approach before implementing/opening a PR.

I’ve looked through the current OSV-Scanner architecture and the previous attempts around #698 and #746. My understanding is that the preferred direction is to integrate this with the existing verbosity/CLI design rather than introducing a separate experimental mechanism.

My proposed approach is:

Add a --dry-run option to the scan commands.
Use an http.RoundTripper wrapper around the existing HTTP client to intercept requests before they are sent.
In dry-run mode, display the HTTP method, URL, relevant headers, and request body (formatted when JSON).
Redact sensitive headers such as Authorization and Cookie.
[...]

PR #304, opened 2023-03-17
Adds concurrency to hydration requests

another-rex:
An attempt at concurrency for hydration requests. I'm not sure if this is the most concise way to do so, it seems pretty long for what should be a simple concurrency logic.

### DIFF
--- a/pkg/osv/osv.go
+++ b/pkg/osv/osv.go
@@ -187,13 +187,6 @@ func Get(id string) (*models.Vulnerability, error) {
 	return &vuln, nil
 }
 
-type hydrateResponse struct {
-	batchIdx  int
-	resultIdx int
-	vuln      *models.Vulnerability
-	err       error
-}
-
 // Hydrate fills the results of the batched response with the full
 // Vulnerability details.
 func Hydrate(resp *BatchedResponse) (*HydratedBatchedResponse, error) {
--- a/pkg/osv/osv.go
+++ b/pkg/osv/osv.go
@@ -23,7 +23,8 @@ const (
 	BaseVulnerabilityURL = "https://osv.dev/"
 	// maxQueriesPerRequest splits up querybatch into multiple requests if
 	// number of queries exceed this number
-	maxQueriesPerRequest = 1000
+	maxQueriesPerRequest  = 1000
+	maxConcurrentRequests = 25
 )
 
 // Package represents a package identifier for OSV.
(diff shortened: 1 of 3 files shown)


pkg/osv/osv.go
```diff
@@ -185,23 +187,74 @@ func Get(id string) (*models.Vulnerability, error) {
...
 
+	responseChan := make(chan hydrateResponse, maxConcurrentRequests)
+	rateLimiter := make(chan struct{}, maxConcurrentRequests)
+
+	// Run in separate goroutine so that we can immediately start
+	// listening to the responseChan, and it will be not blocked
+	go func(resp *BatchedResponse) {
+		for batchIdx, response := range resp.Results {
+			for resultIdx, vuln := range response.Vulns {
+				rateLimiter <- struct{}{}
```
oliverchang, line 210: nit: can we use https://pkg.go.dev/golang.org/x/sync/semaphore#example-package-WorkerPool instead?

pkg/osv/osv.go
```diff
@@ -185,23 +187,74 @@ func Get(id string) (*models.Vulnerability, error) {
...
+				go func(id string, batchIdx int, resultIdx int) {
+					vuln, err := Get(id)
+					if err != nil {
+						responseChan <- hydrateResponse{
```
oliverchang, line 214: I don't think we need this channel. 

Can we just pre-allocate the hydrated structure (i.e. all indexes are accessible) and have each "worker" write to the specified index directly? 

I believe this is safe, per the example in https://pkg.go.dev/golang.org/x/sync/semaphore#example-package-WorkerPool:

```golang
	// Compute the output using up to maxWorkers goroutines at a time.
	for i := range out {
		// When maxWorkers goroutines are in flight, Acquire blocks until one of the
		// workers finishes.
		if err := sem.Acquire(ctx, 1); err != nil {
			log.Printf("Failed to acquire semaphore: %v", err)
			break
		}

		go func(i int) {
			defer sem.Release(1)
			out[i] = collatzSteps(i + 1)
		}(i)
	}
```

And this: https://go.dev/ref/spec#Variables:

> Structured variables of [array](https://go.dev/ref/spec#Array_types), [slice](https://go.dev/ref/spec#Slice_types), and [struct](https://go.dev/ref/spec#Struct_types) types have elements and fields that may be [addressed](https://go.dev/ref/spec#Address_operators) individually. Each such element acts like a variable.

And all the nested structures in `HydratedBatchedResponse` are slices.

oliverchang (commented):
nice! is this in response to any particular performance complaints in particular?

another-rex:
Not any issue in particular, I just realized the sequential calls are a major part of the run time when looking at implementing #303 .

pkg/osv/osv.go
another-rex: I did see that package, but since semaphores are under the /x/ experimental package I'm not sure whether we should use it.

oliverchang: Should be fair game to use it. There's a lot of prominent packages using it already: https://deps.dev/go/golang.org%2Fx%2Fsync/v0.1.0/dependents

pkg/osv/osv.go
another-rex: Good idea, done

oliverchang (approved):

oliverchang:
@another-rex are we ready to merge?

G-Rath (approved):

PR #309, opened 2023-03-22
feat: verify provenance

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>
Co-authored-by: Furkan Türkal <furkan.turkal@trendyol.com>

PTAL @laurentsimon @ianlewis

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -24,10 +24,10 @@ jobs:
         uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f # v3.4.0
         with:
           fetch-depth: 0
-      - name: Run Tests
-        uses: ./.github/workflows/test.yaml
-      - name: Run Lints
-        uses: ./.github/workflows/lint.yaml
+      # - name: Run Tests
+      #   uses: ./.github/workflows/test.yaml
+      # - name: Run Lints
+      #   uses: ./.github/workflows/lint.yaml
       - uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18 # v2
       - uses: docker/setup-buildx-action@4b4e9c3e2d4531116a6f8ba8e71fc6e2cb6e6c8c # v2
       - name: Set up Go
@@ -49,9 +49,11 @@ jobs:
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OWNER: ${{ github.repository_owner }}
       - name: Generate subject
         id: hash
         env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ARTIFACTS: "${{ steps.run-goreleaser.outputs.artifacts }}"
         run: |
           set -euo pipefail
@@ -63,7 +65,49 @@ jobs:
       actions: read # To read the workflow path.
       id-token: write # To sign the provenance.
       contents: write # To add assets to a release.
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.5.0
+    uses: developer-guy/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.5.0
     with:
       base64-subjects: "${{ needs.goreleaser.outputs.hashes }}"
       upload-assets: true # upload to a new release
+      draft-release: true
+  verification:
+    needs: [goreleaser, provenance]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # To add assets to a release.
+    steps:
+      - name: Install SLSA verifier
+        uses: slsa-framework/slsa-verifier/actions/installer@v2.1.0
+      - name: Download assets
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PROVENANCE: "${{ needs.provenance.outputs.provenance-name }}"
+        run: |
+          set -euo pipefail
+          TAG=${GITHUB_REF#refs/tags/}
+          echo "Downloading assets for tag $TAG"
+          for i in $(gh api "repos/$GITHUB_REPOSITORY/releases" | jq -r '.[] | select(.tag_name=='\"$TAG\"').assets[].id')
+          do
+            echo "Downloading asset $i"
+            curl -L -o $(gh api "repos/$GITHUB_REPOSITORY/releases/assets/$i" | jq -r '.name') -H "Authorization: token $GH_TOKEN" -H "Accept: application/octet-stream" $(gh api "repos/$GITHUB_REPOSITORY/releases/assets/$i" | jq -r '.url')
+          done
+          echo "Downloading provenance $PROVENANCE"
+          gh -R $GITHUB_REPOSITORY release download --clobber $GITHUB_REF_NAME -p $PROVENANCE
+      - name: Verify assets
+        env:
+          CHECKSUMS: ${{ needs.goreleaser.outputs.hashes }}
+          PROVENANCE: "${{ needs.provenance.outputs.provenance-name }}"
+        run: |
+          set -euo pipefail
+          checksums=$(echo "$CHECKSUMS" | base64 -d)
+          while read -r line; do
+              fn=$(echo $line | cut -d ' ' -f2)
+              echo "Verifying $fn"
+              slsa-verifier verify-artifact "$fn" \
+                            --provenance-path "$PROVENANCE" \
+                            --source-uri "github.com/$GITHUB_REPOSITORY" \
+                            --source-tag "$GITHUB_REF_NAME" \
+                            --builder-id "https://github.com/developer-guy/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v1.5.0"
+          done <<<"$checksums"
+  
+
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -32,7 +32,7 @@ builds:
 dockers:
   # Arch: amd64
   - image_templates:
-      - 'ghcr.io/google/osv-scanner:{{ .Tag }}-amd64'
+      - 'ghcr.io/{{ .Env.OWNER }}/osv-scanner:{{ .Tag }}-amd64'
     dockerfile: goreleaser.dockerfile
     use: buildx
     build_flag_templates:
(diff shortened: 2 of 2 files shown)


.github/workflows/goreleaser.yml
```diff
@@ -24,10 +24,10 @@ jobs:
         uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f # v3.4.0
         with:
           fetch-depth: 0
-      - name: Run Tests
-        uses: ./.github/actions/test.yaml
-      - name: Run Lints
-        uses: ./.github/actions/lint.yaml
+      # - name: Run Tests
```
oliverchang, line 27: Can we bring these back? This is an important part of the release process.

.github/workflows/goreleaser.yml
```diff
@@ -67,3 +71,43 @@ jobs:
...
+    steps:
+      - name: Install SLSA verifier
+        uses: slsa-framework/slsa-verifier/actions/installer@v2.1.0
+      - name: Download assets
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PROVENANCE: "${{ needs.provenance.outputs.provenance-name }}"
+        run: |
+          set -euo pipefail
+          TAG=${GITHUB_REF#refs/tags/}
```
oliverchang, line 88: There's a fair bit of bash here to maintain :) 

@laurentsimon @ianlewis is this recommended practice? If so, would it be possible at all to generalize this in a more re-usable way "upstream" somewhere as part of the slsa-framework repo?

oliverchang (commented):
thanks for this contribution! just some questions for both you and @laurentsimon and @ianlewis :)

.github/workflows/goreleaser.yml
ianlewis: Yeah, it might be a good idea to have an action for verifying artifacts on a GitHub release. i.e. Given a release download all the artifact and verify them.

This would only work for a situation like osv-scanner where you provide binary artifacts along with the provenance (as opposed to some other language package) but I think it could still be useful.

We would also need to be careful to provide a way to provide the list of expected artifacts. We would need to catch the situation where an artifact was excluded from the provenance even though it was expected to be there (in this case `goreleaser.outputs.hashes`)

/cc @laurentsimon

.github/workflows/goreleaser.yml
```diff
@@ -49,12 +49,16 @@ jobs:
...
+          ACTOR: ${{ github.actor }}
       - name: Generate subject
         id: hash
         env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ARTIFACTS: "${{ steps.run-goreleaser.outputs.artifacts }}"
         run: |
           set -euo pipefail
+          gh api "repos/developer-guy/osv-scanner/releases"
+          echo $GITHUB_REPOSITORY
```
ianlewis, line 61: I assume this is debugging code and you'll want to remove it?

.goreleaser.yml
```diff
@@ -32,7 +32,7 @@ builds:
 dockers:
   # Arch: amd64
   - image_templates:
-      - 'ghcr.io/google/osv-scanner:{{ .Tag }}-amd64'
+      - 'ghcr.io/{{ .Env.ACTOR }}/osv-scanner:{{ .Tag }}-amd64'
```
ianlewis, line 35: Is this right? I'm not sure that the `github.actor` is going to be right if the repo is owned by an organization.

.github/workflows/goreleaser.yml
```diff
@@ -67,3 +71,43 @@ jobs:
...
+          PROVENANCE: "${{ needs.provenance.outputs.provenance-name }}"
+        run: |
+          set -euo pipefail
+          TAG=${GITHUB_REF#refs/tags/}
+          echo "Downloading assets for tag $TAG"
+          for i in $(gh api "repos/$GITHUB_REPOSITORY/releases" | jq -r '.[] | select(.tag_name=='\"$TAG\"').assets[].id')
+          do
+            echo "Downloading asset $i"
+            curl -L -o $(gh api "repos/$GITHUB_REPOSITORY/releases/assets/$i" | jq -r '.name') -H "Authorization: token $GH_TOKEN" -H "Accept: application/octet-stream" $(gh api "repos/$GITHUB_REPOSITORY/releases/assets/$i" | jq -r '.url')
+          done
```
ianlewis, line 94: Is there a reason you can't just do `gh release download "$TAG"` ?

.github/workflows/goreleaser.yml
```diff
@@ -67,3 +71,43 @@ jobs:
...
+          echo "Downloading provenance $PROVENANCE"
+          gh -R $GITHUB_REPOSITORY release download --clobber $GITHUB_REF_NAME -p $PROVENANCE
```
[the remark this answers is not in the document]
laurentsimon: I think the code from https://github.com/google/go-containerregistry/blob/main/.github/workflows/release.yml#L48-L76 should work just fine, no?

.github/workflows/goreleaser.yml
developer-guy: I can work on this 🤩

.github/workflows/goreleaser.yml
developer-guy: the reason why I did this is that workflow didn't worked as it didn't find these action files under `.github/actions`.

.goreleaser.yml
developer-guy: I might use `github.repository_owner` instead, will it work?

.github/workflows/goreleaser.yml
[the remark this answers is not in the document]
developer-guy: here is the PR for that: _https://github.com/google/go-containerregistry/pull/1611_

`gh release download` didn't work with draft releases as osv-scanner marks its release as draft and I looked for a solution for `gh release download` to enable downloading assets from draft release but I couldn't find anything useful, then I turned the solution back to the GitHub API calls via `gh api`. But let me try again to ensure that I didn't work.

.github/workflows/goreleaser.yml
developer-guy: for the same reason I explained above: _https://github.com/google/osv-scanner/pull/309#discussion_r1146679828_

.github/workflows/goreleaser.yml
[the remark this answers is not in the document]
developer-guy: it worked btw with the right permissions, so to download releases it requires you to put write permissions for content. Great!

.github/workflows/goreleaser.yml
developer-guy: it worked @ianlewis with right permisssions.

.github/workflows/goreleaser.yml
another-rex: Ah my bad, fixing this in #311.

.goreleaser.yml
developer-guy: I renamed it as `OWNER` and used `github.repository_owner`.

.github/workflows/goreleaser.yml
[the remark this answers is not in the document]
laurentsimon: that's interesting. Maybe file an issue on GitHub.. sounds like a bug... they should have a REST flag to allow for draft release to be fetched, without the need for a write token.

.github/workflows/goreleaser.yml
```diff
@@ -67,3 +69,36 @@ jobs:
     with:
       base64-subjects: "${{ needs.goreleaser.outputs.hashes }}"
       upload-assets: true # upload to a new release
+  verification:
+    needs: [goreleaser, provenance]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # To add assets to a release.
```
laurentsimon, line 76: please file a bug on https://github.com/actions/runner and cc me. Requiring write permission is really odd

laurentsimon (approved):
LGTM

.github/workflows/goreleaser.yml
developer-guy: let me try it again, I changed back it to "read".

.github/workflows/goreleaser.yml
developer-guy: @another-rex I think the problem still continues. I also do not think that this is the way of referencing different workflows within the workflow.

Screen Shot 2023-03-25 at 3 29 07 PM

.github/workflows/goreleaser.yml
developer-guy: see it didn't worked: _https://github.com/developer-guy/osv-scanner/actions/runs/4519224365/jobs/7959640924_

developer-guy: it didn't work with "write" permission either: 
_https://github.com/developer-guy/osv-scanner/actions/runs/4519341906/jobs/7959797410#step:1:17_

but it worked once what is going on 🤔 
_https://github.com/developer-guy/osv-scanner/actions/runs/4504207210/jobs/7928457868#step:1:17_

developer-guy: this script is just working fine with `write` permission: 

```bash
set -euo pipefail
  TAG=${GITHUB_REF#refs/tags/}
  echo "Downloading assets for tag $TAG"
  for i in $(gh api "repos/$GITHUB_REPOSITORY/releases" | jq -r '.[] | select(.tag_name=='\"$TAG\"').assets[].id')
  do
    echo "Downloading asset $i"
    curl -L -o $(gh api "repos/$GITHUB_REPOSITORY/releases/assets/$i" | jq -r '.name') -H "Authorization: token $GH_TOKEN" -H "Accept: application/octet-stream" $(gh api "repos/$GITHUB_REPOSITORY/releases/assets/$i" | jq -r '.url')
  done
  echo "Downloading provenance $PROVENANCE"
  gh -R $GITHUB_REPOSITORY release download --clobber $GITHUB_REF_NAME -p $PROVENANCE
```

developer-guy: we even don't need this line:

```bash
gh -R $GITHUB_REPOSITORY release download --clobber $GITHUB_REF_NAME -p $PROVENANCE
```

and I also disabled uploading assets in the generic generator settings:

```bash
upload-assets: false
```

.github/workflows/goreleaser.yml
another-rex: You're right, I misunderstood how the uses line works. I updated and tested it in #315, which seems to work in my fork.

.github/workflows/goreleaser.yml
oliverchang: @ianlewis @laurentsimon thoughts on this going into some slsa-framework repo instead (either a new one or existing?)

 I would very much prefer that approach as it makes this work a lot more useful and we (OSV-Scanner maintainers) have less things we're not as familiar with to maintain :)

Spiritaine (approved):

PR #310, opened 2023-03-23
Improve the OSV models to allow for 3rd party use of the library.

calebbrown:
- splits the models into separate structs
- defines a bunch of constants
- adds Credit and Severity types


another-rex (approved):
Thanks! Looks much cleaner.

Issue #324, opened 2023-04-02
PNPM lock file v6.0+ broken support


dbtedman:
PNPM has updated their lock file format such that it can no longer be processed correctly by osv-scanner.

The issue appears two fold, one in the version specifier, previously `lockfileVersion: 5.4` now `lockfileVersion: '6.0'` as well as the way it defines package name and version pairs, previously `/acorn/8.7.0` now `/acorn@8.7.0`.

PR #325, opened 2023-04-02
Support PNPM v6+ Lockfile

dbtedman:
A proposed fix for https://github.com/google/osv-scanner/issues/324 issue.

### DIFF
--- a/pkg/lockfile/fixtures/pnpm/one-package-v6-lockfile.yaml
+++ b/pkg/lockfile/fixtures/pnpm/one-package-v6-lockfile.yaml
@@ -1,0 +1,14 @@
+lockfileVersion: '6.0'
+
+dependencies:
+  acorn:
+    specifier: 8.7.0
+    version: 8.7.0
+
+packages:
+
+  /acorn@8.7.0:
+    resolution: {integrity: sha512-V/LGr1APy+PXIwKebEWrkZPwoeoF+w1jiOBUmuxuiUIaOHtob8Qc9BTrYo7VuI5fR8tqsy+buA2WFooR5olqvQ==}
+    engines: {node: '>=0.4.0'}
+    hasBin: true
+    dev: false
--- a/pkg/lockfile/parse-pnpm-lock.go
+++ b/pkg/lockfile/parse-pnpm-lock.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"gopkg.in/yaml.v3"
@@ -23,8 +24,32 @@ type PnpmLockPackage struct {
 }
 
 type PnpmLockfile struct {
+	Version  float64                    `yaml:"lockfileVersion"`
+	Packages map[string]PnpmLockPackage `yaml:"packages,omitempty"`
+}
+
+type pnpmLockfileV6 struct {
 	Version  string                     `yaml:"lockfileVersion"`
 	Packages map[string]PnpmLockPackage `yaml:"packages,omitempty"`
+}
+
+func (l *PnpmLockfile) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var lockfileV6 pnpmLockfileV6
+
+	if err := unmarshal(&lockfileV6); err != nil {
+		return err
+	}
+
+	parsedVersion, err := strconv.ParseFloat(lockfileV6.Version, 64)
+
+	if err != nil {
+		return err
+	}
+
+	l.Version = parsedVersion
+	l.Packages = lockfileV6.Packages
+
+	return nil
 }
 
 const PnpmEcosystem = NpmEcosystem
--- a/pkg/lockfile/parse-pnpm-lock.go
+++ b/pkg/lockfile/parse-pnpm-lock.go
@@ -83,7 +83,7 @@ func extractPnpmPackageNameAndVersion(dependencyPath string) (string, string) {
 
 func parseNameAtVersion(value string) (name string, version string) {
 	// look for pattern "name@version", where name is allowed to contain zero or more "@"
-	matches := regexp.MustCompile("^(.+)@([^@/]+)$").FindStringSubmatch(value)
+	matches := regexp.MustCompile("^(.+)@([\\d.]+)$").FindStringSubmatch(value)
 
 	if len(matches) != 3 {
 		return name, ""
--- a/pkg/lockfile/parse-pnpm-lock.go
+++ b/pkg/lockfile/parse-pnpm-lock.go
@@ -64,11 +64,8 @@ func extractPnpmPackageNameAndVersion(dependencyPath string) (string, string) {
 		version = parts[0]
 	}
 
-	// handle pnpm v6.0+ package specifier that uses name@version pattern
-	atName := strings.Split(name, "@")
-
-	if version == "" && len(atName) == 2 {
-		return atName[0], atName[1]
+	if version == "" {
+		name, version = parseNameAtVersion(name)
 	}
 
 	if version == "" || !startsWithNumber(version) {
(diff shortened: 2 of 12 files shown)


pkg/lockfile/fixtures/pnpm/one-package-v6-lockfile.yaml
```diff
@@ -0,0 +1,14 @@
+lockfileVersion: '6.0'
+
+dependencies:
+  acorn:
+    specifier: 8.7.0
+    version: 8.7.0
+
+packages:
+
+  /acorn@8.7.0:
```
oliverchang, line 10: Can you add a test to make sure we handle packages with "@" in the name properly? 

e.g. https://github.com/privatenumber/minification-benchmarks/blob/c2a2e3e4c2fc3b286a5ad4ef8b3b0d68afa49f68/pnpm-lock.yaml#L208

pkg/lockfile/parse-pnpm-lock.go
```diff
@@ -64,6 +64,13 @@ func extractPnpmPackageNameAndVersion(dependencyPath string) (string, string) {
 		version = parts[0]
 	}
 
+	// handle pnpm v6.0+ package specifier that uses name@version pattern
+	atName := strings.Split(name, "@")
```
oliverchang, line 68: Does this handle packages with "@" in the name?

oliverchang (commented):
thank you for the contribution!! some comments/questions

pkg/lockfile/parse-pnpm-lock.go
G-Rath: That should be handled due to line 53, but we should double check with a v6 specific test

dbtedman: Looks like it was a good idea to add the additional test case as it did have a bug when the package name contained an "@".

I have added this test case and updated my implementation to match the "name@version" pattern explicitly, but allow name to contain zero or more "@" letters in it.

Many of the fixtures used in `parse-pnpm-lock_test.go` tests are using lock file version `5.3`. To ensure there are no further bugs with the implementation, should I create a duplicate of each using current PNPM version? (or would this be going overboard as it would be adding an additional 12 or so test cases)

G-Rath: What I usually do is put a handful of the different types in the single test, focusing on the ones that are most likely to be impacted (in particular, we probably don't need to have multiple tests for different number of packages)

dbtedman: I think that the two test cases included in this PR are sufficient for now (please let me know if there are other test scenarios you would like me to add). Additionally, I have built and tested the osv-scanner with a OSS project I am working on that uses the v6.0 lock file and lots of dependencies and everything looks ok.

pkg/lockfile/fixtures/pnpm/one-package-v6-lockfile.yaml
dbtedman: See the resolved conversation below for details on the updated test cases.

dbtedman:
@oliverchang @G-Rath anything else you need from my end to be able to review this PR?

pkg/lockfile/parse-pnpm-lock.go
```diff
@@ -77,6 +81,17 @@ func extractPnpmPackageNameAndVersion(dependencyPath string) (string, string) {
 	return name, version
 }
 
+func parseNameAtVersion(value string) (name string, version string) {
+	// look for pattern "name@version", where name is allowed to contain zero or more "@"
+	matches := regexp.MustCompile("^(.+)@([^@/]+)$").FindStringSubmatch(value)
```
oliverchang, line 86: Can we be a bit stricter here to avoid potentially matching on too many things? Some things we can do:

- Require colon at the end of this. i.e. ":$" 
- Match on `[\d\.]` instead of `[^@/]` for the version bits. 
- Anything else?

oliverchang (commented):
thank you @dbtedman! 

Sorry for the slow reviews -- a lot of us here in APAC were out due to public holidays.

pkg/lockfile/parse-pnpm-lock.go
```diff
@@ -23,7 +23,7 @@ type PnpmLockPackage struct {
 }
 
 type PnpmLockfile struct {
-	Version  float64                    `yaml:"lockfileVersion"`
+	Version  string                     `yaml:"lockfileVersion"`
```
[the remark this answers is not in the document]
dbtedman: Looks like with latest PNPM lock file version, they changed the format of the version from a float to string. The float64 type here cannot be mapped to the v6 lock file. The string type however maps successfully to both lock file formats.

One option would be to define a new type for the v6+ lock file version but that does not seem elegant.

Happy for your suggestions on this one (I am still fairly new to Go so I may not be aware of a nicer solution to handle this).

pkg/lockfile/parse-pnpm-lock.go
dbtedman: Sounds like a good idea, I will see what I can do to make this match more specific.

dbtedman: The `name` parameter does not include the `:` at the end in this case as its already been removed so I can't add that match however.

pkg/lockfile/parse-pnpm-lock.go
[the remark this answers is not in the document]
dbtedman: To avoid breaking that public interface, a new local interface could be added, and then inside the `ParsePnpmLock` function, explicitly convert the version to `float64`.

Something like the following maybe: (I have included just the key changes below)

(The following passes the test suite)

```go
type PnpmLockfile struct {
	Version  float64                    `yaml:"lockfileVersion"`
	Packages map[string]PnpmLockPackage `yaml:"packages,omitempty"`
}

type pnpmLockfileV6 struct {
	Version  string                     `yaml:"lockfileVersion"`
	Packages map[string]PnpmLockPackage `yaml:"packages,omitempty"`
}

func ParsePnpmLock(pathToLockfile string) ([]PackageDetails, error) {
	var parsedLockfileV6 *pnpmLockfileV6
	var parsedLockFile *PnpmLockfile

	lockfileContents, err := os.ReadFile(pathToLockfile)

	if err != nil {
		return []PackageDetails{}, fmt.Errorf("could not read %s: %w", pathToLockfile, err)
	}

	err = yaml.Unmarshal(lockfileContents, &parsedLockfileV6)

	if err != nil {
		return []PackageDetails{}, fmt.Errorf("could not parse %s: %w", pathToLockfile, err)
	}

	parsedVersion, err := strconv.ParseFloat(parsedLockfileV6.Version, 64)

	if err != nil {
		return []PackageDetails{}, fmt.Errorf("could not parse version %s: %w", parsedLockfileV6.Version, err)
	}

	parsedLockFile = &PnpmLockfile{
		Version:  parsedVersion,
		Packages: parsedLockfileV6.Packages,
	}

	return parsePnpmLock(*parsedLockFile), nil
}
```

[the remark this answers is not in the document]
oliverchang: Thanks! That looks pretty reasonable to me. @G-Rath any thoughts?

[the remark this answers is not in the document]
G-Rath: I'd probably be tempted to try using `UnmarshalYAML` instead so that way `PnpmLockfile` by itself can "parse" v6 lockfiles - otherwise this only fixes the problem for when you're using `ParsePnpmLock` (which arguably is not useful since that doesn't expose the `PnpmLockfile` in anyway)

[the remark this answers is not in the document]
dbtedman: That's done, I believe I got the interface implemented correctly. Tests seem to be all passing.

oliverchang (approved):
Looks great, thanks!

Issue #335, opened 2023-04-11
SPDX parsing improvements


oliverchang:
Tracking bug for a number of improvements to our current SPDX parsing

- [ ] We should warn with a more specific error message when there are no PURLs available in the SBOM.
- [x] Currently for passing SPDX JSON files to `-L` we expect ".spdx.json" in the filename. Otherwise we will silently report that there are no packages. We should be more flexible here since we've seen a few instances/complaints for this.  
- [x] We don't propagate detailed errors messages back to the user when a malformed SPDX file is provided (https://github.com/google/osv-scanner/blob/cfaf9f8205ac0e2a8f1fb93ea792bc648831d0fd/internal/sbom/spdx.go#L66)

anotherbridge:
@oliverchang For your second point I would suggest that first an appropriate error message should be placed, since if the tool reports that no packages were found the first point one would think of is that something within the SBOM is incorrect and not that the filename is non conformant.

oliverchang:
Thanks @anotherbridge. We're actually going to be more relaxed on the filenames here (in https://github.com/google/osv-scanner/pull/339) to make the scanner easier to use. Point 3 in this bug (more detailed parsing errors) was also implemented.

This should go out in our next release.

PR #339, opened 2023-04-12
SBOM parsing improvements.

oliverchang:
Addresses part of #335

1. Be more relaxed about filename matching the spec when the SBOM is explicitly passed via --sbom since we've seen multiple complaints from different users about this.
2. Surface all parse errors when SBOM parsing fails (and we're explicitly scanning via --sbom).

This will look something like:

```
Failed to parse SBOM using all supported formats:
failed to parse SPDX:
        failed trying json: failed to parse Supplier 'Google LLC'
        failed trying rdf: found extra chars before tag start
        failed trying tv: no colon found in '{'
failed to parse CycloneDX:
        failed trying json: invalid BOMFormat
        failed trying xml: XML syntax error on line 74: expected attribute name in element
exit status 128
```

### DIFF
--- a/.github/workflows/lint-action/action.yml
+++ b/.github/workflows/lint-action/action.yml
@@ -23,3 +23,5 @@ runs:
       with:
         # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
         version: v1.51.1
+        # https://github.com/golangci/golangci-lint-action/issues/135
+        skip-pkg-cache: true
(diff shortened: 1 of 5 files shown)


michaelkedar (approved):

oliverchang:
@another-rex does this looks reasonable to you?

another-rex (approved):
LGTM, just FYI the SPDX 0.5.0 has also been recently released which changes the interface up again, so this might need further tweaking when we upgrade that dependency.

oliverchang:
> 

yep, will address that separately later!

PR #341, opened 2023-04-13
Make the reporter public

another-rex:
Make the reporter package public so that it can be passed into the public DoScan function.

### DIFF
--- a/pkg/reporter/reporter.go
+++ b/pkg/reporter/reporter.go
@@ -1,10 +1,11 @@
-package output
+package reporter
 
 import (
 	"fmt"
 	"io"
 	"strings"
 
+	"github.com/google/osv-scanner/internal/output"
 	"github.com/google/osv-scanner/pkg/models"
 )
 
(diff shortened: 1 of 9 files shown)


G-Rath (approved):

oliverchang (approved):

pkg/reporter/reporter.go
```diff
@@ -60,11 +61,11 @@ func (r *Reporter) PrintText(msg string) {
...
 	case "json":
-		return PrintJSONResults(vulnResult, r.stdout)
+		return output.PrintJSONResults(vulnResult, r.stdout)
 	case "markdown":
-		PrintMarkdownTableResults(vulnResult, r.stdout)
+		output.PrintMarkdownTableResults(vulnResult, r.stdout)
 	case "table":
-		PrintTableResults(vulnResult, r.stdout)
+		output.PrintTableResults(vulnResult, r.stdout)
 	}
```
michaelkedar, line 69: nit:
Since it's public, I think it'd make sense to make `Reporter` an interface, with individual json/markdown/table/void reporter implementations.

That way, an external program could write/wrap their own logging/formatting outputs.

(Also, it avoids this awkward format checking)

michaelkedar (approved):

pkg/reporter/reporter.go
another-rex: This sounds like a interesting idea, I'll make the changes and push it in a separate PR.

PR #345, opened 2023-04-13
Refactor reporter to interfaces

another-rex:
The reporter has been refactored to be an interface with various implementations. Based on the idea from Michael here: https://github.com/google/osv-scanner/pull/341#discussion_r1164891532 

Want to get some feedback on whether this is worth changing or not.

I'm also not sure if we should have the different reporter types public here, or make them private and have the users of the library provide their own implementations of Reporter.


oliverchang (approved):
looks a lot cleaner to me!

michaelkedar (approved):
LGTM

cmd/osv-scanner/main.go
```diff
@@ -20,10 +20,10 @@ var (
 )
 
 func run(args []string, stdout, stderr io.Writer) int {
-	var r *reporter.Reporter
+	var r reporter.Reporter
 
 	cli.VersionPrinter = func(ctx *cli.Context) {
-		r = reporter.NewReporter(ctx.App.Writer, ctx.App.ErrWriter, "")
+		r = reporter.NewTableReporter(stdout, stderr, false)
```
G-Rath, line 26: This should not have been changed because it causes tests to be flakey - see #166

PR #346, opened 2023-04-14
Move a specific regex to static variable

### LINKED ISSUE #333: Regexp compilation taking a long time
It was noted that in some workloads regexp compliation is taking up to 60% of osv-scanner's runtime.

Especially this line: https://github.com/google/osv-scanner/blob/26c9df8ad0e891dda69f9a620b638710bcd49da2/pkg/lockfile/parse-yarn-lock.go#L114

A solution would be to store the result of the compilation, rather than compiling every time. 

FYI @spencerschrock

another-rex:
Fixes #333 

I measured the time taken to compile the regexp at startup and it's unnoticeable for these regexp, so I don't think we need the added complexity with `sync.Once`.


G-Rath:
imo the principle is what's important - as @spencerschrock right now it happens that scorecard is scanning a large amount of yarn lockfiles relative to other lockfiles, so that's why this _seems_ the slowest but that could very easily change and I know there's a huge regexp for `requirements.txt` which will have an impact.

So while I agree it's not necessarily worth the extra complexity of `sync.Once` for this case, I think it would be worth considering the principle on "should this regexp be global or not?" (to which `sync.Once` should mean no one has to actually debate over the benchmarks, as the rule can be just always use that and don't worry about it).

another-rex:
@G-Rath Good point, I added a utility function that'll cache regex's globally and replaced most instances of regexp.MustCompile with it, so it should be relatively painless to use in the future as well. PTAL

internal/utility/regex.go
```diff
@@ -0,0 +1,17 @@
+package utility
```
oliverchang, line 1: Per https://go.dev/blog/package-names, let's avoid generic package names like these.

Let's just call this cachedregexp or something similar. We can make the function name match the one in the canonical `regexp` as well.

spencerschrock: +1. Would also allow you to simplify the other names in the file

possibilities:
`compiledRegexCache` -> `cache`
`CachedRegexMustCompile` -> `MustCompile`

spencerschrock:
I like the cached regex package approach, provides a simple replacement call without sacrificing readability or usability conventions that a caller would be responsible for following.

Wasn't aware of `sync.Map` but it seems like a good fit:
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow

G-Rath:
fwiw it looks like [this package](https://github.com/umisama/go-regexpcache/tree/master) provides exactly what you want, and doesn't have any other dependencies

another-rex:
It does look like basically what is implemented here, but since this is just a couple of lines, I don't think it's necessary to pull in a dependency.

oliverchang (approved):

oliverchang:
> It does look like basically what is implemented here, but since this is just a couple of lines, I don't think it's necessary to pull in a dependency.

+1. It's pretty trivial, and not worth adding this to our supply chain.

Issue #347, opened 2023-04-14
misleading document or code implementation


trganda:
As the document example from [specify-sbom](https://google.github.io/osv-scanner/usage/#specify-sbom),

We can specific the sbom file to osv-scanner.

```
osv-scanner --sbom=/path/to/your/sbom.json
```

But the code from v1.3.1, [func (c *CycloneDX) MatchesRecognizedFileNames(path string) bool](https://github.com/google/osv-scanner/blob/v1.3.1/internal/sbom/cyclonedx.go#L25) was expected the file name like

```go
expectedGlobs := []string{
  "bom.xml",
  "bom.json",
  "*.cdx.json",
  "*.cdx.xml",
}
```

May be we should update the use case of document or re-implement the `MatchesRecognizedFileNames`

another-rex:
Thanks for pointing this out, you're right the documentation is outdated/misleading for v1.3.1, I'll create a PR to fix this.

Though FYI recently #339 was merged which will relax the code implementation so that the example in the documentation will be correct again in the next release.

trganda:
That's good, hope the tools be better.😃

Issue #290, opened 2023-03-13
SBOM scanning documentation improvements and more helpful error messages


another-rex:
Create clearer documentation on why SBOMs generated by some tools might not work with osv-scanner scanning, as they don't generate package URLs, or don't include what ecosystem the package comes from. 

The scanner's error should also be more helpful error in clarifying this, or maybe link to the documentation. 

See #93 for additional context.

Issue #93, opened 2022-12-21
No package sources found Error when scanning SBOMs


MFaisalZaki:
I have generated a valid `spdx` file from VCPKG, then I try to use it with `osv-scanner`, but it generates no package sources found.
Can you please help me with this issue?
[vcpkg.spdx.zip](https://github.com/google/osv-scanner/files/10277401/vcpkg.spdx.zip)

Rohan-Kar:
SPDX file format may not be supported . you need to tried with JSON output once .

Rohan-Kar:
> I have generated a valid `spdx` file from VCPKG, then I try to use it with `osv-scanner`, but it generates no package sources found. Can you please help me with this issue? [vcpkg.spdx.zip](https://github.com/google/osv-scanner/files/10277401/vcpkg.spdx.zip)

You need to tried once with JSON output . SPDX file may not be supported

MFaisalZaki:
I have tried [spdx-tool](h
[converted.zip](https://github.com/google/osv-scanner/files/10298602/converted.zip)
ttps://github.com/spdx/tools-java/releases/tag/v1.1.3) to convert to JSON. Yet the same issue. I am attaching both the original file and its converted version.

Rohan-Kar:
> I have tried [spdx-tool](https://github.com/spdx/tools-java/releases/tag/v1.1.3) to convert to JSON. Yet the same issue. I am attaching both the original file and its converted version.
> 
> [converted.zip](https://github.com/google/osv-scanner/files/10298602/converted.zip)

you can try with cyclonedx plugin .  https://github.com/CycloneDX/cyclonedx-maven-plugin and generate the bom.json . Hope this will fixed your problem .

MFaisalZaki:
@Rohan-Kar Thanks for the tip. To ensure I use this to generate SBOM from my SPDX file? And if you have any idea how I can use vcpkg with CycloneDX directly, this would be great.

MFaisalZaki:
Update, I have used [cyclonedx-cli](https://github.com/CycloneDX/cyclonedx-cli) to convert from `spdx` to `cdx` using the following command: 
```
cyclonedx-win-x64.exe convert --input-file vcpkg.spdx.json --output-file vcpkg.json --input-format spdxjson --output-format json
```
And I received this output:
```
Scanned CycloneDX SBOM
No package sources found, --help for usage information.
```
The problem still exists. 
[...]

PR #349, opened 2023-04-17
Better SBOM documentation and error message

### LINKED ISSUE #335: SPDX parsing improvements

another-rex:
Closes #335 #290 #93 #347 (Turns out we have a bunch of duplicate issues all roughly tracking the same thing)

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -39,7 +39,7 @@ osv-scanner --sbom=/path/to/your/sbom.spdx.json
 [SPDX] and [CycloneDX] SBOMs using [Package URLs] are supported. The format is
 auto-detected based on the input file contents and the file name.
 
-When scanning a directory, only SBOMs following the specification filename will be scanned. See [SPDX Filenames]; [CycloneDX Filenames].
+When scanning a directory, only SBOMs following the specification filename will be scanned. See the specs for [SPDX Filenames] and [CycloneDX Filenames].
 
 [SPDX]: https://spdx.dev/
 [SPDX Filenames]: https://spdx.github.io/spdx-spec/v2.3/conformance/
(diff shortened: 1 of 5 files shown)


oliverchang (approved):
thanks!!

docs/usage.md
```diff
@@ -33,13 +33,17 @@ The `--no-ignore` flag can be used to force the scanner to scan ignored files.
...
 ```bash
-osv-scanner --sbom=/path/to/your/sbom.json
+osv-scanner --sbom=/path/to/your/sbom.spdx.json
 ```
 
 [SPDX] and [CycloneDX] SBOMs using [Package URLs] are supported. The format is
-auto-detected based on the input file contents.
+auto-detected based on the input file contents and the file name.
+
+When scanning a directory, only SBOMs following the specification filename will be scanned. See [SPDX Filenames]; [CycloneDX Filenames].
```
hayleycd, line 42:
```suggestion
When scanning a directory, only SBOMs following the specification filename will be scanned. See the specs for [SPDX Filenames] and [CycloneDX Filenames].
```

Issue #336, opened 2023-04-05
[improvement] osv output summary in cli should also contain 'vulnerability fixed version' column


dineshr93:
Hello,

Right now the ` osv --L requirements.txt`
produces the following output
![image](https://user-images.githubusercontent.com/4498415/230174787-34606691-f2b5-4f8e-9d0f-e0aa7eeb38fc.png)

I am forced to go through each and every link and pick the version for which vulnerabilities is fixed.

It would be ideal to show the fixed into the summary.

dineshr93:
fix for this provided in pull request https://github.com/google/osv-scanner/pull/343

dineshr93:
Added improved message in this pull request https://github.com/google/osv-scanner/pull/350

![image](https://user-images.githubusercontent.com/4498415/232407295-742650e7-99b0-408c-9939-3036b0a4e99b.png)

PR #350, opened 2023-04-17
(enhancement) Table output: detailed fixed version

dineshr93:
Added fixed version column as seen in the below screenshot
image

fixes https://github.com/google/osv-scanner/issues/336


dineshr93:
closed previous related pull request https://github.com/google/osv-scanner/pull/343 because this pull request contains improved details in the column

dineshr93:
> Adding more columns to our table output is expensive, and screen real estate is limited. We spent a lot of time earlier to reduce the horizontal width to make the table easier to render on smaller screens/terminals.

Hi @oliverchang Sure.  

Finding the vulnerability is one aspect and mitigating those issue is another.

By including this **Fixed versions** column, makes the osv-scanners life easier, to take action without going through each link and pick the fixed version to upgrade the library.  (for applicability check vex might be used)

I tried to include only the fix version in this discarded PR https://github.com/google/osv-scanner/pull/343 . But there is a multiple fixed version within a pkg  and people would not know the fix is for what package.  So I was forced to include purl and ecosystem details
for ex: 
image


But I agree with the screen realestate. I have also below shortened format display

image

If you agree with this atleast we can ignore this PR and raise a new PR with shortened format.

dineshr93:
Hi @another-rex  @oliverchang  

I have corrected the lint and updated the unit tests acordingly. 

This is the final image taken
image

Require your review and approval.

Thanks

oliverchang:
Hi @dineshr93. thank you for continuing to work on this. 

Unfortunately I'm not sure the additional column added provides much value to users. We spent a fair bit of effort before to reduce the horizontal width in #85 , and this would increase that again. 

Regarding its utility to users, the fix version(s) on its own is likely not going to be directly helpful to users because these dependencies are likely layers down in the dependency graph and it's not a simple version bump. If a user would like to make use of this information in an accurate/semi-automated way, the JSON output would be the way to go since they contain the full OSV details. For remediation in general, we also have #352 planned. 

We will re-open this if we decide to revisit this. Once again, thanks for proposing this change.

PR #357, opened 2023-04-25
Update pkg/osv to allow overriding the http client / transport.

jeffmendoza:
I don't know if you all intended `pkg/osv` to be a general purpose https://osv.dev/ client library, but we are using it as that here: https://github.com/guacsec/guac/blob/main/pkg/certifier/osv/osv.go

It would be nice to override the http client used to insert custom transports, etc.

### DIFF
--- a/pkg/osv/osv.go
+++ b/pkg/osv/osv.go
@@ -127,6 +127,12 @@ func checkResponseError(resp *http.Response) error {
 
 // MakeRequest sends a batched query to osv.dev
 func MakeRequest(request BatchedQuery) (*BatchedResponse, error) {
+	return MakeRequestWithClient(request, http.DefaultClient)
+}
+
+// MakeRequestWithClient sends a batched query to osv.dev with the provided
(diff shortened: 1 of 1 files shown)


another-rex (commented):
Being able to override the client seems like a good feature to have. 

The main concern with this solution is if two libraries are both using the osv package, (if I understand go packages correctly), and one overrides the HTTPClient this will change the client for all libraries using osv.go, which might not be desirable.

A potentially better solution could be to create two separate functions (e.g. GetWithClient, HydrateWithClient) which takes in a http client as well, and move the logic in there. Then just make the original Get and Hydrate functions wrap the new functions, filling the client argument with http.DefaultClient.

jeffmendoza:
Yes, that is correct. If it were used in two places in the same binary, there could be contention. I was looking to be the least disruptive.

Updated with the "WithClient" strategy.

another-rex (approved):
LGTM!

Issue #354, opened 2023-04-19
cycle in requirements.txt causes infinite recursion


spencerschrock:
The logic that handles [referring to other requirement files](https://pip.pypa.io/en/stable/reference/requirements-file-format/#referring-to-other-requirements-files) can recurse infinitely (or at least until the stack hits the goroutine 1Gb limit).

https://github.com/google/osv-scanner/blob/2c101c1b0e63b8c7cdb1821fcf7e78208aecc544/pkg/lockfile/parse-requirements-txt.go#L96-L113

Consider the following `requirements.txt`:
```
-r requirements.txt
iterpipes3==0.4
numpy==1.16.4
```

Running `osv-scanner` on this file will crash the program (and my VM when trying to replicate)
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

For what it's worth, it's a problematic `requirements.txt` for `pip` too:

```
pip install -r requirements.txt 
ERROR: Exception:
Traceback (most recent call last):
...
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Would some sort of cycle detection in `osv-scanner` make sense? Or is this too much work to handle what is essentially bad input?
The alternative on my end is just removing the offending repo from my dataset, which I'm fine doing.

G-Rath:
I think an easy way to handle this could be to just track what requirement files have been previously included (and include the one that is _about to be_ included) because that should also optimize for more complex tree shapes that result in the same file being included multiple times in a way that is not a cycle e.g.

```
# base.txt
# (other packages)

# test.txt
-r base.txt
# (other packages)

# dev
-r base.txt
-r test.txt
# (other packages)
```

I would also be interested to know if you have any occurrences of that in your dataset - they'd not be noticable on the final output since packages are duplicated.

spencerschrock:
> I think an easy way to handle this could be to just track what requirement files have been previously included (and include the one that is _about to be_ included) because that should also optimize for more complex tree shapes that result in the same file being included multiple times in a way that is not a cycle e.g.

+1, sounds like a straight-forward way to avoid processing the same file twice.

> ```
> # base.txt
> # (other packages)
> 
> # test.txt
> -r base.txt
> # (other packages)
> 
> # dev
> -r base.txt
> -r test.txt
> # (other packages)
> ```
> 
> I would also be interested to know if you have any occurrences of that in your dataset - they'd not be noticable on the final output since packages are duplicated.

[...]

PR #366, opened 2023-05-04
fix: handle cyclical `-r`s in `requirements.txt`

robotdana:
Ported from G-Rath/osv-detector#191

Note that this is not actually supported by `pip` itself, but doing so actually optimizes the parser a bit anyway by only reading each file exactly once regardless of how often it is required

fixes: #354


G-Rath (changes requested):
It looks like all the empty newlines got dropped in the cherry-pick - could you restore those? otherwise looks good!

another-rex (approved):
LGTM, thanks!

PR #368, opened 2023-05-05
add unit tests for results.go

### LINKED ISSUE #356: add unit tests for results.go
I would like to add tests for `pkg/models/results.go`

khareyash05:
Fixes #356


pkg/models/results_test.go
```diff
@@ -0,0 +1,48 @@
...
+						Score: "1",
+					},
+				},
+			},
+		},
+	}
+	source := PackageSource{Source: SourceInfo{Path: "package"}, Packages: []PackageVulns{pkg}}
+	vulns = VulnerabilityResults{Results: []PackageSource{source}}
+	flattened = vulns.Flatten()
+	if len(flattened) != 1 {
```
another-rex, line 33: Instead of comparing each individual field in it's own if statement, can you swap this to use `cmp.Diff(flattened, expectedFlattened)`, and create a new variable with the expected flattened structure. 

Similar to how "pkg/grouper/grouper_test.go" is done.

another-rex (changes requested):
Thank you! Just a comment about changing how the final comparison is done.

pkg/models/results_test.go
khareyash05: Will do that!

another-rex (approved):
Thanks, LGTM!

pkg/models/results_test.go
```diff
@@ -0,0 +1,49 @@
+package models
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestFlatten(t *testing.T) {
```
another-rex, line 9:
```suggestion
func TestFlatten(t *testing.T) {
	t.Parallel()
```

khareyash05:
@another-rex is my PR ready to merge?

khareyash05:
Thank you @another-rex I would like to contribute more to this repo.Is there any communication channel like Slack for the same? So that we can discuss about the repo and that I can get a clear understanding of the code.

Issue #369, opened 2023-05-06
requirements.txt mis-parses lines that contain --hash


alex:
In a `requirements.txt` file, a dependency may be written like:

```
boto3==1.26.121 --hash=sha256:f87d694c351eba1dfd19b5bef5892a1047e7adb09c57c2c00049de209a8ab55d
```

This specifies both the package version as well as the packages hash. However, `Parse` mishandles this and generates:

```
lockfile.PackageDetails{
    Name:"boto3",
    Version:"1.26.121 --hash=sha256:f87d694c351eba1dfd19b5bef5892a1047e7adb09c57c2c00049de209a8ab55d",
    Commit:"",
    Ecosystem:"PyPI",
    CompareAs:"PyPI"
}
```

Note how `Version` just contains the remainder of the line, not merely the version portion.

G-Rath:
ah yup, I missed [this section](https://pip.pypa.io/en/stable/reference/requirements-file-format/#per-requirement-options) when implementing the parser (also https://pip.pypa.io/en/stable/topics/secure-installs/).

I'd say we just want to split on the first space, and use the first element - @robotdana this could be a good one for you to pick up.

PR #379, opened 2023-05-12
Remove --hash from version in requirements.txt

robotdana:
This removes everything after the whitespace of the version to also
catch the other per-requirement options --global-option and
--config-settings and any future options that may be added

https://pip.pypa.io/en/stable/reference/requirements-file-format/#per-requirement-options

fixes: https://github.com/google/osv-scanner/issues/369

While addressing this i noticed the pip documentation example for --hash
used line continuations which weren't currently supported by this parser
so i've added support for this

### DIFF
--- a/pkg/lockfile/fixtures/pip/line-continuation.txt
+++ b/pkg/lockfile/fixtures/pip/line-continuation.txt
@@ -1,0 +1,17 @@
+# unescaped
+foo==\
+\
+ \
+  \
+1.2.3
+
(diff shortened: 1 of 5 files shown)


G-Rath (approved):
nice stuff! I think it would be good to add another line to the `with-hash` fixture that uses either `--global-option` or `--config-settings` just to show that it's not _specifically_ supporting just `--hash`, but that's not blocking

another-rex (approved):
LGTM, thanks!

PR #394, opened 2023-05-24
PURLs on stdin

another-rex:
Read package urls line by line on the Stdin, this allows the user flexibility to connect up their own parser (e.g. for formats we don't support) for their manifest files and pipe the results directly into osv-scanner.

### DIFF
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -125,21 +125,6 @@ func TestRun(t *testing.T) {
 				built at: n/a
 			`, version),
 		},
-		{
-			name:         "PURLs from stdin",
-			args:         []string{"", "--experimental-stdin-purl"},
-			stdin:        "pkg:npm/ansi-html@0.0.1\n",
-			wantExitCode: 1,
-			wantStdout: `
-            Enter Package URLs, one per line, finishing with EOF:
-            +-------------------------------------+-----------+-----------+---------+--------+
-            | OSV URL (ID IN BOLD)                | ECOSYSTEM | PACKAGE   | VERSION | SOURCE |
-            +-------------------------------------+-----------+-----------+---------+--------+
-            | https://osv.dev/GHSA-whgm-jr23-g3j9 | npm       | ansi-html | 0.0.1   |        |
-            +-------------------------------------+-----------+-----------+---------+--------+
-            `,
-			wantStderr: "",
-		},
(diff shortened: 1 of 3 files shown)


cmd/osv-scanner/main_test.go
```diff
@@ -113,15 +115,29 @@ func TestRun(t *testing.T) {
         No package sources found, --help for usage information.
 			`,
 		},
+
 		{
 			name:         "",
-			args:         []string{"", "--version"},
```
G-Rath, line 118: Why is this test being removed?

G-Rath (commented):
Would be good to have some additional tests covering multiple PURLs and blank input

cmd/osv-scanner/main_test.go
another-rex: Oops, replaced the wrong test, undid this change now.

another-rex: Added more tests

cmd/osv-scanner/main_test.go
```diff
@@ -309,6 +310,100 @@ func TestRun(t *testing.T) {
...
+	t.Parallel()
+
+	tests := []cliTestCase{
+		{
+			name:         "1 PURL from stdin",
+			args:         []string{"", "--experimental-stdin-purl"},
+			stdin:        "pkg:npm/ansi-html@0.0.1\n",
+			wantExitCode: 1,
+			wantStdout: `
+            Enter Package URLs, one per line, finishing with EOF:
```
G-Rath, line 323: nit: it would be good to keep the indentation consistent for these (it should be intended two spaces in from the start of `wantStdout`)

pkg/osvscanner/osvscanner.go
```diff
@@ -532,6 +533,12 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
 		}
 	}
 
+	for _, purl := range actions.Purls {
+		query.Queries = append(query.Queries, osv.MakePURLRequest(purl))
+	}
+
+	// r.PrintText(fmt.Sprintf("%v", query.Queries))
+
```
G-Rath, line 541:
```suggestion

```

G-Rath (approved):

another-rex:
Closing this for now in favor of a more generic version using the format of our batch API queries coming in a future PR.

PR #395, opened 2023-05-24
Fix version printer to use app stdout and stderr

another-rex:
Added back in the change made in #166


G-Rath (changes requested):
This is because you've reverted back to using the global `stdout` and `stderr` in #345 when you should be using the one from `ctx.App` (#166)

another-rex:
:facepalm: My bad, fixed that now.

G-Rath (approved):

michaelkedar (approved):

PR #399, opened 2023-05-30
govulncheck / osv-scanner integration

julieqiu:


Spiritaine (approved):

internal/sourceanalysis/go.go
```diff
@@ -1,35 +1,77 @@
...
+	}
+	defer func() {
+		rerr := os.RemoveAll(dir)
+		if err == nil {
+			err = rerr
+		}
+	}()
 
-		return
+	cmd := scan.Command(context.Background(), "-db", fmt.Sprintf("file://%s", dbdir), "-C", dir, "-json", "./...")
```
oliverchang, line 32: Can you add a comment for aht this is doing?

internal/sourceanalysis/go.go
```diff
@@ -1,35 +1,77 @@
...
 	"golang.org/x/exp/slices"
-	"golang.org/x/vuln/exp/govulncheck"
+	"golang.org/x/vuln/scan"
 )
 
-func goAnalysis(r reporter.Reporter, pkgs []models.PackageVulns, source models.SourceInfo) {
-	vulns, vulnsByID := vulnsFromAllPkgs(pkgs)
-	res, err := govulncheckshim.RunGoVulnCheck(filepath.Dir(source.Path), vulns)
+func goAnalysis(dir string, pkgs []models.PackageVulns) (_ []models.PackageVulns, err error) {
+	_, vulnsByID := vulnsFromAllPkgs(pkgs)
```
oliverchang, line 19: Does this mean we no longer supply the actual OSV vulnerabilities to govulncheck to do the analysis on? And the DB that govulncheck uses is managed separately? This could result in some discrepancy between OSV's scanning and govulncheck right? 

Previously we supplied our own list of OSV vulns here: https://github.com/google/osv-scanner/pull/399/files#diff-1ee86035bb7844e7b4b351c7dafa655a088c8d10d7ae446df8faf1453807a506L24

julieqiu:
Moved to https://github.com/google/osv-scanner/pull/431/files

internal/sourceanalysis/go.go
julieqiu: Done in https://github.com/google/osv-scanner/pull/431/files

internal/sourceanalysis/go.go
julieqiu: Updated and left a comment in https://github.com/google/osv-scanner/pull/431/files

Issue #57, opened 2022-09-01
Build GitHub actions for running osv-scanner


oliverchang:
For running osv-scanner

oliverchang:
(Moving issue to the right repo)

Thanks for proposing this @abhisek! We're still trying to figure out all the necessary requirements for this action (and don't want to rush it), as we want this to go beyond the scope of just scanning (but also include tracking/remediation). 

We're happy to accept contributions to make the necessary plumbing/PR changes for these pieces!

abhisek:
One of the key refactoring change that I would propose is to introduce the following inside `PackageDetails` and have the parser retain this information.

```go
type PackageDetailSource struct {
   sourceFile, sourceLine string
}
```

> It may not be trivial for XML / JSON files where an external format parser is used :(

Subsequently correlate the `vulnerabilityResults` with input packages so that the final result can include a source file and line for each vulnerability. This will allow better consumption of results by annotating source lines in PR using Github (or equivalent) Checks API

G-Rath:
@abhisek that sounds like a pretty huge complexity jump since you'd have to implement custom parsers for at least JSON, YAML, TOML, and XML.

I'd recommend checking out [`check-with-osv-detector`](https://github.com/G-Rath/check-with-osv-detector) as prior art - its a straightforward composite action that pulls the binary from GH releases which is more lightweight than using a docker-based action; now that `osv-scanner` is releasing binaries it should be pretty easy to create an action from that.

Dentrax:
Hey, is this still the case? Has anybody implemented the gh-action yet?

oliverchang:
Thanks @slonka! We're also looking at building officially workflows to run osv-scanner as part of GH Actions soon. @another-rex is the one owning this :)

PR #409, opened 2023-06-14
Add option to include severity in table output

giovanni-bozzano:
Add an optional `--include-severity` flag to include a severity column when using table or markdown output, when severity is present in the OSV schema.


G-Rath (changes requested):
Can you please add tests

giovanni-bozzano:
Sure, test added.

From an end user perspective, it would be useful to also have the numeric score computed in the same column, for a quicker overview. Let me know if this sounds good.

another-rex:
Thanks for working on this @giovanni-bozzano! 

Unfortunately we want to be very careful with what we add to the table output as we want to keep the horizontal width down, and also keep the number of CLI flags limited. 

Converting `CVSS:3.1/AV:L/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:L` into a single numeric score will help, but it also loses a decent amount of information. 

A more long term solution is to create a markdown/text based output format, which is more suitable for adding additional information like severity info, fixed versions...etc, We are currently developing this as part of #57, currently estimated to be completed this quarter. 

For now a workaround would be to parse the json output to get the severity value.

another-rex:
Discussed this a bit more with the team, something similar to this could work showing only the computed scores in the table, and only showing the full string (or a subset of it) in the text output I was describing above so if the user wants more info they can get it. 

We can use https://github.com/goark/go-cvss to compute the score relatively easily, if you can replace the current full severity string with just the score it would be great!

internal/output/table.go
```diff
@@ -15,10 +19,15 @@ import (
...
 
 // PrintTableResults prints the osv scan results into a human friendly table.
-func PrintTableResults(vulnResult *models.VulnerabilityResults, outputWriter io.Writer) {
+func PrintTableResults(vulnResult *models.VulnerabilityResults, outputWriter io.Writer, includeSeverity bool) {
 	outputTable := table.NewWriter()
 	outputTable.SetOutputMirror(outputWriter)
-	outputTable.AppendHeader(table.Row{"OSV URL (ID In Bold)", "Ecosystem", "Package", "Version", "Source"})
+	row := table.Row{"OSV URL (ID In Bold)"}
+	if includeSeverity {
+		row = append(row, "Severity")
```
another-rex, line 27: Can you rename severity in the output to "CVSS".

another-rex (changes requested):
Thanks for the change! Agreed with adding to the documentation to clarify that this is the base severity score. 

I think now that it's reasonably short, we can remove the `include-severity` flag and just always show the severity column. (Does mean a couple of the tests in main_test.go will need to be updated)

oliverchang:
> Thanks for the change! Agreed with adding to the documentation to clarify that this is the base severity score.
> 
> I think now that it's reasonably short, we can remove the `include-severity` flag and just always show the severity column. (Does mean a couple of the tests in main_test.go will need to be updated)

CC @hayleycd to see where we can fit this in our current docs.

giovanni-bozzano:
I removed the flag and renamed the column to "CVSS". Tests have been changed accordingly.

internal/output/table.go
```diff
@@ -100,6 +104,36 @@ func tableBuilderInner(vulnResult *models.VulnerabilityResults, addStyling bool,
...
+					for _, vuln := range pkg.Vulnerabilities {
+						if vuln.ID == vulnID {
+							severities = vuln.Severity
+						}
+					}
+					for i, severity := range severities {
+						if i != 0 {
+							outputSeverities = append(outputSeverities, ", ")
+						}
+
```
another-rex, line 119:
```suggestion
```
I don't think this is correct, it'll append a new line to just a comma when you do a join on line 135. I think you just need to add `,`  to line 135

internal/output/table.go
```diff
@@ -100,6 +104,36 @@ func tableBuilderInner(vulnResult *models.VulnerabilityResults, addStyling bool,
...
+							numericSeverity, _ := v3_metric.NewBase().Decode(severity.Score)
+							outputSeverity = fmt.Sprintf("%v", numericSeverity.Score())
+						default:
+							outputSeverity = severity.Score
+						}
+
+						outputSeverities = append(outputSeverities, outputSeverity)
+					}
+				}
+				outputRow = append(outputRow, strings.Join(outputSeverities, "\n"))
```
another-rex, line 135:
```suggestion
				outputRow = append(outputRow, strings.Join(outputSeverities, ",\n"))
```
See other comment

another-rex (commented):
Thanks! Just check if the final comments are correct then we should be good to merge!

internal/output/table.go
giovanni-bozzano: Right, thanks for the suggestion! Changed it in the latest commit.

hayleycd:
So what determines if the severity score is CVSS v2 or CVSS v3? (Clarifying so I can update the docs)

another-rex:
Inside each OSV entry the severity looks like this:
```
  "severity": [
    {
      "type": "CVSS_V3",
      "score": "CVSS:3.1/AV:N/AC:H/PR:L/UI:R/S:U/C:H/I:H/A:H"
    }
  ],
```

The type field shows what version is used, currently the schema defines only two options, v3 and v2: https://ossf.github.io/osv-schema/#severity-field

As for why you would choose one over the other I think it's just newer advisories use cvss v3, and older ones use v2.

another-rex (approved):

hayleycd:
> 

So there is nothing specific to this code, it will just pull the cvss (whether version 2 or 3) from the OSV entry?

another-rex:
Yep

PR #417, opened 2023-06-20
Update the model to better match schema and add YAML tags.

calebbrown:
- YAML tags added to support YAML serialization
- omitempty added to optional fields
- Package changed to a pointer in Affected
- Bitnami added as an ecosystem

This PR is primarily to support the replacement of the same structs in google/osv.dev/vulnfeeds/vulns.


pkg/models/vulnerability.go
```diff
@@ -24,78 +24,79 @@ type Package struct {
...
 //
 // See: https://ossf.github.io/osv-schema/#affected-fields
 type Affected struct {
-	Package           Package                `json:"package"`
-	Severity          []Severity             `json:"severity,omitempty"`
-	Ranges            []Range                `json:"ranges,omitempty"`
-	Versions          []string               `json:"versions,omitempty"`
-	DatabaseSpecific  map[string]interface{} `json:"database_specific,omitempty"`
-	EcosystemSpecific map[string]interface{} `json:"ecosystem_specific,omitempty"`
+	Package           *Package               `json:"package,omitempty" yaml:"package,omitempty"`
```
oliverchang, line 57: Is this a required change? It may break things subtlely for some users expecting copy semantics instead?

oliverchang (commented):
thanks Caleb!

pkg/models/vulnerability.go
calebbrown: I can change it back. It does change the output to always include the package, even if it has been omitted.

The workaround for this would be to re-add `omitempty` to `name` and `ecosystem` in `Package`

calebbrown: Public dependents appear to be only Guac and Scorecard.

calebbrown: I guess the question is whether this is important to get right early even if a few dependents break, or whether it should be done later with a v2 release?

oliverchang: Let's just maintain compatibility for now, and re-add omitempty to `name` and `ecosystem`.

calebbrown: An alternative would be to add a custom MarshalYAML/MarshalJSON that uses reflection to remove the field.

oliverchang: Probably a bit of an overkill at this point. Is there much benefit to not having `omitempty` on the `name` and `ecosystem` fields?

oliverchang: Also I may have a bit of an misunderstanding about how this all works in Go here. If we just revert back to what we had before (with `omitempty` on the fields within the `Package`), would we just end up with:

```
"package": {}
```

Or would `package` itself be omitted too?

oliverchang (approved):
thank you for fixing the serialization issue!!

PR #420, opened 2023-06-21
Github action

### LINKED ISSUE #216: Add support for SARIF output

Blocked by #217

another-rex:
This PR features:

- Refactors the format flag's internal logic so that we can don't need to repeat the format types so much, and we can test when we add a new format entry if we forgot anything.
- Adds a new format "sarif", which returns a SARIF report (closes #216 )
- Adds a Github Action `action.yaml` and it's specialized dockerfile `action.dockerfile`. This docker image runs a bash script wrapping osv-scanner, first by preprocessing the input so the last argument will be split by new line, allowing the workflow user to pass in multiple directories/files they wish to scan. The script also changes exit codes 127 and 128 to 0 as they contain errors that the user can't really do anything about.
- Adds a workflow using this new github action for this repo
- Sorts the grouped ID output.

Example of what workflow sarif output looks like:
image


pkg/reporter/sarif_reporter.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/osv-scanner/internal/output"
+	"github.com/google/osv-scanner/pkg/models"
+	"github.com/owenrumney/go-sarif/v2/sarif"
```
oliverchang, line 12: Are there any alternatives to using this dependency? Is this a significant amount of code to be writing ourselves? 

e.g. Scorecard seems to have implemented their own sarif support: https://github.com/ossf/scorecard/blob/eb941e396499e64e7b83d30e54c0a462c93f61fb/pkg/sarif.go#L155

pkg/reporter/format.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	switch format {
+	case "json":
+		return NewJSONReporter(stdout, stderr), nil
+	case "table":
+		return NewTableReporter(stdout, stderr, false, width), nil
+	case "markdown":
+		return NewTableReporter(stdout, stderr, true, width), nil
+	case "report":
```
oliverchang, line 36: Why not just "sarif" to be clearer?

cmd/osv-scanner/main.go
```diff
@@ -68,21 +71,22 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+					return fmt.Errorf("unsupported output format \"%s\" - must be one of: %s", s, strings.Join(reporter.Format(), ", "))
 				},
 			},
 			&cli.BoolFlag{
 				Name:  "json",
 				Usage: "sets output to json (deprecated, use --format json instead)",
 			},
+			&cli.StringFlag{
+				Name:      "output",
+				Usage:     "save result output to file",
```
oliverchang, line 87: nit: to be more clear, something like "save the result to the given file path" ?

exit_code_redirect.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Get the total number of arguments
+total_args=$#
```
oliverchang, line 4: Can you add more comments to explain that this is working around a GitHub actions limitation with passing array inputs?

cmd/osv-scanner/main.go
```diff
@@ -113,15 +117,15 @@ func run(args []string, stdout, stderr io.Writer) int {
...
-			switch format {
-			case "json":
-				r = reporter.NewJSONReporter(stdout, stderr)
-			case "table":
-				r = reporter.NewTableReporter(stdout, stderr, false)
-			case "markdown":
-				r = reporter.NewTableReporter(stdout, stderr, true)
-			default:
-				return fmt.Errorf("%v is not a valid format", format)
+			outputPath := context.String("output")
```
oliverchang, line 120: Any reason to create an in-memory buffer instead of directly writing to the file ?

oliverchang (commented):
nice!

pkg/reporter/format.go
another-rex: Yep, done

exit_code_redirect.sh
another-rex: Yep

cmd/osv-scanner/main.go
another-rex: Because this way it only writes it to file when osv-scanner succeeds, if it's writing directly to file it'll always create a file, but upon an error (e.g. the file it's pointed to scan doesn't exit, the flags passed in are wrong...etc) it'll exit early and nothing will be written to it. 

Though now thinking about it, it might be surprising behavior, maybe it should just create an empty file in those cases.

another-rex:
- Added a test for the sarif output
- Removed markdowntext.go
- Refactored Sarif reporter so that most of the logic is in the internal output library

pkg/reporter/sarif_reporter.go
another-rex: The scorecard implementation is 700+ lines and 800+ lines of tests. We can try vendoring that implementation in?

another-rex: Discussed offline, seems like the transitive dependencies are relatively minimal so let's just stick with using this library.

cmd/osv-scanner/main.go
oliverchang: Yeah, creating an empty file seems like reasonable behaviour. Would we still need this buffer in that case? 

We could also alternatively just clear the file on an error to avoid this indirection with the buffer in that case?

.github/workflows/osv-scanner.yml
```diff
@@ -0,0 +1,73 @@
...
+          persist-credentials: false
+
+      - name: "Run scanner"
+        uses: ./ # Uses ./action.yaml
+        with:
+          results-format: sarif
+          results-file: results.sarif
+          to-scan: |-
+            ./
+            ./cmd/osv-scanner/fixtures/locks-many/
```
oliverchang, line 55: is the second arg still necessary if we're doing recursive scans?

.github/workflows/osv-scanner.yml
```diff
@@ -0,0 +1,73 @@
...
+name: osv-scanner
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  merge_group:
+    branches: [ main ]
```
oliverchang, line 24: Should we make this a reusable workflow to let users easily use this? https://github.blog/2022-02-10-using-reusable-workflows-github-actions/

Disclaimer: I haven't looked too deeply into this. There's also composite actions which is something different.

internal/output/sarif.go
```diff
@@ -0,0 +1,118 @@
...
+				pv.Package.Version,
+				strings.Join(fixedVersions, "\n")})
+		}
+	}
+
+	return remediationTable
+}
+
+// PrintSarifReport prints SARIF output to outputWriter
+func PrintSarifReport(vulnResult *models.VulnerabilityResults, outputWriter io.Writer) error {
```
oliverchang, line 64: nit: PrintSARIFReport to respect https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/models/results.go
```diff
@@ -82,6 +85,27 @@ func (groupInfo *GroupInfo) IsCalled() bool {
 	return false
 }
 
+func (groupInfo *GroupInfo) IndexString() string {
+	// Assumes IDs is sorted
+	return strings.Join(groupInfo.IDs, ",")
+}
+
+// IsFixed returns a list of fixed versions, or an empty slice if no fixed version is available
+func (vuln *Vulnerability) FixedVersions() []string {
```
oliverchang, line 94: Should we be matching on the exact package name? Given that an OSV can have multiple packages.

another-rex: Good point, fixed now and I added a unit test to the grouping function that uses this function.

PR #423, opened 2023-06-27
test: make `models` tests their own package

G-Rath:
This reduces the diff with #183 a bit, and is good practice

PR #424, opened 2023-06-27
Add withdrawn and fix time serialization to conform to the schema.

calebbrown:
- withdrawn timestamp does not exist in the model - add it to the model.
- fix the timestamp serialization to ensure published and withdrawn are *only* present when the value is not zero
- fix timestamp serialization to use RFC3339 for JSON (it uses RFC3339Nano by default)
- fix the timestamps to be normalized to UTC as per the schema

### DIFF
--- a/pkg/models/vulnerability.go
+++ b/pkg/models/vulnerability.go
@@ -148,7 +148,7 @@ func (v Vulnerability) MarshalJSON() ([]byte, error) {
 	return json.Marshal(raw)
 }
 
-// MarshalJSON implements the json.Marshaler interface.
+// MarshalYAML implements the yaml.Marshaler interface.
 //
 // This method ensures times all times are formated correctly.
 func (v Vulnerability) MarshalYAML() (interface{}, error) {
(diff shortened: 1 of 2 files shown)


G-Rath:
Note this will conflict with #423

calebbrown:
> Note this will conflict with #423

Thanks! Fixed :)

pkg/models/vulnerability.go
```diff
@@ -73,7 +73,7 @@ type Affected struct {
 func (a Affected) MarshalJSON() ([]byte, error) {
 	type rawAffected Affected // alias Affected to avoid recursion during Marshal
 	type wrapper struct {
-		Package *Package `json:"package,omitempty" yaml:"package,omitempty"`
+		Package *Package `json:"package,omitempty"`
```
oliverchang, line 76: Is this no longer needed?

oliverchang (approved):
thanks Caleb!

pkg/models/vulnerability.go
calebbrown: In `MarshalJSON` it is unused and falls back to the tag in the main struct

PR #431, opened 2023-07-02
update govulncheck integration

julieqiu:


internal/sourceanalysis/go.go
```diff
@@ -39,14 +50,18 @@ func matchAnalysisWithPackageVulns(pkgs []models.PackageVulns, gvcResByVulnID ma
...
-				gvcVuln, ok := gvcResByVulnID[vulnID]
+				moduleToCalled, ok := idToModuleToCalled[vulnID]
 				if !ok { // If vulnerability not found, check if it contain any source information
 					fillNotImportedAnalysisInfo(vulnsByID, vulnID, pv, analysis)
 					continue
 				}
 				// Module list is unlikely to be very big, linear search is fine
-				containsModule := slices.ContainsFunc(gvcVuln.Modules, func(module *govulncheck.Module) bool {
-					return module.Path == pv.Package.Name
+				var modulePaths []string
```
another-rex, line 59: This part can be replaced a direct map key check, something like:
```
called, hasModule := moduleToCalled[pv.Package.Name];
(*analysis)[vulnID] = models.AnalysisInfo{
	Called: hasModule && called,
}
```

another-rex (commented):
Thanks! Implementation looks good!, can you move the deleted test-project fixtures into sourceanalysis so we can still have the integration test?

internal/sourceanalysis/go.go
julieqiu: Nice catch! I left out the `hasModule` bit, since we want to set `Called` to `false` regardless of if `pv.Package.Name` is not in the map, or is present but has a value of `false`

julieqiu:
> Thanks! Implementation looks good!, can you move the deleted test-project fixtures into sourceanalysis so we can still have the integration test?

Added

another-rex (commented):
LGTM!

another-rex (approved):

PR #432, opened 2023-07-05
Recreated Github Action PR 

### LINKED ISSUE #57: Build GitHub actions for running osv-scanner

### LINKED ISSUE #216: Add support for SARIF output

another-rex:
This PR features:

- Refactors the format flag's internal logic so that we can don't need to repeat the format types so much, and we can test when we add a new format entry if we forgot anything.
- Adds a new format "sarif", which returns a SARIF report (closes #216 )
- Adds a Github Action `action.yaml` and it's specialized dockerfile `action.dockerfile`. This docker image runs a bash script wrapping osv-scanner, first by preprocessing the input so the last argument will be split by new line, allowing the workflow user to pass in multiple directories/files they wish to scan. The script also changes exit codes 127 and 128 to 0 as they contain errors that the user can't really do anything about.
- Adds two reusable workflows using this new github action for this repo
  - Reusable PR workflow, for using to check if PRs introduce new vulnerabilities.
  - Reusable Scheduled workflow, for use to regularly check for new vulns applying to your existing vulns.
- Adds an experimental flag: `--experimental-diff`, which will only output the difference between a previous run and this run of the osv-scanner. This is for use in the PR workflow. 
- Sorts the grouped ID output.

Closes #57 

Currently the reusable workflow has to point to a specific action which cannot be relative (otherwise it would point to the wrong action when reused in another repo). This means right now it's pointed to this fork/branch instead of the master branch, this will need to be updated once this PR is merged.

Example of what workflow sarif output looks like:
image

Here is an example of the PR reusable workflow working:

https://github.com/another-rex/scorecard-check-osv-e2e/pull/1

That PR adds an additional vulnerability, which causes it to fail. You can see that only the new vuln is showing up in the code scanning report: https://github.com/another-rex/scorecard-check-osv-e2e/security/code-scanning/1

TODO after this PR is merged:
- Change links that point to this PR branch to point to main (and/or a tagged commit of main)
- Add support for annotations
- Add documentation (this is for later, as we want to dogfood it in our own repos first before broadcasting this widely)


another-rex:
Here is an example of the PR reusable workflow working:

https://github.com/another-rex/scorecard-check-osv-e2e/pull/1

That PR adds an additional vulnerability, which causes it to fail. You can see that only the new vuln is showing up in the code scanning report: https://github.com/another-rex/scorecard-check-osv-e2e/security/code-scanning/1

action.yml
```diff
@@ -3,19 +3,31 @@ name: 'osv-scanner'
...
-    description: 'Directory to scan'
-    required: true
-    default: '/github/workspace'
-  version:
-    description: 'osv-scanner version to use'
+    description: 'Directories to scan'
+    required: false
+    default: "./"
+  results-file:
+    description: 'OUTPUT: Path to file to store results'
```
oliverchang, line 10: nit: just "Output path" to be consistent with line 13?

action.yml
```diff
@@ -3,19 +3,31 @@ name: 'osv-scanner'
...
     required: true
-  arch:
-    description: 'osv-scanner architecture'
+  results-format:
+    description: 'Output result format'
+    required: false
+    default: 'sarif'
+  recursive-search:
+    description: 'Recursively scan though subdirectories'
+    required: false
```
oliverchang, line 18: is `required: false` required if there is a `default` ?

cmd/osv-scanner/main.go
```diff
@@ -147,13 +157,26 @@ func run(args []string, stdout, stderr io.Writer) int {
 				return fmt.Errorf("failed to write output: %w", errPrint)
 			}
 
+			if outputPath != "" {
+				file, err := os.Create(outputPath)
```
oliverchang, line 161: Can we just use `https://pkg.go.dev/os#WriteFile` here?

cmd/osv-scanner/main.go
```diff
@@ -147,13 +157,26 @@ func run(args []string, stdout, stderr io.Writer) int {
 				return fmt.Errorf("failed to write output: %w", errPrint)
 			}
 
+			if outputPath != "" {
```
oliverchang, line 160: Can you add a comment for why we do this? I.e. we only want to create a file on success?

pkg/reporter/format.go
```diff
@@ -0,0 +1,41 @@
...
+	"golang.org/x/term"
+)
+
+var format = []string{"table", "json", "markdown", "sarif"}
+
+func Format() []string {
+	return format
+}
+
+func GetReporter(format string, stdout, stderr io.Writer, fileOutput bool) (Reporter, error) {
```
oliverchang, line 17: nit: avoid stutter here by just calling this `Get`. i.e. `reporter.Get`. Or better just `reporter.New`.

internal/output/sarif.go
```diff
@@ -0,0 +1,122 @@
...
+	flattened := vulnResult.Flatten()
+
+	groupFixedVersions := GroupFixedVersions(flattened)
+	workingDir, workingDirErr := os.Getwd()
+
+	for _, source := range vulnResult.Results {
+		// TODO: Support docker images
+
+		var artifactPath string
+		if workingDirErr == nil {
```
oliverchang, line 86: When can `workingDirErr` be non-nill?

cmd/osv-scanner/main.go
```diff
@@ -99,6 +103,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Usage: "attempt call analysis on code to detect only active vulnerabilities",
 				Value: false,
 			},
+			&cli.StringFlag{
+				Name:      "experimental-diff",
```
oliverchang, line 107: Fine as an experimental flag for now, but should we consider breaking this out into its own binary? It seems like something that would only ever be used in the CI/PR scanning case.

pkg/models/results.go
```diff
@@ -82,6 +85,29 @@ func (groupInfo *GroupInfo) IsCalled() bool {
 	return false
 }
 
+func (groupInfo *GroupInfo) IndexString() string {
+	// Assumes IDs is sorted
+	return strings.Join(groupInfo.IDs, ",")
+}
+
+// FixedVersions returns a list of fixed versions, or an empty slice if no fixed version is available
```
oliverchang, line 93: Comment seems wrong? We're returning a map right?

pkg/models/results.go
```diff
@@ -82,6 +85,29 @@ func (groupInfo *GroupInfo) IsCalled() bool {
...
+func (v *Vulnerability) FixedVersions() map[Package][]string {
+	output := map[Package][]string{}
+	for _, a := range v.Affected {
+		cleanedPackaged := a.Package
```
oliverchang, line 97: s/cleanedPackaged/packageKey/ to be clearer?

pkg/reporter/format.go
oliverchang, line 17: `fileOutput` is a bit of a weird interface. Can we have this take in a `terminalWidth` instead, and just have the caller provide it?

pkg/osvscanner/vulnerability_result.go
```diff
@@ -74,3 +75,58 @@ func groupResponseBySource(r reporter.Reporter, query osv.BatchedQuery, resp *os
...
+
+// diffVulnerabilityResults will return any new vulnerabilities that are in `newRes`
+// which is not present in `oldRes`, but not the reverse.
+//
+// Current implementation is O(n^2) on the number of vulns, but can be reduced to linear time
+func diffVulnerabilityResults(oldRes, newRes models.VulnerabilityResults) models.VulnerabilityResults {
+	result := models.VulnerabilityResults{}
+	for _, ps := range newRes.Results {
+		sourceIdx := slices.IndexFunc(oldRes.Results, func(elem models.PackageSource) bool { return elem.Source == ps.Source })
+		if sourceIdx == -1 {
```
oliverchang, line 87: nit: add more comments for the different cases here.

e.g. here: "Entire source finding is new"  or some variation.

pkg/osvscanner/vulnerability_result.go
```diff
@@ -74,3 +75,58 @@ func groupResponseBySource(r reporter.Reporter, query osv.BatchedQuery, resp *os
...
+		}
+		result.Results = append(result.Results, models.PackageSource{
+			Source: ps.Source,
+		})
+		resultPS := &result.Results[len(result.Results)-1]
+		for _, pv := range ps.Packages {
+			pkgs := oldRes.Results[sourceIdx].Packages
+			pkgIdx := slices.IndexFunc(pkgs, func(elem models.PackageVulns) bool { return elem.Package == pv.Package })
+			if pkgIdx == -1 {
+				resultPS.Packages = append(resultPS.Packages, pv)
```
oliverchang, line 99: Add a comment: "Newly introduced package vulnerability".

oliverchang (commented):
nice! some minor comments/questions.

.github/workflows/osv-scanner-reusable-pr.yml
```diff
@@ -0,0 +1,85 @@
...
+      #   run: |
+      #     echo ${{ steps.changed-files.outputs.all_changed_files }}
+      #     echo ------------
+      #     for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+      #       echo "$file was changed"
+      #     done
+      - name: "Checkout target branch"
+        run: git checkout $GITHUB_BASE_REF
+      - name: "Run scanner"
+        uses: another-rex/osv-scanner@markdown-output
```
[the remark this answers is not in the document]
another-rex: Yes we do, but only after this is already merged, otherwise I think it will prevent merging as the action will fail. Maybe a better way is to merge only the code part and have a separate PR for the workflows part.

action.yml
another-rex: I don't think it's required, but I think it's clearer to explicitly state it for all for all inputs.

G-Rath: Random note because frankly its weird imo: [`required` isn't actually enforced "yet"](https://github.com/actions/runner/issues/1070

internal/output/sarif.go
another-rex: Not too sure, I think this might happen when the working directory no longer exists. It depends on the OS though.

cmd/osv-scanner/main.go
another-rex: That's a good point, happy to move it to another binary.

pkg/osvscanner/vulnerability_result.go
another-rex: Done in: 5f1fd7a471c91a6065a92814389ab6399a399bee

pkg/osvscanner/vulnerability_result.go
another-rex: Done in: 5f1fd7a471c91a6065a92814389ab6399a399bee

pkg/reporter/format.go
another-rex: Done in: 5f1fd7a471c91a6065a92814389ab6399a399bee

pkg/models/results.go
another-rex: Correct it in: 5f1fd7a471c91a6065a92814389ab6399a399bee

cmd/osv-scanner/main.go
another-rex: Ah this is circumvented by resolving this: https://github.com/google/osv-scanner/pull/420#discussion_r1244619219

cmd/osv-scanner/main.go
another-rex: Updated it to now write directly to output file as described here: https://github.com/google/osv-scanner/pull/420#discussion_r1244619219

oliverchang (approved):

cmd/osv-scanner/fixtures/locks-many/osv-scanner.toml
```diff
@@ -0,0 +1,4 @@
+[[IgnoredVulns]]
```
G-Rath, line 1: Having this here arguably waters the test, since the test cases don't expect an ignore file - is there a reason why you've using recursive on this repo anyway?

cmd/osv-scanner/main_test.go
```diff
@@ -290,6 +294,101 @@ func TestRun(t *testing.T) {
 				Scanned %%/fixtures/locks-many/composer.lock file and found 1 packages
 			`,
 		},
+		// output format: sarif
+		{
+			name:         "Empty sarif output",
+			args:         []string{"", "--format", "sarif", "./fixtures/locks-many/composer.lock"},
+			wantExitCode: 0,
+			wantStdout: `
+            {
```
G-Rath, line 303: nit: all of these should be intended a few levels less

another-rex: Just changed the indentation from spaces to tabs like the rest of the code, see if this looks correct now?

oliverchang (commented):
Is there anything in particular blocking this PR from merging at this point?

PR #439, opened 2023-07-10
Fix PURLToPackage function and move it

another-rex:
Turns out our `PURLToPackage` function was returning incorrect results for ecosystems that contain a namespace like golang, the returned result was simply missing the full namespace (github.com/author/...). When adding the namespace, there's also some exceptions with some ecosystems (e.g. Maven uses `:`, debian and alpine repeats their name in their namespace, etc).

This also moves the `PURLToPackage` to the `models` package instead of `osvscanner`, deprecating the existing one in `osvscanner` because:
- Makes more sense, it actually has nothing to do with the scanner itself, but is converting between PURLs and a structure under `model`.
- Prevents cyclic imports when being used elsewhere (in the offline scanning PR, and in the upcoming PURL parsing PR that I'm currently working on) 

Also added additional tests to clarify behavior and prevent regressions in the future.

### DIFF
--- a/pkg/models/purl_to_package.go
+++ b/pkg/models/purl_to_package.go
@@ -32,7 +32,14 @@ func PURLToPackage(purl string) (PackageInfo, error) {
 	// PackageInfo expects the full namespace in the name for ecosystems that specify it.
 	name := parsedPURL.Name
 	if parsedPURL.Namespace != "" {
-		name = parsedPURL.Namespace + "/" + parsedPURL.Name
+		if ecosystem == string(EcosystemMaven) { // Maven uses : to separate namespace and package
+			name = parsedPURL.Namespace + ":" + parsedPURL.Name
+		} else if ecosystem == string(EcosystemDebian) || ecosystem == string(EcosystemAlpine) {
+			// Debian and Alpine repeats their namespace in PURL, so don't add it to the name
+			name = parsedPURL.Name
+		} else {
+			name = parsedPURL.Namespace + "/" + parsedPURL.Name
+		}
 	}
 
 	return PackageInfo{
--- a/pkg/models/purl_to_package.go
+++ b/pkg/models/purl_to_package.go
@@ -1,0 +1,43 @@
+package models
+
+import (
+	"github.com/package-url/packageurl-go"
+)
+
+var purlEcosystems = map[string]string{
+	"cargo":    "crates.io",
+	"deb":      "Debian",
+	"hex":      "Hex",
+	"golang":   "Go",
+	"maven":    "Maven",
+	"nuget":    "NuGet",
+	"npm":      "npm",
+	"composer": "Packagist",
+	"generic":  "OSS-Fuzz",
+	"pypi":     "PyPI",
+	"gem":      "RubyGems",
+}
+
+// PURLToPackage converts a Package URL string to models.PackageInfo
+func PURLToPackage(purl string) (PackageInfo, error) {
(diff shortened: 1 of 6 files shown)


pkg/models/purl_to_package.go
```diff
@@ -0,0 +1,50 @@
...
+	"hex":      "Hex",
+	"golang":   "Go",
+	"maven":    "Maven",
+	"nuget":    "NuGet",
+	"npm":      "npm",
+	"composer": "Packagist",
+	"generic":  "OSS-Fuzz",
+	"pypi":     "PyPI",
+	"gem":      "RubyGems",
+}
```
G-Rath, line 19: nit: could we make this a map of `models.Ecosystem` and start using the ecosystem constants?

(I think that would be ideal, but could result a lot more casting so might not be worth it)

pkg/models/purl_to_package.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	ecosystem := purlEcosystems[parsedPURL.Type]
+	if ecosystem == "" {
+		ecosystem = parsedPURL.Type
+	}
+
+	// PackageInfo expects the full namespace in the name for ecosystems that specify it.
+	name := parsedPURL.Name
+	if parsedPURL.Namespace != "" {
+		if ecosystem == string(EcosystemMaven) { // Maven uses : to separate namespace and package
```
G-Rath, line 35: nit: I feel like this would look better as a switch

pkg/models/purl_to_package_test.go
```diff
@@ -0,0 +1,86 @@
+package models
```
G-Rath, line 1: This should be within `models_test`

G-Rath (changes requested):

pkg/models/purl_to_package.go
another-rex: Changed this to use models.Ecosystem. I haven't changed `PackageInfo`'s `Ecosystem` field to be Ecosystem since that'll be a breaking change I believe.

G-Rath: hmm ok I think you should include that as part of considering what to do going forward about ecosystem within the scanner per our meeting earlier

oliverchang (approved):

pkg/models/purl_to_package.go
```diff
@@ -0,0 +1,52 @@
...
+	if parsedPURL.Namespace != "" {
+		switch ecosystem { //nolint:exhaustive
+		case EcosystemMaven:
+			// Maven uses : to separate namespace and package
+			name = parsedPURL.Namespace + ":" + parsedPURL.Name
+		case EcosystemDebian, EcosystemAlpine:
+			// Debian and Alpine repeats their namespace in PURL, so don't add it to the name
+			name = parsedPURL.Name
+		default:
+			name = parsedPURL.Namespace + "/" + parsedPURL.Name
```
oliverchang, line 43: Is this definitely the case in all other ecosystems?

another-rex: I'm pretty sure this is the case for all other ecosystems in osv right now, going by the package_to_purl function in python:

https://github.com/google/osv.dev/blob/24e724430a0c2d6ef23eae7ae7007e7d2771b3d7/osv/purl_helpers.py#L42

If we add more linux distros we'll also need to remove the namespace for them as well.

G-Rath (approved):

PR #441, opened 2023-07-12
Optimize Dockerfile and add .dockerignore

theinfosecguy:
Changes include:

- Removed the `mkdir` command as the `WORKDIR` instruction automatically creates the folder if it doesn't already exist.

- Added a .dockerignore file to prevent unnecessary files (such as README, tests, Dockerfile itself, etc.) from being included in the Docker context, thereby speeding up the build process.

- Combined two `RUN` instructions into one in order to reduce the number of layers in the final image.


another-rex (commented):
Thanks! The dockerfile updates look good, though I don't think the .dockerignore is necessary, as it might hinder us wanting to run tests in docker in the future, while not significantly speeding up the build process (there's not that many files being ignored).

Can you remove the .dockerignore?

theinfosecguy:
> Can you remove the .dockerignore?

Removed the `.dockerignore` file.

another-rex (approved):
LGTM! Thanks!

theinfosecguy:
@another-rex - can we merge this?

Dockerfile
```diff
@@ -12,27 +12,21 @@
...
-RUN git config --global --add safe.directory '*'
+RUN apk --no-cache add ca-certificates git && \
+    git config --global --add safe.directory '*'
 
 WORKDIR /root/
-COPY --from=0 /src/osv-scanner ./
+COPY --from=builder /src/osv-scanner .
 
-ENTRYPOINT ["/root/osv-scanner"]
+ENTRYPOINT ["./osv-scanner"]
```
another-rex, line 32:
```suggestion
ENTRYPOINT ["/root/osv-scanner"]
```

PR #451, opened 2023-07-26
feat: support `io.Reader` based parsers

### LINKED ISSUE #176: Add `io.Reader` variants to `lockfile` package

G-Rath:
Resolves #176

### DIFF
--- a/pkg/lockfile/extractor.go
+++ b/pkg/lockfile/extractor.go
@@ -62,3 +62,15 @@ func OpenLocalDepFile(path string) (NestedDepFile, error) {
 
 var _ DepFile = LocalFile{}
 var _ NestedDepFile = LocalFile{}
+
+func extractFromFile(pathToLockfile string, extractor Extractor) ([]PackageDetails, error) {
+	f, err := OpenLocalDepFile(pathToLockfile)
+
+	if err != nil {
+		return []PackageDetails{}, err
+	}
+
+	defer f.Close()
+
+	return extractor.Extract(f)
+}
--- a/pkg/lockfile/extractor.go
+++ b/pkg/lockfile/extractor.go
@@ -1,0 +1,64 @@
+package lockfile
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+)
+
+// DepFile is an abstraction for a file that has been opened for extraction,
+// and that knows how to open other DepFiles relative to itself.
+type DepFile interface {
+	io.Reader
+
+	// Open opens an NestedDepFile based on the path of the
+	// current DepFile if the provided path is relative.
+	//
+	// If the path is an absolute path, then it is opened absolutely.
+	Open(string) (NestedDepFile, error)
+
+	Path() string
+}
+
+// NestedDepFile is an abstraction for a file that has been opened while extracting another file,
+// and would need to be closed.
+type NestedDepFile interface {
+	io.Closer
+	DepFile
+}
+
+type Extractor interface {
+	// ShouldExtract checks if the Extractor should be used for the given path.
+	ShouldExtract(path string) bool
+	Extract(f DepFile) ([]PackageDetails, error)
+}
+
+// A LocalFile represents a file that exists on the local filesystem.
+type LocalFile struct {
+	io.ReadCloser
+
+	path string
+}
+
+func (f LocalFile) Open(path string) (NestedDepFile, error) {
+	if filepath.IsAbs(path) {
+		return OpenLocalDepFile(path)
+	}
+
+	return OpenLocalDepFile(filepath.Join(filepath.Dir(f.path), path))
+}
+
+func (f LocalFile) Path() string { return f.path }
+
+func OpenLocalDepFile(path string) (NestedDepFile, error) {
+	r, err := os.Open(path)
+
+	if err != nil {
+		return LocalFile{}, err
+	}
+
+	return LocalFile{r, path}, nil
+}
+
+var _ DepFile = LocalFile{}
+var _ NestedDepFile = LocalFile{}
--- a/pkg/lockfile/extractor.go
+++ b/pkg/lockfile/extractor.go
@@ -1,10 +1,13 @@
 package lockfile
 
 import (
+	"errors"
 	"io"
 	"os"
 	"path/filepath"
 )
+
+var ErrNotSupported = errors.New("this file does not support opening files")
 
 // DepFile is an abstraction for a file that has been opened for extraction,
 // and that knows how to open other DepFiles relative to itself.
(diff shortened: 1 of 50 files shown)


G-Rath:
@oliverchang @another-rex I think we need to discuss how you want to handle transitioning from "cli" to "library" given that everything there is currently "parse-as"; I think we could either keep it as is, go for a hybrid approach, or rename everything non-breaking to "extract-as".

The part where the code actually suffers currently is in `FindExtractor`, because the Gradle extractor has two different files but the `Extractor` interface doesn't support convoying that back - for now I've just duplicated the `ShouldExtract` into the `FindExtractor` code.

cc @robotdana

another-rex:
> I think we need to discuss how you want to handle transitioning from "cli" to "library" given that everything there is currently "parse-as"; I think we could either keep it as is, go for a hybrid approach, or rename everything non-breaking to "extract-as".

I think the CLI actually doesn't mention "parse" at all right? To specify a parsing method we are using the `Cargo.lock:/path/to/Cargo.lock`, and taking a look at osvscanner.go, the `parse` and `parseAs` mentions are argument names are all private variables / argument names.

I think we can go ahead and rename non breaking things to `extract`, but haven't looked into it enough to see how practical that is. 

> The part where the code actually suffers currently is in FindExtractor, because the Gradle extractor has two different files but the Extractor interface doesn't support convoying that back - for now I've just duplicated the ShouldExtract into the FindExtractor code.

Hmm I'm not clear on what you mean here. I had a look at `FindExtractor` in this PR, and I don't think `stabilizeFindExtractorName` is necessary, since both gradle files are parsed the same, we can just decide on a parser/extractor name and return that, unless I am missing something?

G-Rath:
> I think the CLI actually doesn't mention "parse" at all right?

It has "parse" in some of its outputs, and the `Lockfile` struct has the `ParsedAs` property.

> we can just decide on a parser/extractor name and return that, unless I am missing something?

Nope, what you've said should be fine - the main potential issue I can imagine is because the extractors are named after their files, someone might assume `buildscript-gradle.lockfile` is not supported if its not on the list (or that they pass it explicitly and it gets rejected) (I'm assuming `gradle.lockfile` will be the preferred name since its shorter).

---

I'm not saying these are big decisions, but they're ones I think depend somewhat on the direction _you_ want to take the scanner 🙂

pkg/lockfile/parse.go
```diff
@@ -158,3 +158,15 @@ func Parse(pathToLockfile string, parseAs string) (Lockfile, error) {
 		Packages: packages,
 	}, err
 }
+
+func parseFile(pathToLockfile string, extractor Extractor) ([]PackageDetails, error) {
```
another-rex, line 162: nit: rename this to `extractFromFile` and move to the `extractor.go` file

pkg/lockfile/parse-gradle-lock.go
```diff
@@ -54,16 +63,27 @@ func ParseGradleLock(pathToLockfile string) ([]PackageDetails, error) {
...
 	}
 
 	return pkgs, nil
 }
+
+var _ Extractor = GradleLockExtractor{}
+
+//nolint:gochecknoinits
+func init() {
+	registerExtractor("gradle.lockfile", GradleLockExtractor{})
```
another-rex, line 84: Would there be any bad side effects if we register the extractor twice? e.g.:
```
	registerExtractor("gradle.lockfile", GradleLockExtractor{})
	registerExtractor("buildscript-gradle.lockfile", GradleLockExtractor{})
```

another-rex (commented):
Looks pretty good! Just some minor comments.

pkg/lockfile/extractor.go
```diff
@@ -0,0 +1,64 @@
...
+// and would need to be closed.
+type NestedDepFile interface {
+	io.Closer
+	DepFile
+}
+
+type Extractor interface {
+	// ShouldExtract checks if the Extractor should be used for the given path.
+	ShouldExtract(path string) bool
+	Extract(f DepFile) ([]PackageDetails, error)
```
oliverchang, line 33: Could we add
```
var ErrNotSupported = errors.New("this file does not support opening files")
```

per https://github.com/google/osv-scanner/issues/176#issuecomment-1637238126  ?

oliverchang (commented):
nice, thanks @G-Rath !!! looks pretty good overall.

pkg/lockfile/extractor.go
oliverchang: @another-rex

pkg/lockfile/parse-gradle-lock.go
G-Rath: Yes, that causes the original problem with extractor names - see https://github.com/google/osv-scanner/pull/451/commits/c9812149f467861aa09252e7e240181708dea46c

pkg/lockfile/parse-pipenv-lock.go
```diff
@@ -17,19 +17,19 @@ type PipenvLock struct {
 
 const PipenvEcosystem = PipEcosystem
 
-func ParsePipenvLock(pathToLockfile string) ([]PackageDetails, error) {
-	var parsedLockfile *PipenvLock
+type PipenvLockExtractor struct{}
 
-	lockfileContents, err := os.ReadFile(pathToLockfile)
+func (e PipenvLockExtractor) ShouldExtract(path string) bool {
+	return filepath.Base(path) == "Pipfile.lock"
```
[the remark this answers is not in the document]
G-Rath: Practically no because I don't see a strong value in doing that now since it's not like it'll make the code faster or anything, and architecturally I'd argue you're conflating "extractor names" with "file paths"

pkg/lockfile/parse-gradle-lock.go
another-rex: Ah I see what you mean now, the parser name being returned won't be consistent. Let's go with what we have now then (i.e. only registering gradle.lockfile). 

This is consistent with us keeping the extractor name separated from what it can extract.

PR #452, opened 2023-07-27
Rust call analysis

another-rex:
Support rust for call analysis, the current version is fully working for any rust binary or library that can be built with `cargo build`

Part of #476

TODO:
- [x] Figure out what if anything is wrong with the go analysis (Nothing is wrong, added another test for the future)
- [x] Add documentation to rust call analysis
- [x] Add tests
  - [x] extractRlibArchive
  - [x] ar
  - [ ] rustBuildSource integration test

Limitations: #464


internal/sourceanalysis/ar/reader.go
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) 2013 Blake Smith <blakesmith0@gmail.com>
```
oliverchang, line 1: Please move this as we discussed offline.

internal/sourceanalysis/rust.go
```diff
@@ -0,0 +1,275 @@
...
+	"strings"
+
+	"github.com/google/osv-scanner/internal/cachedregexp"
+	"github.com/google/osv-scanner/internal/sourceanalysis/ar"
+	"github.com/google/osv-scanner/pkg/models"
+	"github.com/google/osv-scanner/pkg/reporter"
+	"github.com/ianlancetaylor/demangle"
+)
+
+const RustFlagsEnv = "RUSTFLAGS=-C opt-level=3 -C debuginfo=1"
```
oliverchang, line 23: Can we group these using 

```
const (
  RustFlagsEnv = "..."
  RustLibExtension = "...'
)
```

internal/sourceanalysis/rust.go
```diff
@@ -0,0 +1,275 @@
...
+
+	return resultBinaryPaths, nil
+}
+
+// cleanRustFunctionSymbols takes in demanged rust symbols and makes them fit format of
+// the common function level advisory information
+func cleanRustFunctionSymbols(val string) string {
+	// Used to remove generics from functions and types as they are not included in function calls
+	// in advisories:
+	// E.g.: `smallvec::SmallVec<A>::new` => `smallvec::SmallVec::new`
```
oliverchang, line 262: I wonder if we should start a discussion here if the data can be formatted in a more convenient way to begin with? 

Regex seems potentially very error prone here.

internal/sourceanalysis/rust.go
```diff
@@ -0,0 +1,275 @@
...
+
+	outputDir := filepath.Join(projectBaseDir, "target", "debug")
+	entries, err := os.ReadDir(outputDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read \"%s\" dir: %w", outputDir, err)
+	}
+
+	resultBinaryPaths := []string{}
+	for _, de := range entries {
+		// We only want .d files
```
oliverchang, line 236: Can you please add some context for why we only want .d files? What does .d mean?

internal/sourceanalysis/rust.go
```diff
@@ -0,0 +1,275 @@
...
+					//
+					// "affects": {
+					//     "os": [],
+					//     "functions": [
+					//         "smallvec::SmallVec::grow"
+					//     ],
+					//     "arch": []
+					// }
+					ecosystemAffects, ok := a.EcosystemSpecific["affects"].(map[string]interface{})
+					if !ok {
```
oliverchang, line 65: Can we print an error in these cases? Also wonder if we can define these in a struct and ummarshal them from a map somehow.

internal/sourceanalysis/rust.go
```diff
@@ -0,0 +1,275 @@
...
+					}
+					affectedFunctions, ok := ecosystemAffects["functions"].([]interface{})
+					if !ok {
+						continue
+					}
+					for _, f := range affectedFunctions {
+						if funcName, ok := f.(string); ok {
+							_, called := calls[funcName]
+							// Once one advisory marks this vuln as called, always mark as called
+							isCalledMap[v.ID] = isCalledMap[v.ID] || called
```
oliverchang, line 76: nit: can we just track this as a `map[string]struct{}`, and just set this unconditionally without boolean logic make things slightly simpler?

oliverchang (commented):
nice! can you please add a simple test as well?

internal/sourceanalysis/rust.go
oliverchang: (also, should these be just private consts instead?)

internal/sourceanalysis/rust.go
another-rex: Not sure how an error would work here, some ecosystem specific data will either not exist (most e.g. GHSA advisories), or have a different format (GHSA also has different function info compared to RUSTSEC, but since GHSA just copies it from RUSTSEC, just supporting RUSTSEC format seems fine)

internal/sourceanalysis/rust.go
another-rex: Hmm I think that will work, but I feel that's more confusing than the current version? The current way clearly shows 3 states in `isCalledMap`:
- There is function level vuln info, but it **wasn't** called   (true)
- There is function level vuln info, and it **is** called    (false)
- There is **no** functional level vuln info, so we don't know whether it is called (doesn't exist)

another-rex: Added some additional comments and changed variable names to make this clearer.

internal/sourceanalysis/rust.go
oliverchang: was just thinking on if there are ways we can validate that we're unmarshalling things correctly with the correct types. e.g. detecting if a type (e.g. `functions`) has changed from array to something else unexpectedly. 

Since this is `ecosystem_specific`, this should be specific to and defined by Rust and be consistent. 

Not a big deal, just a minor defensive thing to be adding here. Could just be a TODO.

internal/sourceanalysis/rust.go
```diff
@@ -0,0 +1,283 @@
...
+							Called: called,
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+func parseDWARFData(_ reporter.Reporter, binaryPath string) (map[string]struct{}, error) {
```
oliverchang, line 109: nit: can we make the name more descriptive. e.g. `functionsFromDWARF` 

Also, why do we need the `reporter.Reporter` argument if it's not used?

internal/sourceanalysis/rust.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		file, err := os.ReadFile(filepath.Join(outputDir, de.Name()))
+		if err != nil {
+			return nil, fmt.Errorf("failed to read \"%s\": %w", filepath.Join(outputDir, de.Name()), err)
+		}
+
+		fileSplit := strings.Split(string(file), ": ")
+		if len(fileSplit) != 2 {
+			// TODO: this can probably be fixed with more effort
```
oliverchang, line 256: IS this something strings.SplitN  could help with?

another-rex: Not really, the `.d` format is very simple, where there's two paths separated by a `:<space>`. The problem happens when the path itself contains a `:<space>`. It doesn't seem to get escaped by rust compiler. It's going to be pretty rare though, so I'm not sure if it's worth coming up with a better solution.

docs/output.md
```diff
@@ -65,18 +65,18 @@ osv-scanner --format markdown your/project/dir
 **Raw output:**
 
 ```
-| OSV URL | CVSS | Ecosystem | Package | Version | Source |
-| --- | --- | --- | --- | --- | --- |
-| https://osv.dev/GHSA-c3h9-896r-86jm<br/>https://osv.dev/GO-2021-0053 | 8.6 | Go | github.com/gogo/protobuf | 1.3.1 | ../scorecard-check-osv-e2e/go.mod |
-| https://osv.dev/GHSA-m5pq-gvj9-9vr8<br/>https://osv.dev/RUSTSEC-2022-0013 | 7.5 | crates.io | regex | 1.5.1 | ../scorecard-check-osv-e2e/sub-rust-project/Cargo.lock |
+| OSV URL                                                                   | CVSS | Ecosystem | Package                  | Version | Source                                                 |
```
hayleycd, line 68: So adding space here is helpful, as it prints raw, so it will be easier for the user to read. However, I don't really think you need to update the white space of the tables that get rendered. It isn't wrong, but I don't think it is necessary. I also think it degrades over time. Anyone who makes changes in the future may feel inclined to make the lines "work" but they may or may not do it.

docs/output.md
```diff
@@ -65,18 +65,18 @@ osv-scanner --format markdown your/project/dir
...
+| ------------------------------------------------------------------------- | ---- | --------- | ------------------------ | ------- | ------------------------------------------------------ |
+| https://osv.dev/GHSA-c3h9-896r-86jm<br/>https://osv.dev/GO-2021-0053      | 8.6  | Go        | github.com/gogo/protobuf | 1.3.1   | ../scorecard-check-osv-e2e/go.mod                      |
+| https://osv.dev/GHSA-m5pq-gvj9-9vr8<br/>https://osv.dev/RUSTSEC-2022-0013 | 7.5  | crates.io | regex                    | 1.5.1   | ../scorecard-check-osv-e2e/sub-rust-project/Cargo.lock |
 ```
 
 **Rendered:**
 
-| OSV URL | CVSS | Ecosystem | Package | Version | Source |
```
hayleycd, line 76: So when I saw that there were changes in the output file, I was hoping to see output from the SCAs. How does the Go & Rust output look different from the standard ?

hayleycd (changes requested):
I'd like to see a sample SCA output for the docs.

docs/output.md
another-rex: Hmm... this is automatically added by vscode when saving, I'll have a look at disabling it.

another-rex: Undid this change in the output.md file

docs/output.md
another-rex: Added some examples and JSON comments explaining the output more thoroughly.

oliverchang (approved):
LGTM! 

As discussed offline, let's raise some data formatting (re the remaining open comment on the PR around generics) and consistency issues (e.g. macros) with the upstream RustSec folks. It would be nice to show osv-scanner as an example tool that can leverage the DB to its full extent.

hayleycd (approved):

PR #457, opened 2023-07-31
Fix purl mapping

another-rex:
Parsing PURL to OSV-Scanner ecosystems was not accurate for alpine, and would be not accurate for future conversions when we add additional ecosystems.

This PR adds a test for alpine, and also fixes/future proofs the PURL mapping.

(This should also be a good test for the newly merged osv-scanner action. I haven't added in the osv-scanner.toml ignore file for the new alpine vuln, so it should prevent us from merging)


oliverchang (approved):

G-Rath (approved):

cmd/osv-scanner/fixtures/sbom-insecure/alpine.cdx.xml
```diff
@@ -0,0 +1,604 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
michaelkedar, line 1: this is a comment to test something

PR #460, opened 2023-08-01
Add annotations and osv-scanner table in the Github Action output

another-rex:
Changed osv-diff action/binary to be a more general osv results reporter for github actions. This allows the logic to be in go code rather than having that logic in github action yaml lines. 

This PR adds annotations and osv-scanner table output to the github action in addition to the existing SARIF generation and uploading.

After this is merged, we have to point the resusable workflows back to main again rather than to this PR branch.

To see the annotation, you need to press Load Diff.

image

### DIFF
--- a/internal/output/githubannotation.go
+++ b/internal/output/githubannotation.go
@@ -38,7 +38,8 @@ func PrintGHAnnotationReport(vulnResult *models.VulnerabilityResults, outputWrit
 		// This is required as github action annotations must be on the same terminal line
 		// so we URL encode the new line character
 		renderedTable = strings.ReplaceAll(renderedTable, "\n", "%0A")
-		fmt.Fprintf(outputWriter, "::error file=%s::%s", artifactPath, renderedTable)
+		// Prepend the table with a new line to look nicer in the output
+		fmt.Fprintf(outputWriter, "::error file=%s::%s", artifactPath, "%0A"+renderedTable)
 	}
 
 	return nil
--- a/internal/output/githubannotation.go
+++ b/internal/output/githubannotation.go
@@ -1,0 +1,46 @@
+package output
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/osv-scanner/pkg/models"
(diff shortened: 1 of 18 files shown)


internal/output/githubannotation.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+// PrintGHAnnotationReport prints Github specific annotations to outputWriter
+func PrintGHAnnotationReport(vulnResult *models.VulnerabilityResults, outputWriter io.Writer) error {
+	flattened := vulnResult.Flatten()
+
+	// TODO: Also support last affected
+	groupFixedVersions := GroupFixedVersions(flattened)
+	workingDir, workingDirErr := os.Getwd()
```
oliverchang, line 19: How common is it for us to get a workingDirErr? Would it make sense to just panic instead to make the code simpler if this is extremely rare?

internal/output/table.go
```diff
@@ -100,32 +100,7 @@ func tableBuilderInner(vulnResult *models.VulnerabilityResults, addStyling bool,
...
-							outputSeverity = fmt.Sprintf("%v", numericSeverity.Score())
-						default:
-							outputSeverity = severity.Score
-						}
-
-						outputSeverities = append(outputSeverities, outputSeverity)
-					}
-				}
-				outputRow = append(outputRow, strings.Join(outputSeverities, "\n"))
+				outputRow = append(outputRow, strings.Join(Severities(group, pkg), "\n"))
```
oliverchang, line 103: Thinking more here, should we just be more opinionated here and take the max severity rather than presenting multiple which might be confusing?

oliverchang (approved):

internal/output/table.go
another-rex: Good idea, I think the current implementation is also slightly incorrect in certain circumstances where the CVSS doesn't line up with the advisory that's publishing it

internal/output/githubannotation.go
another-rex: Changed it to panic now

PR #463, opened 2023-08-02
refactor: move experimental flags into their own struct

G-Rath:
This pulls across the experimental struct from #183 which I'm assuming folks are happy with given that PR is approved - I think it is good to land this in its own PR as its technically an allowed breaking change (or "experimental change" if you will), and that way it can be explicitly linked to in changelogs etc without having to understand #183 as much.


another-rex (approved):
Good idea!

PR #475, opened 2023-08-11
Scanner action should pass if the vulnerabilities remain the same

another-rex:
Scanner action should pass if the vulnerabilities and vuln count remain the same, even if the source files have moved.


cmd/osv-reporter/main.go
```diff
@@ -92,7 +92,25 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+				// this actually changes the results in some cases, e.g.
+				//
+				// When a lockfile is moved, `DiffVulnerabilityResults` will report the moved lockfile as having
+				// a new vulnerability if the existing lockfile has a vulnerability. However this check will
+				// report no vulnerabilities. This is desired behavior.
+
+				// TODO: This will need to be not empty when we change osv-scanner to report all packages
+				diffVulns = models.VulnerabilityResults{}
+			} else {
+				// TODO: This will need to contain all scanned packages when we change osv-scanner to report all packages
```
oliverchang, line 111: unrelated question: do we want this to be a separate flag/mode, instead of being done for all runs?

internal/ci/vulnerability_result_diff.go
```diff
@@ -65,3 +65,32 @@ func DiffVulnerabilityResults(oldRes, newRes models.VulnerabilityResults) models
...
+
+	for k, oldVulnCount := range oldResMap {
+		// If the new result has less vulnerabilities than the old result remove the entry from the new result.
+		// `map`'s default value is 0 when empty, and delete also works fine when the entry is empty
+		if newResMap[k] <= oldVulnCount {
+			delete(newResMap, k)
+		}
+	}
+
+	return newResMap
```
oliverchang, line 95: nit: should we just return len(newResMap) here? That's what the comment above the function also implies : that we're returning a count and not a map.

internal/ci/vulnerability_result_diff.go
```diff
@@ -65,3 +65,32 @@ func DiffVulnerabilityResults(oldRes, newRes models.VulnerabilityResults) models
 
 	return result
 }
+
+// DiffVulnerabilityResultsByUniqueVulnCount will return a count of unique vulnerabilities that are in `newRes`
```
oliverchang, line 69: Nit: can we better define "unique" in the comment? It seems a bit confusing to me as is.

Alternatively, we can describe this in terms of "occurrences", which seems closer to what we're actually doing. 

e.g. "DiffVulnResultsByOccurences". "Return a count of new occurrences of vulnerabilities, while ignoring the exact source the vulenrability was found in". 

Also probably worth adding a comment here on the use case this serves: "this prevents us reporting "new" vulnerabilities in a PR when a previously vulnerable file is being moved."

oliverchang (approved):
LGTM with some nits on naming and return values.

cmd/osv-reporter/main.go
another-rex: Yeah,  it might be a good idea, it is a sort of breaking change because we are adding additional packages to the JSON output, even though they don't have vulnerabilities, someone might be relying on the fact that all packages in the JSON has some vulnerabilities, and this will break that (and blow up the size of the JSON output).

internal/ci/vulnerability_result_diff.go
another-rex: Reworded it slightly, I still prefer to return the full map just to make testing a bit more accurate.

PR #484, opened 2023-08-14
Add go stdlib version support

### LINKED ISSUE #453: osv-scanner does not catch Go/stdlib vulnerabilities
Running `govulncheck` on a repo directly is producing the following results:

```
# govulncheck ./...
Using go1.20.4 and govulncheck@v1.0.0 with vulnerability data from https://vuln.go.dev (last modified 2023-07-27 20:09:46 +0000 UTC).

Scanning your code and 484 packages across 79 dependent modules for known vulnerabilities...

Vulnerability #1: GO-2023-1878
    Insufficient sanitization of Host header in net/http
  More info: https://pkg.go.dev/vuln/GO-2023-1878
  Standard library
    Found in: net/http@go1.20.4
    Fixed in: net/http@go1.20.6
    Example traces found:
     <redacted>

Vulnerability #2: GO-2023-1840
    Unsafe behavior in setuid/setgid binaries in runtime
  More info: https://pkg.go.dev/vuln/GO-2023-1840
  Standard library
    Found in: runtime@go1.20.4
    Fixed in: runtime@go1.20.5
    Example traces found:
      <redacted>
      
Your code is affected by 2 vulnerabilities from the Go standard library.
```

Running `osv-scanner` however does not catch those. The issue here is, that based on https://github.com/google/osv-scanner/blob/93a2ec9f45208209cbfa1df825d94396d7bcf472/internal/sourceanalysis/go.go#L113
`osv-scanner` is producing a fixed set of vulnerabilities based only on the entries from the `go.mod` file. The result get's injected into `govulncheck` as database. However, GO's stdlib dependencies are not listed within `go.mod` and thus will never get found. 

Does it make sense to include everything from stdlib by default? https://osv.dev/list?ecosystem=Go&q=stdlib

another-rex:
Fixes #453,

As discussed in #453, it does require go to be installed on the system that osv-scanner is being ran on. The go installation can be at a custom location as long as $GOROOT environment var is specified.

Also exclude the depguard check, which has been added in a newer version of golangci (which we haven't updated to, but will cause linting issues when we eventually do).

### DIFF
--- a/pkg/osvscanner/optional_enricher.go
+++ b/pkg/osvscanner/optional_enricher.go
@@ -27,7 +27,7 @@ func getGoVersion() (string, error) {
 }
 
 func postLockfileEnricher(r reporter.Reporter, parsedLockfile *lockfile.Lockfile) {
-	switch parsedLockfile.ParsedAs {
+	switch parsedLockfile.ParsedAs { //nolint:gocritic
 	case "go.mod":
 		goVer, err := getGoVersion()
 		if err != nil {
--- a/pkg/osvscanner/optional_enricher.go
+++ b/pkg/osvscanner/optional_enricher.go
@@ -1,0 +1,43 @@
+package osvscanner
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/google/osv-scanner/pkg/reporter"
+)
+
+// Get the go version of the default environment that osv-scanner is being ran on
+func getGoVersion() (string, error) {
+	// GOROOT will return either the custom go location specified by $GOROOT environment
+	// variable, or it will return the default go location used during the go build,
+	// which will match the default of the platform being ran on.
+	versionBytes, err := os.ReadFile(filepath.Join(runtime.GOROOT(), "VERSION"))
+	if err != nil {
+		return "", err
+	}
+	version := strings.TrimPrefix(string(versionBytes), "go")
+	version, _, _ = strings.Cut(version, " ")
+
+	return version, nil
+}
+
+func postLockfileEnricher(r reporter.Reporter, parsedLockfile *lockfile.Lockfile) {
+	switch parsedLockfile.ParsedAs {
+	case "go.mod":
+		goVer, err := getGoVersion()
+		if err != nil {
+			r.PrintError(fmt.Sprintf("cannot get go standard library version, go might not be installed: %s", err))
+		}
+		parsedLockfile.Packages = append(parsedLockfile.Packages, lockfile.PackageDetails{
+			Name:      "stdlib",
+			Version:   goVer,
+			Ecosystem: lockfile.GoEcosystem,
+			CompareAs: lockfile.GoEcosystem,
+		})
+	}
+}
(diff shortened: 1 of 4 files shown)


pkg/osvscanner/optional_enricher.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/google/osv-scanner/pkg/reporter"
+)
+
+// Get the go version of the default environment that osv-scanner is being ran on
+func getGoVersion() (string, error) {
+	// GOROOT will return either the custom go location specified by $GOROOT environment
+	// variable, or it will return the default go location used during the go build,
+	// which will match the default of the platform being ran on.
+	versionBytes, err := os.ReadFile(filepath.Join(runtime.GOROOT(), "VERSION"))
```
another-rex, line 19: the GOROOT could cause a obscure bug in the future, since the default GOROOT value is the GOROOT value used to **build** the binary, not the default GOROOT for go, if we ever start building osv-scanner from a custom GO location, this will be incorrect. 

I haven't found a good alternative though, I don't think hard-coding the default GOROOT for each platform we support and switch between them is a good idea either.

ecktom: Reading from `runtime` package clearly has the disadvantages you've described.

How about relying on the environment and doing it like eg. `govulncheck` itself?
https://go.googlesource.com/vuln/+/refs/heads/master/internal/scan/run.go#84

another-rex: Good point, updated to match the `govulncheck` implementation.

pkg/osvscanner/optional_enricher.go
```diff
@@ -0,0 +1,41 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/google/osv-scanner/pkg/reporter"
+)
+
+// Get the go version of the default environment that osv-scanner is being ran on
+func getGoVersion() (string, error) {
+	versionBytes, err := exec.Command("go", "env", "GOVERSION").Output()
```
oliverchang, line 14: What's the difference between this and just `go version` ?

pkg/osvscanner/optional_enricher.go
```diff
@@ -0,0 +1,41 @@
...
+	// version format can be:
+	// - go1.20.6
+	// - go1.22-20230729-RC00 cl/552016856 +457721cd52 X:fieldtrack,boringcrypto
+	version := strings.TrimPrefix(string(versionBytes), "go")
+	version, _, _ = strings.Cut(version, " ")
+
+	return version, nil
+}
+
+func postLockfileEnricher(r reporter.Reporter, parsedLockfile *lockfile.Lockfile) {
```
oliverchang, line 27: Rather than naming this a generic enricher, let's just call it what it is. e.g. "addCompilerVersion" or something. We can always rename this later if we need to generalize this since it's private.

pkg/osvscanner/optional_enricher.go
another-rex: go version prints a lot more (unnecessary) text, e.g.
```bash
$ go version
go version go1.20.6 linux/amd64
$ go env GOVERSION
go1.20.6
```

The split is still needed when go is from a custom build of some kind, it mostly keeps it semver sortable, e.g.
```
$ go env GOVERSION
go1.22-20230729-RC00 cl/123456789 +123456789 X:fieldtrack,boringcrypto
```

oliverchang (approved):

PR #501, opened 2023-08-21
License checker feature

josieang:
TODO:
- [x] all the tests
- [x] license allowlist behaviour as detailed in the thread https://github.com/google/osv-scanner/pull/501#discussion_r1306839555
- [x] license table view


cmd/osv-scanner/main.go
```diff
@@ -121,6 +121,14 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Usage:  "sets the path that local databases should be stored",
 				Hidden: true,
 			},
+			&cli.BoolFlag{
+				Name:  "experimental-all-packages",
+				Usage: "when json output is selected, prints all packages",
+			},
+			&cli.BoolFlag{
```
josieang, line 128: I think in the json output when the licenses is specified without an allowlist, we want it to print all packages (and their licenses) by default. If that's the case, then I'm not sure how the all-packages and licenses flags should interact.

pkg/osvscanner/osvscanner.go
```diff
@@ -599,27 +633,109 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
 	compareOffline bool,
-	query osv.BatchedQuery,
 	localDBPath string,
 ) (*osv.HydratedBatchedResponse, error) {
+
+	// Make OSV queries from the packages.
+	var query osv.BatchedQuery
+	for _, p := range packages {
+		switch {
+		// Prefer making paackage requests where possible.
```
G-Rath, line 723:
```suggestion
		// Prefer making package requests where possible.
```

pkg/osvscanner/vulnerability_result.go
```diff
@@ -11,47 +11,52 @@ import (
 	"github.com/google/osv-scanner/pkg/reporter"
 )
 
-// groupResponseBySource converts raw OSV API response into structured vulnerability information
-// grouped by source location.
-func groupResponseBySource(r reporter.Reporter, query osv.BatchedQuery, resp *osv.HydratedBatchedResponse, callAnalysis bool) models.VulnerabilityResults {
+// assembleResult takes the responses from the OSV API and the deps.dev API
+// and converts this into a VulnerabilityResults. As part is this, it groups
+// vulnerability information by source location.
+func assembleResult(r reporter.Reporter, packages []Package, vulnsResp *osv.HydratedBatchedResponse, licensesResp []models.Licenses, allPackages bool, callAnalysis bool, licenses bool) models.VulnerabilityResults {
```
G-Rath, line 17:
```suggestion
func assembleResult(
	r reporter.Reporter,
	packages []Package,
	vulnsResp *osv.HydratedBatchedResponse,
	licensesResp []models.Licenses,
	allPackages bool,
	callAnalysis bool,
	licenses bool,
) models.VulnerabilityResults {
```

pkg/models/licenses.go
```diff
@@ -0,0 +1,3 @@
+package models
+
+type Licenses []string
```
josieang, line 3: I've noticed that we don't really do type aliases in osv-scanner, maybe to safeguard against us ever wanting to add to this Licenses field in the future?

Maybe this should be

```
type Licenses struct {
   Licenses []string
}
```

Also this should probably go in the results.go file.

pkg/osvscanner/osvscanner.go
```diff
@@ -62,7 +72,7 @@ var OnlyUncalledVulnerabilitiesFoundErr = errors.New("only uncalled vulnerabilit
 //   - Any lockfiles with scanLockfile
 //   - Any SBOM files with scanSBOMFile
 //   - Any git repositories with scanGit
-func scanDir(r reporter.Reporter, query *osv.BatchedQuery, dir string, skipGit bool, recursive bool, useGitIgnore bool) error {
+func scanDir(r reporter.Reporter, dir string, skipGit bool, recursive bool, useGitIgnore bool) ([]Package, error) {
```
josieang, line 75: While i think there is some benefit to using the `Package` type instead of the `osv.BatchedQuery` type here because we're no longer just calling the osv API in this PR, I'm not sure there is much benefit to returning `[]Package` everywhere (rather than updating it in place somehow). To me returning the packages makes it clearer what the function's inputs and outputs are, but I am willing to concede that this is just personal preference.

pkg/osvscanner/osvscanner.go
```diff
@@ -599,27 +633,109 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
 		// Otherwise return OnlyUncalledVulnerabilitiesFoundErr
-		return vulnerabilityResults, OnlyUncalledVulnerabilitiesFoundErr
+		return results, OnlyUncalledVulnerabilitiesFoundErr
 	}
 
-	return vulnerabilityResults, nil
+	return results, nil
+}
+
+const DepsdevURL = "api.deps.dev:443"
```
josieang, line 650: Maybe this license fetching function and variables can go in a file pkg/depsdev/license.go?

cmd/osv-scanner/main.go
another-rex: Now that we do have a all-packages flag, maybe licenses (without allowlist) shouldn't print all packages by default, and just fill in the field for packages that already exists? If someone wants full license info for everything, pass the all-packages flag. 

@oliverchang thoughts?

pkg/models/licenses.go
another-rex: I think a
```
type License string
```
could work better, and `Licenses` field will be `[]License`.

pkg/osvscanner/osvscanner.go
another-rex: Returning `[]Package` definitely looks better to me as well

pkg/osvscanner/osvscanner.go
josieang: Done.

pkg/osvscanner/vulnerability_result.go
josieang: Done.

pkg/osvscanner/osvscanner.go
josieang: Moved to pkg/depsdev/license.go

pkg/models/licenses.go
josieang: changed this to a type License string.

pkg/depsdev/license.go
```diff
@@ -0,0 +1,93 @@
...
+	var g errgroup.Group
+	for i := range queries {
+		if queries[i] == nil {
+			// This may be a private package.
+			continue
+		}
+		i := i
+		g.Go(func() error {
+			resp, err := client.GetVersion(ctx, queries[i])
+			if err != nil && status.Code(err) != codes.NotFound {
```
another-rex, line 74: hmm... it might be clearer to move the  
```
= []models.License{"UNKNOWN"}
```

here when you get `codes.NotFound`.

Are there cases where deps.dev won't return not found, but returns an empty array?

pkg/osvscanner/osvscanner.go
```diff
@@ -74,8 +79,9 @@ func scanDir(r reporter.Reporter, query *osv.BatchedQuery, dir string, skipGit b
 	}
 
 	root := true
+	var scannedPackages []Package
 
-	return filepath.WalkDir(dir, func(path string, info os.DirEntry, err error) error {
+	return scannedPackages, filepath.WalkDir(dir, func(path string, info os.DirEntry, err error) error {
```
another-rex, line 84: Can you move filepath.WalkDir to it's own line, e.g.
```
err := filepath.WalkDir(...

return scannedPackages, err
```

pkg/osvscanner/vulnerability_result.go
```diff
@@ -11,47 +11,74 @@ import (
 	"github.com/google/osv-scanner/pkg/reporter"
 )
 
-// groupResponseBySource converts raw OSV API response into structured vulnerability information
-// grouped by source location.
-func groupResponseBySource(r reporter.Reporter, query osv.BatchedQuery, resp *osv.HydratedBatchedResponse, callAnalysis bool) models.VulnerabilityResults {
+// assembleResult takes the responses from the OSV API and the deps.dev API
+// and converts this into a VulnerabilityResults. As part is this, it groups
+// vulnerability information by source location.
+func assembleResult(
```
another-rex, line 17: This seems to be getting pretty long, can you add a TODO comment here to refactor this function in the future?

another-rex (commented):
Added some more comments!

Spiritaine (approved):

Spiritaine (approved):

pkg/depsdev/license.go
josieang: yes, when the package is known but the license is unknown it returns an empty array. I've changed this so that the UNKNOWN is added in the depsdev package.

https://api.deps.dev/v3alpha/systems/go/packages/github.com%2Fpackage-url%2Fpackageurl-go/versions/v0.1.0

cmd/osv-scanner/main.go
oliverchang: Just to be sure I understand the alternatives here, is it:

1. `--licenses` (without `--all-packages`) will just ignore all Unknown licenses?
2. `--licenses` implies `--all-packages` ?

This is complicated by the behaviour for when an allow-list is passed, because to be consistent with the vulnerability scanning behaviour, it makes more sense for `--licenses <ALLOWLIST>` to not imply `--all-packages`. 

Thinking about this a bit more, all packages should in fact have a license right? (as opposed to vulnerabilities). We just consider "Unknown" to be a license that's worthwhile reporting too, because from a license compliance perspective it's just as important. 

From that perspective, 2. above seems to make more sense to me. Not necessarily because it implies `--all-packages` for the sake of it, but rather that we do have something worthwhile to report for all packages when it comes to licenses.

josieang: okay i see.

> This is complicated by the behaviour for when an allow-list is passed, because to be consistent with the vulnerability scanning behaviour it makes more sense for --licenses <ALLOWLIST> to not imply --all-packages.

Thinking about this, I think rather than thinking that --licenses implies --all-packages, we can think of how --licenses with an empty allowlist means every license a violation, which means all packages should be printed in the json output.

This is the solution I'm tending towards:
1. --all-packages has no effect on the table output. It will always be only the vulnerabilities in the vulns table, and if licenses is specified (regardless of allowlist), all the licenses and their counts in addition to the vulns table.
2. in json output --licenses=<allowlist> prints only packages with license violations or packages with vulns. If allowlist is empty then all packages are violating licenses (the license_violations field will be populated for every package)
3. in json output --licenses=<allowlist> with --all-packages prints all packages whether they have a vuln or license violation or not. The license_violations field is populated for every package that has a violation, license field populated for every package.

thoughts?

another-rex: The JSON output makes sense to me.

Can you clarify what happens in the table output when `--licenses` is passed in with an allowlist? 

Would the packages that are violating the allowlist be shown individually? or only a list of violating licenses in a summary format?

josieang: I was imagining that it would be the same as if there was no allowlist specified, it would just show all the licenses with their counts. If we did want to do something, we could add a column in the table when an allowlist is specified like 'violation' with value either true or false.

josieang: and to clarify, all licenses, including the UNKNOWN license would be in that table whether allowlist is specified or not.

another-rex: How would a user know what package violated the allowlist without having to parse the JSON?

josieang: they wouldn't.

my thinking was listing all individual licenses, or all license violations here may clutter the view. But you do make a good point as the vulns table shows the individual vulns, but this table would only shows a summary. it doesn't mirror the behaviour of the vulns perfectly.

we did discuss in the meeting adding a summary/all-packages flag to toggle this, but iirc we concluded that we will wait for feedback first and add flags or change the table format in response to user feedback.

did you have a solution in mind?

josieang: sorry! my bad my memory is wrong. For the table view we agreed to list a summary when allowlist is empty, and violations when allowlist isn't empty.

internal/output/markdowntable.go
```diff
@@ -12,12 +12,20 @@ import (
...
 	outputTable = tableBuilder(outputTable, vulnResult, false)
 
-	if outputTable.Length() == 0 {
+	if outputTable.Length() != 0 {
+		outputTable.RenderMarkdown()
+	}
+
+	outputLicenseTable := table.NewWriter()
+	outputLicenseTable.SetOutputMirror(outputWriter)
+	outputLicenseTable.AppendHeader(table.Row{"License", "No. of Package Versions"})
```
another-rex, line 23: Why is this line needed here but not in table.go?

cmd/osv-scanner/main_test.go
```diff
@@ -1071,6 +1107,287 @@ func TestRun_LocalDatabases(t *testing.T) {
 	}
 }
 
+func TestRun_Licenses(t *testing.T) {
+	t.Parallel()
+	tests := []cliTestCase{
```
another-rex, line 1112: Can you add one test for markdowntable output?

internal/output/table.go
```diff
@@ -156,3 +170,84 @@ func MaxSeverity(group models.GroupInfo, pkg models.PackageVulns) string {
...
+			if len(pkg.LicenseViolations) == 0 {
+				continue
+			}
+			violations := make([]string, len(pkg.LicenseViolations))
+			for i, l := range pkg.LicenseViolations {
+				violations[i] = string(l)
+			}
+			sourcePath, err := filepath.Rel(workingDir, pkgSource.Source.Path)
+			if err == nil { // Simplify the path if possible
+				pkgSource.Source.Path = sourcePath
```
another-rex, line 240: ~~Does this change the JSON output to also be a relative path?~~

Actually probably not, since the Source field is not a pointer.

pkg/osvscanner/osvscanner.go
```diff
@@ -779,19 +797,50 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
-	return vulnerabilityResults, nil
+		switch {
+		case !vuln && !onlyUncalledVuln && !licenseViolation:
+			// There is no error.
+			return results, nil
+		case vuln && !onlyUncalledVuln && !licenseViolation:
+			return results, VulnerabilitiesFoundErr
+		case !vuln && onlyUncalledVuln && !licenseViolation:
+			// Impossible state.
+			return results, nil
```
another-rex, line 828: Just panic here if this should be impossible

pkg/osvscanner/osvscanner.go
```diff
@@ -779,19 +797,50 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
+		case vuln && onlyUncalledVuln && !licenseViolation:
+			return results, OnlyUncalledVulnerabilitiesFoundErr
+		case !vuln && !onlyUncalledVuln && licenseViolation:
+			return results, LicenseViolationsErr
+		case vuln && !onlyUncalledVuln && licenseViolation:
+			return results, VulnerabilitiesFoundAndLicenseViolationsErr
+		case !vuln && onlyUncalledVuln && licenseViolation:
+			// Impossible state.
+			return results, nil
```
another-rex, line 837: Also panic here

pkg/osvscanner/vulnerability_result_internal_test.go
```diff
@@ -4,31 +4,467 @@ import (
...
+		{models.License("MIT"), models.License("0BSD")},
+		{models.License("MIT")},
+		{models.License("UNKNOWN")},
+	}
+
 	tests := []struct {
 		name string
 		args args
 		want models.VulnerabilityResults
-	}{}
```
another-rex, line 24: Thanks for adding these tests!

another-rex (changes requested):
Looks pretty good! Added some comments.

cmd/osv-scanner/main.go
```diff
@@ -188,20 +200,23 @@ func run(args []string, stdout, stderr io.Writer) int {
 		if r == nil {
 			r = reporter.NewTableReporter(stdout, stderr, false, 0)
 		}
-		if errors.Is(err, osvscanner.VulnerabilitiesFoundErr) {
+		switch {
```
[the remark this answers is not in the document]
josieang: oh great! thank you.

cmd/osv-scanner/main.go
```diff
@@ -161,25 +169,29 @@ func run(args []string, stdout, stderr io.Writer) int {
...
 				ExperimentalScannerActions: osvscanner.ExperimentalScannerActions{
-					LocalDBPath:    context.String("experimental-local-db-path"),
-					CallAnalysis:   context.Bool("experimental-call-analysis"),
-					CompareLocally: context.Bool("experimental-local-db"),
-					CompareOffline: context.Bool("experimental-offline"),
+					LocalDBPath:       context.String("experimental-local-db-path"),
+					CallAnalysis:      context.Bool("experimental-call-analysis"),
+					CompareLocally:    context.Bool("experimental-local-db"),
+					CompareOffline:    context.Bool("experimental-offline"),
+					AllPackages:       context.Bool("experimental-all-packages"),
```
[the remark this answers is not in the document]
josieang: yes, removing this.

cmd/osv-scanner/json
```diff
@@ -0,0 +1,10 @@
+Scanning dir fixtures/locks-many/package-lock.json
```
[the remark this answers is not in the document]
josieang: oh yes it is. removed.

cmd/osv-scanner/osv-scanner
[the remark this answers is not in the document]
josieang: unintentional

internal/output/markdowntable.go
josieang: that's a mistake, removed this line

pkg/osvscanner/osvscanner.go
josieang: done

cmd/osv-scanner/main.go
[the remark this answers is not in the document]
josieang: Good point, done.

cmd/osv-scanner/main.go
```diff
@@ -161,25 +169,29 @@ func run(args []string, stdout, stderr io.Writer) int {
...
 			}, r)
 
 			if err != nil &&
 				!errors.Is(err, osvscanner.VulnerabilitiesFoundErr) &&
-				!errors.Is(err, osvscanner.OnlyUncalledVulnerabilitiesFoundErr) {
-				//nolint:wrapcheck
+				!errors.Is(err, osvscanner.OnlyUncalledVulnerabilitiesFoundErr) &&
+				!errors.Is(err, osvscanner.LicenseViolationsErr) &&
+				!errors.Is(err, osvscanner.VulnerabilitiesFoundAndLicenseViolationsErr) &&
+				!errors.Is(err, osvscanner.OnlyUncalledVulnerabilitiesFoundAndLicenseViolationsErr) {
```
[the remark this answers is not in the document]
josieang: good idea

cmd/osv-scanner/main.go
```diff
@@ -188,20 +200,23 @@ func run(args []string, stdout, stderr io.Writer) int {
...
 			return 2
-		}
-
-		if errors.Is(err, osvscanner.NoPackagesFoundErr) {
+		case errors.Is(err, osvscanner.LicenseViolationsErr):
+			return 4
+		case errors.Is(err, osvscanner.VulnerabilitiesFoundAndLicenseViolationsErr):
+			return 5
+		case errors.Is(err, osvscanner.OnlyUncalledVulnerabilitiesFoundAndLicenseViolationsErr):
+			return 6
```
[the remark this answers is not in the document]
josieang: good idea also

internal/output/table.go
josieang: I think theoretically this could change the JSON output, as vulnResult is a pointer. But to produce that error, it would need to be possible to produce JSON and table output in the same osv-scanner invocation, and I don't think that's possible?

either way, I've changed this so that we don't modify vulnResult, I think that's bad practice. good catch.

josieang:
Ah I see, looks like we might need to update the linter and the go version in the checks. `errors.Join` was added in go 1.20.

cmd/osv-scanner/main_test.go
josieang: done

cmd/osv-scanner/main.go
```diff
@@ -121,11 +121,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Hidden: true,
 			},
 			&cli.BoolFlag{
-				Name:  "experimental-all-packages",
+				Name:  "experimental-show-all-packages",
```
another-rex, line 124: Oh I mean the CLI names can be kept as is, just change the struct names. We already do that for some of the other flags.

another-rex (approved):
LGTM!

cmd/osv-scanner/main.go
```diff
@@ -188,20 +202,23 @@ func run(args []string, stdout, stderr io.Writer) int {
...
-		}
-
-		if errors.Is(err, osvscanner.NoPackagesFoundErr) {
+		switch {
+		case errors.Is(err, osvscanner.VulnerabilitiesFoundErr):
+			return 0b0001 // 1
+		case errors.Is(err, osvscanner.OnlyUncalledVulnerabilitiesFoundErr):
+			// TODO: Discuss whether to have a different exit code
+			// now that running call analysis is not default.
+			return 0b0010 // 2
```
oliverchang, line 211: I wonder if we're complicating the return codes here by having so many cases and different error types. If users do need to distinguish these, this should be done through another mechanism (e.g. the JSON output or `models.VulnerabilityResults` output). 

I suggest we just return 1 if there is any error at all. 

@G-Rath @another-rex

pkg/osvscanner/osvscanner.go
```diff
@@ -779,19 +797,49 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
 			}
 		}
-		// Otherwise return OnlyUncalledVulnerabilitiesFoundErr
-		return vulnerabilityResults, OnlyUncalledVulnerabilitiesFoundErr
-	}
+		onlyUncalledVuln = onlyUncalledVuln && vuln
 
-	return vulnerabilityResults, nil
+		switch {
+		case !vuln && !onlyUncalledVuln && !licenseViolation:
```
oliverchang, line 821: Returning to this :) Sorry I didn't get a chance to review this earlier. 

Similar to my other comment, this is getting very complicated. I suggest we just fold this into a single `VulnerabilitiesFoundErr` if there are any errors (including license errors). 

If a user wants to distinguish between these errors more, they can just look at the `models.VulnerabilityResults` result. 

@another-rex

cmd/osv-scanner/main.go
G-Rath: yyeaahhh I'd say for now best to keep it simple and find out if people have a strong need for them - as you've eluded to @oliverchang, I think as well the scanner will grow to a point that fine-grain errors should be handled via e.g. JSON just because of all the different facets

josieang: I'm happy to return 0 or 1 here and sync with @another-rex so that these changes can be done alongside his changes to make call analysis return 0 or 1 as well.

I may not get to implementing error details in the `models.VulnerabilityResults` output soon but I'll file an issue for this.

josieang: https://github.com/google/osv-scanner/issues/675

oliverchang: Thanks! Note that I don't think #675 is necessary -- all the data is there in models.VulnerabilityResults already.

PR #505, opened 2023-08-25
Add osv output lockfile + refactor

another-rex:
Introduces the osv-scanner results lockfile format (feel free to suggest a better name for this), 

This also refactors the models folder slightly to move some of the helper functions out to a separate internal crate. This stops the models package from pulling in the lockfile package, causing cyclic imports.

I also added another field to `PackageDetails`, `Source`, a currently optional string to specify the location the dependency is actually specified. The only times it will be different from the lockfile location currently is:

- Python requirements.txt `-r`
- This new osv-scanner format, it stores the lockfile path parsed in the output results.


Spiritaine (approved):

Spiritaine (approved):

G-Rath:
I think it would be good if you split the refactor out into its own PR, and consider if anything else should be refactored for consistency since removing the receiver from a method is a breaking change so best to do it now than regret it later (you can also always make a method public in future in a non-breaking version too)

internal/modelsutil/vulnerabilities.go
```diff
@@ -0,0 +1,20 @@
+package modelsutil
```
oliverchang, line 1: How about naming this packgage `vulns` instead? This is similar to e.g. `bytes`, `strings` in the Go std library. 

This way we also avoid a lot of stutter in the funciton names. 

e.g. vulns.Includes() vs modeslutil.VulnsIncludes()

pkg/lockfile/osv-vuln-results.go
```diff
@@ -0,0 +1,58 @@
+package lockfile
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+const OSVScannerEcosystem Ecosystem = "OSV-Scanner-Results"
```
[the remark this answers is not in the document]
another-rex: Good point, removed it now.

pkg/lockfile/helpers_test.go
```diff
@@ -31,10 +32,32 @@ func packageToString(pkg lockfile.PackageDetails) string {
...
+func hasPackage(t *testing.T, packages []lockfile.PackageDetails, pkg lockfile.PackageDetails) bool {
+	t.Helper()
+	// Store source here since original source is set to empty string to do equal comparison
+	pkgSource := pkg.Source
 	for _, details := range packages {
-		if details == pkg {
-			return true
+		// Custom source equality check to not be too path specific
+		// areEqual is not symmetrical, so compare both ways
+		if areEqual(t, pkgSource, details.Source) || areEqual(t, details.Source, pkgSource) {
```
G-Rath, line 55: I'll dig into this more later and maybe it'll then become clear, but I would expect you shouldn't need to do any `%%`ing for `Source` (aka the path should be determinable and easy to generate for each test) so I'd say do that rather than use `%%` because it should be a more robust test.

pkg/lockfile/helpers_test.go
```diff
@@ -55,11 +78,12 @@ func expectPackage(t *testing.T, packages []lockfile.PackageDetails, pkg lockfil
 	}
 }
 
-func findMissingPackages(actualPackages []lockfile.PackageDetails, expectedPackages []lockfile.PackageDetails) []lockfile.PackageDetails {
+func findMissingPackages(t *testing.T, actualPackages []lockfile.PackageDetails, expectedPackages []lockfile.PackageDetails) []lockfile.PackageDetails {
```
G-Rath, line 81: This change is good by itself along with the `hasPackage` one, so if you can be bothered I'd favor landing it in its own PR so that this diff is reduced rather than have it wait around for the rest of the changes to get reviewed

pkg/lockfile/helpers_test.go
another-rex: Hmm I'm not sure if I understand what you mean here. This is for requirements.txt because the result is an absolute path which will change depending on where the test is being ran. 

I guess we can expand `%%` to the CWD instead for more accuracy.

G-Rath: You should be able to determine what that absolute path is via `filepath.Abs` - frankly we could do that in the CLI tests too, I just didn't because `%%` was available and I was being a bit lazy (and it came from the detector which needed it because it printed the db count and update time which couldn't be easily determined).

another-rex: Ah I see what you mean, I'll try that out.

another-rex: Done in a5fc1b5

pkg/lockfile/helpers_test.go
```diff
@@ -41,10 +45,20 @@ func hasPackage(packages []lockfile.PackageDetails, pkg lockfile.PackageDetails)
 	return false
 }
 
+func absPathNoErr(path string) string {
```
G-Rath, line 48: nit: generally I'd consider calling this something like `absPathOrPanic`, or `mustBeAbsPath` if you want to be a true Go traditionalist 😅 - but that way it gives a stronger indication of what happens if it does error

another-rex: Went with `absPathOrPanic`, thanks for the name suggestion, originally I had `absPathButPanicOnErr` and thought that was way too long haha.

pkg/lockfile/parse-requirements-txt.go
```diff
@@ -167,6 +167,7 @@ func parseRequirementsTxt(f DepFile, requiredAlready map[string]struct{}) ([]Pac
 		}
 
 		detail := parseLine(line)
+		detail.Source = f.Path()
```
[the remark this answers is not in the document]
oliverchang: Thanks for adding a comment. Another question -- How will `Source` be used by consumers?

Assuming this is needed by consumers, I wonder if we should just be consistent and **always** include this field?

oliverchang (commented):
Change itself LGTM, but I have some remaining questions about the `Source` field we added.

pkg/lockfile/parse-requirements-txt.go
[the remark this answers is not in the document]
another-rex: We probably should always include this field. it is really only optional because 

1. If Source is empty, it can be assumed to be the same value as the Lockfile (this is the case for all our lockfiles apart from requirements.txt and OSV-Scanner)
2. It'll be a lot more lines of change. 

I'm happy to go ahead and implement this for everything.

> How will Source be used by consumers?

Three examples:

1. OSV-Scanner is used in two phases. first it's ran on a system to get a full list of packages without any scanning (e.g. vm, docker container, offline/firewalled system...etc), then the osv-scanner output is loaded onto a system that can scan it later, it would be nice to preserve where each package actually comes from.  (This PR doesn't do that yet, but it provides the ability for future PR to add that capability, and enables it in the API)
2. Someone writes a custom extractor (e.g. for DEPs files) and outputs the OSV format, this allows the path to the original file to be preserved after scanning.
3. Helps find where the actual dependency is specified in lockfiles that can span multiple files (like requirements.txt)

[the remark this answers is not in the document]
oliverchang: Got it. Makes sense now. 

Keeping it optional seems fine, and avoids a bit of bloat, but let's document this and make it explicit that it's only required in cases where it's different from the original source lockfile?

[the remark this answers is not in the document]
G-Rath: Personally I'd actually say:
  1. if there's no immediate use-case, consider omitting it for now as it can always be added later (that's why we didn't add this when support for `-r` was added in the first place)
  2. make it required, because otherwise tools have to handle "if else" whereas if its always required then they can just use that if needed, and it shouldn't cost anything for the scanner to include it in its output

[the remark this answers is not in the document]
oliverchang: I'd argue there is a hidden cost to including more details in the output -- it includes a little bit of cognitive load when there's more duplication in the JSON output that a human has to handle when writing code to deal with our output. 

+1 on just removing this if there are no immediate users. For the DEPs/custom extractors case, can people just re-use the existing `packageSource` field? 

i.e.
```
"packageSource": {
        "path": "/absolute/path/to/DEPS",
        "type": "custom"
      },
```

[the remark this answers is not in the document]
another-rex: > +1 on just removing this if there are no immediate users. For the DEPs/custom extractors case, can people just re-use the existing packageSource field?

That is where the Source field is populated from.

This `Source` field only exists in the lockfile API, and the plan is to use this in the output eventually. Happy to remove this for now and we can add this back in, in the future when we make use of it.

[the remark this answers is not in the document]
G-Rath: > I'd argue there is a hidden cost to including more details in the output

I'd argue you're not removing that cost by making it conditional, but in fact _adding_ to it because now the human has to understand _when_ the field will actually be present; plus the conditional handling has to be forever carried around in the code and understood by every reader whereas if the field is just always valid for the most part I would expect known or ignored based on if the tool in question uses the field (i.e. when you read the code you'll either see the property being used and thus know it exists and understand it, or you won't see any reference to it in which case you don't know it exists until you do actually need it).

(related, this is why I've found `omitempty` so strange and having empty slices serialize as nil rather than an empty array - because that then requires humans to check if the output is an array or `null` instead of being able to just call their array function and have no iterations occur if its empty 🤷)

but this is getting off topic - happy to discuss separately if you'd like 🙂 

> Happy to remove this for now and we can add this back in, in the future when we make use of it.

Sounds like we're all in agreement to go with that.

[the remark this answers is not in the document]
another-rex: Removed source field for now :)

oliverchang (approved):

cmd/osv-scanner/main_test.go
```diff
@@ -91,6 +91,8 @@ type cliTestCase struct {
 func testCli(t *testing.T, tc cliTestCase) {
 	t.Helper()
 
+	t.Logf("Running test: [%s]", tc.name)
```
G-Rath, line 94: I think this should be removed as Go already prints the name of the test if it has one, otherwise it prints its index so having this just adds extra noise:

```
--- FAIL: TestRun_LocalDatabases (0.00s)
    --- FAIL: TestRun_LocalDatabases/blah (1.01s)
        main_test.go:843: Running test: [blah]
        main_test.go:843: stdout
             got:
            Scanning dir ./fixtures/locks-many/composer.lock
            Scanned /home/jones/workspace/projects-oss/osv-scanner/cmd/osv-scanner/fixtures/locks-many/composer.lock file and found 1 package
            Loaded Packagist local db from /tmp/osv-scanner-test-4097718912/osv-scanner/Packagist/all.zip
            No vulnerabilities found

             want:
            Scanning dir ./fixtures/locks-many/composer.lock
            Scanned %%/fixtures/locks-many/composer.lock file and found 1 package
             Loaded Packagist local db from %%/osv-scanner/Packagist/all.zip
            No vulnerabilities found
FAIL
```

vs

```
--- FAIL: TestRun_LocalDatabases (0.00s)
    --- FAIL: TestRun_LocalDatabases/#11 (1.50s)
        main_test.go:843: Running test: []
        main_test.go:841: stdout
             got:
            Scanning dir ./fixtures/locks-many/composer.lock
            Scanned /home/jones/workspace/projects-oss/osv-scanner/cmd/osv-scanner/fixtures/locks-many/composer.lock file and found 1 package
            Loaded Packagist local db from /tmp/osv-scanner-test-1140406433/osv-scanner/Packagist/all.zip
            No vulnerabilities found

             want:
            Scanning dir ./fixtures/locks-many/composer.lock
            Scanned %%/fixtures/locks-many/composer.lock file and found 1 package
             Loaded Packagist local db from %%/osv-scanner/Packagist/all.zip
            No vulnerabilities found
FAIL
```

G-Rath (changes requested):

G-Rath (approved):

cmd/osv-scanner/main_test.go
another-rex: Huh, I must have missed that when debugging, removed.

Issue #442, opened 2023-07-13
Unify Ecosystem type and constants


another-rex:
The current Ecosystem type and constants are duplicated in multiple packages, and in multiple places ecosystem also only has a type of `string`. For V2 of the API we should combine these to the same ecosystem type.

another-rex:
This is now done, Ecosystem now uses the osv-schema package.

PR #510, opened 2023-08-28
Refactor models

another-rex:
Pulled refactor out to a separate PR as suggested here: #505 

In this PR:
- Refactored models package to no longer depend on lockfile
- Moved functions that required lockfile types into separate internal package under `utility/vulns`
- Copied over OSV url from `osv` package to avoid dependency on `osv` from `output`/`reporter`

I went through and graphed out the dependencies, 

- `models` no longer depends on any internal dependencies
- `lockfile` doesn't depend on any internal dependencies apart from `cacheexp`
- `osv` brings in both `lockfiles` and `models`. This is probably the most problematic one. Ideally `osv` should only be referencing `models`, but it'll be a breaking change to change this API. This is one of the main things we should change for V2 (related to #442) (TODO: Make separate issue for this
- `reporter` depends on the internal `output`, which depends on `osv`. This dependency causes `lockfile` to be pulled into everywhere that needs to print out anything. This can be decoupled without too much trouble by copying in the OSV url. I did so in this PR.
- `config` package references `reporter`,  with the above change it also no longer depends on `lockfile`.


internal/utility/vulns/vulnerabilities.go
```diff
@@ -0,0 +1,20 @@
+package vulns
+
+import "github.com/google/osv-scanner/pkg/models"
+
+func VulnsInclude(vs models.Vulnerabilities, vulnerability models.Vulnerability) bool {
```
oliverchang, line 5: nit: just `Include` ? to avoid stutter?

oliverchang (approved):
LGTM. I wonder if there's a way to enforce these dependency visibility rules somehow.

G-Rath:
idk about enforcing, but you could probably use something like https://github.com/kisielk/godepgraph to post a comment on the PR with a dependency graph - though that'd probably be pretty noisy so might be better as an item on the "before release" checklist, to do a final review of the dependency tree in case you want to make any changes that would be breaking if not done otherwise.

PR #517, opened 2023-09-02
Adding Offline info

hayleycd:
Please see [offline docs](https://hayleycd.github.io/osv-scanner/usage/offline/)

I also reorganized the [usage docs](https://hayleycd.github.io/osv-scanner/usage/) generally, because the usage doc was getting long. I can redo it if we don't like it. 

Addresses #515


docs/usage/offline.md
```diff
@@ -0,0 +1,30 @@
...
+## Offline mode
+
+{: .note }
+Features and flags with the `experimental` prefix might change or be removed with only a minor version update.
+
+OSV-Scanner now supports offline scanning as an experimental feature. Offline scanning checks your project against a local database instead of calling the OSV.dev API.
+
+### Local Database Option
+
+The local database flag `--experimental-local-db` causes OSV-Scanner to download or update your local database and then scan your project against it. `--experimental-local-db-path` determines the location of the database.
```
G-Rath, line 18: `--experimental-local-db-path` is an internal flag so should not be documented

G-Rath (changes requested):
I've had a quick look through and it's looking good so far but will have a more detailed look later in the week

docs/usage/offline.md
hayleycd: So the user won't need to set it?

G-Rath: Correct - it's available for the library (which isn't documented yet) but not for the CLI; that stores the databases in the user's cache directory within a directory named `osv-scanner`.

The flag ended up still existing because otherwise tests can't be run in parallel easily, but it's marked as hidden so it'll never show up in the help output and will be considered internal-only.

Btw I'll share the design doc with you to give you a bit more context

hayleycd: Thanks @G-Rath

another-rex:
It could be good to also list the paths that OSV-Scanner uses to look for the database, and some basic instructions on how to manually download that database themselves.

The paths right now:
> 1. The value returned by [`os.UserCacheDir`](https://pkg.go.dev/os#UserCacheDir), if possible
> 2. The value returned by [`os.TempDir`](https://pkg.go.dev/os#UserCacheDir)

To download themselves, follow osv data dumps doc, then place the `all.zip` files in this directory:

> ```
> {local_db_dir}/
>   osv-scanner/
>     npm/all.zip
>     PyPI/all.zip
>     …
>     {ecosystem}/all.zip
> ```
(Copied from the design doc, see it for more context)

another-rex (changes requested):
Good idea with splitting the call analysis into a separate page!

The docker page, let's hold that off for now, the current feature is very basic, and we'll be working on building additional docker capabilities soon, probably better to wait until then to create a specific page for it.

docs/usage/offline.md
```diff
@@ -0,0 +1,39 @@
...
+  {: .text-delta }
+- TOC
+{:toc}
+</details>
+## Offline mode
+
+{: .note }
+Features and flags with the `experimental` prefix might change or be removed with only a minor version update.
+
+OSV-Scanner now supports offline scanning as an experimental feature. Offline scanning checks your project against a local database instead of calling the OSV.dev API.
```
G-Rath, line 23: Couple of things I should are important to note:
  * where the databases are sourced from, for greater transparency (I imagine also folks dealing with air gapped environments will also appreciate being able to easily know what traffic to look for / allow in their general networks)
  * commits are not supported
  * local tools/libraries are not required aka version comparison is done using Go-based implementation of each ecosystems version spec, rather than trying to callout to dependency managers
  * people are encouraged to report if they think versions are being compared incorrectly (most of the ecosystems should be pretty robust, but some like Maven have a lot of grey areas and ultimately real-world data is going to be the best feedback)

G-Rath (changes requested):
Looking good - I've listed a few things I think that should be noted; I also wonder if it would be worth having an "experimental" section that these could live under to make it more clearer and save having to repeat the general "here there be dragons" since its expected to be a general usage pattern for future

hayleycd:
@G-Rath @another-rex 

Please see updated documentation: 
- [Offline mode](https://hayleycd.github.io/osv-scanner/experimental/#offline-mode)
- [Usage](https://hayleycd.github.io/osv-scanner/Usage)

Rex, if you could specifically check the [manual download section](https://hayleycd.github.io/osv-scanner/experimental/#manual-database-download)

G-Rath (changes requested):
Looking good! main thing missing is documenting where the databases are actually stored which follows this logic:

> {local_db_dir}  would be resolved based on the following logic:
>   1. The value of the `OSV_SCANNER_LOCAL_DB_CACHE_DIRECTORY` environment variable, if set
>   2. The value returned by [`os.UserCacheDir`](https://pkg.go.dev/os#UserCacheDir), if possible
>   3. The value returned by [`os.TempDir`](https://pkg.go.dev/os#TempDir)

I think that should be fine to stick almost as is after the block showing the database directory structure

docs/experimental.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Example
+```bash
+osv-scanner --experimental-call-analysis ./my/project/path
+```
+
+## Offline mode
+
+{: .note }
+Features and flags with the `experimental` prefix might change or be removed with only a minor version update.
```
another-rex, line 64: Since this is already in the experimental page, we probably don't need a second banner

docs/experimental.md
```diff
@@ -0,0 +1,120 @@
...
+```bash
+gsutil cp gs://osv-vulnerabilities/PyPI/all.zip .
+```
+
+You can also download over HTTP via https://osv-vulnerabilities.storage.googleapis.com/<ECOSYSTEM>/all.zip
+
+A list of all current ecosystems is available at 
+[`gs://osv-vulnerabilities/ecosystems.txt`](https://osv-vulnerabilities.storage.googleapis.com/ecosystems.txt)
+
+When run with the `--experimental-local-db` flag, OSV-Scanner downloads the database into the following file structure:
```
another-rex, line 105: Move this section where we describe where the files are located above the manual download section. That way you can refer to this section directly for where to place the downloaded files.

docs/experimental.md
```diff
@@ -0,0 +1,120 @@
...
+osv-scanner --experimental-offline ./path/to/your/dir
+```
+
+### Manual database download
+Instead of using the `--experimental-local-db` flag to download the database, it is possible to manually download the database. 
+
+A downloadable copy of the OSV database is stored in a GCS bucket maintained by OSV:
+[`gs://osv-vulnerabilities`](https://osv-vulnerabilities.storage.googleapis.com)
+
+This bucket contains individual entries of the format
```
another-rex, line 89: I think it'll be better to remove the part about the `json` files and be more focused on getting the `all.zip` part since that's what's needed for osv-scanner. Add a link to the data dump section on the osv.dev docs for more details.

another-rex (commented):
Looks pretty good in general, added some minor comments!

docs/experimental.md
hayleycd: good point

hayleycd:
@another-rex @G-Rath Thanks for your comments. Please review.

docs/experimental.md
```diff
@@ -0,0 +1,132 @@
...
+- Does not support dependencies that link external non-rust code
+
+### Example
+```bash
+osv-scanner --experimental-call-analysis ./my/project/path
+```
+
+## Offline mode
+
+OSV-Scanner now supports offline scanning as an experimental feature. Offline scanning checks your project against a local database instead of calling the OSV.dev API. Local copies of dependencies are not required because version comparison is done using Go-based implementation of each ecosystems version specificiation. OSV-Scanner does not callout to dependency managers when using `--experimental-local-db` or `--experimental-offline` flags.
```
another-rex, line 63:
```suggestion
OSV-Scanner now supports offline scanning as an experimental feature. Offline scanning checks your project against a local database instead of calling the OSV.dev API.
```

Removed those lines as it's describing the same behavior as non-local scanning, and the custom version comparison is just an implementation detail.

another-rex, line 63: I would add a line (or expand the existing one) emphasizing that no information about the dependencies being scanned is sent anywhere when using this mode. I think that's a main attraction of having a local scanning mode is to avoid leaking what dependencies are being used, by using osv-scanner.

docs/experimental.md
```diff
@@ -0,0 +1,132 @@
...
+### Local database option
+
+The local database flag `--experimental-local-db` causes OSV-Scanner to download or update your local database and then scan your project against it. 
+
+```bash
+osv-scanner --experimental-local-db ./path/to/your/dir
+```
+
+### Offline option
+The offline database flag `--experimental-offline` causes OSV-Scanner to scan your project against a previously downloaded local database. OSV-Scanner will not download or update the local database. When a local database is not present, you will get an error message.
```
another-rex, line 102:
```suggestion
The offline database flag `--experimental-offline` causes OSV-Scanner to scan your project against a previously downloaded local database. OSV-Scanner will not download or update the local database. When a local database is not present, you will get an error message. No network connection is required when using this flag.
```

another-rex (commented):
Final comments, everything else LGTM!

docs/usage.md
```diff
@@ -4,8 +4,6 @@ title: Usage
 permalink: /usage/
 nav_order: 3
 ---
-# Usage
```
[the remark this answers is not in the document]
hayleycd: Not intended! Thanks for the catch.

docs/experimental.md
```diff
@@ -0,0 +1,132 @@
+---
+layout: page
+title: Experimental Features
+permalink: /experimental/
+nav_order: 6
+---
+{: .no_toc }
+
```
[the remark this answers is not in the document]
hayleycd: Thanks

hayleycd:
@another-rex please see changes. I think this may be ready.

PR #524, opened 2023-09-05
test: add cases for extra coverage

G-Rath:


G-Rath:
@another-rex we might as well land this since its an improvement - there could be more, but I've not gotten back to this so lets just land what we have.

another-rex (approved):
Thanks!

PR #525, opened 2023-09-05
ci: change how coverage is collected

G-Rath:


G-Rath:
@oliverchang @another-rex I'm not sure why codecov isn't reporting the actual coverage change here, but I think this is what we want  - compare https://app.codecov.io/gh/google/osv-scanner/tree/change-coverage vs https://app.codecov.io/gh/google/osv-scanner/tree/main/pkg, you'll see this branch has a coverage rating of 77% whereas main has 68%, and there's just a lot more coverage being reported through out the packages.

Note that at least the coverage for `cmd` has actually dropped, but that's because without this change only packages that actually have tests get included at all - so now the coverage report includes packages like `cmd/osv-reporter` which have a coverage of 0.

I suspect that codecov hasn't posted a comment because of how different the two reports are, and that it thinks something like the reporting format itself has changed meaning it can't/shouldn't compare what it currently has to this new report. I say we merge this and we can use #524 to check its working properly.

another-rex (approved):

G-Rath:
@another-rex @oliverchang yup this looks to be better - have a look over the report before our next catchup; previously [this function](https://app.codecov.io/gh/google/osv-scanner/blob/main/pkg%2Fconfig%2Fconfig.go#L36) was completely uncovered, but now its reported as mostly covered (and #524 has it fully covered)

PR #534, opened 2023-09-14
Update SARIF format

### LINKED ISSUE #216: Add support for SARIF output

another-rex:
Fixes #216 with a new format that separates out individual vulnerabilities. 

Each vulnerability is now it's own rule violation. The aliased vulnerabilities are grouped together as one rule violation, with an ID picked in this priority (CVE -> [Eco Specific] -> GHSA).


internal/output/identifiers.go
```diff
@@ -0,0 +1,40 @@
...
+			return 1
+		}
+	}
+
+	// Neither is CVE
+	aIsGHSA := strings.HasPrefix(strings.ToUpper(a), "GHSA")
+	bIsGHSA := strings.HasPrefix(strings.ToUpper(b), "GHSA")
+	if aIsGHSA || bIsGHSA {
+		if aIsGHSA == bIsGHSA {
+			// Both are CVEs, order by alphanumerically
```
oliverchang, line 27: nit: s/CVE/GHSA/

internal/output/identifiers.go
```diff
@@ -0,0 +1,40 @@
+package output
+
+import "strings"
+
+// idSortFunc sorts IDs ascending by CVE < [ECO-SPECIFIC] < GHSA
+func idSortFunc(a, b string) int {
```
oliverchang, line 6: I think we can simplify this a bit? 

```go

func prefixOrder(prefix string) int {
  if prefix == "CVE" {
    // Highest precedence
    return 2
  } else if prefix == "GHSA" {
    // Lowest precedence
    return 0
  }
  return 1
}

prefixAOrd = prefixOrder(strings.Split(a, "-")[0])
prefixBOrd = prefixOrder(strings.Split(b, "-")[0])

if prefixAOrd != prefixBOrd {
  return intComparson(prefixAOrd, prefixBOrd) // not sure if there's a builtin available
}

return strings.Compare(a, b)
}
```

internal/output/result.go
```diff
@@ -0,0 +1,126 @@
...
+		groupFixedVersions[k] = slices.Compact(fixedVersions)
+	}
+
+	return groupFixedVersions
+}
+
+// groupByVulnGroups creates a map over all vulnerability IDs, with aliased vuln IDs
+// pointing to the same groupVulns object
+func groupByVulnGroups(vulns *models.VulnerabilityResults) map[string]*groupedVuln {
+	// Map of Vuln IDs to
```
oliverchang, line 79: finish comment?

internal/output/result.go
```diff
@@ -0,0 +1,126 @@
...
+
+	*pss = make(pkgSourceSet)
+	for _, pws := range aux {
+		(*pss)[pws] = struct{}{}
+	}
+
+	return nil
+}
+
+type groupedVuln struct {
```
oliverchang, line 42: nit: let's name this `sarifFinding` or something similar since it's very specific to that? Probably have some of the terminology a bit wrong here, but hopefully you get the idea :) 

`groupedVuln` on its own is otherwise a bit confusing? Unless you think this is something that generalizes to other output formats?

internal/output/result.go
```diff
@@ -0,0 +1,126 @@
...
+		slices.Sort(fixedVersions)
+		groupFixedVersions[k] = slices.Compact(fixedVersions)
+	}
+
+	return groupFixedVersions
+}
+
+// groupByVulnGroups creates a map over all vulnerability IDs, with aliased vuln IDs
+// pointing to the same groupVulns object
+func groupByVulnGroups(vulns *models.VulnerabilityResults) map[string]*groupedVuln {
```
oliverchang, line 78: nit: this function name on its own is pretty confusing with two instances of "group". 

Perhaps rename this as `mapVulnIDsToSARIFFinding` or something similar?

cmd/osv-scanner/main_test.go
```diff
@@ -349,59 +343,95 @@ func TestRun(t *testing.T) {
...
+                            "markdown": "\n**Your dependency is vulnerable to [GHSA-whgm-jr23-g3j9](https://osv.dev/vulnerability/GHSA-whgm-jr23-g3j9)** \n.\n\n\n\u003e ## [GHSA-whgm-jr23-g3j9](https://osv.dev/vulnerability/GHSA-whgm-jr23-g3j9)\n\u003e \n\u003e This affects all versions of package ansi-html. If an attacker provides a malicious string, it will get stuck processing the input for an extremely long time.\n\u003e \n\n\n---\n\n### Affected Packages\n| Source | Package Name | Package Version |\n| --- | --- | --- |\n| lockfile:%%/fixtures/locks-many/package-lock.json | ansi-html | 0.0.1 |\n\n"
+                          },
+                          "properties": {
+                            "deprecatedIds": [
+                              "GHSA-whgm-jr23-g3j9"
```
oliverchang, line 371: This is the same as the actual ID? Should this be the case?

cmd/osv-scanner/main_test.go
```diff
@@ -349,59 +343,95 @@ func TestRun(t *testing.T) {
...
+              "$schema": "https://json.schemastore.org/sarif-2.1.0.json",
+              "runs": [
+                {
+                  "tool": {
+                    "driver": {
+                      "informationUri": "https://github.com/google/osv-scanner",
+                      "name": "osv-scanner",
+                      "rules": [
+                        {
+                          "id": "GHSA-whgm-jr23-g3j9",
```
oliverchang, line 357: Should we have an end-to-end test here that includes CVE, Ecosystem ID examples etc?

cmd/osv-scanner/main_test.go
another-rex: Yes, I'm thinking deprecated Ids will contain all the IDs separately to avoid people using it to have separate logic to append displayId to it. This indicates that all the IDs are the same.

oliverchang: Ack.

cmd/osv-scanner/main.go
```diff
@@ -215,6 +216,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 	return 0
 }
 
+func initVersion() {
+	output.OSVVersion = version
```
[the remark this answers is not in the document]
another-rex: Created a new internal package `version` which everything uses now

internal/output/identifiers.go
another-rex: This looks much cleaner, thanks!

cmd/osv-scanner/main_test.go
another-rex: Added a bigger test with CVE's and multiple ecosystems, not as an end-to-end test but a unit test since it makes the path resolution easier. It already found some problems with CVEs not actually being the display ID, which should be fixed now. 

Do you still want a bigger end-to-end test as well?

oliverchang (approved):

PR #542, opened 2023-09-18
Attempt at multiline action

another-rex:
Update the github actions to pass arguments in a multi line fashion to allow much more customisability.

### DIFF
--- a/cmd/osv-reporter/main.go
+++ b/cmd/osv-reporter/main.go
@@ -22,6 +22,8 @@ var (
 	date    = "n/a"
 )
 
+// splitLastArg splits the last argument by new lines and appends the split
+// elements onto args and returns it
 func splitLastArg(args []string) []string {
 	lastArg := args[len(args)-1]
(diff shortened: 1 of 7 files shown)


cmd/osv-reporter/main.go
```diff
@@ -16,14 +17,24 @@ import (
 
 var (
 	// Update this variable when doing a release
-	version = "1.3.5"
+	version = "1.4.0"
```
oliverchang, line 20: Can we use that new internal package you added for this?

oliverchang (approved):
Nice! I assume for the scanner the existing splitting in the bash wrapper still works?

cmd/osv-reporter/main.go
another-rex: That's still in the SARIF branch, which hasn't been merged, I'll update it in that branch so this variable uses the new internal package.

another-rex:
> Nice! I assume for the scanner the existing splitting in the bash wrapper still works?

Yep, that still works.

PR #547, opened 2023-09-25
SARIF output fixes.

oliverchang:
- Use relative path for URIs.
- Fix description/summary stability (due to map insertion order not being preserved), and sort by `<ECOSYSTEM> < <GHSA> < <CVE>` for extracting descriptions.
- Remove an extraneous period.

image


oliverchang:
Sample report: https://github.com/oliverchang/osv-fork/security/code-scanning/53

internal/output/sarif.go
```diff
@@ -29,7 +29,7 @@ const SARIFTemplate = `
 **Your dependency is vulnerable to [{{.ID}}](https://osv.dev/vulnerability/{{.ID}})** 
 {{if gt (len .AliasedVulns) 1 -}}
 (Also published as: {{range .AliasedVulns -}} {{if ne .ID $.ID}} [{{.ID}}](https://osv.dev/vulnerability/{{.ID}}) {{end}}{{end}})
-{{- end}}.
+{{- end}}
```
another-rex, line 32: I think we still want to keep the period. The issue of it being on a new line is because the if statement on line 30 does not have a `-`. E.g. `{{- if gt (len ...` to remove the newline/whitespace before it.

another-rex (commented):
Hmm about changing to relative file paths, this does not work for outputs where we only have absolute paths (e.g. the example in fixtures/test-vuln-result-a.sarif), it is not a valid `uri`, which is why I changed it to always be an absolute path.

It seems like GitHub does support absolute paths: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#specifying-the-root-for-source-files

another-rex:
I can't see the sample report because I think the code scanning tab is only visible to maintainers.

oliverchang (commented):
> Hmm about changing to relative file paths, this does not work for outputs where we only have absolute paths (e.g. the example in fixtures/test-vuln-result-a.sarif), it is not a valid uri, which is why I changed it to always be an absolute path.
> 
> It seems like GitHub does support absolute paths: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#specifying-the-root-for-source-files

Yeah, but it didn't work for us :/ 

Even GitHub recommends using relative paths: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#physicallocation-object

I think for absolute paths to work, `invocation.workingDirectory.uri` has to be set: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#run-object

Should we just scope this SARIF support explicitly to GitHub for now? Trying to make this work more generally seems pretty complicated, because I'm not sure what path we'd set in other environments (or we'll have to get the user to provide their own).

oliverchang:
> I can't see the sample report because I think the code scanning tab is only visible to maintainers.

Added a screenshot to the description.

another-rex:
> > Hmm about changing to relative file paths, this does not work for outputs where we only have absolute paths (e.g. the example in fixtures/test-vuln-result-a.sarif), it is not a valid uri, which is why I changed it to always be an absolute path.
> > It seems like GitHub does support absolute paths: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#specifying-the-root-for-source-files
> 
> Yeah, but it didn't work for us :/
> 
> Even GitHub recommends using relative paths: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#physicallocation-object
> 
> I think for absolute paths to work, `invocation.workingDirectory.uri` has to be set: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#run-object
> 
> Should we just scope this SARIF support explicitly to GitHub for now? Trying to make this work more generally seems pretty complicated, because I'm not sure what path we'd set in other environments (or we'll have to get the user to provide their own).

A quick solution for now is to do the stripGithubWorkspace function, then check if the resulting path is still an absolute path, if it is, add the `file://` in, otherwise, just put the relative path in. This should make it be valid for most cases, while still working on github actions (if whoever is running it don't check the repo out to a custom location).

oliverchang:
> > > Hmm about changing to relative file paths, this does not work for outputs where we only have absolute paths (e.g. the example in fixtures/test-vuln-result-a.sarif), it is not a valid uri, which is why I changed it to always be an absolute path.
> > > It seems like GitHub does support absolute paths: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#specifying-the-root-for-source-files
> > 
> > 
> > Yeah, but it didn't work for us :/
> > Even GitHub recommends using relative paths: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#physicallocation-object
> > I think for absolute paths to work, `invocation.workingDirectory.uri` has to be set: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#run-object
> > Should we just scope this SARIF support explicitly to GitHub for now? Trying to make this work more generally seems pretty complicated, because I'm not sure what path we'd set in other environments (or we'll have to get the user to provide their own).
> 
> A quick solution for now is to do the stripGithubWorkspace function, then check if the resulting path is still an absolute path, if it is, add the `file://` in, otherwise, just put the relative path in. This should make it be valid for most cases, while still working on github actions (if whoever is running it don't check the repo out to a custom location).

Done!

another-rex (approved):

Issue #516, opened 2023-08-31
Document GitHub actions


oliverchang:

hayleycd:
Process has begun in #541

hayleycd:
#541 has been merged! The basic docs are ready to be merged into the docs branch whenever you would like them to go live. However, I am keeping this open because there are upcoming pieces of additional documentation. 

- [ ] Documenting how to use flags, once that is incorporated into the functionality

- [ ] Writing copy for getting the action up on marketplace

another-rex:
Flags have been added to the reusable actions. Let's add some examples for how to scan custom targets, what the defaults are, where to find all the flags.
E.g.:
```
# Scan specific lockifle
jobs:
  scan-pr:
    uses: "./.github/workflows/osv-scanner-reusable-pr.yml"
    with: 
      scan-args: |-
        ./path/to/lockfile1
        ./path/to/lockfile2
```

```
# Default
jobs:
  scan-pr:
    uses: "./.github/workflows/osv-scanner-reusable-pr.yml"
    with: 
      scan-args: |-
        -r
        --skip-git
        ./
```

hayleycd:
Thanks @another-rex

PR #552, opened 2023-09-26
Add custom scan arguments

another-rex:
Default's are maintained.

After this is merged we need to add more detailed documentation on how to use this as part of #516

### DIFF
--- a/.github/workflows/osv-scanner-reusable-pr.yml
+++ b/.github/workflows/osv-scanner-reusable-pr.yml
@@ -16,6 +16,14 @@ name: OSV-Scanner PR scanning reusable
 
 on:
   workflow_call:
+    inputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/osv-scanner-reusable-pr.yml
```diff
@@ -16,6 +16,14 @@ name: OSV-Scanner PR scanning reusable
 
 on:
   workflow_call:
+    inputs:
+      custom-scan-args:
```
oliverchang, line 20: nit: can  we just call this "scan-args"? From the user's perspective, "custom" seems unnecessary.

another-rex: Done

oliverchang (approved):

PR #553, opened 2023-09-26
ci: run tests on Windows

### LINKED ISSUE #603: Run tests against Windows
The test suite should be run against Windows

G-Rath:
Resolves #603


another-rex:
I think I fixed the (first) govulncheck problem with 734094c, the error it revealed I have to guess is because of line endings (I'm not sure why the line endings would change, but the +12 differences matches up with the line number/num of new lines). I think we just have to have two different test values for different platforms?

another-rex:
Oh I know why it has windows line endings, it's the same reason the SARIF test is failing, the github action git checkout is configured to automatically change line endings to the windows version when checking out on windows, just configuring it to not do that will fix both the SARIF error and the govulncheck test error

G-Rath:
> just configuring it to not do that will fix both the SARIF error and the govulncheck test error

I don't think doing that is the end of the world, but also is this something Go should actually be handling 🤔

I've inlined `json.MarshalIndent` and had it output `\r` if on Windows and the tests are now looking better (still not passing but for different reasons) - I'm not sure if it is valid for Go to just always output `\n` or if it should internally be sorting this; looking around though it also sounds like Go has decided this our problem to deal with not theirs...

PR #555, opened 2023-09-26
chore: move scripts into the `scripts` directory

G-Rath:
We spoke about this a few weeks ago but I never got around to actioning it - overall I've tried to kept it light since this is more about keeping things tidy than committing to a long-term structure.


another-rex (approved):
Thanks!

PR #558, opened 2023-09-27
test: use `cmp.Diff` for comparing output

G-Rath:
This makes it easier to determine what is wrong in CLI tests by showing an actual diff; while in most cases this is an improvement, I have found it's sometimes useful to have the raw output printed so I've included an env variable to allow switching easily with `cmp.Diff` being the default.

In order to reduce the diff noise when a test does fail, this also switches to replacing occurrences of the current working directory in the _actual_ output with `<rootdir>`; this also means that the output will be what should be in the test cases, rather than the absolute path that people would have to replace with `<rootdir>`.

While this could be used throughout the whole test suite, I've just applied this to the CLI tests for now because I think they've got the most to gain whereas it'd be a lot of tedious work to switch to using it everywhere; it should be easy to switch to using it in other places over time.

I have also confirmed that `dedent` correctly handles both spaces and tabs - you can mix and match them without issue (and in fact a few of the tests are using spaces instead of tabs).

### DIFF
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -105,6 +105,19 @@ func expectAreEqual(t *testing.T, subject, actual, expect string) {
 	}
 }
 
+// normalizeRootDirectory attempts to replace references to the current working
+// directory with "<rootdir>", in order to reduce the noise of the cmp diff
+func normalizeRootDirectory(t *testing.T, str string) string {
+	t.Helper()
+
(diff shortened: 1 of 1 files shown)


another-rex (approved):

another-rex:
I think the dedent assumes you're tab sizing of 2 (it replaces tab with 2 spaces), but I think we often have a tab sizing of 4, which is where we run into the issue

PR #559, opened 2023-09-28
SARIF with fixed version

another-rex:
Add fixed version in help text + unit tests

Also add `osv-reporter` to scheduled scanning, this allows scheduled scanning to also print out the vulnerability table as well as the SARIF.

Updated the unit tests to show a git like diff output for text comparisons to make it easier to see where things changed from the last snapshot. This does pull in 1 additional dependency however: github.com/hexops/gotextdiff


internal/output/fixtures/sarif-output.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+---
+
+### Affected Packages
+| Source | Package Name | Package Version |
+| --- | --- | --- |
+| lockfile:/path/to/sub-rust-project/Cargo.lock | regex | 1.5.1 |
+
+
```
oliverchang, line 111: Can you add some text around "remediation instructions" around here? 

This should say something along the lines of, update the offending package past the listed fix versions, or add this to the OSV config (link to documentation) to silence this.

another-rex: Done, @hayleycd can you take a look at the `sarif-output.md` file and see if there's anything you think would be better reworded? Thanks!

hayleycd: Hey @another-rex, can you give me some context on what this file is? Is this a sample of what a sarif output would look like? 

Does it change depending on what vulnerabilities are in my specific project? 

I am trying to understand the conditions under which a user would see this file. It doesn't look like we link to it in the docs at all?

another-rex: So this file is what you will see when looking at the output in `Security > Code Scanning` tab, e.g. 
image

This doc is an example of what will be displayed between "Your dependency is vulnerable to..." and "Show less ^". In this case the `sarif-output.md` is just a specific output we are using to test the behavior of the code generating the SARIF file, the actual "template" is located at the top of the `sarif.go` file (https://github.com/google/osv-scanner/pull/559/files#diff-d35339a2a0d00fb1bdef55bde8021cf88c25c1817a1d1f8afcfd734a1e80a784L29)

internal/output/fixtures/sarif-output.md
```diff
@@ -0,0 +1,136 @@
...
+If you believe these vulnerabilities do not affect your code and wish to ignore them, add them to the ignore list in an
+`osv-scanner.toml` file located in the same directory as the lockfile containing the vulnerable dependency.
+
+See the format and more options in our documentation here: https://google.github.io/osv-scanner/configuration/
+
+Example config to ignore this vulnerability:
+
+`path/to/lockfile-parent-dir/osv-scanner.toml`
```
oliverchang, line 131: Can we auto-populate these? 

We know where the lockfile lives, and we also know which IDs to ignore.

oliverchang (commented):
nice! Just one more minor suggestion to nail the UX.

internal/output/fixtures/sarif-output.md
another-rex: > Does it change depending on what vulnerabilities are in my specific project?

Yep, so the vulnerability ID, link, the contents of the Source table and Fixed table will change depending on what vulnerability this help entry is for and what lockfiles are found in the scan.

another-rex: >  Is this a sample of what a sarif output would look like?

Yep, though only a partial sample. A SARIF file can contain multiple rules (in our case ), and each rule can have a "helpText". This is what the helpText contains

internal/output/fixtures/sarif-output.md
another-rex: Yep, updated this so the path is automatically filled.

oliverchang (approved):

internal/output/fixtures/sarif-output.md
```diff
@@ -0,0 +1,138 @@
...
+> Unfortunately a bug was discovered in the mitigations designed to prevent
+> untrusted regexes to take an arbitrary amount of time during parsing, and it's
+> possible to craft regexes that bypass such mitigations. This makes it possible
+> to perform denial of service attacks by sending specially crafted regexes to
+> services accepting user-controlled, untrusted regexes.
+> 
+> ## Affected versions
+> 
+> All versions of the `regex` crate before or equal to 1.5.4 are affected by this
+> issue. The fix is include starting from  `regex` 1.5.5.
```
hayleycd, line 39: This should be included, not include. 

Not sure if you can adjust this line or if you need to adjust a different file because this is a sample output.

internal/output/fixtures/vuln-grouped.json
```diff
@@ -0,0 +1,206 @@
...
...signed to prevent untrusted regexes to take an arbitrary amount of time during parsing, and it's possible to craft regexes that bypass such mitigations. This makes it possible to perform denial of service attacks by sending specially crafted regexes to services accepting user-controlled, untrusted regexes.\n\n## Affected versions\n\nAll versions of the `regex` crate before or equal to 1.5.4 are affected by this issue. The fix is include starting from  `regex` 1.5.5.\n\n## Mitigations\n\nWe recommend everyone accepting user-controlled regexes to upgrade immediately to the latest version of the `regex` crate.\n\nUnfortunately there is no fixed set of problematic regexes, as there are practically infinite regexes that could be crafted to exploit this vulnerability. Because of this, we do not recommend denying known problematic regexes.\n\n## Acknowledgements\n\nWe want to thank Addison Crump for responsibly disclosing this to us according to the [Rust security policy](https://www.rust-lang.org/policies/security), and for helping review the fix.\n\nWe also want to thank Andrew Gallant for developing the fix, and Pietro Albini for coordinating the disclosure and writing this advisory.",
```
hayleycd, line 27: Maybe this is where you can fix the tense? Should be included, not include.

internal/output/fixtures/vuln-grouped.json
```diff
@@ -0,0 +1,206 @@
...
...trusted regexes to take an arbitrary amount of time during parsing, and it's\npossible to craft regexes that bypass such mitigations. This makes it possible\nto perform denial of service attacks by sending specially crafted regexes to\nservices accepting user-controlled, untrusted regexes.\n\n## Affected versions\n\nAll versions of the `regex` crate before or equal to 1.5.4 are affected by this\nissue. The fix is include starting from  `regex` 1.5.5.\n\n## Mitigations\n\nWe recommend everyone accepting user-controlled regexes to upgrade immediately\nto the latest version of the `regex` crate.\n\nUnfortunately there is no fixed set of problematic regexes, as there are\npractically infinite regexes that could be crafted to exploit this\nvulnerability. Because of this, we do not recommend denying known problematic\nregexes.\n\n## Acknowledgements\n\nWe want to thank Addison Crump for responsibly disclosing this to us according\nto the [Rust security policy][1], and for helping review the fix.\n\nWe also want to thank Andrew Gallant for developing the fix, and Pietro Albini\nfor coordinating the disclosure and writing this advisory.\n\n[1]: https://www.rust-lang.org/policies/security",
```
hayleycd, line 141: Wrong tense for include.

hayleycd (changes requested):
Fix the verb tenses noted and it'll be ready to go.

internal/output/fixtures/sarif-output.md
another-rex: We can't adjust any of the lines beginning with `>` since those are just quotes from the actual advisory. Only the lines that don't start with `>` can be changed. See https://github.com/google/osv-scanner/pull/559/files#diff-d35339a2a0d00fb1bdef55bde8021cf88c25c1817a1d1f8afcfd734a1e80a784L29 for the template that is used to generate these files.

oliverchang (approved):

PR #561, opened 2023-09-29
Block release on vuln scan

another-rex:
`osv-scanner-reusable-scheduled` actually doesn't have anything to do with having a schedule, so just removed the scheduled part from the name. 

This reusable workflow is now also used in the release pipeline to block releases if there are vulnerabilities.


hayleycd (approved):
The documentation section looks good to me.

oliverchang (approved):

PR #575, opened 2023-10-06
Improve documentation for github actions

another-rex:
Attempt at the 3 remaining items mentioned in https://github.com/google/osv-scanner/issues/516. 

Feel free to do any rewording or rearranging @hayleycd to make things clearer!

Also fix the SARIF test since it was still calling `createTextFixture` instead of `assertTextFixture`

### DIFF
--- a/docs/github-action.md
+++ b/docs/github-action.md
@@ -14,6 +14,78 @@ OSV-Scanner is offered as a GitHub Action. We currently have two different GitHu
 1. An action that performs a single vulnerability scan, which can be configured to scan on a [regular schedule](./github-action.md#scheduled-scans), or used as a check [on releases](./github-action.md#scan-on-release) to prevent releasing with known vulnerabilities in dependencies.
 2. An action that triggers a scan with each [pull request](./github-action.md#scans-on-prs) and will only check for new vulnerabilities introduced through the pull request.
 
+## Scans on PRs
+
+Scanning your project on each pull request can help you keep vulnerabilities out of your project. This GitHub Action compares a vulnerability scan of the main branch to a vulnerability scan of the feature branch. You will be notified of any new vulnerabilities introduced through the feature branch. You can also choose to [prevent merging](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging) if new vulnerabilities are introduced through the feature branch.
+
+### Instructions
+
+In your project repository, create a new file `.github/workflows/osv-scanner-pr.yml`.
+
+Include the following in the [`osv-scanner-pr.yml`](https://github.com/google/osv-scanner/blob/main/.github/workflows/osv-scanner-pr.yml) file:
+
+```yml
+name: OSV-Scanner PR Scan
+
+# Change "main" to your default branch if you use a different name, i.e. "master"
+on:
+  pull_request:
+    branches: [ main ]
+  merge_group:
+    branches: [ main ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  scan-pr:
+    uses: "google/osv-scanner/.github/workflows/osv-scanner-reusable-pr.yml@main"
+```
+
+### View results
+
+Results may be viewed by clicking on the details of the failed action, either from your project's actions tab or directly on the PR. Results are also included in GitHub annotations on the "Files changed" tab for the PR.
+
+### Customization
+
+`osv-scanner-reusable.yml` takes two optional inputs:
+
+- `scan-args`: This value is passed to `osv-scanner` CLI after being split by each line. See the [usage]()./usage) page for the available options.
+  Importantly `--format` and `--output` flags are already set by the reusable workflow and should not be overridden here.
+  Default:
+  ```
+    --recursive # Recursively scan subdirectories
+    --skip-git=true # Skip commit scanning to focus on dependencies
+    ./ # Start the scan from the root of the repository
+  ```
+- `results-file-name`: This is the name of the final SARIF file uploaded to Github.
+  Default: `results.sarif`
+
+#### Examples
+```yml
+# Scan specific lockfiles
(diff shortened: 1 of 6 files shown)


hayleycd:
Can you build the docs from this branch so we can preview them? @another-rex

docs/github-action.md
```diff
@@ -11,7 +11,7 @@ nav_order: 6
 
 OSV-Scanner is offered as a GitHub Action. We currently have two different GitHub Actions:
 
-1. An action that performs a vulnerability scan on a [regular schedule](./github-action.md#scheduled-scans).
+1. An action that performs a single vulnerability scan, which can be configured to scan on a [regular schedule](./github-action.md#scheduled-scans), or used as a check [on releases](./github-action.md#scan-on-release) to prevent releasing with known vulnerabilities in dependencies.
 2. An action that triggers a scan with each [pull request](./github-action.md#scans-on-prs) and will only check for new vulnerabilities introduced through the pull request.
 
 ## Scheduled scans
```
hayleycd, line 17: Maybe we want to rename this section, as we now run scans on either releases or on a schedule.

docs/github-action.md
```diff
@@ -11,7 +11,7 @@ nav_order: 6
 
 OSV-Scanner is offered as a GitHub Action. We currently have two different GitHub Actions:
 
-1. An action that performs a vulnerability scan on a [regular schedule](./github-action.md#scheduled-scans).
+1. An action that performs a single vulnerability scan, which can be configured to scan on a [regular schedule](./github-action.md#scheduled-scans), or used as a check [on releases](./github-action.md#scan-on-release) to prevent releasing with known vulnerabilities in dependencies.
```
hayleycd, line 14: Hey @another-rex . 

I tried to add commits to the PR but I don't think I have the right permissions, so I am going to comment instead. 

Please either:
1. Make these changes before merging
2. Update permissions so I can add commits. I tried to follow [these instructions] (https://tighten.com/insights/adding-commits-to-a-pull-request/) and got an error message that said "fatal: Could not read from remote repository.Please make sure you have the correct access rights and the repository exists." 
3. Merge it anyway and I can do another PR with my updates. Nothing is wrong, just not optimal so it would be ok to merge.

hayleycd, line 14: It isn't good practice that we have the scheduled scan and the release scan grouped together in this first item in the list and then they aren't grouped that way in the instructions below which are in this order:
1. Scheduled scan
2. PR scan
3. Release scan

Would you prefer to update the order of the instructions, or how they are grouped in lines 14 & 15?

docs/github-action.md
```diff
@@ -47,9 +47,39 @@ jobs:
 
 As written, the scanner will run on 12:12 pm UTC every Monday. You can change the schedule by following the instructions [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).
 
+### Customization
+
+`osv-scanner-reusable.yml` takes the `scan-args` parameter as input, this value is passed to `osv-scanner` CLI after being split by each line. See the [usage]()./usage) page for the available options.
```
hayleycd, line 52: I think we may want to be a little more direct about what `osv-scanner-reuseable.yml` is. If I was newer to GitHub Actions I might be confused because I don't need this file in my project.

docs/github-action.md
```diff
@@ -47,9 +47,39 @@ jobs:
...
+
+Importantly `--format` and `--output` flags are already set by the reusable workflow and should not be overridden here.
```
hayleycd, line 54: I think it would also be good practice to explain the defaults.

hayleycd:
The GitHub Action page is getting a little long, I would include a table of contents like some of the other pages have. I can put this in if you would like me to make my changes in a separate PR.

hayleycd (approved):
I am approving these.

I have commented some changes I would like to see, but they don't necessarily need to go in this PR. Please see the options in [this comment](https://github.com/google/osv-scanner/pull/575/files/c75ac35506e6dbd8b0002b502a89f86a3bc0cae5#r1349755229) and let me know what you prefer.

another-rex:
Should now be previewing this branch here: https://another-rex.github.io/osv-scanner/

another-rex:
@hayleycd Updated your permissions to maintainer, you should be able to push to PR branches now.

docs/github-action.md
another-rex: That makes sense, let's change it to this order so that scheduled and release is together:

1. PR scan
2. Scheduled scan
3. Release scan

Should we still separate out the two entries on line 14&15? We'll probably have to reword it to something like 
> We currently support 3 different ways to use Github Actions:
>
> 1. PR Scan ...
> 2. Scheduled Scan ...
> 3. Release Scan ...
>

another-rex: Thanks for the comments! Added permissions to allow you to push, do you want me to make the changes you proposed, or would you like to push the changes on your side, and I'll just review them?

docs/github-action.md
another-rex: Good point, maybe adding something like

> In your github action file (`.github/workflows/osv-scanner-scheduled.yml`)...

could point the user at the file they are meant to edit.

docs/github-action.md
another-rex:
```suggestion
Importantly `--format` and `--output` flags are already set by the reusable workflow and should not be overridden here.

The default behavior will recursively scan your repository from the root, but skip commit scanning to focus on dependencies.
```

docs/github-action.md
another-rex: Updated the ordering in f62f6c3

docs/github-action.md
another-rex: Added defaults value as a code block, and added inline comments describing what each line does

hayleycd:
@another-rex , I tried to add commits again with no luck. I think it would be best to merge this, and I will follow up with a few small changes in a PR tomorrow.

PR #590, opened 2023-10-13
fix: use the first version when a dependency is listed multiple times in `pom.xml` files

### LINKED ISSUE #589: Extract multiple versions of the same package from pom.xml
Some `pom.xml` files contain multiple versions of the same package. For example:
```
<dependency>
  <groupId>com.fasterxml.jackson.core</groupId>
  <artifactId>jackson-databind</artifactId>
  <version>2.13.3</version>
</dependency>

<dependency>
  <groupId>com.fasterxml.jackson.core</groupId>
  <artifactId>jackson-databind</artifactId>
  <version>2.13.4</version>
</dependency>

<dependency>
  <groupId>com.fasterxml.jackson.core</groupId>
  <artifactId>jackson-databind</artifactId>
  <version>2.13.4.1</version>
</dependency>
```

Since it's hard to know which will end up being installed / used, it might be best to extract all of them. Currently, only the last package is retained because extraction is keyed on `groupId` and `artifactId`: https://github.com/google/osv-scanner/blob/main/pkg/lockfile/parse-maven-lock.go#L124

G-Rath:
Resolves #589

Per https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies

> Note that if two dependency versions are at the same depth in the dependency tree, the first declaration wins.

It doesn't seem to explicitly mention what happens when there are duplicate packages for dependency management so I've assumed it follows the same behaviour.

### DIFF
--- a/pkg/lockfile/parse-maven-lock.go
+++ b/pkg/lockfile/parse-maven-lock.go
@@ -68,6 +68,28 @@ type MavenLockFile struct {
 	ManagedDependencies []MavenLockDependency `xml:"dependencyManagement>dependencies>dependency"`
 }
 
+func (mlf *MavenLockFile) collectPackages(dependencies []MavenLockDependency) map[string]PackageDetails {
+	details := map[string]PackageDetails{}
+
+	for _, lockPackage := range dependencies {
+		finalName := lockPackage.GroupID + ":" + lockPackage.ArtifactID
+
+		// the first declaration of a dependency at a specific depth wins
+		if _, ok := details[finalName]; ok {
+			continue
+		}
+
+		details[finalName] = PackageDetails{
+			Name:      finalName,
+			Version:   lockPackage.ResolveVersion(*mlf),
+			Ecosystem: MavenEcosystem,
+			CompareAs: MavenEcosystem,
+		}
+	}
+
+	return details
+}
+
 const MavenEcosystem Ecosystem = "Maven"
 
 type MavenLockProperties struct {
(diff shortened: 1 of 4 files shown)


pkg/lockfile/parse-maven-lock.go
```diff
@@ -68,6 +68,28 @@ type MavenLockFile struct {
 	ManagedDependencies []MavenLockDependency `xml:"dependencyManagement>dependencies>dependency"`
 }
 
+func (mlf *MavenLockFile) collectPackages(dependencies []MavenLockDependency) map[string]PackageDetails {
```
cuixq, line 71: maybe add comments stating that our implementation follows Maven's documentation but it is different from how `mvn` actually does

oliverchang: I think we should follow `mvn`s rules over documentation, and reflect real world behaviour (and document this).

G-Rath: I did this before I knew that `mvn` doesn't actually follow this - I think that's a far bigger thing that should be focused on first; if `mvn` doesn't follow its own spec here then I think 1. that's a huge bug in `mvn` (or its docs) which someone should engage with the community to get at least clarified and 2. we should probably just say the current implementation is fine as I imagine it's more realistic to be doing what `mvn` actually _does_ rather than what it says

pkg/lockfile/parse-maven-lock.go
```diff
@@ -116,29 +138,11 @@ func (e MavenLockExtractor) Extract(f DepFile) ([]PackageDetails, error) {
...
-		details[finalName] = PackageDetails{
-			Name:      finalName,
-			Version:   lockPackage.ResolveVersion(*parsedLockfile),
-			Ecosystem: MavenEcosystem,
-			CompareAs: MavenEcosystem,
-		}
-	}
+	details := parsedLockfile.collectPackages(parsedLockfile.Dependencies)
 
 	// managed dependencies take precedent over standard dependencies
```
cuixq, line 143: I am curious about why managed dependencies take precedent over standard dependencies?

G-Rath: Because [they take precedent](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-management):

> * b is defined in B's parent's dependency management section and since **dependency management takes precedence over dependency mediation for transitive dependencies**, version 1.0 will be selected should it be referenced in a or c's POM. b will also have compile scope.
> * Finally, since d is specified in B's dependency management section, should d be a dependency (or transitive dependency) of a or c, version 1.0 will be chosen - again **because dependency management takes precedence over dependency mediation** and also because the current POM's declaration takes precedence over its parent's declaration.

G-Rath:
Closing per https://github.com/google/osv-scanner/pull/590#discussion_r1359983607

pkg/lockfile/parse-maven-lock.go
cuixq: my understanding is that dependency management takes precedence
 - for *transitive* dependencies
 - when version requirement is not specified for a direct dependencies

So for the dependencies listed in `with-dependency-management-and-duplicates.xml`, versions of direct dependencies should be reported.

feel free to correct me if this is not accurate!

G-Rath: I don't use Maven and don't really do a lot of Java work so I'm only going off the docs and #214 - I agree that it sounds like there could be cases where the behaviour should be different, but I'm not in a good position to accurately confirm.

I think let's open a new issue to discuss and go from there - ultimately, I'm happy to just be told what the logic should be and do the implementation if you've got a better understanding of how Maven is doing things 🙂

(also just to make sure we're on the same page: this is _replacing_ dependencies, not removing them - so the result currently is the _combination_ of both groups of dependencies, and that if the same dependency is in both then the version in `dependencyManagement` wins)

cuixq: sounds good - thank you!

PR #593, opened 2023-10-17
Update docs to add in saving to file option

another-rex:
- Update docs to add in saving to file option
- Fix the output page toc to not include the example's headers


hayleycd (approved):
LGTM

Assuming you confirmed that removing the examples worked.

another-rex:
Yep, ran it locally and it worked.

PR #598, opened 2023-10-23
Gate extended tests

another-rex:
Put tests that require additional dependencies beyond the go toolchain behind the CI flag to not force all developers to install all dependencies (this becomes more important as more source analysis support is added).


G-Rath:
Actually thinking about it maybe it would be better to leverage the fact that the scanner should be failing gracefully anyway - this PR could still be fine to land for now, but in the long-run I think the scanner should probably have some sort of `ToolNotPresentErr` that gets raised early for these situations, and then instead of just skipping the tests you could have logic that says something like:

```
if errors.Is(err, ToolNotPresentErr) {
  if testutils.IgnoreToolNotPresentErrors {
    t.Skipf("Skipping extended test since tools are not present: %v", err)
  } else {
    t.Errorf("Cannot run extended test since tools are not present: %v", err)
  }
}
```

internal/testutility/utility.go
```diff
@@ -100,3 +100,12 @@ func AssertMatchFixtureText(t *testing.T, path string, actual string) {
 		}
 	}
 }
+
+// AcceptanceTests marks this test function as a extended that require additional dependencies
+// automatically skipped unless running in a CI environment
+func AcceptanceTests(t *testing.T, reason string) {
+	t.Helper()
+	if os.Getenv("Acceptance_Tests") == "" {
```
G-Rath, line 108: env variables traditionally are all upper case

G-Rath (changes requested):

internal/testutility/utility.go
```diff
@@ -105,7 +105,7 @@ func AssertMatchFixtureText(t *testing.T, path string, actual string) {
 // automatically skipped unless running in a CI environment
 func AcceptanceTests(t *testing.T, reason string) {
 	t.Helper()
-	if os.Getenv("Acceptance_Tests") == "" {
+	if os.Getenv("ACCEPTANCE_TESTS") == "" {
```
G-Rath, line 108: I think we should probably actually check for a value e.g. `1`:

```suggestion
	if os.Getenv("ACCEPTANCE_TESTS") == "1" {
```

G-Rath (approved):

internal/testutility/utility.go
michaelkedar: I think this is meant to be checking that it's not set, though [`os.LookupEnv`](https://pkg.go.dev/os#LookupEnv) seems to be the more correct way to do it

michaelkedar (approved):

internal/testutility/utility.go
G-Rath: Right good point, have updated my suggestion - `os.LookupEnv` is not inherently the correct way, because it checks if its set at all, not if it's both set _and_ set to an empty value.

I personally think at least in this case `os.LookupEnv` isn't what we want because that typically makes it more work to disable i.e. I have to completely remove `ACCEPTANCE_TESTS` from my commands rather than just setting it to an alternative value

another-rex: I set it to `true` to match the `TEST_NO_DIFF` env we currently have.

G-Rath: actually that's a good point 😅 probably best to call it `TEST_ACCEPTANCE` for consistency...

PR #602, opened 2023-10-24
Update release pipeline

another-rex:
Update the release pipeline to have it be manually triggered, and create a tag itself if checks and scans are successful.

It looks something like this:
image

The release doc has been updated to follow this new workflow.


michaelkedar (approved):

suzuki-shunsuke:
This change caused an issue.

- https://github.com/google/osv-scanner/issues/632

PR #606, opened 2023-10-24
fix: trim leading and trailing newlines off SARIF output

G-Rath:
I don't know a lot about SARIF so maybe these are required but that would be surprising to me whereas trimming these off make it a bit consistent for editors and such.

Obviously if these are required then lets close this PR


another-rex (approved):

G-Rath:
image

😞

another-rex:
Oh, did you used to just press the delete branch button?

G-Rath:
Yeah that's what I've always been doing on this repo and e.g. what I did after posting that screenshot - I guess you were right it just doesn't work across forks 🤷

Issue #588, opened 2023-10-13
NPM aliased packages are not handled correctly


michaelkedar:
In NPM, you can install packages under a different name e.g. from [@isaacs/cliui](https://www.npmjs.com/package/@isaacs/cliui/v/8.0.2):
```json
"dependencies": {
  "string-width": "^5.1.2",
  "string-width-cjs": "npm:string-width@^4.2.0",
  "strip-ansi": "^7.0.1",
  "strip-ansi-cjs": "npm:strip-ansi@^6.0.1",
  "wrap-ansi": "^8.1.0",
  "wrap-ansi-cjs": "npm:wrap-ansi@^7.0.0"
},
```

Which seems to show up in the package-lock.json as
```json
"node_modules/string-width-cjs": {
  "name": "string-width",
  "version": "4.2.3",
  "resolved": "https://registry.npmjs.org/string-width/-/string-width-4.2.3.tgz",
  ...
},
```
~~(not sure what it looks like in the v1 format)~~ **Edit:** seems like it shows up in the old format as
```json
"string-width-cjs": {
  "version": "npm:string-width@4.2.3",
  "resolved": "https://registry.npmjs.org/string-width/-/string-width-4.2.3.tgz",
  ...
},
```
(but I don't think version of npm that only use this format actually allows aliases...)

I believe osv-scanner will use the aliased name when looking for vulnerabilities, which can potentially cause false negatives (or false positives maybe, if the alias is a vulnerable package?)

We should also look into if other ecosystems have a similar issue.

oliverchang:
@cuixq this is a simple first issue if you're looking for a smaller code task to tackle.

PR #610, opened 2023-10-25
handle npm aliased packages

cuixq:
Fixes https://github.com/google/osv-scanner/issues/588

### DIFF
--- a/pkg/lockfile/parse-npm-lock.go
+++ b/pkg/lockfile/parse-npm-lock.go
@@ -70,7 +70,7 @@ func parseNpmLockDependencies(dependencies map[string]NpmLockDependency) map[str
 
 		// If the package is aliased, get the name and version
 		if strings.HasPrefix(detail.Version, "npm:") {
-			i := strings.Index(detail.Version, "@")
+			i := strings.LastIndex(detail.Version, "@")
 			name = detail.Version[4:i]
 			finalVersion = detail.Version[i+1:]
 		}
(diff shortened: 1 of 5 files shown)


another-rex (commented):
LGTM, though can you add the test case in the issue where there's two dependencies of the same package (`string-width: ^5.1.2` and `string-width-cjs: npm:string-width@^4.2.0`)

pkg/lockfile/parse-npm-lock.go
```diff
@@ -65,6 +68,13 @@ func parseNpmLockDependencies(dependencies map[string]NpmLockDependency) map[str
 		finalVersion := version
 		commit := ""
 
+		// If the package is aliased, get the name and version
+		if strings.HasPrefix(detail.Version, "npm:") {
+			i := strings.Index(detail.Version, "@")
```
michaelkedar, line 73: This will need to use `strings.LastIndex` to handle packages with scopes e.g. `npm:@foo/bar@1.2.3`

michaelkedar (commented):
LGTM, just one thing

pkg/lockfile/parse-npm-lock.go
cuixq: good point! I have also added a test case for this.

another-rex (approved):
LGTM, thanks!

michaelkedar (approved):

PR #614, opened 2023-10-27
Refactor package scanning to produce packages instead of queries

josieang:
this is in preparation for the license scanning feature. the queries are structured around making requests to the osv API, we also will want to make requests to the deps.dev api.
https://github.com/google/osv-scanner/pull/501


pkg/osvscanner/osvscanner.go
```diff
@@ -398,28 +410,27 @@ func scanDebianDocker(r reporter.Reporter, query *osv.BatchedQuery, dockerImageN
...
-		}
-		query.Queries = append(query.Queries, pkgDetailsQuery)
-		packages += 1
+		packagesCount += 1
 	}
 	r.PrintText(fmt.Sprintf(
 		"Scanned docker image with %d %s\n",
-		packages,
-		output.Form(packages, "package", "packages"),
+		packagesCount,
```
another-rex, line 429: Remove packageCount and just use `len(packages)`

pkg/osvscanner/osvscanner.go
```diff
@@ -350,45 +362,45 @@ func getCommitSHA(repoDir string) (string, error) {
...
-	return scanGitCommit(query, commit, repoDir)
+	return scanGitCommit(commit, repoDir), nil
 }
 
-func scanGitCommit(query *osv.BatchedQuery, commit string, source string) error {
-	gitQuery := osv.MakeCommitRequest(commit)
-	gitQuery.Source = models.SourceInfo{
-		Path: source,
-		Type: "git",
+func scanGitCommit(commit string, source string) Package {
```
another-rex, line 375: rename this to `createCommitQueryPackage`

pkg/osvscanner/osvscanner.go
```diff
@@ -275,12 +286,13 @@ func scanSBOMFile(r reporter.Reporter, query *osv.BatchedQuery, path string, fro
...
-			}
-			query.Queries = append(query.Queries, purlQuery)
+			packages = append(packages, Package{
+				PURL: id.PURL,
+				Source: models.SourceInfo{
+					Path: path,
+					Type: "sbom",
+				},
+			})
 			count++
```
another-rex, line 296: `count` is just `len(packages)` now.

pkg/osvscanner/osvscanner.go
```diff
@@ -617,11 +639,30 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
 	compareOffline bool,
-	query osv.BatchedQuery,
-	localDBPath string,
-) (*osv.HydratedBatchedResponse, error) {
+	localDBPath string) (*osv.HydratedBatchedResponse, error) {
+	// Make OSV queries from the packages.
+	var query osv.BatchedQuery
+	for _, p := range packages {
+		switch {
+		// Prefer making paackage requests where possible.
```
another-rex, line 650:
```suggestion
		// Prefer making package requests where possible.
```

pkg/osvscanner/osvscanner.go
```diff
@@ -505,6 +516,15 @@ func parseLockfilePath(lockfileElem string) (string, string) {
 	return splits[0], splits[1]
 }
 
+type Package struct {
```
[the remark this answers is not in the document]
another-rex: Hmm... now that I think about it, do we even need this intermediate type? It looks like what we're looking for is just the `osv.Query` type, but rather than being appended to a pointer being passed in, we return it as an array. 

This should reduce the number of changes while still accomplishing what you want right?

another-rex (commented):
Thanks! Added some comments, mainly about whether we need the Package struct or if we can reuse the Query struct, but still follow the return pattern in this refactor rather than the original appending to a big slice.

pkg/osvscanner/osvscanner.go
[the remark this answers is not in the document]
josieang: I make a separate Package object because it doesn't make semantic sense to use an osv.Query object to create a depsdev query. While it might be possible to do so, I think it's a misuse of the meaning of the osv.Query object and we should be making osv and depsdev queries from an abstracted package object instead.

In the case where commit information and package information are available, it's valid to just construct an osv.Query that just contains just the commit [1]. in this case we wouldn't be able to construct a depsdev query from the osv query. I'm not sure whether this scenario exists yet, but I think it illustrates why it's cleaner to have a separate package object.

As discussed I'll make this a private type and submit it as is!

another-rex (approved):
LGTM!

PR #615, opened 2023-10-29
fix: handle yarn aliased packages

### LINKED ISSUE #588: NPM aliased packages are not handled correctly

G-Rath:
Resolves #588
Follows up #610

Because of how PNPM structures its lockfile, aliases are already supported there


another-rex (approved):
LGTM thanks!

PR #627, opened 2023-11-01
test: don't ignore anything from coverage

G-Rath:
I'm honestly sure CodeCov isn't bugged (or that somehow we're uploading results from a branch...), but removing this file/ignore increases coverage a little so why not 🤷

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,3 @@
-ignore: []
-#  - internal/testutility
+ignore:
+  - internal/testutility
 #  - 'internal/**'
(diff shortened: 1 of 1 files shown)


another-rex:
Probably need to update the branch to show correct coverage change.

G-Rath:
Na it's been a bit weird in general - no matter what I do it seems want to go down a fair bit, and it was like that on e.g. #628 too

I suspect it might be we have to land this on `main` before the % will be properly accurate; alternatively I'm thinking it might actually just be completely tossing out the coverage while still including these in the line count which would always result in that worse outcome...

another-rex (approved):

PR #628, opened 2023-11-02
test: check against error type rather than message

G-Rath:
Cherry-picked from #553

---

It's required for testing against Windows because it has a different error message, but it's also just a good overall change and landing it separately removes ~25 files from the main PR 😅


another-rex (approved):
LGTM!

PR #637, opened 2023-11-03
upgrade version of Go in GitHub checks

josieang:
I want to use errors.Join in the following PR:
- License checker feature #501

It is a method added in go 1.20.


another-rex:
Can you also change the go version in the go.mod?

G-Rath:
Isn't this a breaking change though for the library?

another-rex:
I was thinking about that, but it looks like go 1.19 is out of support anyway? https://endoflife.date/go

I'm actually not sure how libraries are suppose to bump versions if we have to do a major version change every go release.

G-Rath:
I thought it that libraries don't need to worry EOL because that only matters for the version of Go that is doing the compiling - i.e. osv-scanner (the cli) should be being built with an in-life version of Go, but osv-scanner (the library) can remain compatible with EOL version of Go

josieang:
I opt for upgrading the version of Go. And that this shouldn't require a major version bump.
- Go is backwards compatible, upgrading is encouraged by the Go project. Upgrading should be a good thing and easy to do. I don't think it makes sense for osv-scanner to support people stuck at Go 1.19.
- I don't think osv-scanner should only use old features of Go until a major version bump. This is much more like a minor change.

However, we might like to make a policy that osv-scanner will user versions of Go at least two release behind or something.
https://github.com/google/osv-scanner/issues/638

another-rex (approved):
LGTM, we'll increase them minor version of osv-scanner when this releases.

PR #643, opened 2023-11-06
Filter local packages from scanning, and report the filtering.

### LINKED ISSUE #639: Scanner fails on C# Solution with "does not have a commit, PURL or ecosystem/name/version identifier"
Since the update to v1.4.3, the scanner fails to correctly scan a C# solution with many projects that depend on each other. The error is the following (full log):

```
Scanning dir .
Scanning /src/ at commit fa0451a287645ac3fdc9f11635150fac4fd73802
Scanning submodule Bricks.Module.Core at commit e62e6bc84277e5347a424282aea2f47328bb874e
Scanned /src/Brixit.Flows/Brixit.Apis/packages.lock.json file and found 7 packages
Scanned /src/Brixit.Flows/Brixit.Apis.Generator/packages.lock.json file and found 90 packages
Scanned /src/Brixit.Flows/Brixit.Apis.Hrworks/packages.lock.json file and found 65 packages
Scanned /src/Brixit.Flows/Brixit.Apis.Hrworks.Tests/packages.lock.json file and found 111 packages
...
Scanned /src/Common/Common.Extensions/packages.lock.json file and found 0 packages
package { brixit.apis NuGet   lockfile:/src/Brixit.Flows/Brixit.Apis.Hrworks/packages.lock.json} does not have a commit, PURL or ecosystem/name/version identifier
```

The submodule is currently not checked out. If I check out the submodule first, the scanner instead reports the same problem on one of the submodule packages.

```
Scanning dir .
Scanning /src/ at commit fa0451a287645ac3fdc9f11635150fac4fd73802
Scanning submodule Bricks.Module.Core at commit e62e6bc84277e5347a424282aea2f47328bb874e
Scanned /src/Bricks.Module.Core/Bricks.Module.Core/packages.lock.json file and found 90 packages
Scanned /src/Bricks.Module.Core/Bricks.Test.Module.Core/packages.lock.json file and found 214 packages
Scanned /src/Brixit.Flows/Brixit.Apis/packages.lock.json file and found 7 packages
...
Scanned /src/Common/Common.Extensions/packages.lock.json file and found 0 packages
package { bricks.module.core NuGet   lockfile:/src/Bricks.Module.Core/Bricks.Test.Module.Core/packages.lock.json} does not have a commit, PURL or ecosystem/name/version identifier
```

I downgraded to v1.4.2 and there it works again. Can reproduce it with multiple projects, both locally on macOS and in the CI with an ubuntu cloud machine.

another-rex:
Filter local packages from scanning, and report the filtering happened. Fixes #639 

Also added a test for this case in main_test.go

Added another rust test of package on a local path, we currently have no way to differentiate local rust packages and ones from the repository. Perhaps something to consider in the future.


pkg/osvscanner/osvscanner.go
```diff
@@ -864,7 +890,7 @@ func makeRequest(
 		case p.PURL != "":
 			query.Queries = append(query.Queries, osv.MakePURLRequest(p.PURL))
 		default:
-			return nil, fmt.Errorf("package %v does not have a commit, PURL or ecosystem/name/version identifier", p)
+			log.Panic("packages received in makeRequest should have already been filtered")
```
oliverchang, line 893: nit: this panic message is a little tied to the implementation details of osvscanner as a whole -- i.e. this assumes knowledge about the caller's behaviour. 

Should we just return an error here insteaD?

another-rex: Good point, changed back to returning  the error

pkg/osvscanner/osvscanner.go
```diff
@@ -842,6 +848,25 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
+// filterUnscannablePackages removes packages that don't have enough information to be scanned
+// e,g, local packages that specified by path
+func filterUnscannablePackages(packages []scannedPackage) []scannedPackage {
+	out := make([]scannedPackage, 0, len(packages))
+	for _, p := range packages {
+		switch {
+		// If none of the cases match, skip this package since it's not scannable
+		case p.Ecosystem != "" && p.Name != "" && p.Version != "":
+		case p.Commit != "":
+		case p.PURL != "":
```
oliverchang, line 860: nit: maybe we should be validating that the PURL has a version as well? Or is this not possible in practice with our current extraction?

oliverchang (approved):
LGTM

josieang (approved):

another-rex:
Talked offline with Josie about the license filtering, the ideal case would be to have the vendored/local/packages without version be reported as Unknown when reporting license information and when --experimental-all-packages is used, and this change will filter out those packages. 

Because of the complexity of doing that, I'll create an issue for now for it and come back to fix it later.

pkg/osvscanner/osvscanner.go
another-rex: We probably can, though we'll need to do purl parsing. I'll do this with the proper change in #644 .

PR #645, opened 2023-11-06
fix: properly handle file/url paths on Windows

G-Rath:
Turns out that file -> url translation on Windows is busted, and that this is a hard problem that Go has an internal util for that has not yet been made public - I've done what apparently a number of other packages have done which is copying that helper into here and hoping one day it actually becomes public 😢

Note that until #646 is landed, there is no way to actually verify this is fixing the problem - #553 shows the result of both PRs being merged.

### DIFF
--- a/internal/output/sarif.go
+++ b/internal/output/sarif.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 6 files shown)


another-rex (approved):
Lgtm!

PR #647, opened 2023-11-06
Fix filter ids

### LINKED ISSUE #634: Github action still reports aliases of suppressed vulnerabilities
After adding the OSV-Scanner github action to [Package Analysis](https://github.com/ossf/package-analysis), ([PR action config](https://github.com/ossf/package-analysis/blob/main/.github/workflows/osv-scanner-pr.yml), [scheduled action config](https://github.com/ossf/package-analysis/blob/main/.github/workflows/osv-scanner-scheduled.yml)), the action picked up the CVE-2020-8911 vulnerability.

Since this does not actually affect us, we followed the suggested remediation steps and suppressed it using [this osv-scanner.toml config](https://github.com/ossf/package-analysis/blob/main/osv-scanner.toml).

However, the OSV-scanner job still fails on subsequent workflow runs and reports the presence of [`GO-2022-0646`](https://osv.dev/vulnerability/GO-2022-0646), which is an alias of `CVE-2020-8911`.

another-rex:
Fixes #634 

The actual change is just adding an `Aliases` field to the Group output, that combines all the IDs and aliases together. A lot of fixtures had to be updated though. 

Added an additional test for this in `main_test`, and also modified a test in `osvscanner_internal_tests.go`

Also added `omitempty` tag to `PackageInfo.commit` which it should have contained in the first place.

### DIFF
--- a/cmd/osv-scanner/fixtures/locks-test-ignore/osv-scanner.toml
+++ b/cmd/osv-scanner/fixtures/locks-test-ignore/osv-scanner.toml
@@ -1,0 +1,3 @@
+[[IgnoredVulns]]
+id = "CVE-2021-23424"
+reason = "Test manifest file (alpine.cdx.xml)"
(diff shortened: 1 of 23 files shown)


oliverchang (approved):

oliverchang:
(sorry for the delay. completely dropped the ball on this one).

Issue #332, opened 2023-04-04
Distinguish between normal and dev dependencies.


michaelkedar:
There is no current way to identify whether a vulnerable dependency only affects the development environment e.g. whether it comes from `devDependencies` or `dependencies` in `package.json`.

We should add a field to the output to present this information, if we can extract it.

aniruth37:
I tried adding a column called `DevDependency` to the output table. Would the below output resolve this issue?

![image](https://github.com/google/osv-scanner/assets/36035280/7221c0d7-63ee-48ef-b2a2-424f8b3dd733)

another-rex:
We are thinking of placing the devDependencies in a separate table under the normal dependencies. 

So if we identify any dep dependencies, we will print two tables, with the first table containing only main deps, and a second table containing dev deps.

G-Rath:
@another-rex probably better to have support for marking "groups" that packages belong to, to support arbitrary groupings and have the default of no group take priority (since that is typically what are considered production dependencies); then in your output you just print-by-group and naturally sort "no group" to the top

oliverchang:
@cuixq This might be an interesting starter project to tackle.

PR #655, opened 2023-11-15
scan and report dependency groups of vulnerabilities

cuixq:
Issue https://github.com/google/osv-scanner/issues/332

Non-default dependency groups are recorded in string as per eco-system:
 - **Composer:** development dependencies in `packages-dev`
 - **Conan:** dependencies in `build-requires` and `python-requires`
 - **Maven:** `<scope/>` in `<dependency/>`
 - **npm:** `dev` and `optional` dependencies
 - **pipenv:** development dependencies in `develop`
 - **pnpm:** development dependencies with `dev` as true
 - **Poetry:** optional dependencies with `optional = true`
 - **Pubspec:** development dependencies marked with `dev`
 - **requirements.txt:** development requirements are defined in `dev.txt`

Reporters:
 - **table:** non-default groups are appended to the end of package name, for example: `abc (development)`
 - **json:** non-default group information in `dependencyGroup`
 - **sarif:** non-default groups are appended to the end of package name and version, for example: `Package 'abc@1.2.3 (development)'`

### DIFF
--- a/internal/output/table.go
+++ b/internal/output/table.go
@@ -128,7 +128,11 @@ func tableBuilderInner(vulnResult *models.VulnerabilityResults, addStyling bool,
 					outputRow = append(outputRow, "GIT", pkgCommitStr, pkgCommitStr)
 					shouldMerge = true
 				} else {
-					outputRow = append(outputRow, pkg.Package.Ecosystem, pkg.Package.Name, pkg.Package.Version)
+					name := pkg.Package.Name
+					if pkg.Package.DepGroup != "" {
+						name += fmt.Sprintf(" (%s)", pkg.Package.DepGroup)
+					}
+					outputRow = append(outputRow, pkg.Package.Ecosystem, name, pkg.Package.Version)
 				}
 
 				outputRow = append(outputRow, source.Path)
--- a/pkg/lockfile/parse-pnpm-lock.go
+++ b/pkg/lockfile/parse-pnpm-lock.go
@@ -23,6 +23,7 @@ type PnpmLockPackage struct {
 	Resolution PnpmLockPackageResolution `yaml:"resolution"`
 	Name       string                    `yaml:"name"`
 	Version    string                    `yaml:"version"`
+	Dev        string                    `yaml:"dev"`
 }
 
 type PnpmLockfile struct {
@@ -119,6 +120,8 @@ func parseNameAtVersion(value string) (name string, version string) {
 	return matches[1], matches[2]
 }
 
+const PnpmDevDependency string = "dev"
+
 func parsePnpmLock(lockfile PnpmLockfile) []PackageDetails {
 	packages := make([]PackageDetails, 0, len(lockfile.Packages))
 
@@ -152,12 +155,18 @@ func parsePnpmLock(lockfile PnpmLockfile) []PackageDetails {
 			}
 		}
 
+		depGroup := ""
+		if strings.TrimSpace(pkg.Dev) == "true" {
+			depGroup = PnpmDevDependency
+		}
+
 		packages = append(packages, PackageDetails{
 			Name:      name,
 			Version:   version,
 			Ecosystem: PnpmEcosystem,
 			CompareAs: PnpmEcosystem,
 			Commit:    commit,
+			DepGroup:  depGroup,
 		})
 	}
 
--- a/internal/output/sarif.go
+++ b/internal/output/sarif.go
@@ -295,13 +295,19 @@ func PrintSARIFReport(vulnResult *models.VulnerabilityResults, outputWriter io.W
 				alsoKnownAsStr = fmt.Sprintf(" (also known as '%s')", strings.Join(gv.AliasedIDList[1:], "', '"))
 			}
 
+			depGroup := ""
+			if pws.Package.DepGroup != "" {
+				depGroup = fmt.Sprintf(" (%s)", pws.Package.DepGroup)
+			}
+
 			run.CreateResultForRule(gv.DisplayID).
 				WithLevel("warning").
 				WithMessage(
 					sarif.NewTextMessage(
 						fmt.Sprintf(
-							"Package '%s' is vulnerable to '%s'%s.",
+							"Package '%s'%s is vulnerable to '%s'%s.",
 							results.PkgToString(pws.Package),
+							depGroup,
 							gv.DisplayID,
 							alsoKnownAsStr,
 						))).
--- a/pkg/lockfile/fixtures/maven/with-scope.xml
+++ b/pkg/lockfile/fixtures/maven/with-scope.xml
@@ -1,0 +1,10 @@
+<project>
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
--- a/pkg/lockfile/fixtures/npm/optional-package.v1.json
+++ b/pkg/lockfile/fixtures/npm/optional-package.v1.json
@@ -1,0 +1,22 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "wrappy": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/wrappy/-/wrappy-1.0.2.tgz",
+      "integrity": "sha1-tSQ9jz7BqjXxNkYFvA0QNuMKtp8=",
+      "dev": true,
+      "optional": true
+    },
+    "supports-color": {
+      "version": "5.5.0",
+      "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-5.5.0.tgz",
+      "integrity": "sha512-QjVjwdXIt408MIiAqCX4oUKsgU2EqAGzs2Ppkm4aQYbjm+ZEWEcW4SfFNTr4uMNZma0ey4f5lgLrkB0aX0QMow==",
+      "optional": true,
+      "requires": {
+        "has-flag": "^3.0.0"
+      }
+    }
+  }
+}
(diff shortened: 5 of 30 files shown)


G-Rath:
I've not had a full look over this, but fwiw I think it would be better to make groups an array rather than a single string because some ecosystems support packages belonging to multiple groups i.e. at least Bundler and Poetry.

While I don't think that's expressed in pretty much any of the lockfiles, given the accuracy goals of the scanner I don't think it's unreasonable to expect (/hope) that one day the scanner will be able to determine that (such as by looking at related files) and so having an array now should make it easier to support (otherwise it'll be a breaking change).

I'm also not sure if it's worth making public constants for all the different types of groups, but I'm less fussed about that 🤷

cuixq:
We actually had a discussion about if we should make the group info an array (my first proposal was to make it an array). Considering the purpose is to help users prfioritize what vulnerabilities to handle, we may only need to know if a dependency is of the default group or not. Then I tried to just use a bool value to represent that, but this makes it harder to provide more information to users in the reporters, so I finally made this a string. However I think if we do care about the exact group names, we should make it an array.

Regarding the constants, it is of the similar consideration - I would like to display the information as close as to the ecosystem terminology since this dependency group information is quite different among ecosystems. I am happy to remove them and just use the strings.

Also when I was researching the dependency groups for different ecosystems, I realised that some ecosystems only have this information available in manifest but not lockfile, so if we want to have a better support on this, we should also support parsing manifests.

G-Rath:
Yeah it comes down to what your priorities are - personally I think at least in the lockfile/library space the scanner should be trying to capture info in a generally neutral manner i.e. it should not decide which group of dependency takes priority, it should just capture all groups a package is part of.

Instead deciding how to treat the different groups should be a downstream responsibility (e.g. being handled by the scanner CLI component, scorecard who are just using the library, etc).

Case and point, NPM records if a package is a dev dependency, peer dependency, and optional dependency, and you demonstrate in your test cases its possible for a dependency to be both dev and optional - you have then had to choose one to favor because you cannot record both (incidentally I believe you should be favoring "dev" not "optional" because the latter will always be installed if possible, where the former might be excluded in production type builds).

cuixq:
Agree with the point that the parser should capture all the relevant information and let the downstream to decide. 

This reminds me of one thing about making groups as an array: how we record the default group?
 - Nothing: then this may cause problem if the package is of the default group and another group since only the non-default one is recorded
 - An empty string: sounds like a better option, but this means the array will be of length one with an empty string very often (default group packages). will this cause confusion?
 - Still string, but with the default group name: similar to above, in addition with the information that makes more sense. Will this cause any problem of using too much space somewhere? I feel is no, since this is not stored somewhere, but just pass to the users

cuixq:
Also regarding the parsers, feel free to point out anything that I am not doing it correctly! 

I am not an expert on all the ecosystems, and some of the implementation is just based on my brief research. 

There is also a TODO for Gradle: I saw `developmentOnly` in the lockfile, but not sure if this indicates the development dependencies - I cannot find any official documentation on this.

G-Rath:
For now I think it should be fine to just capture groups as they are - so in the case of e.g. NPM which doesn't record `production: true`, nothing gets added to the array.

I'm pretty sure the only ecosystem currently where that might cause confusion is with `requirements.txt` but that would be resolvable by saying the rule there is the filename sans the extension is the group, because at best you'd end up with `base`, `production`, etc and at worse you'll have `requirements.txt` but because the downstream knows the name of the original lockfile it should be able to easily filter that out if needed.

> Also regarding the parsers, feel free to point out anything that I am not doing it correctly!

You're doing fine 🙂 while I wrote most of them and all of `semantic`, that was done more by being comfortable hacking with the package managers than knowing each of them in depth - especially ones like for Java and C++ I'm pretty happy to be just be told if something is wrong and if it should be done another way for the ecosystem 🤷

internal/ci/vulnerability_result_diff.go
```diff
@@ -26,7 +28,7 @@ func DiffVulnerabilityResults(oldRes, newRes models.VulnerabilityResults) models
 		resultPS := &result.Results[len(result.Results)-1]
 		for _, pv := range ps.Packages {
 			pkgs := oldRes.Results[sourceIdx].Packages
-			pkgIdx := slices.IndexFunc(pkgs, func(elem models.PackageVulns) bool { return elem.Package == pv.Package })
+			pkgIdx := slices.IndexFunc(pkgs, func(elem models.PackageVulns) bool { return reflect.DeepEqual(elem.Package, pv.Package) })
```
oliverchang, line 31: Can we instead define a method on `models.PackageVulns` for this equality comparison? Is the main difference here that we want "groups" to be included in the comparison?

internal/output/table.go
```diff
@@ -128,7 +128,11 @@ func tableBuilderInner(vulnResult *models.VulnerabilityResults, addStyling bool,
 					outputRow = append(outputRow, "GIT", pkgCommitStr, pkgCommitStr)
 					shouldMerge = true
 				} else {
-					outputRow = append(outputRow, pkg.Package.Ecosystem, pkg.Package.Name, pkg.Package.Version)
+					name := pkg.Package.Name
+					if len(pkg.DepGroups) > 0 {
+						name += fmt.Sprintf(" (%s)", strings.Join(pkg.DepGroups, ","))
```
oliverchang, line 133: Hmm, can we perhaps be more opinionated here in the table output, and just output e.g. "(dev)" in the cases where it is a dev dependency?

Users may not care about optional etc most of the time, and it would just clutter this table. If they do want the exact group details, they can use the JSON output.

We can define perhaps some kind of per-ecosystem function which just does:

```
type Ecosystem interface {
  IsDevGroup(groups []string) bool
}
```
or something similar per ecosystem? 

@another-rex @G-Rath wdyt?

cuixq: Ecosystem is now a `string` can we still make it an interface?

Or, we can have something like this?
```
func (s Ecosystem) IsDevGroup(groups []string) bool {
  switch s {
  case CargoEcosystem: 
     ...
  case NpmEcocystem:
    ...
  }
  ...
  return false;
}
```

internal/ci/vulnerability_result_diff.go
cuixq: Actually the code change here is not needed - I used to put "groups" in `PackageInfo` and that causes `==` not working. 
since I moved "groups" to `PackageVulns` so we can use `==` to compare `PackageInfo`.

internal/output/table.go
another-rex: Agree we should just stick to a single output in the table output, but I think we should do it with a separate column.

I'm thinking "prod" column with a checkmark if it's a production dependency, and empty if not a production dep. This way it makes it applicable to all the other not "dev" named dependencies (e.g. optional in npm).

pkg/lockfile/dpkg-status.go
```diff
@@ -122,7 +124,7 @@ func (e DpkgStatusExtractor) Extract(f DepFile) ([]PackageDetails, error) {
 
 		// PackageDetails does not contain any field that represent a "not installed" state
 		// To manage this state and avoid false positives, empty struct means "not installed" so skip it
-		if (PackageDetails{}) == pkg {
+		if reflect.DeepEqual(PackageDetails{}, pkg) {
```
G-Rath, line 127: my understanding is that pulling in `reflect` can be a big deal in Go, and at at least in the past with some methods (which tbh probably don't include `DeepEqual`) could trigger sizeable deoptimizations.

I don't think this check is anything special and very internal so could be easily replaced with something like returning a pointer (so that we can then return `nil`), returning an `error` that we can eat here, or setting the `Name` to something like `<not installed>`.

I think that's probably worth doing either way so that we're not as dependent on the _whole_ structure of `PackageDetails`

pkg/lockfile/parse-pnpm-lock.go
```diff
@@ -23,6 +23,7 @@ type PnpmLockPackage struct {
 	Resolution PnpmLockPackageResolution `yaml:"resolution"`
 	Name       string                    `yaml:"name"`
 	Version    string                    `yaml:"version"`
+	Dev        string                    `yaml:"dev"`
```
G-Rath, line 26: shouldn't we be able to parse this as a bool directly?

internal/output/table.go
oliverchang: Re column, I think the reason we went with a "(dev)" marker for now is that it because the column becomes redundant whenever we do/support any kind of filtering, and makes things inconsistent.

This is also somethign we can change at any time, given that there's no stability guarantees for the table/human readable output.

cuixq: I have added a method on `Ecosystem` to check if there is `dev` group in the provided groups.

pkg/lockfile/dpkg-status.go
cuixq: I modified the code here to check each field in `PackageDetails` and if they are all "empty". wdty?

pkg/lockfile/parse-pnpm-lock.go
cuixq: good catch - I remember `bool` didn't work so I put `string` here but I just tried this again and `bool` is perfectly fine.

G-Rath (approved):

another-rex (approved):
LGTM!

oliverchang (commented):
Thanks @cuixq!! This looks awesome. Sorry for the delay in reviewing this.

spencerschrock:
Out of curiosity, what's the difference between [PackageVulns](https://pkg.go.dev/github.com/google/osv-scanner@v1.6.0/pkg/models#PackageVulns) and [VulnerabilityFlattened](https://pkg.go.dev/github.com/google/osv-scanner@v1.6.0/pkg/models#VulnerabilityFlattened)?

I see `Licenses` and `LicenseViolations` exist in both, but `DepGroups` was only added to `PackageVulns`. Is this intentional?

cuixq:
good find - we should add `DepGroups` to `VulnerabilityFlattened` as well.

Issue #642, opened 2023-11-05
Support for scanning renv lockfiles for the R language


dijitali:
Following the addition of CRAN for the R language to the OSV database (https://github.com/google/osv.dev/pull/1477), it would be helpful if this CLI supported scanning [`renv` lockfiles](https://github.com/rstudio/renv/).

A workaround at present is to parse the lockfile with jq and [generate a custom lockfile](https://google.github.io/osv-scanner/usage/#custom-lockfiles):

```shell
jq '{
    results: [
      {
        packages: [
          .Packages | to_entries[] | {
            package: {
              name: .key,
              version: .value.Version,
              ecosystem: "CRAN"
            }
          }
        ]
      }
    ]
  }' renv.lock > osv-renv-lock.json

osv-scanner --lockfile osv-scanner:./osv-renv-lock.json
```

G-Rath:
I'd actually been looking into this a few weeks ago; would you mind providing some example lockfiles for fixtures? I can dig these out from open source projects, but if you know of any special cases that would be worth having tests for that'd be great to hear (e.g. the NPM ecosystem supports the same package multiple times with different versions and peer dependencies, ruby supports different OSs, Go has `replace`, etc; it doesn't look like `renv` has that kind of complexity, but would be good to confirm that with folks who actually use the tool!)

dijitali:
I'm not too familiar with any special cases but perhaps a good foundation would be the renv project's tests?  

For example:
https://github.com/rstudio/renv/blob/main/tests/testthat/test-lockfile.R

Other than that, I'm aware there are changes in the works for things like dev dependencies (https://github.com/rstudio/renv/issues/1695).  But actually from a quick look, it sounds like this isn't something osv-scanner supports yet either (https://github.com/google/osv-scanner/issues/332).

PR #656, opened 2023-11-15
feat: add support for comparing CRAN versions

G-Rath:
Part of #642

See https://github.com/G-Rath/osv-detector/pull/235 for the journey I went on with R for this


scripts/generators/generate-cran-versions.R
```diff
@@ -0,0 +1,183 @@
...
+      package <- affected$package$name
+
+      if (!(package %in% names(result))) {
+        result[[package]] <- list()
+      }
+
+      for (version in affected$versions) {
+        tryCatch(
+          {
+            as.package_version(version)
```
cuixq, line 36: what does `as` here mean?

G-Rath: From what I understand (which is very little in this space 😅) it's R's version of casting: https://www.rdocumentation.org/packages/methods/versions/3.0.3/topics/as

So R has this [`package_version`](https://astrostatistics.psu.edu/su07/R/html/base/html/package_version.html) class, and we can make one of them by doing `as.package_version(...)` which... apparently we can also do by just calling `package_version(...)`?

It sounds like among other things [`as` is more performant as it does a quick return if the arg is already the right type](https://stackoverflow.com/questions/39279238/why-use-as-factor-instead-of-just-factor)

internal/semantic/version-cran.go
```diff
@@ -0,0 +1,46 @@
+package semantic
```
oliverchang, line 1: Is there an upstream documentation page for the versioning that we can link from here?

oliverchang (commented):
very nice!!

oliverchang (approved):

G-Rath:
also just a note: from what I can tell this is actually more "R" rather than "CRAN" aka this is exactly the same for the Bioconductor ecosystem, but I've not worried about that at all because 1. `semantic` is internal and 2. there are advisories for any Bioconductor package.

I think it should be pretty easy if just slightly "special-case-y" to handle if/when there's Bioconductor advisories (this will also apply to #668)

Issue #151, opened 2023-01-17
Exclude Packages by Regex Value from Scan


gromhacks:
This is an Enhancement request.

First of all the tool is fantastic, but it could be better with the following enhancement.

Summary - Include the ability to exclude packages based on a regex pattern.

Reasoning - Some developers and companies consider "internal" package names sensitive information especially when it comes to dependency confusion attacks, and do not want to share them accidentally with any third-party service. Giving users the ability to exclude packages that match a certain regex pattern(s) from the batch request prevents those package names from being leaked.

Example something like: `osv-scanner -r <Target Directory> --exclude=@mycompanyrepoid,mycompany.git,mycompany.domain,randomthingiwantexcluded`

Bonus points if you print out a preview of the package list before it's sent.

G-Rath:
Discussed offline: initially I thought we could support this via the existing field but then I realised that while rare there are some ecosystems whose package names can include regex characters (such as Python whereby "extras" are noted with square brackets i.e. `somepackage[bcrypt]`).

We still want to support this, so we're going to instead have a new field that is used to control if the name field should be treated as a full string or a regexp - I can't think of any other "modes" that could be useful, so it'll probably be a boolean though I still need to think of a name: `name_partial`, `name_mode`, `name_regex`?

majiayu000:
Hi, I've been looking into this and traced the root cause.

**Proposed approach:** 1. Add a new field (e.g., `name_is_regex bool`) to PackageOverrideEntry in internal/config/config.go. 2. Modify the `matches()` method to compile and match the name as regex when the flag is true. 3. Handle regex compilation errors gracefully. 4. Add unit tests for regex matching including edge cases (invalid regex, special chars in ecosystem names like Python extras). 5. Update osv-scanner.toml documentation. Comment on issue first to confirm field naming with G-Rath before coding.

I'll open a draft PR shortly. Happy to adjust based on your preference.

PR #658, opened 2023-11-16
✨ Add --exclude-packges Flag to osv-scanner CLI

toddtee:
This PR adds an `--exclude-packages` flag to the osv-scanner CLI. This flag allows users to specify a list of regex patterns to which matches of collected package names will be matched against.

If the packages are found to match the regex patterns, they will be excluded from the query to the OSV database.

Exclusions work for both lockfile and SBOM package definitions.

Excluded-packages flag will address issue #151.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -144,6 +144,18 @@ Then pass this to `osv-scanner` with this:
 osv-scanner --lockfile osv-scanner:/path/to/osv-scanner.json
 ```
 
+## Specify Package(s) to Exclude from OSV Query
+
+If you want to exclude a package or a range of packages from the query to the OSV database, you can pass a list of regex patterns with the '--exclude-packages' flag. OSV-Scanner will check for packages matching the specified regex patterns and exclude these packages from being checked for known vulnerabilities. The exclusions apply to both lockfile specified packages and SBOM specified packages.
+
+The exclusion regex patterns can be applied with the following flag:
+
+```bash
+osv-scanner --exclude-packages='^example.com/.*,^react$' <rest-of-cli-interaction>
+```
+
+Please note! It is best practice to specify the regex patterns in single quotes to avoid shell expansion of the regex patterns.
+
 ## Scanning a Debian based docker image packages
 Preview
 {: .label }
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -10,8 +10,10 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/google/osv-scanner/internal/exclusions"
 	"github.com/google/osv-scanner/internal/local"
 	"github.com/google/osv-scanner/internal/output"
 	"github.com/google/osv-scanner/internal/sbom"
@@ -37,6 +39,7 @@ type ScannerActions struct {
 	NoIgnore             bool
 	DockerContainerNames []string
 	ConfigOverridePath   string
+	Exclusions           []*regexp.Regexp
 
 	ExperimentalScannerActions
 }
@@ -88,7 +91,7 @@ const (
 //   - Any SBOM files with scanSBOMFile
 //   - Any git repositories with scanGit
 
-func scanDir(r reporter.Reporter, dir string, skipGit bool, recursive bool, useGitIgnore bool, compareOffline bool) ([]scannedPackage, error) {
+func scanDir(r reporter.Reporter, dir string, skipGit bool, recursive bool, useGitIgnore bool, compareOffline bool, exclusionRegexes []*regexp.Regexp) ([]scannedPackage, error) {
 	var ignoreMatcher *gitIgnoreMatcher
 	if useGitIgnore {
 		var err error
@@ -145,7 +148,7 @@ func scanDir(r reporter.Reporter, dir string, skipGit bool, recursive bool, useG
 
 		if !info.IsDir() {
 			if extractor, _ := lockfile.FindExtractor(path, ""); extractor != nil {
-				pkgs, err := scanLockfile(r, path, "")
+				pkgs, err := scanLockfile(r, path, "", exclusionRegexes)
 				if err != nil {
 					r.PrintError(fmt.Sprintf("Attempted to scan lockfile but failed: %s\n", path))
 				}
@@ -154,7 +157,7 @@ func scanDir(r reporter.Reporter, dir string, skipGit bool, recursive bool, useG
 			// No need to check for error
 			// If scan fails, it means it isn't a valid SBOM file,
 			// so just move onto the next file
-			pkgs, _ := scanSBOMFile(r, path, true)
+			pkgs, _ := scanSBOMFile(r, path, true, exclusionRegexes)
 			scannedPackages = append(scannedPackages, pkgs...)
 		}
 
@@ -318,7 +321,7 @@ func (m *gitIgnoreMatcher) match(absPath string, isDir bool) (bool, error) {
 
 // scanLockfile will load, identify, and parse the lockfile path passed in, and add the dependencies specified
 // within to `query`
-func scanLockfile(r reporter.Reporter, path string, parseAs string) ([]scannedPackage, error) {
+func scanLockfile(r reporter.Reporter, path string, parseAs string, exclusionRegexes []*regexp.Regexp) ([]scannedPackage, error) {
 	var err error
 	var parsedLockfile lockfile.Lockfile
 
@@ -360,6 +363,15 @@ func scanLockfile(r reporter.Reporter, path string, parseAs string) ([]scannedPa
 		output.Form(len(parsedLockfile.Packages), "package", "packages"),
 	))
 
+	// If exclusions are flagged, check if any packages should be excluded.
+	if exclusionRegexes != nil {
+		r.PrintText("Exclusions have been flagged, searching for packages to exclude...\n")
+		_, err := exclusions.ExcludePackages(r, exclusionRegexes, &parsedLockfile)
+		if err != nil {
+			r.PrintError(fmt.Sprintf("Failed to analyse lockfile packages for exclusions: %s\n", err))
+		}
+	}
+
 	packages := make([]scannedPackage, len(parsedLockfile.Packages))
 	for i, pkgDetail := range parsedLockfile.Packages {
 		packages[i] = scannedPackage{
@@ -379,7 +391,7 @@ func scanLockfile(r reporter.Reporter, path string, parseAs string) ([]scannedPa
 
 // scanSBOMFile will load, identify, and parse the SBOM path passed in, and add the dependencies specified
 // within to `query`
-func scanSBOMFile(r reporter.Reporter, path string, fromFSScan bool) ([]scannedPackage, error) {
+func scanSBOMFile(r reporter.Reporter, path string, fromFSScan bool, exclusionRegexes []*regexp.Regexp) ([]scannedPackage, error) {
 	var errs []error
 	var packages []scannedPackage
 	for _, provider := range sbom.Providers {
@@ -407,6 +419,18 @@ func scanSBOMFile(r reporter.Reporter, path string, fromFSScan bool) ([]scannedP
 				ignoredCount++
 				//nolint:nilerr
 				return nil
+			}
+			// If exclusions are flagged, check if the package should be excluded.
+			if exclusionRegexes != nil {
+				excluded, err := exclusions.ExcludeSBOMPackages(r, exclusionRegexes, &id)
+				if err != nil {
+					r.PrintError(fmt.Sprintf("Failed to analyse SBOM packages for exclusions: %s\n", err))
+					return nil
+				}
+				if excluded {
+					r.PrintText(fmt.Sprintf("Excluding package %s from OSV query\n", id.PURL))
+					return nil
+				}
 			}
 			packages = append(packages, scannedPackage{
 				PURL: id.PURL,
@@ -726,7 +750,7 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
 			r.PrintError(fmt.Sprintf("Failed to resolved path with error %s\n", err))
 			return models.VulnerabilityResults{}, err
 		}
-		pkgs, err := scanLockfile(r, lockfilePath, parseAs)
+		pkgs, err := scanLockfile(r, lockfilePath, parseAs, actions.Exclusions)
 		if err != nil {
 			return models.VulnerabilityResults{}, err
 		}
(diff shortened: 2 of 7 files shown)


pkg/exclusions/exclusions.go
```diff
@@ -0,0 +1,103 @@
...
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// isValidRegex checks if a regex pattern is valid.
+func isValidRegex(pattern string) bool {
+	_, err := regexp.Compile(pattern)
```
G-Rath, line 87: Regexp compiling should be done in `cachedregexp` - currently that only has `MustCompile` but this is a reason for `Compile`, and is extra nice because we know its going to get `Compile`'d at least twice

toddtee: Hey mate, just going to play back my understanding to make sure I fully understand your suggestion.

Thanks for pointing out the `cachedregex` package; definitely makes sense to move the regex handling to that package. So for that reason:

- I will also create a `Compile` function for the `cachedregex` package which can be called by the `checkForRegexMatch` function:
```
// Compile takes a regex pattern and returns a list of Regexp objects.
func Compile(pattern string) (*regexp.Regexp, error) {

	re, err := regexp.Compile(pattern)
	if err != nil {
		return nil, fmt.Errorf("invalid regex pattern: %s", pattern)
	}
	return re, nil
}
```

-  I will move `exclusions.isValidRegex` to `cachedregex.IsValidRegex`, which will also make use of the new `Compile` function.

```
// IsValidRegex checks if a regex pattern is valid.
func IsValidRegex(pattern string) bool {
	_, err := Compile(pattern)
	return err == nil
}
```

I have also updated the `parseExclusions` function to utilise the `cachedregexp.MustCompile` fucntionality as that function doesn't care about errors?

```
// parseExclusions parses a list of exclusion patterns and returns a list of Regexp objects.
func parseExclusions(patterns []string) ([]*regexp.Regexp, error) {
	exclusionRegexes := make([]*regexp.Regexp, len(patterns))
	for i, pattern := range patterns {
		re := cachedregexp.MustCompile(pattern)
		exclusionRegexes[i] = re
	}

	return exclusionRegexes, nil
}
```

Am I roughly in the ballpark or way off?

Thanks for your time.

G-Rath: Just create the `Compile` function in `cachedregexp` - `isValidRegex` should remain where it is

toddtee: Thanks @G-Rath, updated now. 👌

docs/usage.md
```diff
@@ -144,6 +144,20 @@ Then pass this to `osv-scanner` with this:
 osv-scanner --lockfile osv-scanner:/path/to/osv-scanner.json
 ```
 
+## Specify Package(s) to Exclude from OSV Query
+
+If you want to exclude a package or a range of packages from the query to the
+OSV database, you can pass a list of regex patterns. The `osv-scanner` will
```
G-Rath, line 150: nit: I feel like we should be mentioning the flag at the end of this sentence:

```suggestion
OSV database, you can pass a list of regex patterns with the `--exclusions` flag. The `osv-scanner` will
```

docs/usage.md
```diff
@@ -144,6 +144,20 @@ Then pass this to `osv-scanner` with this:
...
+check for packages matching the specified regex patterns and exclude these
+packages from being checked for known vulnerabilities. The exclusions apply to
+both lockfile specified packages and SBOM specified packages.
+
+The exclusion regex patterns can be applied with the following flag:
+
+```bash
+osv-scanner --exclusions="^example.com/.*,^react$" <rest-of-cli-interaction>
```
G-Rath, line 158: nit: I think it would be better to use single quotes as (at least in Bash) they don't do interpolation and the nature of regex syntax makes me paranoid of that happening by mistake 😅 

```suggestion
osv-scanner --exclusions='^example.com/.*,^react$' <rest-of-cli-interaction>
```

(it might even be worth adding a warning about that?)

pkg/exclusions/exclusions.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/google/osv-scanner/internal/cachedregexp"
+	"github.com/google/osv-scanner/internal/sbom"
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/google/osv-scanner/pkg/reporter"
+)
+
+// ExcludePackages checks for matches in a Lockfile's Packages against the list of provided exclusion regex patterns.
+// If a match is found, the package is removed from the Lockfile's Packages.
+// The updated Lockfile is returned.
```
G-Rath, line 15: nit: while I like comments, I feel like a few of them in this file are redundant

e.g. I think it should be pretty obvious given the return type that the updated lockfile is what is being returned, and that a private function called `isRegexValid` checks if strings are valid regexps...

Would you mind doing re-check with that thinking in your head and seeing if there are any comments you're comfortable with dropping? I know this can be very subjective so happy if you come back saying you think we need all these comments 🙂

pkg/osvscanner/osvscanner.go
```diff
@@ -37,6 +38,7 @@ type ScannerActions struct {
 	NoIgnore             bool
 	DockerContainerNames []string
 	ConfigOverridePath   string
+	Exclusions           []string
```
G-Rath, line 41: I'm softly leaning towards saying we should be taking a slice of regexps here because that is the most accurate (and thus least-error-prone) representation of the data; the reason I'm not completely sold is that I expect a number of library consumers are probably taking user input aka dynamic regexps, but for us it should mean the library doesn't have to handle "what if the string isn't a valid regexp"...

This would be a reason to keep `exclusions` as a public package as then downstream users could just pull that in like the CLI does rather than have to implement their own parser (then again they could also just copy and paste, which is the Go way 🤷)

pkg/exclusions/exclusions.go
G-Rath, line 1: This should be `internal` by default since public packages have to follow semver

G-Rath (changes requested):
This is on the right track - I've done a general review but not gone over every line yet as I think the next main question is about if the library should be accepting a slice of regexp or strings, as changing that will probably mean other code can be changed which'd make other parts of my review moot 😅

docs/usage.md
toddtee: Bash in general makes me paranoid, so I get it... 🤣

docs/usage.md
```diff
@@ -144,6 +144,21 @@ Then pass this to `osv-scanner` with this:
 osv-scanner --lockfile osv-scanner:/path/to/osv-scanner.json
 ```
 
+## Specify Package(s) to Exclude from OSV Query
+
+If you want to exclude a package or a range of packages from the query to the
+OSV database, you can pass a list of regex patterns with the '--exclusions'
+flag. The `osv-scanner` will check for packages matching the specified regex
```
hayleycd, line 151:
```suggestion
flag. OSV-Scanner will check for packages matching the specified regex
```
This phrasing is more consistent with our documentation.

toddtee: Sure thing!

pkg/exclusions/exclusions.go
toddtee: Completely agree. Comments are good, but superfluous commenting can reduce readability in some cases. I do believe it is good practice to leave comments on public functions though.

oliverchang (commented):
Thanks for this! Our team has considered this in the past as a config file rule (https://google.github.io/osv-scanner/configuration/).

Could we understand if a config file would also work for your use case? Or are there reasons why you'd prefer a CLI flag?

docs/usage.md
```diff
@@ -144,6 +144,24 @@ Then pass this to `osv-scanner` with this:
 osv-scanner --lockfile osv-scanner:/path/to/osv-scanner.json
 ```
 
+## Specify Package(s) to Exclude from OSV Query
```
hayleycd, line 147: Thank you for updating the documentation for the --exclusions flag. I appreciate you including this as part of your PR. 

Our GitHub Page theme automatically wraps text. Each paragraph should be included in a single line. This makes future changes easier and is more consistent with the majority of the documentation (though there are a few places where I need to fix this).

hayleycd:
Docs look good with a couple of comments.

docs/usage.md
toddtee: Updated.

docs/usage.md
toddtee: Hi @hayleycd did you need me to update the paragraphs to be a single line to fit in with your Github Pages theme? Or it is fine to leave the paragraph wrapped?

toddtee:
Hi @oliverchang, thanks for checking out my PR!

> Could we understand if a config file would also work for your use case? Or are there reasons why you'd prefer a CLI flag?

Ideally, I think having the `exclusions` parameter both as a CLI flag and a configuration parameter would be ideal. 

I like the idea of having the CLI flag as I personally believe that CLI tools should have good levels of interactivity for a human user. For example, if I as a developer want to quickly update a project and I have a particular package that is giving me issues that I am blocked on updating or don't want to query for whatever reason, it is nice to have the flag as a quick option to interact with the tool as I need it. Or I could even use the flag to test the scan results for when I am developing the regex patterns that will eventually end up in the configuration file! 🤣 

However... let's now talk automation, not human interaction. If I incorporate the `osv-scanner` as part of my project's CI/CD pipeline or automated security checks, having the flag available is good, however having a configuration parameter is better. That way, maintainers of the project can configure what exclusions they like and perhaps even have comments in the config file as to why they are there. I see the configuration parameter for more permanent and/or governing exclusion purposes.

**TL;DR, I see the CLI flag for quick developer use for once-offs or short term purposes and the configuration file more for automated processes and more permanent solutions.** 

I don't know how challenging incorporating the configuration file into my current offering would be in the short term...

However, as I am a supporter of one PR including one change only. I would propose that this PR maintain adding the functionality of a CLI flag exclusively, and perhaps create a related Issue and follow-up PR to add exclusions as a configuration option as well. 

The benefit of this is the tool can still offer the ability to exclude packages from the query in the short term and deliver value to those that need it. A configuration parameter will extend that functionality offering further.

I hope that answers your question 🙏

docs/usage.md
hayleycd: I added a commit. Docs are good to go now.

pkg/exclusions/exclusions.go
toddtee: Yep agreed, ain't no one got time for that.

docs/usage.md
toddtee: Thanks @hayleycd, I have squashed this into the main commit.

pkg/osvscanner/osvscanner.go
toddtee: Hey @G-Rath, given in my brain, the user was always going to be input from the user, I naturally just assumed it was going to be a list of string. If it were going to be regexp, we would still need to take the list of string from the flag and parse them as regexp _before_ we assign them to the ScannerActions struct? 🤔

G-Rath: What you've said is correct but the code base is made up of both the CLI and the library - and it's the CLI that handles the fact the input is coming in as a string (which is because that's just how data comes in from terminals)

The library on the other hand can just accept regexp so it doesn't need to be responsible for handling the edge cases that come with parsing strings into regexp

toddtee: Thanks @G-Rath that makes sense. Let me shuffle some things and see what happens. 🔀

toddtee: Updated @G-Rath.

The `ScannerActions` struct now has the `Exclusions` field as a `[]*regex.regexp`.

oliverchang:
> Hi @oliverchang, thanks for checking out my PR!
> 
> > Could we understand if a config file would also work for your use case? Or are there reasons why you'd prefer a CLI flag?
> 
> Ideally, I think having the `exclusions` parameter both as a CLI flag and a configuration parameter would be ideal.
> 
> I like the idea of having the CLI flag as I personally believe that CLI tools should have good levels of interactivity for a human user. For example, if I as a developer want to quickly update a project and I have a particular package that is giving me issues that I am blocked on updating or don't want to query for whatever reason, it is nice to have the flag as a quick option to interact with the tool as I need it. Or I could even use the flag to test the scan results for when I am developing the regex patterns that will eventually end up in the configuration file! 🤣

Thanks, that makes sense. We've generalised shied away from having multiple ways to achieve the same thing out of a desire to keep the CLI simple to understand, but having a way to quickly iterate/test does make sense. 

Could we perhaps rename the flag to `--exclude-packages` though? `--exclude` is a little ambiguous on its own. 

@another-rex @G-Rath Do you think there's value in maintaining this as a CLI flag? 

> 
> However... let's now talk automation, not human interaction. If I incorporate the `osv-scanner` as part of my project's CI/CD pipeline or automated security checks, having the flag available is good, however having a configuration parameter is better. That way, maintainers of the project can configure what exclusions they like and perhaps even have comments in the config file as to why they are there. I see the configuration parameter for more permanent and/or governing exclusion purposes.

Big +1 to this. Having this stored as the repo would help maintain state around exclusions and reduce duplicate triage efforts for the project.

> 
> **TL;DR, I see the CLI flag for quick developer use for once-offs or short term purposes and the configuration file more for automated processes and more permanent solutions.**
> 
> I don't know how challenging incorporating the configuration file into my current offering would be in the short term...
> 
> However, as I am a supporter of one PR including one change only. I would propose that this PR maintain adding the functionality of a CLI flag exclusively, and perhaps create a related Issue and follow-up PR to add exclusions as a configuration option as well.

Yep! I've filed https://github.com/google/osv-scanner/issues/670 for this. 
> 
> The benefit of this is the tool can still offer the ability to exclude packages from the query in the short term and deliver value to those that need it. A configuration parameter will extend that functionality offering further.
> 
> I hope that answers your question 🙏

toddtee:
@oliverchang, I think `--excluded-packages` is better as it is more explicit to what the functionality actually is. I have updated the flag and documentations to reflect this. 👍

G-Rath:
> Do you think there's value in maintaining this as a CLI flag?

I'm not sure - I'm not getting super strong "omg yes this'd be great" vibes from my gut, but that doesn't meant it still wouldn't be useful.

However...

>  For example, if I as a developer want to quickly update a project and I have a particular package that is giving me issues that I am blocked on updating or don't want to query for whatever reason, it is nice to have the flag as a quick option to interact with the tool as I need it

I'm not sure that is as useful as you think: since we're agreed that it's best to capture this in config long-term, in this situation you'd be running the scanner locally in most cases, meaning you could just mentally ignore the results too - I know that is not as nice as having them omitted but not sure if its "support a new flag forever" nice.

> I could even use the flag to test the scan results for when I am developing the regex patterns that will eventually end up in the configuration file!

That does seem like a useful reason to have the flag, though I don't think it would be that much overhead to just be using the config file; that would also mean less risk that a specific shell might get confused by the regexp.

---

I don't think this should be taken as "never will we ever have this flag", but it feels like it would probably be best to start life via the config file

Issue #513, opened 2023-08-30
Stabilize go call analysis and make it default behavior


another-rex:
Go call analysis should be the default behavior if the go toolchain is detected.

TODO:
Figure out the best way to organize call graph analysis flags to allow disabling/enabling, and experimental-ness on a per language level.

oliverchang:
@hogo6002 this is another potential starter project for you. @another-rex WDYT?

G-Rath:
Could we have the flags take a comma list of langs, with the presence of a flag meaning "all"? e.g.

```
# this has all stable call analysis being run (also same if `--call-analysis` is passed without a value)
osv-scanner .

# this enables stable call analysis for just rust and java
osv-scanner --call-analysis=rust,java

# this disables all stable call analysis
osv-scanner --no-call-analysis

# this disables stable call analysis for rust, but all others are still run
osv-scanner --no-call-analysis=rust
```

And then the same again for experimental.

I know it seems a bit verbose but I think logically that should be very straightforward to determine - I think something like:

```
func updateEnabledCallAnalysis(supportedCallAnalysis *map[string]bool, enabledFlag any, disabledFlag any) {
	langsToEnable := strings.Split(enabledFlag)
	langsToDisable := strings.Split(disabledFlag)

	if enabledFlag == "*" {
		langsToEnable = getMapKeys(supportedCallAnalysis)
	}

	if disabledFlag == "*" {
		langsToDisable = getMapKeys(supportedCallAnalysis)
	}

	for _, s := range langsToEnable {
		if _, ok := supportedCallAnalysis[s]; ok {
			supportedCallAnalysis[s] = true
		}
	}

	for _, s := range langsToDisable {
		if _, ok := supportedCallAnalysis[s]; ok {
			supportedCallAnalysis[s] = false
		}
	}
}

func main() {
	stableCallAnalysisStates := map[string]bool{
		"go":   true,
		"rust": true,
		"java": true,
	}

	experimentalCallAnalysisStates := map[string]bool{
		"ruby": true,
	}


	updateEnabledCallAnalysis(&stableCallAnalysisStates, enabledStableFlag, disabledStableFlag)
	updateEnabledCallAnalysis(&experimentalCallAnalysisStates, enabledExperimentalFlag, disabledExperimentalFlag)
}
```

(That's really rough pseudo code - but I think the most fiddly bit is probably handling the actual flag parsing to support what is typically mapped to a bool but also supporting string values).

another-rex:
[...]

PR #665, opened 2023-11-17
Set Go call analysis to default behaviour

hogo6002:
Fix issue https://github.com/google/osv-scanner/issues/513
- Replace `experimental-call-analysis` with `call-analysis`. (`--call-analysis=all`, `--call-analysis=rust`)
- Adding a `--no-call-analysis` to disable call analysis. (`--no-call-analysis=all`, `--no-call-analysis=go`). This overrides `call-analysis`.
- Delete `call-analysis` from `experimental-config` in result report.
- Call analysis for non-experimental languages (e.g. go) is auto enabled.


G-Rath:
This looks like a good start but I don't think it's quite the right approach - I've written up [a comment on the issue](https://github.com/google/osv-scanner/issues/513#issuecomment-1817609642) on how I think this should be handled, as it's more of a design thing at this point

cmd/osv-scanner/main.go
```diff
@@ -128,6 +128,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Name:  "experimental-licenses",
 				Usage: "report on licenses",
 			},
+			&cli.BoolFlag{
+				Name:  "disable-call-analysis",
```
oliverchang, line 132: @another-rex @G-Rath we discussed making this "no-call-analysis" to be consistent with "no-ignore". 

@hogo6002 could you please make this change?

hogo6002: Sure, just updated.

cmd/osv-scanner/main.go
```diff
@@ -97,16 +97,20 @@ func run(args []string, stdout, stderr io.Writer) int {
...
-				Value: false,
-			},
 			&cli.BoolFlag{
 				Name:  "no-ignore",
 				Usage: "also scan files that would be ignored by .gitignore",
 				Value: false,
 			},
+			&cli.StringSliceFlag{
+				Name:    "call-analysis",
+				Aliases: []string{"ca"},
```
another-rex, line 107: Let's remove `ca` alias for now, we can add it in later if necessary. 
```suggestion
```

another-rex (commented):
Looks pretty good! Just some minor nits. 

Can you also update the user documentation to reflect the change? This can be done in a follow up PR.

hogo6002:
> Can you also update the user documentation to reflect the change? This can be done in a follow up PR.
Sure! I will create another PR to update the user documentation.

cmd/osv-scanner/main.go
```diff
@@ -97,16 +97,19 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Usage:   "check subdirectories",
 				Value:   false,
 			},
-			&cli.BoolFlag{
```
G-Rath, line 100: question: should we keep this for a release or two and print a deprecation warning about it going away in a non-major?

I know we don't _have_ to, but it's pretty easy and only costs a few bytes in the binary

cmd/osv-scanner/main.go
```diff
@@ -168,9 +171,10 @@ func run(args []string, stdout, stderr io.Writer) int {
 				NoIgnore:             context.Bool("no-ignore"),
 				ConfigOverridePath:   context.String("config"),
 				DirectoryPaths:       context.Args().Slice(),
+				NoCallAnalysis:       context.StringSlice("no-call-analysis"),
+				CallAnalysis:         context.StringSlice("call-analysis"),
```
G-Rath, line 175: I think the library should just take a map of bools and then use the default for a language if its not set in the map; how the flags are structured as an implementation detail of the CLI that doesn't need to be inflicted on the library

cmd/osv-scanner/main.go
hogo6002: Thanks for pointing out. It definitely will be more user friendly.

cmd/osv-scanner/main.go
hogo6002: Updated. I moved the createCallAnalysisStates function a new file to keep main.go less cluttered

cmd/osv-scanner/flaghelpers.go
```diff
@@ -0,0 +1,43 @@
+package main
```
oliverchang, line 1: nit: helpers is generlaly not a good name. Can we rename this to callanalysis_state.go instead?

cmd/osv-scanner/main.go
```diff
@@ -99,14 +99,25 @@ func run(args []string, stdout, stderr io.Writer) int {
 			},
 			&cli.BoolFlag{
 				Name:  "experimental-call-analysis",
-				Usage: "attempt call analysis on code to detect only active vulnerabilities",
+				Usage: "[Deprecated] attempt call analysis on code to detect only active vulnerabilities",
```
oliverchang, line 102: Can we just remove this flag? We make no guarantees around the stability of experimental flags anyway.

cmd/osv-scanner/flaghelpers.go
```diff
@@ -0,0 +1,43 @@
...
+
+var stableCallAnalysisStates = map[string]bool{
+	"go":   true,
+	"rust": false,
+}
+
+// Creates a map to record if languages are enabled or disabled for call analysis.
+func createCallAnalysisStates(enabledCallAnalysis []string, disabledCallAnalysis []string) map[string]bool {
```
oliverchang, line 9: Can you add some unit tests for this function in particular?

oliverchang (commented):
very nice!!

cmd/osv-scanner/main.go
another-rex: I think Gareth made a good point that it'll be nice to give at least one release gap before deleting something for people to switch to the new flag while being able to update to the latest smoothly. We can remove it after next release without increasing complexity much.

oliverchang: Makes sense!

cmd/osv-scanner/callanalysis_parser.go
```diff
@@ -0,0 +1,43 @@
+package main
+
+var stableCallAnalysisStates = map[string]bool{
+	"go":   true,
+	"rust": false,
+}
+
+// Creates a map to record if languages are enabled or disabled for call analysis.
+func createCallAnalysisStates(enabledCallAnalysis []string, disabledCallAnalysis []string) map[string]bool {
+	enableAll, disableAll := false, false
```
G-Rath, line 10: you could avoid the conditions in the first two loops by expecting "all" in the actual `callAnalysisStates` - while I think you'd still need to have a variable outside the third loop, I feel like it would still be simpler (effectively it should let you shift all the logic to the end of the function) 🤔 

(not a blocker, just worth playing with if you havn't already)

cmd/osv-scanner/main.go
```diff
@@ -99,14 +99,25 @@ func run(args []string, stdout, stderr io.Writer) int {
...
 				Value: false,
+				Action: func(context *cli.Context, b bool) error {
+					return fmt.Errorf("the experimental-call-analysis flag is no longer supported. Please use the call-analysis and no-call-analysis flags instead")
```
G-Rath, line 105: nit: I think it would be better to say something like "has been replaced" than "is no longer supported", since we're still _supporting_ call analysis we're just now upgrading it (effectively)

cmd/osv-scanner/callanalysis_parser.go
hogo6002: Thanks Gareth! Updated.

cmd/osv-scanner/main.go
```diff
@@ -99,14 +99,25 @@ func run(args []string, stdout, stderr io.Writer) int {
 			},
 			&cli.BoolFlag{
 				Name:  "experimental-call-analysis",
-				Usage: "attempt call analysis on code to detect only active vulnerabilities",
+				Usage: "[Deprecated] attempt call analysis on code to detect only active vulnerabilities",
 				Value: false,
+				Action: func(context *cli.Context, b bool) error {
```
another-rex, line 104: Does this keep the behavior the same? or just return an error?

I think we want to emulate the previous behavior (basically, if this flag is passed in, pretend --call-analysis=all is passed in), and print a warning message.

another-rex (commented):
One last thing, otherwise LGTM!

cmd/osv-scanner/main.go
hogo6002: Updated. Thanks Rex!

another-rex (approved):
LGTM, thanks!

cuixq (approved):

oliverchang (approved):

cmd/osv-scanner/callanalysis_parser.go
```diff
@@ -0,0 +1,35 @@
...
+		callAnalysisStates[language] = false
+	}
+
+	enableAll, containsAll := callAnalysisStates["all"]
+	for language, isStable := range stableCallAnalysisStates {
+		if _, exists := callAnalysisStates[language]; !exists {
+			callAnalysisStates[language] = isStable
+			if enableAll {
+				callAnalysisStates[language] = true
+			}
```
G-Rath, line 26: you can use `||` here, since really `enableAll` is for enabling non-stable languages:

```suggestion
			callAnalysisStates[language] = isStable || enableAll
```

cmd/osv-scanner/callanalysis_parser.go
```diff
@@ -0,0 +1,35 @@
...
+		callAnalysisStates[language] = true
+	}
+
+	for _, language := range disabledCallAnalysis {
+		callAnalysisStates[language] = false
+	}
+
+	enableAll, containsAll := callAnalysisStates["all"]
+	for language, isStable := range stableCallAnalysisStates {
+		if _, exists := callAnalysisStates[language]; !exists {
```
G-Rath, line 22: nit: I _think_ this is equivalent, which feels a bit simpler: 

```
if containsAll {
  callAnalysisStates[language] = enableAll
} else if _, exists := callAnalysisStates[language]; !exists {
  callAnalysisStates[language] = isStable
}
```

up to you though as I don't think it's that much simpler

cmd/osv-scanner/main.go
```diff
@@ -159,6 +167,14 @@ func run(args []string, stdout, stderr io.Writer) int {
 				return err
 			}
 
+			var callAnalysisStates map[string]bool
+			if context.Bool("experimental-call-analysis") {
```
G-Rath, line 171: You should check if the flag is set since you can explicitly provide booleans i.e. this won't print the warning for `--experimental-call-analysis=false`

hogo6002: Thanks!

another-rex (approved):
LGTM!

PR #672, opened 2023-11-22
make license allowlist matching case insensitive

josieang:
this just makes it easier for our users to use.


another-rex (approved):
Good idea!

PR #677, opened 2023-11-27
Simplify return codes to return 1 if any vulnerability related error

### LINKED ISSUE #676: Simplify return codes
The current (unreleased) return code setup is a little too complicated ([context](https://github.com/google/osv-scanner/pull/501#discussion_r1401646226)).

https://github.com/google/osv-scanner/blob/8c5634d41241f1786a8fe0b56a0fe828da97e871/cmd/osv-scanner/main.go#L206

I propose we simplify all of this to just:

- Return 0 if there are no findings or errors.
- Return 1 if there are **any** findings (license or vulns). 
- Return 128 if no packages are found. 

As part of this we can delete the new public error types added [here](https://github.com/google/osv-scanner/pull/501#discussion_r1401649796) also. 

For govulncheck / call analysis, we should also return 0 if any findings are uncalled (this matches `govulncheck` CLI behaviour). This may end up being a per-ecosystem default as we add support for more languages that are less robust when it comes to reachability analysis.

another-rex:
Fixes #676 

> - Return 0 if there are no findings or errors.
> - Return 1 if there are any findings (license or vulns).
> - Return 128 if no packages are found.

### DIFF
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -62,17 +62,10 @@ var NoPackagesFoundErr = errors.New("no packages found in scan")
 //nolint:errname,stylecheck // Would require version major bump to change
 var VulnerabilitiesFoundErr = errors.New("vulnerabilities found")
 
+// deprecated: This error is no longer returned, check the results to determine if this is the case
+//
 //nolint:errname,stylecheck // Would require version bump to change
 var OnlyUncalledVulnerabilitiesFoundErr = errors.New("only uncalled vulnerabilities found")
-
-//nolint:errname,stylecheck // Would require version bump to change
-var LicenseViolationsErr = errors.New("license violations found")
-
-//nolint:errname,stylecheck // Would require version bump to change
-var VulnerabilitiesFoundAndLicenseViolationsErr = errors.New("vulnerabilities found and license violations found")
-
-//nolint:errname,stylecheck // Would require version bump to change
-var OnlyUncalledVulnerabilitiesFoundAndLicenseViolationsErr = errors.New("only uncalled vulnerabilities found and license violations found")
 
 var (
 	vendoredLibNames = map[string]struct{}{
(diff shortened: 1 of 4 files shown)


pkg/osvscanner/osvscanner.go
```diff
@@ -827,25 +820,11 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
 		}
 		onlyUncalledVuln = onlyUncalledVuln && vuln
 
-		switch {
-		case !vuln && !onlyUncalledVuln && !licenseViolation:
+		if !vuln && !onlyUncalledVuln && !licenseViolation {
```
oliverchang, line 823: Is this logic correct? i.e. if we only have uncalled vulns, 

```
vulns = true
onlyUncalledVuln = true
```

then this will fail this check? 

This this be:

```
if (!vuln || onlyUncalledVuln) && !licenseViolation {
}
```

?

another-rex: Yeah you're right, I'll make that change

pkg/osvscanner/osvscanner.go
```diff
@@ -62,18 +62,11 @@ var NoPackagesFoundErr = errors.New("no packages found in scan")
 //nolint:errname,stylecheck // Would require version major bump to change
 var VulnerabilitiesFoundErr = errors.New("vulnerabilities found")
```
josieang, line 63: Could you leave a comment here saying that this includes uncalled vulns, and license violations? The name is misleading but I get that changing it would require a major version bump change.

another-rex: Done.

josieang (commented):
Thanks for doing this!

josieang (approved):

pkg/osvscanner/osvscanner.go
```diff
@@ -59,21 +59,17 @@ type ExperimentalScannerActions struct {
...
+//
 //nolint:errname,stylecheck // Would require version major bump to change
 var VulnerabilitiesFoundErr = errors.New("vulnerabilities found")
 
+// Deprecated: This error is no longer returned, check the results to determine if this is the case
+//
 //nolint:errname,stylecheck // Would require version bump to change
 var OnlyUncalledVulnerabilitiesFoundErr = errors.New("only uncalled vulnerabilities found")
 
-//nolint:errname,stylecheck // Would require version bump to change
```
josieang, line 68: is it okay to delete these? I suppose they all pertain to experimental features so I am okay with it.

josieang (approved):

oliverchang (approved):
thanks Rex!

PR #682, opened 2023-11-29
update docs for call analysis.

hogo6002:
Previous PR https://github.com/google/osv-scanner/pull/665 updated '--experimental-call-analysis' to '--call-analysis' and '--no-call-analysis'.
Updating docs to reflect the changes.


another-rex:
Can you add the experimental tag to the rust call analysis, and change the description to indicate that all the experimental analysis is off by default (rust), while the non experimental languages (go) is turned on by default.

docs/usage.md
```diff
@@ -145,3 +145,43 @@ OSV-Scanner's C/C++ support is based on commit-level data. OSV's commit-level da
...
+
+Call stack analysis can be performed on some languages to check if the 
+vulnerable code is actually being executed by your project. If the code
+is not being executed, these vulnerabilities will be marked as unexecuted.
+
+To enable call analysis in all languages, call OSV-Scanner with the `--call-analysis=all` flag. By default, call analysis in Go is enabled, but you can disable it using the `--no-call-analysis=go` flag.
+
+### Call analysis in Go
+
+OSV-Scanner uses the `govulncheck` library to analyze Go source code to identify called vulnerable functions.
```
hayleycd, line 159:
```suggestion
OSV-Scanner uses the [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) library to analyze Go source code to identify called vulnerable functions.
```

docs/usage.md
```diff
@@ -145,3 +145,43 @@ OSV-Scanner's C/C++ support is based on commit-level data. OSV's commit-level da
...
+
+#### Additional Dependencies
+
+`go` compiler needs to be installed and available on `PATH`
+
+### Call analysis in Rust
+
+OSV-Scanner compiles Rust source code and analyzes the output binary's DWARF debug information to identify called vulnerable functions.
```
hayleycd, line 167:
```suggestion
Experimental {: .label }

OSV-Scanner compiles Rust source code and analyzes the output binary's DWARF debug information to identify called vulnerable functions.
```

docs/usage.md
```diff
@@ -145,3 +145,43 @@ OSV-Scanner's C/C++ support is based on commit-level data. OSV's commit-level da
...
+
+OSV-Scanner uses the [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) library to analyze Go source code to identify called vulnerable functions.
+
+#### Additional Dependencies
+
+`go` compiler needs to be installed and available on `PATH`
+
+### Call analysis in Rust
+
+OSV-Scanner compiles Rust source code and analyzes the output binary's DWARF debug information to identify called vulnerable functions.
```
hayleycd, line 167:
```suggestion
Experimental {: .label }
Call analysis in Rust is still considered experimental. 

OSV-Scanner compiles Rust source code and analyzes the output binary's DWARF debug information to identify called vulnerable functions.
```

hayleycd:
I'm confused. Why do we have both a `--call-analysis` and a `--no-call-analysis` flag? shouldn't it be either one or another?

hogo6002:
> I'm confused. Why do we have both a `--call-analysis` and a `--no-call-analysis` flag? shouldn't it be either one or another?

Hi Hayley, it is because we want go call analysis to be enabled by default. So we need to add one more flag to turn it off. More details: https://github.com/google/osv-scanner/issues/513

hayleycd:
But if call analysis is the default behavior, we wouldn't need a flag for it right? Or is it because it is not the default for rust?

Either way, the behavior is not currently clear in the docs.

oliverchang:
> > I'm confused. Why do we have both a `--call-analysis` and a `--no-call-analysis` flag? shouldn't it be either one or another?
> 
> Hi Hayley, it is because we want go call analysis to be enabled by default. So we need to add one more flag to turn it off. More details: #513

And also because not all languages are enabled by default. Only Go is enabled so far. If someone wants rust, then they need to specify `--call-analysis=rust`.

hogo6002:
> And also because not all languages are enabled by default. Only Go is enabled so far. If someone wants rust, then they need to specify `--call-analysis=rust`.
> 
> And if we want **only** rust, you'd do:
> 
> `--call-analysis=rust --no-call-analysis=go`
> 
> @hogo6002 is that right?

Yeah that's right. With this implementation, we are able to enable/disable call analysis per language.

hogo6002:
> But if call analysis is the default behavior, we wouldn't need a flag for it right? Or is it because it is not the default for rust?
> 
> Either way, the behavior is not currently clear in the docs.

Call analysis in Go is enabled by default (don't need to put any flags) as it is stable. But users may still want to disable it, so we added a '--no-call-analysis' flag. 
Usages:

```
# this has all non-experimental call analysis being run, same with `--call-analysis=default` is passed (only go now)
osv-scanner .

# this enables all call analysis (both go and rust)
osv-scanner --call-analysis=all

# this enables all default call analysis and rust (call analysis for go is also enabled as it is by default.)
osv-scanner --call-analysis=rust

# this disables all call analysis
osv-scanner --disable-call-analysis=all

# this disables call analysis for go, but rust still runs
osv-scanner --call-analysis=rust --disable-call-analysis=go
```

hayleycd (approved):
I took a closer look at things and I think things are looking alright, especially with the 'experimental' label on the rust section. 

Thanks for clarifying about the different flags.

another-rex (approved):
LGTM

oliverchang (approved):

PR #651, opened 2023-11-10
feat: CVSS v4.0 support and replace cvss implementation to comply with the specifications

pandatix:
## Why this PR

[CVSS v4.0](https://www.first.org/cvss/v4-0/) has been released lately, and the OSV will most probably add its support (the first CVSS v4.0 vector known to the FIRST.ORG SIG CVSS has been published [by Palo Alto Networks for the CVE-2023-3282](https://security.paloaltonetworks.com/CVE-2023-3282)).

As a FIRST.ORG SIG CVSS member and [Go CVSS implementation](https://github.com/pandatix/go-cvss) maintainer, I'm looking forward to improve its adoption and understanding in the Open-Source Ecosystem.
Moreover, there exist issues with the currently used CVSS implementation, such as [invalid scoring computation](https://github.com/goark/go-cvss/issues/33), and [CVSS v4.0 is currently not planned for support](https://github.com/goark/go-cvss/issues/37#issuecomment-1706334126).

## What it brings

With the current PR, I provide multiple direct improvements:
- proper CVSS v2.0 scoring computation (only affect the environmental score computation, but has been an unresolved issue for months)
- add support of CVSS v4.0 in the OSV schema
- performance improvements according to [benchmarks](https://github.com/pandatix/go-cvss#comparison)

Given ossf/osv-schema#166 the CVSS v4.0 key will most likely be `CVSS_V4` to align with the previous CVSS versions support.

## Is it breaking ?

For the code, no, but for the Go version, yes :tada:

Issue #638, opened 2023-11-03
Policy on how recent the go version in the go.mod file can be


josieang:
This point was brought up in In https://github.com/google/osv-scanner/pull/637.

We might like to make a policy that osv-scanner will only use versions of Go, say at least two release behind. This will help consumers know how up-to-date their version of Go should be to be able to use this as a library and as a CLI tool.

spencerschrock:
Based on my understanding of this blog, this problem may get less severe in the future for those with default settings:
https://go.dev/doc/toolchain

> The default GOTOOLCHAIN setting is auto

> When GOTOOLCHAIN is set to <name>+auto or <name>+path (or the shorthands auto or path), the go command selects and runs a newer Go version as needed. Specifically, it consults the toolchain and go lines in the current workspace’s go.work file or, when there is no workspace, the main module’s go.mod file. If the go.work or go.mod file has a toolchain <tname> line and <tname> is newer than the default Go toolchain, then the go command runs <tname> instead. If the file has a toolchain default line, then the go command runs the default Go toolchain, disabling any attempt at updating beyond <name>. Otherwise, if the file has a go <version> line and <version> is newer than the default Go toolchain, then the go command runs go<version> instead.

another-rex:
This will be completed in #1156

PR #690, opened 2023-12-05
refactor: switch golang.org/x/exp/slices usages to stdlib

spencerschrock:
`golang.org/x/exp/slices` is now in the standard library as `slices`: https://go.dev/doc/go1.21#slices.
The two libraries were [made identical](https://go-review.googlesource.com/c/exp/+/511895), so not expecting any breakages and all the tests pass.

A similar conversion can be done for golang.org/x/exp/maps in Go 1.22, which would remove the dependency on golang.org/x/exp. It can't be done now as `maps.Keys` was removed from Go 1.21 (see [related issue](https://www.github.com/golang/go/issues/61538)).


spencerschrock:
Note: I also had to bump the `.go-version` file to 1.21, in order to match the version declared in your `go.mod` file (which was bumped in https://github.com/google/osv-scanner/pull/651).

another-rex (approved):
Thanks Spencer!

spencerschrock:
> Why doing this? Was it so important to break builds?

This wasn't an important change, just a cleanup. But this PR wasn't the one to bump the go.mod version, so perhaps #638 is the more applicable issue to ask in.

PR #691, opened 2023-12-06
Fix go mod

another-rex:
Current go mod file does not compile #688 (I am actually unsure how it managed to get merged in the first place, maybe caused by our CI being 1.20 rather than 1.22). 
Latest experimental package requires golang.org/x/tools v0.16.0, which removed a module required by govulncheck.


josieang (approved):

michaelkedar (approved):
> Latest experimental package requires golang.org/x/tools v0.16.0, which removed a module required by govulncheck.

Should probably change govulncheck to not depend on the removed module

PR #693, opened 2023-12-06
refactor: setup `prettier` for formatting files

G-Rath:
This uses `prettier` to format non-Go files using my (opinioned) config; note that I'd recommend having `proseWrap` set to `always` but that has a _huge_ number of changes so I'll do a seperate PR for that.

Resolves #144
Resolves #129

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /osv-scanner
 /coverage.out
 /coverage.html
+fixtures/
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+**/fixtures/**
+**/fixtures-go/**
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
 **/fixtures/**
 **/fixtures-go/**
+/docs/vendor/**
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json.schemastore.org/prettierrc",
-  "singleQuote": true,
+  "singleQuote": false,
   "proseWrap": "preserve",
   "endOfLine": "lf",
   "arrowParens": "avoid",
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,9 @@
+{
+  "$schema": "https://json.schemastore.org/prettierrc",
+  "singleQuote": true,
+  "proseWrap": "preserve",
+  "endOfLine": "lf",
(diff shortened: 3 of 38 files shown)


.prettierrc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://json.schemastore.org/prettierrc",
+  "singleQuote": true,
```
another-rex, line 3: I think we want double quote here, it seems to change *slightly* fewer files. We somehow managed to have even number of single and double quotes usage.

.prettierignore
```diff
@@ -0,0 +1,2 @@
+**/fixtures/**
+**/fixtures-go/**
```
another-rex, line 2:
```suggestion
**/fixtures-go/**
/docs/vendor/**
```

another-rex (commented):
Thanks! Let's just switch single quotes to double quotes, then LGTM!

.prettierrc.json
G-Rath: While I do personally tend to prefer single quotes I don't mind changing this if we really want to, but I think doing so because of the number of files changed isn't the best reason since this is a "do once" operation and you can exclude commits in `git blame` and other places using a setting I'll dig up in a bit

another-rex: The reason I prefer double quotes is it matches go strings being double quotes so it keeps things slightly more consistent.

G-Rath: Now _that_ is a good reason 😄

another-rex (approved):

.gitignore
```diff
@@ -5,3 +5,4 @@
 /osv-scanner
 /coverage.out
 /coverage.html
+fixtures/
```
another-rex, line 8: @G-Rath Do you remember why this line was added?

another-rex: I think this ignores all fixtures/ folders, even the ones we have in cmd/osv-scanner/fixtures for example.

G-Rath: Yes, because the fixtures are crafted from different sources whose formatting might be different from what Prettier wants and (more importantly) that includes files that are deliberately not parsable i.e. invalid JSON files which are for testing behaviour in that scenario.

cuixq: but this PR also adds `.prettierignore` which has `fixtures/` folders listed?

G-Rath: This was a mistake - I've opened https://github.com/google/osv-scanner/pull/825 to remove it

PR #694, opened 2023-12-06
Add support for NuGet lock files version 2

giovanni-bozzano:
If [NuGet central package management](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management) is enabled, V2 lock files will be generated when restoring projects.
Check out [NuGet source](https://github.com/NuGet/NuGet.Client/blob/a59e64507383b64bcfbe9bf63b34aca946ab0da9/src/NuGet.Core/NuGet.Commands/PackagesLockFileBuilder.cs#L119-L128) for reference.

Since version V2 is backward compatible with V1, the required change to support it is straightforward.


another-rex (approved):
Thanks!

PR #698, opened 2023-12-09
Add experimental support for dry run

kemzeb:
Works towards #303.

### DIFF
--- a/cmd/osv-scanner/main.go
+++ b/cmd/osv-scanner/main.go
@@ -116,6 +116,10 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Usage: "disables call graph analysis",
 			},
 			&cli.BoolFlag{
+				Name:  "experimental-dry-run",
+				Usage: "Perform a dry run of OSV-Scanner, which prints API calls without actually sending anything",
+			},
+			&cli.BoolFlag{
 				Name:  "experimental-local-db",
 				Usage: "checks for vulnerabilities using local databases",
 			},
@@ -204,6 +208,7 @@ func run(args []string, stdout, stderr io.Writer) int {
 					LocalDBPath:    context.String("experimental-local-db-path"),
 					CompareLocally: context.Bool("experimental-local-db"),
 					CompareOffline: context.Bool("experimental-offline"),
+					DryRun:         context.Bool("experimental-dry-run"),
 					// License summary mode causes all
 					// packages to appear in the json as
 					// every package has a license - even
--- a/pkg/osv/osv.go
+++ b/pkg/osv/osv.go
@@ -330,6 +330,10 @@ func makeRetryRequest(action func() (*http.Response, error)) (*http.Response, er
 }
 
 func MakeDetermineVersionRequest(name string, hashes []DetermineVersionHash) (*DetermineVersionResponse, error) {
+	return MakeDetermineVersionRequestWithClient(name, hashes, http.DefaultClient)
+}
+
+func MakeDetermineVersionRequestWithClient(name string, hashes []DetermineVersionHash, client *http.Client) (*DetermineVersionResponse, error) {
 	var buf bytes.Buffer
 
 	request := determineVersionsRequest{
@@ -351,7 +355,6 @@ func MakeDetermineVersionRequest(name string, hashes []DetermineVersionHash) (*D
 		req.Header.Set("User-Agent", RequestUserAgent)
 	}
 
-	client := http.DefaultClient
 	resp, err := client.Do(req)
 	if err != nil {
 		return nil, err
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -2,10 +2,13 @@ package osvscanner
 
 import (
 	"bufio"
+	"bytes"
 	"crypto/md5" //nolint:gosec
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
+	"net/http"
 	"os"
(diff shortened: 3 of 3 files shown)


kemzeb:
Decided to make it an experimental feature for now (I guess if user feedback has it change for some reason, though I don't believe users will throw this into scripts or anything like that).

Took quite some time trying to figure out how to implement this without making many non-DRY changes or having to resort to changing public API; my current approach is to "mock" the `http.Client` instance that performs these requests by just having an implementation of the `http.RoundTripper` interface print out data and return a fake HTTP response.

It seems to work fine for the `v1/querybatch` and `v1expermiental/determineversion` endpoints, but I wish to have a discussion about the following questions that came up:

1. How should we represent these API calls when we print it out? Should I use the Reporter here? Also, should I print out the body for POST requests?
2.  Consuming `/v1/vulns/:id` requires the batch request to give us a list of vulnerability IDs for each query, but since we're doing a dry run we can't get this information. What should we do here?
4. When I manually tested this change by running it against osv-scanner, it seems that the reporter calls a `PrintError()` somewhere that will resort in a non-zero exit code. Should we just exit early when we are done with dry run mode?
5. Not quite sure how to test this. Should I create unit tests for `osvscanner.DoScan()`?
6. Should I consider the API calls related to the experimental features (e.g. perform HEAD/GET request on zip db archive, using gRPC with the deps.dev API)?

another-rex:
Thanks! I think the current implementation for dry run looks pretty good!

However this does add quite a bit of complexity, and the problem where there are dependencies in the requests like you mentioned in point 2. 

That's why I was thinking maybe going with a verbose but not dry run option, which would be easier to implement, but still work for someone trying to learn what info is being sent when running osv-scanner. For users with very sensitive dependencies they will have to use the `--local-db` / `--offline` option anyway, so it seems like there isn't a big incentive for a --dry-run option.


> 1. How should these API calls look when we print it out? Should I use the Reporter here? Also, should I print out the body for POST requests?

Yes, reporter is the way to go. If we don't need to do the dry-run option, another way to implement this is to extend the reporter to print different verbosity levels (err, warn, info, verbose...) (which I want to do anyway in the future), this could make the implementation for a verbose mode quite straightforward - store the verbosity level in reporter, and add print statements to all the `makeRequest` functions. This will also make it easy to cover the gRPC requests

Yes I think POST bodies should be printed out, they are probably what someone trying to understand what information osv-scanner is sending out want to see. 

> 2. Consuming `GET /v1/vulns/:id` requires the batch request to give us a list of vulnerability IDs for each query, but since we're doing a dry run we can't get this information. What should we do here?
> 3. When I manually tested this change by running it against osv-scanner, it seems that the reporter calls a `PrintError()` somewhere that will resort in a non-zero exit code. Should we just exit early when we are done with dry run mode?

See above about still logging the queries, just without a dry-run. I'm not sure the dry-run feature is worth it to have to cover these cases. 

> 4. Not quite sure how to test this. Should I create unit tests for `osvscanner.DoScan()`?

The best way to test this is probably to add tests to `cmd/osv-scanner/main_test.go` using the fixtures there show the output.

> 5. Should I consider the API calls made in the experimental features (e.g. we perform a HEAD/GET request on zip db archive, use gRPC with the deps.dev API)?

Yes, I think we should also print out what data  queries are sending too.

another-rex:
By the way, do you also personally have a specific use case in mind for needing dry-run instead of just a verbosity flag as well?

kemzeb:
Thanks for reviewing this @another-rex!

> Yes, reporter is the way to go. If we don't need to do the dry-run option, another way to implement this is to extend the reporter to print different verbosity levels (err, warn, info, verbose...) (which I want to do anyway in the future), this could make the implementation for a verbose mode quite straightforward - store the verbosity level in reporter, and add print statements to all the makeRequest functions. This will also make it easy to cover the gRPC requests

I think defining verbosity levels is a great idea compared to the limitations I currently see with implementing a dry run mode (i.e. point 2). I agree with you that a dry run mode may not be useful given that users can now opt to perform their scan locally (it seems this feature was made long after the issue I'm trying to solve was made).

> By the way, do you also personally have a specific use case in mind for needing dry-run instead of just a verbosity flag as well?

No use case for wanting a dry run mode, just saw an open issue and wanted to see if I could resolve it :^).

If there is an interest for a verbose option, I could try tackling this in a different PR. From what I understand what should be done here is 2-fold:
1. Modify the `Reporter` interface to allow callers to specify a verbosity level (maybe have `PrintError()` and `PrintText()` use an "Error" level and "Info" level respectfully; not sure if you would want to remove them since its public API). Expose a new CLI option argument for this ( maybe like `--verbose-level info`)
2. Just add print statements to the `makeRequest` functions and set each to the highest verbose level.

another-rex:
Technically adding additional functions to the `Reporter` interface is a breaking change, but I believe there are no other implementations (at least public on github, from a quick code search) of this interface, and there are no good use cases for implementing this manually instead of using one of the preset implementations we provide. 

Because of the additional complexity required to keep compatibility let's just go ahead and add those methods directly onto Reporter and update the various reporter presets we have to also implement the new methods. We probably should make the reporter interface itself internal as well to make sure no other implementations exist in the future, but I'll do that in a separate PR.

kemzeb:
Gotcha, I'll close this PR and set my sights on implementing verbosity levels for the `Reporter` interface and update those who implement it when I get the time.

kemzeb:
Hey @another-rex, FYI I haven't forgot about this but decided to wait until #706 is merged 👍

Issue #679, opened 2023-11-27
golang `stdlib` scan inconsistency for fresh go toolchain versions: bare metal vs docker


tommiv:
I've recently faced a weird issue in our CI process – it reports some vulnerabilities for go v1.20, while the go.mod file claims `go 1.21.4`. At the same time, my local version on a dev machine reports "no vulnerabilities found".

So, given this `/tmp/osv-issue/go.mod` file (not much to see here, the original one is way bigger, but the issue persists even with a minimal one):

```go
module some_module_name

go 1.21.4

require (
)
```

I have the following results:

#### [MacOS v14.1.1] [clean] bare metal binary, installed via homebrew:
```bash
osv-scanner -v
osv-scanner version: 1.4.3
commit: n/a
built at: n/a

Scanned /tmp/osv-issue/go.mod file and found 1 package
No vulnerabilities found
```

#### [MacOS v14.1.1] [clean] bare metal binary, manual download darwin_amd64 from github releases:
```bash
curl -OL https://github.com/google/osv-scanner/releases/download/v1.4.3/osv-scanner_1.4.3_darwin_amd64
chmod +x osv-scanner_1.4.3_darwin_amd64
./osv-scanner_1.4.3_darwin_amd64 -v
osv-scanner version: 1.4.3
commit: 6316373e47d7e3e4b4fd3630c4bbc10987738de6
built at: 2023-11-02T00:53:14Z

./osv-scanner_1.4.3_darwin_amd64 -L /tmp/osv-issue/go.mod
Scanned /tmp/osv-issue/go.mod file and found 1 package
No vulnerabilities found
```

#### [vulnerable] official docker way (copypasted some commands from docs)
```bash
docker run -it -v ${PWD}:/src ghcr.io/google/osv-scanner:v1.4.3 -v            
osv-scanner version: 1.4.3
commit: 6316373e47d7e3e4b4fd3630c4bbc10987738de6
built at: 2023-11-02T00:53:14Z

 docker run -it -v /tmp/osv-issue:/src ghcr.io/google/osv-scanner:v1.4.3 -L /src/go.mod
Scanned /src/go.mod file and found 1 package
╭──────────────────────────────┬──────┬───────────┬─────────┬─────────┬────────────╮
│ OSV URL                      │ CVSS │ ECOSYSTEM │ PACKAGE │ VERSION │ SOURCE     │
├──────────────────────────────┼──────┼───────────┼─────────┼─────────┼────────────┤
[...]

PR #704, opened 2023-12-11
Get go stdlib version from go.mod

hogo6002:
Fix https://github.com/google/osv-scanner/issues/679


pkg/lockfile/parse-go-lock.go
```diff
@@ -82,6 +82,20 @@ func (e GoLockExtractor) Extract(f DepFile) ([]PackageDetails, error) {
 		}
 	}
 
+	if parsedLockfile.Go != nil && parsedLockfile.Go.Version != "" {
+		goVersion := parsedLockfile.Go.Version
+		versionParts := strings.Split(goVersion, ".")
```
cuixq, line 87: maybe we should use some Go package that deals with the versioning here?
I found this package: https://pkg.go.dev/golang.org/x/mod/semver, but this requires a "v" prefix in the version string.

G-Rath: If only we had a robust semantic based version parsing and comparing library already... 😁

(`internal/semantic` should have everything you need for this)

hogo6002: Updated, thanks!

another-rex (approved):
LGTM

PR #706, opened 2023-12-16
feat: support `PrintTextf` and `PrintErrorf` on `Reporter`

G-Rath:
When I originally implemented `Reporter`, IDEs such as GoLand didn't support custom `Printf` functions so I stuck with plain methods and did the `fmt` formatting on the string; that's changed as of GoLand 2023.3 via [GO-5841](https://youtrack.jetbrains.com/issue/GO-5841) 🎉 

Technically adding to `Reporter` is a breaking change but as covered [in this comment](https://github.com/google/osv-scanner/pull/698#issuecomment-1853318857):

> I believe there are no other implementations (at least public on github, from a quick code search) of this interface, and there are no good use cases for implementing this manually instead of using one of the preset implementations we provide.

Either way I think it's better to land these ASAP to reduce the blast radius then to carry them around for possibly a lot longer - note that I'm not strictly against deprecating/removing `PrintText` and `PrintError` though I don't think there's a lot of value in keeping them.

As penance, I've also added rich method comments for the interface.

(having said that, since this is a breaking change already maybe we should just remove `PrintText` and `PrintError` right now)

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -42,6 +42,12 @@ linters:
     - unused
 
 linters-settings:
+  govet:
+    settings:
+      printf:
+        funcs:
+          - (github.com/google/osv-scanner/pkg/reporter.Reporter).PrintErrorf
+          - (github.com/google/osv-scanner/pkg/reporter.Reporter).PrintTextf
   depguard:
     rules:
       regexp:
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -66,6 +66,9 @@ linters-settings:
 
 issues:
   exclude-rules:
+    - path: pkg/reporter
+      linters:
+        - dupl
     - path: _test\.go
       linters:
         - goerr113
--- a/cmd/osv-reporter/main.go
+++ b/cmd/osv-reporter/main.go
@@ -179,11 +179,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 		}
 
 		if errors.Is(err, osvscanner.NoPackagesFoundErr) {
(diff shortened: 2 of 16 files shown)


.golangci.yaml
```diff
@@ -67,6 +67,9 @@ linters-settings:
 
 issues:
   exclude-rules:
+    - path: pkg/reporter
```
G-Rath, line 70: These changes made the reporters big enough to trigger `dupl` - I've just disabled it for the whole package because while it's technically not wrong, the only way I know that we could address that would be by having a `standardReporter` that gets used by each struct with the basic "print to stdout and error" behaviour (like what I've done in `osv-detector` with the ["memory database"](https://github.com/G-Rath/osv-detector/blob/main/pkg/database/mem-check.go)), but from what I understand that would be a breaking change (because when initializing the struct you have to pass in the inner struct) and it won't actually result in much savings

G-Rath:
> Attention: 65 lines in your changes are missing coverage. Please review.

(because they're code that we just don't have tests for, and adding them would be a lot of work for this PR 😅)

another-rex (approved):
LGTM, thanks!

G-Rath:
@another-rex let me know if you have thoughts about removing `PrintText` and `PrintError` entirely right now vs waiting

another-rex:
I'm thinking we remove them now since it's breaking anyway, and we'll want to have the changes discussed in #698 to be added before we make a release to keep all the breaking changes in one release.

PR #716, opened 2023-12-19
Adjusting docs

hayleycd:
Adjusting docs to account for the new combined action. 

[Preview](https://hayleycd.github.io/osv-scanner/github-action/)


hayleycd:
Need updated repository information.

docs/github-action.md
```diff
@@ -223,3 +133,85 @@ jobs:
...
+<summary>
+Examples
+</summary>
+
+#### Scan specific lockfiles
+
+```yml
+jobs:
+  scan-pr:
+    uses: "google/osv-scanner/.github/workflows/osv-scanner-reusable.yml"
```
another-rex, line 163:
```suggestion
    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@v1.5.0"
```

docs/github-action.md
```diff
@@ -223,3 +133,85 @@ jobs:
...
+        --lockfile=./path/to/lockfile1
+        --lockfile=requirements.txt:./path/to/python-lockfile2.txt
+```
+
+#### Default arguments
+
+```yml
+jobs:
+  scan-pr:
+    uses: "google/osv-scanner/.github/workflows/osv-scanner-reusable.yml"
```
another-rex, line 175:
```suggestion
    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@v1.5.0"
```

docs/github-action.md
```diff
@@ -223,3 +133,85 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: converted-OSV-Scanner-deps
+          path: osv-scanner-deps.json
+          retention-days: 2
+  vuln-scan:
+    name: Vulnerability scanning
+    # makes sure the extraction step is completed before running the scanner
+    needs: extract-deps
+    uses: "google/osv-scanner/.github/workflows/osv-scanner-reusable.yml@main"
```
another-rex, line 202:
```suggestion
    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@v1.5.0"
```

another-rex (commented):
Looks good! I added some suggested change comments with the new path/format, we'll need to change it in the manual install paths as well (e.g. line 54 and line 85)

hayleycd:
@another-rex Thanks for the new url. I have accepted the changes and used the new url in the other two spots you mentioned. 

I'm not sure what is going on with the prettier check, I thought it would specify a line that was failing, but it doesn't. Otherwise I think this is ready.

docs/github-action.md
```diff
@@ -223,3 +133,85 @@ jobs:
...
+        --lockfile=osv-scanner:osv-scanner-deps.json
+    permissions:
+      # Needed to upload the SARIF results to code-scanning dashboard.
+      security-events: write
+      contents: read
+```
+
+</details>
+
+````
```
G-Rath, line 217: This is what Prettier doesn't like - it wants it to be three backticks, but actually it should be removed entirely because it doesn't have a matching pair:

```suggestion

```

(Prettier doesn't show a diff because that would require more code/complexity and the whole point of Prettier is it can fix everything it reports and it's always safe, so you should just `npx prettier --write` and commit)

G-Rath:
Now Prettier will be unhappy about the final line ending - I tried to craft my suggestion to keep the right number of line endings, but not surprised it didn't work.

Best to run `npx prettier --write docs/github-action.md` and commit the result

another-rex (approved):

hayleycd:
@another-rex @G-Rath thank you both!

PR #717, opened 2023-12-19
test: use snapshot-based testing

G-Rath:
This replaces all our bulky string output comparison tests with nice automated snapshot tests - sample output:

image

Snapshots are generated automatically if they don't exist already when the test suite is run - otherwise they can be regenerated by setting `UPDATE_SNAPS=true`:

```
UPDATE_SNAPS=true scripts/run_scripts.sh
```

I will update the contributing documentation at some point, but the code should be good to go - I've tried not to go too overboard with helper methods 😅


G-Rath:
@another-rex scanning is failing due to #740 though I'm not sure the best way to fix it - would you mind weighing in? (I'm still not across the new go 1.21 `toolchain` stuff...)

go.mod
```diff
@@ -1,15 +1,17 @@
 module github.com/google/osv-scanner
 
-go 1.21
+go 1.21.0
```
another-rex, line 3:
```suggestion
go 1.21
```

another-rex: Just removing the .0 should make osv-scanner assume it's the latest version with #740

go.mod
```diff
@@ -1,15 +1,17 @@
...
+
+toolchain go1.21.5
```
another-rex, line 5: I'm not sure if we need `toolchain` here.

go.mod
G-Rath: This change was required by Go - it demanded I run `go mod tidy` and this was the result 🤷‍♂️

I think it might be related to `go-snaps`?

another-rex: ah, I think it might just be go1.21 compiler as well. Let's update this back to `1.21.5` then, since we only support go 1.21 to compile osv-scanner anyway, not supporting older compilers doesn't matter too much here.

cmd/osv-scanner/main_test.go
```diff
@@ -144,438 +91,156 @@ func testCli(t *testing.T, tc cliTestCase) {
 	ec := run(tc.args, stdoutBuffer, stderrBuffer)
 	// ec := run(tc.args, os.Stdout, os.Stderr)
 
-	stdout := normalizeRootDirectory(t, normalizeFilePaths(t, stdoutBuffer.String()))
-	stderr := normalizeRootDirectory(t, normalizeFilePaths(t, stderrBuffer.String()))
+	stdout := normalizeErrors(t, normalizeTempDirectory(t, normalizeRootDirectory(t, normalizeFilePaths(t, stdoutBuffer.String()))))
```
another-rex, line 94: This is getting a bit long, can we make a "normalize" function where we compose them together.

go.mod
```diff
@@ -26,6 +26,19 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
+	github.com/gkampitakis/ciinfo v0.3.0 // indirect
```
another-rex, line 30: Are these separate in the go.mod file because it's just for the test dependency? (If so can you add a comment above the require)

internal/output/result_test.go
```diff
@@ -40,13 +51,24 @@ func Test_mapIDsToGroupedSARIFFinding(t *testing.T) {
...
-			args:     testutility.LoadJSONFixture[models.VulnerabilityResults](t, "fixtures/test-vuln-results-a.json"),
-			wantPath: "fixtures/test-vuln-results-a-grouped.json",
+			args: testutility.LoadJSONFixtureWithWindowsReplacements[models.VulnerabilityResults](t,
+				"fixtures/test-vuln-results-a.json",
+				map[string]string{
+					"/path/to/sub-rust-project/Cargo.lock": "D:\\\\path\\\\to\\\\sub-rust-project\\\\Cargo.lock",
+					"/path/to/go.mod":                      "D:\\\\path\\\\to\\\\go.mod",
+				},
+			),
+			want: testutility.NewSnapshot().WithWindowsReplacements(
```
another-rex, line 66: There seems to be a lot of these path replacements that's very similar. 

Let's change this to take a map of  regex's replacements instead of strings,  this way we can add a "default" regex replace rule that covers the majority of the path replacements.

pkg/osvscanner/osvscanner_internal_test.go
```diff
@@ -59,17 +47,14 @@ func Test_filterResults(t *testing.T) {
...
-					"Full json output written to %s", diff, out)
-				//nolint:staticcheck
-				testutility.CreateJSONFixture(t, out, got)
-			}
-			if filtered != tt.testCase.numFiltered {
-				t.Errorf("filterResults() = %v, want %v", filtered, tt.testCase.numFiltered)
+
+			testutility.NewSnapshot().MatchJSON(t, got)
+
+			if filtered != tt.want {
```
another-rex, line 56: Will this get called when MatchJSON does not match? If it doesn't, we might want to move this up above line 54, as this could be more useful information at a glance compared to the full snapshot diff.

another-rex (commented):
Thanks! Added some more comments.

go.mod
G-Rath: I have not done anything manually with `go.mod`, so this was all generated by `go mod tidy` - I'll see if just moving these into the other `require` makes it happy, otherwise I'd prefer not to touch (or comment) it since it's using some internal go semantics

cmd/osv-scanner/main_test.go
G-Rath: yeah good point I had that on my todo for refactoring but forgot about it

internal/output/result_test.go
G-Rath: I'll explore this but my concern is these are typically pretty hard to craft since most folks only have one OS on hand so I think it's better to go the "boring but long" route than "complex but short" especially since these are tests.

But I'll explore it first to see how complex it actually ends up being before I officially take a stance 🙂

pkg/osvscanner/osvscanner_internal_test.go
G-Rath: This is just the count of items that got filtered, so I think it would be worse to have it fail first because it wouldn't tell you _which_ items were unexpected missing or present.

go.mod
G-Rath: ok yup it just seems to be `go.mod` preserves groups if you create them, and I must have somehow (probably during a rebase or something)

pkg/osvscanner/osvscanner_internal_test.go
another-rex: I think because we use Errorf here it will flow through to the MatchJSON and we'll get both outputs. I'm just not sure if snapshot uses Fatalf or Errorf when it fails to match.

G-Rath: ah yeah, `go-snaps` uses `t.Errorf` too:

```
--- FAIL: Test_filterResults (0.00s)
    --- FAIL: Test_filterResults/filter_partially (0.00s)
        osvscanner_internal_test.go:54:
            - Snapshot - 0
            + Received + 4

            @@ -547,5 +547,9 @@

                  }
                ],
                "experimental_config": {
            +     "licenses": {
            +       "summary": false,
            +       "allowlist": null
            +     }
                }
              }

            at ../__snapshots__/osvscanner_internal_test.snap:1476

        osvscanner_internal_test.go:57: filterResults() = 10, want 17
FAIL
coverage: 7.0% of statements in ./...

Snapshot Summary

✓ 2 snapshots passed
✕ 1 snapshot failed
```

internal/output/result_test.go
G-Rath: Yeah so playing with this locally I'm not sure if this is worth the extra complexity, as you still need to specify the full path - you've effectively trading vertical length for horizontal length.

fwiw the design of this btw is that while yes you'll need to specify the exact line, that is what you can just copy & paste from the snapshot line without much thinking (again, perfect imo because people don't have multiple OSs so they'll be checking the diff from CI); if you were using a regexp you'd have to think more about how you could normalize things and the overlap, and if you mucked up a regexp (or two) it would be harder to see which line is the problem.

I think we should stick with the current layout for now, and revisit after living with it for a bit to see if it is as troublesome as the complexity

another-rex: Makes sense to me, let's stick with this then.

another-rex (approved):

PR #723, opened 2023-12-27
Add formatting instructions to docs contribution

another-rex:
I just realized that our contributing to docs instructions is circular (CONTRIBUTING.md points to docs/README.md which points back to CONTRIBUTING.md). I removed that line and added in the formatting instructions.


docs/README.md
```diff
@@ -11,9 +11,19 @@ Here are other [pre-requisites] and instructions for running the [docs locally].
 [pre-requisites]: https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/testing-your-github-pages-site-locally-with-jekyll#prerequisites
 [docs locally]: https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/testing-your-github-pages-site-locally-with-jekyll#building-your-site-locally
 
-## Contributing to the docs
+## Formatting docs
 
-Please see [CONTRIBUTING.md](https://github.com/google/osv-scanner/blob/main/CONTRIBUTING.md/#contributing-documentation) for information on contributing documentation.
+### Dependencies
+
+- [Prettier](https://prettier.io/docs/en/install)
```
G-Rath, line 18: The actual dependency is Node (and NPM, though that ships with Node by default so I think its safe to omit) - because we're using `npx` to invoke `prettier`, it will prompt to install if `prettier` is not already available.

hayleycd (approved):
This looks alright to me.

PR #724, opened 2023-12-27
Fix vulnerabilities that OSV-Scanner found

another-rex:
Update stdlib to 1.21.1 and update crypto package.


G-Rath:
Do we actually want to explicitly fix the development version to a patch? Keep in mind that means you can't use a higher version either (at least for `goenv` - I wouldn't be entirely surprised if `go.mod` still allows higher versions).

I find it useful to be able to use a potentially slightly different patch version given I work across multiple (and unrelated) Go projects.

All that really matters for the vulns is that the build is using the specific patch version right?

another-rex:
> Do we actually want to explicitly fix the development version to a patch? Keep in mind that means you can't use a higher version either (at least for goenv - I wouldn't be entirely surprised if go.mod still allows higher versions).

So according to [this](https://go.dev/doc/modules/gomod-ref) from 1.21 onwards go.mod version sets the minimum go version, so it sounds like we can use higher versions. 

I think we do want to follow our own logic in #704 / https://github.com/google/osv-scanner/issues/679#issuecomment-1831217041 where we are using the go.mod version to determine the stdlib version. 

> All that really matters for the vulns is that the build is using the specific patch version right?

Yes, during testing it doesn't really matter what exact patch version we use. 

> I find it useful to be able to use a potentially slightly different patch version given I work across multiple (and unrelated) Go projects.

Good point, not sure if there's a good solution here to limit the go version restriction only to the final build.

oliverchang (approved):

docs/installation.md
```diff
@@ -61,7 +61,7 @@ Alternatively, you can install this from source by running:
 go install github.com/google/osv-scanner/cmd/osv-scanner@v1
 ```
 
-This requires Go 1.21+ to be installed.
+This requires Go 1.21.1+ to be installed.
```
another-rex, line 64:
```suggestion
This requires Go 1.21.4+ to be installed.
```

G-Rath (approved):
lets just get this landed...

PR #727, opened 2023-12-28
Add support for verbosity levels

kemzeb:
Inspired by what was discussed in #698, this PR makes the necessary (breaking) changes to the `reporter` package such that the `Reporter` interface now provides methods for printing at certain levels (i.e. ErrorLevel, WarnLevel, InfoLevel, and VerboseLevel). It also adds a --verbosity option so that users/automated tools can take advantage of this too.

### DIFF
--- a/cmd/osv-scanner/main.go
+++ b/cmd/osv-scanner/main.go
@@ -116,6 +116,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Name:  "no-call-analysis",
 				Usage: "disables call graph analysis",
 			},
+			&cli.StringFlag{
+				Name:  "verbosity",
+				Usage: fmt.Sprintf("specify the level of information that should be provided during runtime; value can be: %s", strings.Join(reporter.VerbosityLevels(), ", ")),
+				Value: "info",
+			},
 			&cli.BoolFlag{
 				Name:  "experimental-local-db",
 				Usage: "checks for vulnerabilities using local databases",
@@ -182,7 +187,12 @@ func run(args []string, stdout, stderr io.Writer) int {
 				}
 			}
 
-			if r, err = reporter.New(format, stdout, stderr, reporter.InfoLevel, termWidth); err != nil {
+			verbosityLevel, err := reporter.ParseVerbosityLevel(context.String("verbosity"))
+			if err != nil {
+				return err
+			}
+
+			if r, err = reporter.New(format, stdout, stderr, verbosityLevel, termWidth); err != nil {
 				return err
 			}
 
--- a/cmd/osv-scanner/main.go
+++ b/cmd/osv-scanner/main.go
@@ -28,8 +28,8 @@ func run(args []string, stdout, stderr io.Writer) int {
 
 	cli.VersionPrinter = func(ctx *cli.Context) {
 		// Use the app Writer and ErrWriter since they will be the writers to keep parallel tests consistent
-		r = reporter.NewTableReporter(ctx.App.Writer, ctx.App.ErrWriter, false, 0)
-		r.PrintTextf("osv-scanner version: %s\ncommit: %s\nbuilt at: %s\n", ctx.App.Version, commit, date)
+		r = reporter.NewTableReporter(ctx.App.Writer, ctx.App.ErrWriter, reporter.InfoLevel, false, 0)
+		r.Infof("osv-scanner version: %s\ncommit: %s\nbuilt at: %s\n", ctx.App.Version, commit, date)
 	}
 
 	osv.RequestUserAgent = "osv-scanner/" + version.OSVVersion
@@ -182,14 +182,14 @@ func run(args []string, stdout, stderr io.Writer) int {
 				}
 			}
 
-			if r, err = reporter.New(format, stdout, stderr, termWidth); err != nil {
+			if r, err = reporter.New(format, stdout, stderr, reporter.InfoLevel, termWidth); err != nil {
 				return err
 			}
 
 			var callAnalysisStates map[string]bool
 			if context.IsSet("experimental-call-analysis") {
 				callAnalysisStates = createCallAnalysisStates([]string{"all"}, context.StringSlice("no-call-analysis"))
-				r.PrintTextf("Warning: the experimental-call-analysis flag has been replaced. Please use the call-analysis and no-call-analysis flags instead.\n")
+				r.Infof("Warning: the experimental-call-analysis flag has been replaced. Please use the call-analysis and no-call-analysis flags instead.\n")
 			} else {
 				callAnalysisStates = createCallAnalysisStates(context.StringSlice("call-analysis"), context.StringSlice("no-call-analysis"))
 			}
@@ -234,21 +234,21 @@ func run(args []string, stdout, stderr io.Writer) int {
 
 	if err := app.Run(args); err != nil {
 		if r == nil {
-			r = reporter.NewTableReporter(stdout, stderr, false, 0)
+			r = reporter.NewTableReporter(stdout, stderr, reporter.InfoLevel, false, 0)
 		}
 		switch {
 		case errors.Is(err, osvscanner.VulnerabilitiesFoundErr):
 			return 1
 		case errors.Is(err, osvscanner.NoPackagesFoundErr):
-			r.PrintErrorf("No package sources found, --help for usage information.\n")
+			r.Errorf("No package sources found, --help for usage information.\n")
 			return 128
 		}
-		r.PrintErrorf("%v\n", err)
+		r.Errorf("%v\n", err)
 	}
 
 	// if we've been told to print an error, and not already exited with
 	// a specific error code, then exit with a generic non-zero code
-	if r != nil && r.HasPrintedError() {
+	if r != nil && r.HasErrored() {
 		return 127
 	}
 
--- a/pkg/reporter/gh-annotations_reporter.go
+++ b/pkg/reporter/gh-annotations_reporter.go
@@ -9,30 +9,46 @@ import (
 )
 
 type GHAnnotationsReporter struct {
-	hasPrintedError bool
-	stdout          io.Writer
-	stderr          io.Writer
+	hasErrored bool
+	stdout     io.Writer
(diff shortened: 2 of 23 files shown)


kemzeb:
Not sure if I resolved all the problem lines that Codecov complained about as it has been silent with my recent force pushes; making this PR ready for review

pkg/reporter/table_reporter.go
```diff
@@ -17,22 +18,23 @@ type TableReporter struct {
...
 
-func NewTableReporter(stdout io.Writer, stderr io.Writer, markdown bool, terminalWidth int) *TableReporter {
+func NewTableReporter(stdout io.Writer, stderr io.Writer, level VerbosityLevel, markdown bool, terminalWidth int) *TableReporter {
 	return &TableReporter{
-		stdout:          stdout,
-		stderr:          stderr,
-		hasPrintedError: false,
-		markdown:        markdown,
-		terminalWidth:   terminalWidth,
+		VerbosityLevelPrinter: NewVerbosityLevelPrinter(level),
```
kemzeb, line 23: I'm unsure if I should use an embedded struct; `VerbosityLevelPrinter` is just a wrapper over `fmt.Fprintf(...)` , where we check to make sure we can print at the provided level before calling it. We could alternatively make it a field or find some other approach to avoid having to duplicate this check across all reporters

kemzeb: Also, is there any value in writing unit tests for these new `Reporter` methods? If so, would it be better to create a *_test.go for each `Reporter` implementation to test them separately. Or, could I create a reporter_test.go and just test all the implementations here (I guess the value of these tests would be to make sure they are printing in their respective level e.g. a `Verbosef()` implementation is actually printing in the `VerboseLevel`)?

another-rex: > I'm unsure if I should use an embedded struct; VerbosityLevelPrinter is just a wrapper over fmt.Fprintf(...) , where we check to make sure we can print at the provided level before calling it. We could alternatively make it a field or find some other approach to avoid having to duplicate this check across all who implement Reporter

Hmm I don't think it's worth the complexity of having the embeded struct, it makes the code a bit harder to understand.

Personally I think there's not that much repetition to have the if statement in every print statement, and makes it very clear what is happening. 
```
func (r *TableReporter) Warnf(format string, a ...any) {
    if WarnLevel <= r.level {
        fmt.Fprintf(writer, format, a...)
    }
}
```

This part of the code also doesn't have to change if we want to add additional levels in the future.

another-rex: > Also, is there any value in writing unit tests for these new Reporter methods? If so, would it be better to create a *_test.go for each Reporter implementation to test them separately. Or, could I create a reporter_test.go and just test all the implementations here (I guess the value of these tests would be to make sure they are printing in their respective level e.g. a Verbosef() implementation is actually printing in the VerboseLevel)?

Let's have them in separate _test.go files, I think it's worth testing now that there's additional logic here.

cmd/osv-scanner/main.go
```diff
@@ -115,6 +115,11 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Name:  "no-call-analysis",
 				Usage: "disables call graph analysis",
 			},
+			&cli.StringFlag{
+				Name:  "verbosity-level",
```
another-rex, line 119: Let's just call this verbosity

another-rex (commented):
Thanks! Overall looks good to me.

another-rex:
I just merged #722 which would have caused some conflicts, should be straightforward to resolve as it only removes unused print functions.

kemzeb:
I believe I made all the necessary changes. I also changed some of the `Reporter` documentation because I felt `Infof()`'s comment would apply to all the other verbosity-related methods (well besides `Errorf()`). Please let me know if there are any further problems!

kemzeb:
Recent change has pkg/reporter/gh-annotations_reporter_test.go create the correct reporter (was creating JSONReporter). It does seem like there is a lot of test code duplication; let me know if this may be a problem

kemzeb:
Also just realized that CodeCov bot just modifies its comment when I make changes to my branch instead of creating a new comment.  

Since the changes I made to the reporter has also changed older areas of the codebase, it is complaining about these areas not be covered. Not sure what I should do; should I try to write coverage tests for these older areas?

another-rex:
> Also just realized that the CodeCov bot just modifies its comment when I make changes to my branch instead of creating a new comment.
> 
> Since the changes I made to the reporter has also changed older areas of the codebase, it is complaining about these areas not being covered. Not sure what I should do; should I try to write coverage tests for these older areas?

No need, as long as the new code has good coverage you're good, and it looks like it does :).

another-rex:
> Recent change has pkg/reporter/gh-annotations_reporter_test.go create the correct reporter (was creating JSONReporter). It does seem like there is a lot of test code duplication; let me know if this may be a problem

Yeah it is quite a bit of duplication, I think eventually we want to decouple the results reporting from the general info/warn/err printing, but that is something for V2.

For now I don't have any ideas for reducing this without over complicating the tests.

another-rex (approved):

pkg/reporter/gh-annotations_reporter.go
```diff
@@ -12,27 +12,45 @@ type GHAnnotationsReporter struct {
...
 	}
 }
 
-func (r *GHAnnotationsReporter) PrintErrorf(msg string, a ...any) {
-	fmt.Fprintf(r.stderr, msg, a...)
-	r.hasPrintedError = true
+func (r *GHAnnotationsReporter) Errorf(format string, a ...any) {
+	if ErrorLevel <= r.level {
+		fmt.Fprintf(r.stderr, format, a...)
+		r.hasPrintedError = true
```
G-Rath, line 30: This should always be set, regardless of if the error is actually printed anyway - it's used to determine if there were any errors that mean the exit code should not be 0.

Though I also wonder if we should be having a condition for this? it doesn't matter much here except maybe letting you simplify tests, but the only time I expect we might want to not print a fatal error would be if we supported a "silent" level, and even then tbh I'd be on the fence since we're talking about errors that completely exploded the tool.

pkg/reporter/reporter.go
```diff
@@ -4,26 +4,31 @@ import (
...
 	// Where the error is actually printed (if at all) is entirely up to the actual
 	// reporter, though generally it will be to stderr.
-	PrintErrorf(msg string, a ...any)
-	// HasPrintedError returns true if there have been any calls to PrintErrorf.
+	Errorf(format string, a ...any)
+	// HasPrintedError returns true if there have been any calls to Errorf.
 	//
 	// This does not actually represent if the error was actually printed anywhere
 	// since what happens to the error message is up to the actual reporter.
 	HasPrintedError() bool
```
G-Rath, line 25: I think we should change this to `HasErrored` to mirror the rename from `PrintErrorf`, and that should also help make it clearer this isn't actually about if an error was printed but just "did an error happen at all?"

G-Rath (changes requested):
ooh love getting a new bikeshed for me to lure people into through follow up PRs :D

Just got a bug to fix and a recommended change, and then this should be good to go - I'm also pretty sure I know a clean way to deduplicate the tests, but it's not that important

pkg/reporter/gh-annotations_reporter.go
G-Rath: ah right this actually isn't a bug because the condition is always true - I do not like that though because it feels too easy to mistakenly bug out; i.e. if you change `ErrorLevel` to something else, then the only test that will fail is the `Reporter` one even though it actually breaks a bunch of other tests throughout the codebase which are using the default verbosity

kemzeb: > This should always be set, regardless of if the error is actually printed anyway - it's used to determine if there were any errors that mean the exit code should not be 0.

Thanks for the review, and nice catch!

> Though I also wonder if we should be having a condition for this? 

Yeah I thought of this too but believed adding an if statement in these `Errorf()` implementations would avoid future problems if levels lower than `ErrorLevel` were introduced. Let me know if you feel differently!

pkg/reporter/reporter.go
kemzeb: I also renamed the field associated to it (in all `Reporter` implementations that have this field) from `hasPrintedError` -> `hasErrored` to keep it consistent 👍

kemzeb:
(Recent change updates some of my test error messages that are out-of-date)

pkg/reporter/gh-annotations_reporter.go
G-Rath: > Let me know if you feel differently!

I do very slightly for two reasons: 1. I really cannot imagine a situation where we'd not want to report an error even with a `--silent` flag I'd make a argument that the errors we're expecting to report with this shouldn't be silenced, and 2. GoLand flags this condition as always being true - those two combined ever-so-slightly push the needle for me towards removing the condition, because less IDE warnings is always nice (and it's technically slightly less code too, which I guess arguably is a third reason)

I won't block over this though, so up to you :)

kemzeb: Makes sense to me, I'll make the necessary changes (though I'm not sure if the tests for `Errorf()` should also change?)

G-Rath: Na the tests should be fine because they're testing the behaviour not the implementation - they shouldn't care about what logic actually goes on, just that the outcome is what's expected for their cases.

G-Rath (approved):
Looks good to me, though it'd be great if you could chuck in a couple of basic CLI tests (`cmd/osv-scanner/main_test.go`) with different verbosity levels - among other things, that'll catch if setting `HasErrored` is ever mucked up

kemzeb:
> Looks good to me, though it'd be great if you could chuck in a couple of basic CLI tests (cmd/osv-scanner/main_test.go) with different verbosity levels - among other things, that'll catch if setting HasErrored is ever mucked up

Sure thing, making the changes now

another-rex (approved):
Thanks!

PR #729, opened 2024-01-01
verify the spdx licenses passed in to the license allowlist

josieang:


pkg/spdx/gen.go
josieang, line 1: code for reading spdx expressions from https://github.com/spdx/license-list-data/blob/main/json/licenses.json into a file using go generate

another-rex (approved):
LGTM! Thanks!

another-rex:
Can you add a test for the unrecognized code path?

josieang:
> Can you add a test for the unrecognized code path?

done!

PR #732, opened 2024-01-04
Add option to not fail on vulnerability being found for github action

another-rex:
Add an option for the github action reporter to not return a failing code when vulnerabilities are found. This can be used to switch the osv-scanner action into a more "informative" mode rather than forcing strictly 0 vulnerabilities. 

The success/failure of a workflow can now indicate whether the workflow itself completely successfully and uploaded scan results, rather than showing the contents of the results. 

A followup PR will be created adding the option to the workflows once this is merged.

### DIFF
--- a/cmd/osv-reporter/main.go
+++ b/cmd/osv-reporter/main.go
@@ -165,11 +165,8 @@ func run(args []string, stdout, stderr io.Writer) int {
 				}
 			}
 
-			failOnVuln := true // Default true
-			if context.IsSet("fail-on-vuln") && !context.Bool("fail-on-vuln") {
-				// If flag is set to false, only then set to false
-				failOnVuln = false
-			}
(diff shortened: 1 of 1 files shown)


cmd/osv-reporter/main.go
```diff
@@ -160,9 +165,14 @@ func run(args []string, stdout, stderr io.Writer) int {
 				}
 			}
 
+			failOnVuln := true // Default true
```
oliverchang, line 168: nit: can we just simplify this to:

```golang
failOnVuln := !context.IsSet("fail-on-vuln") || context.Bool("fail-on-vuln")
```

?

oliverchang (approved):

cmd/osv-reporter/main.go
another-rex: Yep

PR #736, opened 2024-01-05
Show ecosystem and version even if git is shown if the info exists.

### LINKED ISSUE #735: Issue with Blank Package Versions in JSON Output when Scanning composer.lock File
### Description
When scanning the `composer.lock` file using JSON output, the package version information is returned as blank.

### Attachments
- [composer.lock](https://github.com/google/osv-scanner/files/13833609/composer.lock.zip)
- [JSON Output File](https://github.com/google/osv-scanner/files/13833603/test.json)

### Environment
- Operating System: CentOS
- Scanner Version: v1.5.0

another-rex:
Fixes #735 

Also update composer.lock in our integration tests fixture to contain a vulnerability to catch this issue in the future.

### DIFF
--- a/cmd/osv-scanner/fixtures/locks-insecure/composer.lock
+++ b/cmd/osv-scanner/fixtures/locks-insecure/composer.lock
@@ -1,1 +1,104 @@
-{}
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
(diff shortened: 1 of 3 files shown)


oliverchang (approved):

G-Rath (approved):
I personally don't mind this change, but fwiw I'm pretty sure it means we're favoring versions over commits - while that'll generally _look_ nicer it is arguably less accurate i.e. I can give an NPM package whatever name and version I like but I can't "forge" a commit.

PR #737, opened 2024-01-05
Add option to not fail on vuln to workflow files

another-rex:
Follow up PR to #732 to add the option to the actual workflow files.

### DIFF
--- a/.github/workflows/osv-scanner-unified-action.yml
+++ b/.github/workflows/osv-scanner-unified-action.yml
@@ -49,3 +49,4 @@ jobs:
         --skip-git
         ./
         ./docs/
+      fail-on-vuln: false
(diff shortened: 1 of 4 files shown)


oliverchang (approved):

.github/workflows/osv-scanner-unified-action.yml
```diff
@@ -49,3 +49,4 @@ jobs:
         --skip-git
         ./
         ./docs/
+      fail-on-vuln: false
```
oliverchang, line 52: actually one question: what's the motivation for false in this case?

another-rex: Ah, good catch!, it shouldn't be false, I accidentally left that in from me testing this workflow. Removed now.

PR #739, opened 2024-01-09
fix: ensure are sorted consistently

G-Rath:
I've found on my snapshotting work (#717) that the `postgres-stretch.cdx.xml` based local database test sometimes fails due to row order being slightly different.

It looks like this is only an issue with the `local` package, but I think it's better to always sort especially since there's no good way to test all sources to catch if their order ever changes.

You can see the changes in the sorting more easily here: https://github.com/google/osv-scanner/pull/717/commits/fa666485bc5307fd068c762e248ff0d61bf774a6

### DIFF
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/google/osv-scanner/internal/local"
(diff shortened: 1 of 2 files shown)


pkg/osvscanner/osvscanner.go
```diff
@@ -789,6 +790,13 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
 		return models.VulnerabilityResults{}, err
 	}
 
+	// ensure results are sorted for consistency
+	for _, results := range vulnsResp.Results {
```
another-rex, line 794: Can you move this into `buildVulnerabilityResults()` in vulnerability_result.go next to the code that does the results output sorting?

another-rex (commented):
I thought we already did sorting, turns out it was only for source path. Let's just add this bit to the existing sorting that is being done.

pkg/osvscanner/osvscanner.go
G-Rath: Doing that at least initially doesn't give the same output - I'll do a bit more digging

G-Rath:
I think this was just a fluke so I'll close it and revisit if I run into it again

Issue #731, opened 2024-01-02
Vulnerabilities in docker image ghcr.io/google/osv-scanner found using syft


jayvdb:
Using osv-scanner, there are no errors:
```bash
 osv-scanner --docker ghcr.io/google/osv-scanner
Scanned docker image with 0 packages
No package sources found, --help for usage information.
```
However using https://github.com/anchore/syft, there are lots
```bash
> syft packages ghcr.io/google/osv-scanner:latest -o cyclonedx-xml > sbom.cdx.xml
 ✔ Pulled image                    
 ✔ Loaded image                                                                                                                                                                           ghcr.io/google/osv-scanner:latest
 ✔ Parsed image                                                                                                                                     sha256:6b736c19e37d8d3c116dd78aa04f20e2b9715daaacd916a3ae5073e61d9190e0
 ✔ Cataloged packages              [130 packages]  
> osv-scanner --sbom sbom.cdx.xml 
Scanned sbom.cdx.xml as CycloneDX SBOM and found 128 packages
╭─────────────────────────────────────┬──────┬───────────┬─────────────────────────────┬─────────┬──────────────╮
│ OSV URL                             │ CVSS │ ECOSYSTEM │ PACKAGE                     │ VERSION │ SOURCE       │
├─────────────────────────────────────┼──────┼───────────┼─────────────────────────────┼─────────┼──────────────┤
│ https://osv.dev/GHSA-mw99-9chc-xw7r │ 7.5  │ Go        │ github.com/go-git/go-git/v5 │ v5.10.1 │ sbom.cdx.xml │
│ https://osv.dev/GHSA-45x7-px36-x8w8 │ 5.9  │ Go        │ golang.org/x/crypto         │ v0.16.0 │ sbom.cdx.xml │
│ https://osv.dev/GO-2023-2402        │      │           │                             │         │              │
│ https://osv.dev/GO-2023-2185        │      │ Go        │ stdlib                      │ 1.20.11 │ sbom.cdx.xml │
│ https://osv.dev/GO-2023-2382        │      │ Go        │ stdlib                      │ 1.20.11 │ sbom.cdx.xml │
│ https://osv.dev/GO-2023-2185        │      │ Go        │ stdlib                      │ 1.20.11 │ sbom.cdx.xml │
│ https://osv.dev/GO-2023-2382        │      │ Go        │ stdlib                      │ 1.20.11 │ sbom.cdx.xml │
[...]

PR #740, opened 2024-01-10
Assume latest patch version if version does not exist

another-rex:
While the change made in #704 is the ideal solution, it only works for people using go 1.21 or later. Adding a patch version to go.mod will cause go1.20 or earlier to fail to build. Since 1.20 is still fully supported and is still the default version for many workflows, this causes too many breakages for users to be acceptable.

There's a few alternative fixes:  

- Assume latest patch version (This PR)
  - This means osv-scanner is no longer checking whether a project is updating to the latest patch version to resolve vulnerabilities, but will still catch users using an out of support go minor version that has no patch available. 
- Remove stdlib vulnerability check entirely / move stdlib vuln check behind a experimental flag
- Add a way to hint to osv-scanner the go version (e.g. .go-version file, or a comment in the go.mod file specifically for osv-scanner to check).

I am leaning towards assuming the latest patch as a fix in the mean time until go1.22 releases (at which point 1.20 will no longer be supported), and less people will be using 1.20, and perhaps reintroducing the patch version check.

---

This implementation also means when a stdlib vulnerability is found, the version that is printed will be 1.xx.99, which is not quite accurate.

### DIFF
--- a/pkg/lockfile/parse-go-lock_test.go
+++ b/pkg/lockfile/parse-go-lock_test.go
@@ -132,7 +132,7 @@ func TestParseGoLock_TwoPackages(t *testing.T) {
 		},
 		{
 			Name:      "stdlib",
-			Version:   "1.17.0",
+			Version:   "1.17.9999",
 			Ecosystem: lockfile.GoEcosystem,
 			CompareAs: lockfile.GoEcosystem,
 		},
@@ -181,7 +181,7 @@ func TestParseGoLock_IndirectPackages(t *testing.T) {
 		},
 		{
 			Name:      "stdlib",
-			Version:   "1.17.0",
+			Version:   "1.17.9999",
 			Ecosystem: lockfile.GoEcosystem,
 			CompareAs: lockfile.GoEcosystem,
 		},
--- a/pkg/lockfile/parse-go-lock_test.go
+++ b/pkg/lockfile/parse-go-lock_test.go
@@ -132,7 +132,7 @@ func TestParseGoLock_TwoPackages(t *testing.T) {
 		},
 		{
 			Name:      "stdlib",
-			Version:   "1.17.9999",
+			Version:   "1.17",
 			Ecosystem: lockfile.GoEcosystem,
 			CompareAs: lockfile.GoEcosystem,
 		},
@@ -181,7 +181,7 @@ func TestParseGoLock_IndirectPackages(t *testing.T) {
 		},
 		{
 			Name:      "stdlib",
-			Version:   "1.17.9999",
+			Version:   "1.17",
 			Ecosystem: lockfile.GoEcosystem,
 			CompareAs: lockfile.GoEcosystem,
 		},
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -862,7 +862,14 @@ func filterUnscannablePackages(packages []scannedPackage) []scannedPackage {
 // patchPackageForRequest modifies packages before they are sent to osv.dev to
 // account for edge cases.
 func patchPackageForRequest(pkg scannedPackage) scannedPackage {
-	// Remove stdlib from go
+	// Assume Go stdlib patch version as the latest version
+	//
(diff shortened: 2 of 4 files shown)


G-Rath (approved):
Looks fine from a code POV, though fwiw my ideal* would be the opposite as I feel false positives are better than false negatives in this space.

Maybe a good middle ground would be support `.go-version` and/or maybe `go version` output? I was hoping `go version` would just give a number, but it doesn't not - but still I think this idea isn't bad: if you supported reading that, then people in CI could do `go version > .go-version` before running the scanner, and then it would know the exact version being used 🤔 

*: with things as they current are - obviously if I had a magic wand my ideal would be something a lot more accurate

another-rex:
Well we had `go version` output support in #484 , which caused #679 and #731 and some other issues I think.

Tldr is that the CI running osv-scanner might not be the CI building the project, and it's doubly confusing when running it locally changes the vulnerabilities being shown.

another-rex:
I agree with .go-version support though, I think that'll be a good option to go with. Maybe we pick the latest from go.mod and .go-version and say that's the go version being used to compile the project.

pkg/lockfile/parse-go-lock_test.go
```diff
@@ -132,7 +132,7 @@ func TestParseGoLock_TwoPackages(t *testing.T) {
 		},
 		{
 			Name:      "stdlib",
-			Version:   "1.17.0",
+			Version:   "1.17.9999",
```
oliverchang, line 135: Hmm, if we surface this to users, this might look a bit weird and raise a lot of questions.

I wonder if we should just  assume .0 here  to avoid false negatives as well. To mitigate false positives, users could either:

1. Just bump the Go version to another minor version.  This should be fairly "easy" to do for most users given that Go is backwards compatible?
2. Add an exclusion via an OSV toml config.

In  any case,  this problem will go away after some time as newer Go versions are adopted.  

This does have another implication though: Go library developers may be incentivised to have a lower go directive in go.mod, in maximise the compatibility of their library (since after go 1.21, the go toolchain will refuse to use modules declaring newer versions). Generally speaking, library developers may still need a way to turn the stdlib scanning off some other way, but I'm not sure what that should look like. Perhaps we can wait to see if we get any feedback on this.

another-rex: > 1. Just bump the Go version to another minor version. This should be fairly "easy" to do for most users given that Go is backwards compatible?

Even the latest go minor version (1.21.0) has 10+ vulnerabilities that's fixed over 1.21.1 - 1.21.5, 

> Add an exclusion via an OSV toml config.
Adding individual vulnerability exclusions in OSV toml config is not practical because of the sheer number of stdlib vulnerabilities. I do remember talking about adding package exclusions as an option to osv-scanner.toml. This will turn off all stdlib scanning though, even for out of support versions..

another-rex: I can make a better fix than adding `9999` though, it's just the easiest/cleanest way to implement this.
I'll just "inject" the patch version when sending to osv.dev API,

oliverchang: Sgtm! Given the volume of vulns, injecting this patch version seems the most reasonable.

pkg/osvscanner/osvscanner.go
```diff
@@ -858,6 +859,25 @@ func filterUnscannablePackages(packages []scannedPackage) []scannedPackage {
 	return out
 }
 
+// patchPackageForRequest modifies packages before they are sent to osv.dev to
+// account for edge cases.
+func patchPackageForRequest(pkg scannedPackage) scannedPackage {
+	// Remove stdlib from go
```
oliverchang, line 865: Is this comment accurate? 

We should add some more details/justification inline in a comment on why we do this:

another-rex: Oops, I don't know why I wrote that haha.

I clarified this and added additional reasoning for why we are doing this.

oliverchang (approved):

Issue #478, opened 2023-08-11
osv-scanner subcommands


oliverchang:
We need to introduce a concept of subcommands to OSV-Scanner. e.g.

```
osv-scanner scan ... # with the current set of flags 
osv-scanner update ...
osv-scanner fix ...
```

We should do this in a backwards compatible way -- if no subcommand is specified, we fall back to "scan".

oliverchang:
@hogo6002 this might be an interesting one to design and implement! 

@cuixq @michaelkedar FYI

andrewpollock:
Drive by signal boost of https://clig.dev/

PR #748, opened 2024-01-16
Add OSV-Scanner subcommands

hogo6002:
Add subcommands. https://github.com/google/osv-scanner/issues/478


cmd/osv-scanner/main.go
```diff
@@ -35,211 +34,148 @@ func run(args []string, stdout, stderr io.Writer) int {
...
-			if errPrint := r.PrintResult(&vulnResult); errPrint != nil {
-				return fmt.Errorf("failed to write output: %w", errPrint)
-			}
+	}
 
-			// This may be nil.
-			return err
-		},
+	if len(args) >= 2 {
+		if args[1] != "scan" && args[1] != "update" && args[1] != "fix" && args[1] != "--help" && args[1] != "--version" {
```
another-rex, line 174: Add these against to an array variable, and populate that variable with a function that loops through the top level commands in the cli struct to make sure these 2 are always synced.

cmd/osv-scanner/main.go
```diff
@@ -35,211 +34,148 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+			args = append(args, "")
+			copy(args[2:], args[1:])
```
another-rex, line 176: Add a comment about what this is doing (e.g. defaulting to the scan function)

another-rex (commented):
Thanks! Added some comments. It looks like it this PR doesn't include the warning when the file exists on disk?

go.mod
```diff
@@ -58,4 +58,4 @@ require (
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20231002182017-d307bd883b97 // indirect
 	google.golang.org/protobuf v1.31.0 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
-)
+)
```
cuixq, line 61: is this a format thing?

cuixq, line 61: is this a format thing?

cmd/osv-scanner/main.go
```diff
@@ -265,6 +202,22 @@ func run(args []string, stdout, stderr io.Writer) int {
 	return 0
 }
 
+func getAllCommands(commands []*cli.Command) []string {
```
cuixq, line 205: any reason not returning a map? since we are checking if it contains something later

cmd/osv-scanner/main_test.go
```diff
@@ -1520,6 +1520,48 @@ Filtered 2 vulnerabilities from output
 	}
 }
 
+func TestRun_SubCommands(t *testing.T) {
+	t.Parallel()
+	tests := []cliTestCase{
```
cuixq, line 1525: shall we also test without any subcommand and default to `scan`?

cmd/osv-scanner/main_test.go
```diff
@@ -1520,6 +1520,48 @@ Filtered 2 vulnerabilities from output
 	}
 }
 
+func TestRun_SubCommands(t *testing.T) {
```
cuixq, line 1523: maybe also add some comment that we should test other sub-commands as well if there will be more in the future

cmd/osv-scanner/main.go
```diff
@@ -35,211 +34,149 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+	}
+	allCommands := getAllCommands(app.Commands)
+	if len(args) >= 2 {
+		// Insert the default command to args if no command is specified.
+		if !slices.Contains(allCommands, args[1]) {
+			// Avoid in-place change to args as it is not a pointer.
+			argsTmp := make([]string, len(args)+1)
+			copy(argsTmp[2:], args[1:])
+			argsTmp[1] = app.DefaultCommand
+			args = argsTmp
```
michaelkedar, line 173: I think this can be
```go
args = slices.Insert(args, 1, app.DefaultCommand)
```

But I have a question: shouldn't setting `app.DefaultCommand` handle this? or does it just not work as we want it to?

michaelkedar (commented):
A few small comments:

cmd/osv-scanner/main_test.go
hogo6002: I didn't add tests for without subcommand as all other existing tests are doing that.

go.mod
hogo6002: yeah, it's by `go mod tidy`

cmd/osv-scanner/main.go
```diff
@@ -35,211 +34,149 @@ func run(args []string, stdout, stderr io.Writer) int {
...
-						return nil
-					}
+		Name:           "osv-scanner",
+		Version:        version.OSVVersion,
+		Usage:          "scans various mediums for dependencies and matches it against the OSV database",
+		Suggest:        true,
+		Writer:         stdout,
+		ErrWriter:      stderr,
+		DefaultCommand: "scan",
+		Commands: []*cli.Command{
```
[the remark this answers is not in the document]
hogo6002: moved subcommand into a function in its respective folder

cmd/osv-scanner/main_test.go
hogo6002: Good suggestion. Thanks!

cmd/osv-scanner/main.go
hogo6002: Yeah, map does give better performance for checking. But, since we only need to get all valid commands here (which is a very small amount), I think it's fine to use a slice as it offers better readability.

cmd/osv-scanner/main.go
```diff
@@ -35,211 +34,149 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+		// Insert the default command to args if no command is specified.
+		if !slices.Contains(allCommands, args[1]) {
+			// Avoid in-place change to args as it is not a pointer.
+			argsTmp := make([]string, len(args)+1)
+			copy(argsTmp[2:], args[1:])
+			argsTmp[1] = app.DefaultCommand
+			args = argsTmp
+		} else if _, err := os.Stat(args[1]); err == nil {
+			if r == nil {
+				r = reporter.NewTableReporter(stdout, stderr, reporter.InfoLevel, false, 0)
```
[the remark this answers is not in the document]
hogo6002: Updated, now the warning message goes to stderr.

cmd/osv-scanner/main.go
hogo6002: `app.DefaultCommand` only works when we don't pass any other args in. Once we pass other args, like flags or filename, it will treat them as a command. 
Some of our UTs are using the same args to run multiple times and expect the same results, so I didn't do in-place change here.

cmd/osv-scanner/osv-scanner
kemzeb, line 1: Not sure if this was mentioned already, but it looks like you included your executable here by accident

hogo6002: Thanks Kemzeb!

michaelkedar (approved):
LGTM

cmd/osv-scanner/scan/main.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"github.com/google/osv-scanner/pkg/osvscanner"
+	"github.com/google/osv-scanner/pkg/reporter"
+	"github.com/google/osv-scanner/pkg/spdx"
+	"golang.org/x/term"
+
+	"github.com/urfave/cli/v2"
+)
+
+func GetCommand(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
```
oliverchang, line 19: nit: no need for "Get" in the name. see https://google.github.io/styleguide/go/decisions#getters

cmd/osv-scanner/main.go
```diff
@@ -265,6 +69,51 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+	}
+
+	return allCommands
+}
+
+// Inserts the default command to args if no command is specified.
+func insertDefaultCommand(argsPointer *[]string, stdout, stderr io.Writer, app *cli.App, r *reporter.Reporter) {
+	allCommands := getAllCommands(app.Commands)
+	args := *argsPointer
+	if len(args) >= 2 {
```
oliverchang, line 100: nit: invert this condition and early return instead (to avoid lots of indentation that follows.

cmd/osv-scanner/main.go
```diff
@@ -265,6 +69,51 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+	for _, flag := range cli.VersionFlag.Names() {
+		allCommands = append(allCommands, "-"+flag)
+		allCommands = append(allCommands, "--"+flag)
+	}
+
+	return allCommands
+}
+
+// Inserts the default command to args if no command is specified.
+func insertDefaultCommand(argsPointer *[]string, stdout, stderr io.Writer, app *cli.App, r *reporter.Reporter) {
```
oliverchang, line 97: nit: can we just return the resulting command instead of passing a pointer here to overwrite?

cmd/osv-scanner/main.go
```diff
@@ -265,6 +69,53 @@ func run(args []string, stdout, stderr io.Writer) int {
...
+		// Avoids modifying args in-place, as some unit tests rely on its original value for multiple calls.
+		argsTmp := make([]string, len(args)+1)
+		copy(argsTmp[2:], args[1:])
+		argsTmp[1] = app.DefaultCommand
+
+		// Executes the cli app with the new args.
+		return argsTmp
+	} else if _, err := os.Stat(args[1]); err == nil {
+		r := reporter.NewJSONReporter(stdout, stderr, reporter.InfoLevel)
+		r.Warnf("Warning: '%v' exists as both a subcommand of OSV-Scanner and as a file in the filesystem. It operates as a command here. If you intend to scan the file, please specify a subcommand.\n", args[1])
```
oliverchang, line 113: Nit: "It operates as a command here." seems a bit confusing to me. I would reword this to somethin like:

```
Warning: `%s` exists as both a subcommand of OSV-Scanner and as a file on the filesystem. `%s` is assumed to be a subcommand here. If you intended for `%s` to be an argument to `%s`, you must specify `%s %s` in your commandline.
```

Also do we need "%v"? can we use "%s" here?

cmd/osv-scanner/main_test.go
```diff
@@ -1520,6 +1520,48 @@ Filtered 2 vulnerabilities from output
...
 }
 
+// Tests all subcommands here.
+func TestRun_SubCommands(t *testing.T) {
+	t.Parallel()
+	tests := []cliTestCase{
+		// with scan subcommand
+		{
+			name:         "with scan subcommand",
+			args:         []string{"", "scan", "./fixtures/locks-many/composer.lock"},
```
oliverchang, line 1530: Can we add more tests to test when e.g. "scan" isn't specified?

cmd/osv-scanner/main_test.go
```diff
@@ -1520,6 +1520,48 @@ Filtered 2 vulnerabilities from output
...
+			name:         "scan with a flag",
+			args:         []string{"", "scan", "--recursive", "./fixtures/locks-one-with-nested"},
+			wantExitCode: 0,
+			wantStdout: `
+				Scanning dir ./fixtures/locks-one-with-nested
+				Scanned <rootdir>/fixtures/locks-one-with-nested/nested/composer.lock file and found 1 package
+				Scanned <rootdir>/fixtures/locks-one-with-nested/yarn.lock file and found 1 package
+				No issues found
+			`,
+			wantStderr: "Warning: 'scan' exists as both a subcommand of OSV-Scanner and as a file in the filesystem. It operates as a command here. If you intend to scan the file, please specify a subcommand.",
```
oliverchang, line 1551: Can we add some testcases for when this error isn't output?

oliverchang (commented):
looks great! just some minor comments to address.

cmd/osv-scanner/main_test.go
hogo6002: When we specify the 'scan' subcommand, the error always shows here as we do have a folder called `./scan`. So I added some tests for `insertDefaultCommand()` to test this type of case.

oliverchang (approved):

cuixq (approved):

another-rex (approved):

Issue #352, opened 2023-04-18
Automated/guided remediation


oliverchang:
Tracking issue for building a guided remediation feature as part of OSV-Scanner. 

Some ideas:
- Suggesting direct dependency updates to remediate transitive vulns. 
- Minimal re-locks to avoid known dependencies. 
- Graph visualisations.

abhisek:
@oliverchang This will be a VERY useful feature. I have some thoughts here:

Suggest minimal number of direct dependencies to upgrade to specific version to remediate maximum risk. 

There is ofcourse a lot of detail to be considered here like how to measure aggregated `risk`. But it will be very useful in driving remediation, which many a times is not feasible due to too many findings or version conflicts in transitive dependencies when one or more direct dependency is updated.

Also another relatively simpler but valuable feature would be to have the ability to identify direct dependency upgrade to specific version to remediate vulnerabilities in transitive dependencies. I think you already suggested that. I am not sure if it is possible to build a `graph` by parsing flat lockfiles like `gradle.lockfile` or `requirements.txt`

In fact, just by having a dependency graph representation instead of a list will enable a lot of mitigation related analysis capability. I am not sure if its already done, but it **may be** feasible to build a usable graph based on data from `deps.dev` API.

Is there any plan to combine OSV and `deps.dev` data in future to perform both vulnerability detection and effective remediation?

oliverchang:
Thanks for the feedback @abhisek ! We are definitely looking at ways to make the mechanisms flexible enough for users to decide on what they want to prioritise. Please stay tuned here for once we have some things ready to try. 

We are working closely with the `deps.dev` team here.

agmond:
Hi @oliverchang,
Will the remediation feature include a command that automatically updates the manifest file and the lock file (like `npm audit fix`), or will it only have the guidelines for a manual upgrade?

oliverchang:
@agmond yes: this tool will give you the changed manifest and lockfile at the end of of the guided remediation.

agmond:
Hi @oliverchang, I have another question here.
Will I be able to remediate only a single vulnerability at a time?
For example, let's say I have several issues in my dependencies file, but I want to fix them one by one (and not all of them at once).
Will this be possible?

abhisek:
[...]

PR #757, opened 2024-01-19
Guided Remediation: Add manifest resolution

michaelkedar:
Starting to make guided remediation public #352 :tada:  

This PR has the code used to resolve the dependency graph of a manifest (i.e. a `package.json`) using the deps.dev resolvers and find vulnerabilities within it.

Doesn't include the code to actually parse/write `package.json` files - will probably add that in the next PR.

Much of this has been reviewed internally already, but I've made some significant changes/refactoring to `dependency_chain.go` and the `computeVulns()` function in `resolve.go`, so please take a more careful look at those.


go.mod
```diff
@@ -1,9 +1,12 @@
 module github.com/google/osv-scanner
 
-go 1.21
+go 1.21.1
+
+toolchain go1.21.5
```
michaelkedar, line 5: Hm, `go mod tidy` changes these automatically - what should I be doing about this?

oliverchang: @another-rex  ?

another-rex: Update it to 1.21.5, I think when dependencies start adding in patch versions in their go.mod file, it forces it on us as well when we import it.

oliverchang (approved):

internal/resolution/client/depsdev_client.go
```diff
@@ -0,0 +1,96 @@
...
+				System:  pb.System(vk.System),
+				Name:    vk.Name,
+				Version: vk.Version,
+			},
+		})
+		if err != nil {
+			continue
+		}
+
+		// Send off queries to cache the packages in the dependency tree
```
another-rex, line 56: Is this caching remote (on deps.dev servers) or local (guessing from datasource/depsdev_api.go)?  Trying to figure out how this caches packages.

cuixq (approved):
🎉

internal/resolution/datasource/depsdev_api.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// DepsDevAPIClient is a wrapper for InsightsClient that caches requests.
+type DepsDevAPIClient struct {
+	pb.InsightsClient
+
+	// cache fields
+	mu                sync.Mutex
+	cacheTimestamp    *time.Time
+	packageCache      map[packageKey]*pb.Package
```
another-rex, line 22: Is it possible to `sync.Map` instead of locking and unlocking here?

another-rex (commented):
What's a good way to test this manually before we get automated tests?

internal/resolution/client/depsdev_client.go
michaelkedar: This is local - The [deps.dev implementation of the client](https://github.com/google/deps.dev/blob/main/util/resolve/api.go) calls the relevant `datasource/depsdeb_api.go` methods, which will fill in its own maps.

oliverchang:
> What's a good way to test this manually before we get automated tests?

Re tests, it would be easiest to run the internal version for now. Check with @michaelkedar on this.

internal/resolution/datasource/depsdev_api.go
michaelkedar: I *could*, it's just annoying having to cast from `any` every time when using it.
Plus it's probably better to have a single lock for the `GobEncode` / `GobDecode` functions

another-rex: Fair enough, let's stick with lock then!

michaelkedar:
Yeah, manually testing this as-is would be clumsy - you have to manually fill in the `Manifest` struct.

another-rex (approved):

PR #765, opened 2024-01-23
Guided Remediation: Add dependency relaxation & re-resolution

michaelkedar:
The remediation part of guided remediation :)
Adds the functionality to attempt to fix vulnerabilities in a manifest by relaxing the requirements of its direct dependencies.

I've simplified & refactored `tryRelaxRemediate` (née `RelaxResolve`), so PTAL at that. 

I still need to migrate the part that uses this to make the list of possible 'patches' to the manifest (hence the `//nolint:unused` everywhere), but I didn't want to keep bringing more and more bits into one PR.

### DIFF
--- a/internal/remediation/relax.go
+++ b/internal/remediation/relax.go
@@ -63,13 +63,8 @@ func reqsToRelax(res *resolution.ResolutionResult, vulnIDs []string, opts Remedi
 		if !slices.Contains(vulnIDs, v.Vulnerability.ID) || (!opts.DevDeps && v.DevOnly) {
 			continue
 		}
-		chains := v.ProblemChains
-		if len(chains) == 0 {
-			// Just in case something is wrong with the problem heuristic
-			chains = v.NonProblemChains
-		}
 		// Only relax dependencies if their chain length is less than MaxDepth
-		for _, ch := range chains {
+		for _, ch := range v.ProblemChains {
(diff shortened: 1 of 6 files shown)


internal/remediation/relax.go
```diff
@@ -0,0 +1,90 @@
...
+func reqsToRelax(res *resolution.ResolutionResult, vulnIDs []string, opts RemediationOptions) []int {
+	toRelax := make(map[resolve.VersionKey]string)
+	for _, v := range res.Vulns {
+		// Don't do a full opts.MatchVuln() since we know we don't need to check every condition
+		if !slices.Contains(vulnIDs, v.Vulnerability.ID) || (!opts.DevDeps && v.DevOnly) {
+			continue
+		}
+		chains := v.ProblemChains
+		if len(chains) == 0 {
+			// Just in case something is wrong with the problem heuristic
```
oliverchang, line 68: Is this something we can do at the place where we can setting/computing these chains?

cuixq (approved):
LGTM!

oliverchang (approved):

internal/remediation/relax.go
michaelkedar: Moved this check into `computeVulns`, where the chains are being classified.

oliverchang (approved):

PR #766, opened 2024-01-24
Guided Remediation: Add computation for all relaxation patches

michaelkedar:
Following on from #765, adds `ComputeRelaxPatches` for generating the possible remediation options after a relock.
Also added a new(ish) cache for OSV API requests, which speeds up the above quite a bit.

### DIFF
--- a/internal/resolution/resolve_vulns.go
+++ b/internal/resolution/resolve_vulns.go
@@ -1,0 +1,149 @@
+package resolution
+
+import (
+	"context"
+	"maps"
+	"sync"
+
+	"deps.dev/util/resolve"
+	"github.com/google/osv-scanner/internal/utility/vulns"
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/google/osv-scanner/pkg/models"
+	"github.com/google/osv-scanner/pkg/osv"
+)
+
+// computeVulns scans for vulnerabilities in a resolved graph and populates res.Vulns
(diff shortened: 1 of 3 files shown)


internal/resolution/resolve_vulns.go
```diff
@@ -0,0 +1,149 @@
...
+		res.Vulns = append(res.Vulns, rv)
+	}
+
+	return nil
+}
+
+// vulnCache caches all vulnerabilities affecting any versions of particular packages.
+// We cache call vulns & manually check affected, rather than querying the affected versions directly
+// since remediation needs to query for OSV vulnerabilities multiple times for the same packages.
+var vulnCache sync.Map // map[resolve.PackageKey][]models.Vulnerability
```
oliverchang, line 73: is there a way for us to avoid this being a global? this could be surpising for future users of this as a library.

michaelkedar: I could pass it in as part of the arguments, it needs to be passed in all of `ComputeRelaxPatches` -> `tryRelaxRemediate` -> `Resolve` -> `computeVulns`.

It probably makes sense to bundle an `OSVClient` together with the `resolve.Client` (which would also make supporting local dbs easier) - I might do that in another PR.

oliverchang: Ack.  Please  add a TODO here  to  do  that.

oliverchang (approved):

cuixq (approved):

another-rex (approved):

PR #767, opened 2024-01-25
Parse severities for guided remediation

michaelkedar:
Factored out the CVSS Score parsing from the table output into its own function to reuse it in guided remediation. The new function also returns the human-readable rating string ("LOW", "HIGH", etc.) which I will end up using for the interactive guided remediation mode.

I also made some changes to the table output of the scores:
- Always render the scores to 1 decimal place, so `6.0` instead of just `6`
- Display `0.0` if the CVSS score actually evaluates to 0, vs nothing when there is no severity listed

### DIFF
--- a/internal/remediation/remediation.go
+++ b/internal/remediation/remediation.go
@@ -4,6 +4,7 @@ import (
 	"slices"
 
 	"github.com/google/osv-scanner/internal/resolution"
+	"github.com/google/osv-scanner/internal/utility/severity"
 )
 
 type RemediationOptions struct {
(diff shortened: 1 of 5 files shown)


internal/remediation/remediation.go
```diff
@@ -35,8 +36,16 @@ func (opts RemediationOptions) MatchVuln(v resolution.ResolutionVuln) bool {
...
+	maxScore := -1.0
+	// TODO: also check Vulnerability.Affected[].Severity
+	for _, sev := range v.Vulnerability.Severity {
+		if score, _, _ := severity.CalculateScore(sev); score > maxScore {
+			maxScore = score
+		}
+	}
+
+	return maxScore < 0 || // Always include vulns with unknown severities
+		int(10*maxScore) >= int(10*opts.MinSeverity) // CVSS scores are only to 1 decimal place
```
oliverchang, line 48: Any reason in particvular for this int(10*X) bit? Add a comment.

michaelkedar: Clarified comment on why (& used `math.Round` instead)

oliverchang (approved):

internal/utility/severity/severity_test.go
```diff
@@ -0,0 +1,93 @@
...
+			want: result{
+				score:  -1,
+				rating: "UNKNOWN",
+			},
+		},
+		{
+			name: "CVSS v2.0",
+			sev: models.Severity{
+				Type:  models.SeverityCVSSV2,
+				Score: "AV:L/AC:M/Au:N/C:N/I:P/A:C/E:H/RL:U/RC:C/CDP:LM/TD:M/CR:L/IR:M/AR:H",
```
cuixq, line 35: just curious - how these strings are generated?

cuixq (approved):

internal/utility/severity/severity_test.go
michaelkedar: I clicked through a bunch of options on these calculators: [v2.0](https://nvd.nist.gov/vuln-metrics/cvss/v2-calculator), [v3.0](https://www.first.org/cvss/calculator/3.0), [v3.1](https://www.first.org/cvss/calculator/3.1), [v4.0](https://www.first.org/cvss/calculator/4.0)

another-rex (approved):

PR #768, opened 2024-01-25
Add pictures to github action docs

another-rex:
Add some pictures to github action docs. 

The specific version of the workflow referenced in the docs does not yet exist, but will soon be tagged.


oliverchang (approved):
nice!

PR #776, opened 2024-01-29
Add pdm lockfile support

jtt:
Add support for parsing package information from  `pdm.lock` -files used by `pdm`, package and dependency manager for Python (https://pdm-project.org/latest/)


pkg/lockfile/parse-pdm-lock.go
G-Rath, line 1: would you mind restructuring these to use individual functions to be consistent with the rest of our lockfile tests

pkg/lockfile/parse-pdm-lock.go
```diff
@@ -0,0 +1,80 @@
...
+
+type PdmLockFile struct {
+	Version  string           `toml:"lock-version"`
+	Packages []PdmLockPackage `toml:"package"`
+}
+
+const PdmEcosystem = PipEcosystem
+const PdmLockFileName = "pdm.lock"
+const pdmGroupDefault = "default"
+const pdmGroupDev = "dev"
```
G-Rath, line 25: we don't typically use constants for these, and you're only referencing them each once so lets just inline them

pkg/lockfile/parse-pdm-lock.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"path/filepath"
+
+	"github.com/BurntSushi/toml"
+)
+
+type PdmLockPackage struct {
+	Name     string   `toml:"name"`
+	Version  string   `toml:"version"`
+	Groups   []string `toml:"groups"`
```
G-Rath, line 13: just noting for others: it looks like pdm only _very_ recently started recording dependency groups at the package level in its lockfiles - I couldn't actually find any locks searching GitHub, but the PRs on pdm support this feature as being added

pkg/lockfile/parse-pdm-lock.go
```diff
@@ -0,0 +1,80 @@
...
+			details.Commit = pkg.Revision
+		}
+
+		packages = append(packages, details)
+	}
+
+	return packages, nil
+}
+
+//nolint:gochecknoinits
```
G-Rath, line 73: we should ensure the extractor matches the interface:

```suggestion
}

var _ Extractor = PdmLockExtractor{}

//nolint:gochecknoinits
```

pkg/lockfile/parse-pdm-lock.go
```diff
@@ -0,0 +1,80 @@
...
+	Version  string           `toml:"lock-version"`
+	Packages []PdmLockPackage `toml:"package"`
+}
+
+const PdmEcosystem = PipEcosystem
+const PdmLockFileName = "pdm.lock"
+const pdmGroupDefault = "default"
+const pdmGroupDev = "dev"
+
+type PdmExtractor struct{}
```
G-Rath, line 27: This matches the naming we use for the other extractors:

```suggestion
type PdmLockExtractor struct{}
```

G-Rath (changes requested):
Looks good, just got a few changes - would you mind also adding an "empty.lock" (which is "no packages", not "completely empty file") and "two-packages.lock" fixtures as well? (I know technically your other tests cover this, but it's consistent with the standard set of fixtures we have for our other lockfiles)

pkg/lockfile/parse-pdm-lock.go
jtt: Restructured

pkg/lockfile/parse-pdm-lock.go
jtt: Removed constants

pkg/lockfile/parse-pdm-lock.go
jtt: Added.

pkg/lockfile/parse-pdm-lock.go
jtt: Changed the name

jtt:
> Looks good, just got a few changes - would you mind also adding an "empty.lock" (which is "no packages", not "completely empty file") and "two-packages.lock" fixtures as well? (I know technically your other tests cover this, but it's consistent with the standard set of fixtures we have for our other lockfiles)

Added requested tests. Amended fixes to the existing commit.

pkg/lockfile/parse-pdm-lock_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+func expectNilErr(t *testing.T, err error) {
+	t.Helper()
+	if err != nil {
+		t.Errorf("got unexpected error: %v", err)
+	}
+}
+
+func TestParsePdmLock_FileDoesNotExist(t *testing.T) {
+	t.Parallel()
```
G-Rath, line 76: nit: these are pretty squashed together

usually I like to have a new line between every function, after `t.Parallel()`, and before the first `expect*` call

G-Rath (approved):
Awesome stuff! It'd be great if you could space out the tests a bit, but I think this is good to go!

pkg/lockfile/parse-pdm-lock_test.go
jtt: Added newlines to test functions.

another-rex (approved):
Thanks!

PR #778, opened 2024-01-30
Guided Remediation: add npm registry clients & `.npmrc` parsing

michaelkedar:
The datasource and `DependencyClient` for querying the npm registry API directly, instead of relying on deps.dev.
Also, parses `.npmrc` configs to allow resolution of requirements from private registries.

Practically unchanged from what we had internally, besides a bunch of linting complaints.


internal/resolution/client/npm_registry_client.go
```diff
@@ -0,0 +1,267 @@
...
+		// Get the preferred version of the import requirement
+		vks, err := c.MatchingVersions(ctx, im.VersionKey)
+		if err != nil || len(vks) == 0 {
+			continue
+		}
+
+		vk := vks[len(vks)-1]
+
+		// Make a request for the precomputed dependency tree
+		// TODO: avoid relying on DepsDevClient internals
```
cuixq, line 219: is this TODO done?

cuixq (approved):

internal/resolution/client/npm_registry_client.go
michaelkedar: Just changed the code a bit - now it's done.

internal/resolution/client/npm_registry_client.go
```diff
@@ -0,0 +1,282 @@
...
+
+	if isNpmBundle(vk.PackageKey) { // bundled dependencies, fallback to deps.dev client
+		return c.fallback.Requirements(ctx, vk)
+	}
+	dependencies, err := c.api.Dependencies(ctx, vk.Name, vk.Version)
+	if err != nil {
+		return nil, err
+	}
+
+	depCount := len(dependencies.Dependencies) + len(dependencies.DevDependencies) +
```
oliverchang, line 102: Can you add a comment explaining this formula a bit? In particular the 2* part.

oliverchang (approved):

oliverchang (approved):

internal/resolution/datasource/npm_registry.go
```diff
@@ -0,0 +1,173 @@
...
+			BundleDependencies:   jsonToStringSlice(data.Get("bundleDependencies")),
+		}
+	}
+	pkgData = npmRegistryPackageDetails{
+		Versions: versions,
+		Tags:     jsonToStringMap(jsonData.Get("dist-tags")),
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
another-rex, line 143: Is there a specific reason this deferred rather than placed on the next line?

internal/resolution/datasource/npm_registry.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/tidwall/gjson"
+	"golang.org/x/exp/maps"
+)
+
+type NpmRegistryAPIClient struct {
+	// registries from the npmrc config
+	registries npmRegistries
+
+	// cache fields
+	mu             sync.Mutex
```
another-rex, line 21: Can you add a comment here that this mutex is only for locking the bottom 2 fields and not for the `registries` field (and therefore `registries` shouldn't be modified from other threads ) (if this is the case)?

internal/resolution/client/npm_registry_client.go
```diff
@@ -0,0 +1,285 @@
...
+				VersionType: resolve.Concrete,
+			},
+		}
+	}
+
+	return resolve.MatchRequirement(vk, resVersions), nil
+}
+
+func isNpmBundle(pk resolve.PackageKey) bool {
+	return strings.Contains(pk.Name, ">")
```
another-rex, line 221: Add an example here of npm bundle key

internal/resolution/client/npm_registry_client.go
```diff
@@ -0,0 +1,285 @@
...
+	if concVer, ok := versions.Tags[vk.Version]; ok {
+		// matched a tag, return just the concrete version of the tag
+		return []resolve.Version{{
+			VersionKey: resolve.VersionKey{
+				PackageKey:  vk.PackageKey,
+				Version:     concVer,
+				VersionType: resolve.Concrete,
+			},
+		},
+		}, nil
```
another-rex, line 203:
```suggestion
			},
		}}, nil
```
nit: the indents here are a bit confusing

another-rex (commented):
Just some minor nits, otherwise LGTM!

PR #786, opened 2024-02-04
ci: remove unneeded `setup-go` step and pin `actions/download-artifact`

G-Rath:
I was having a look at the openssf scorecard score as I've not really dug into that before, and saw it was flagging some unpinned actions.

Most of them were about the `osv-scanner` action itself which I don't think can be pinned since the point is to use the latest version for self testing right? but we can pin `actions/download-artifact` and `actions/setup-go` _though_ I'm pretty sure that isn't needed anymore because it was a workaround for https://github.com/github/codeql/issues/13992 and we were just waiting on a new version? either way we'll find out.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -43,7 +43,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       # Update go to the latest version to support minor go versions is go.mod file
       - name: Install Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
           go-version-file: go.mod
       # Initializes the CodeQL tools for scanning.
--- a/.github/workflows/osv-scanner-reusable.yml
+++ b/.github/workflows/osv-scanner-reusable.yml
@@ -55,7 +55,7 @@ jobs:
         with:
           persist-credentials: false
       - name: "Download custom artifact if specified"
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@6b208ae046db98c579e8a3aa621ab581ff575935 # v4.1.1
         if: "${{ inputs.download-artifact != '' }}"
         with:
           name: "${{ inputs.download-artifact }}"


another-rex:
We need  that `setup-go` step to workaround a CodeQL problem, it was added here: https://github.com/google/osv-scanner/pull/724

G-Rath:
@spencerschrock I wouldn't mind picking your brain a little if that's ok about the "Token-Permissions" check - most notably the docs say for remediation:

> Set top-level permissions as read-all or contents: read as described in GitHub's [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).

But the docs also say you can use `{}` to disable all permissions for all scopes which is what I would have thought would be the most ideal setup for the top-level permissions? That might just be a new thing that the docs are missing (in which case, happy to open a PR) but I'd prefer to know for sure what the optimal setup is before opening a PR changing anything 😅

G-Rath:
> We need that setup-go step to workaround a CodeQL problem

Yeah that's for https://github.com/github/codeql/issues/13992 which I was hoping had been fixed in the latest release but apparently not so I'll add it back in and pin it

spencerschrock:
> @spencerschrock I wouldn't mind picking your brain a little if that's ok about the "Token-Permissions" check - most notably the docs say for remediation:
> 
> > Set top-level permissions as read-all or contents: read as described in GitHub's [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).
> 
> But the docs also say you can use `{}` to disable all permissions for all scopes which is what I would have thought would be the most ideal setup for the top-level permissions? That might just be a new thing that the docs are missing (in which case, happy to open a PR) but I'd prefer to know for sure what the optimal setup is before opening a PR changing anything 😅

Yes, `read-all` is probably overkill, especially given that `security-events` is included in that list. It was added explicitly to the docs in https://github.com/ossf/scorecard/pull/552, but the rationale is vague. The code probably only checked for top-level writes, so the doc was edited to match.

If I had to guess (Laurent is OOO currently), it's a usability trade-off. Easier to tell people to set to `read-all` than have them figure out which read permissions are needed. But I would agree `{}` is ideal from a least privilege perspective.

G-Rath:
@hogo6002 want to land this while you're at it?

another-rex (approved):

PR #789, opened 2024-02-06
Guided Remediation: Compute in-place updates

michaelkedar:
Adding in-place update computation, for finding 'drop-in' fixes of vulns in indirect dependencies.

I did end up rewriting a lot of this, so I'd appreciate a more thorough review of this PR.

Something to note for the future: unlike what happens internally, the returned patches can be incompatible with each other (i.e. could possibly suggest two different versions for the same package). I'll need to make sure the caller handles this possibility.

### DIFF
--- a/internal/remediation/in_place.go
+++ b/internal/remediation/in_place.go
@@ -1,0 +1,300 @@
+package remediation
+
+import (
+	"cmp"
+	"context"
+	"errors"
+	"slices"
+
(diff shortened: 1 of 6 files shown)


cuixq (approved):
LGTM!

another-rex (approved):

oliverchang (approved):

PR #792, opened 2024-02-09
Start setting up guided remediation subcommand

michaelkedar:
Just the `osv-scanner fix` flag parsing, the command itself doesn't do anything.
I'd appreciate opinions on the naming/aliasing of the flags.

### DIFF
--- a/cmd/osv-scanner/fix/main.go
+++ b/cmd/osv-scanner/fix/main.go
@@ -106,6 +106,7 @@ func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 			},
 
 			&cli.BoolFlag{
+				// TODO: allow for finer control e.g. specific packages, major/minor/patch
 				Category: upgradeCategory,
 				Name:     "disallow-major-upgrades",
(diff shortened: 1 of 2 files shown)


cmd/osv-scanner/fix/main.go
```diff
@@ -0,0 +1,220 @@
...
+				Usage:    "apply the top N patches",
+				Value:    -1,
+			},
+
+			&cli.BoolFlag{
+				Category: upgradeCategory,
+				Name:     "disallow-major-upgrades",
+				Usage:    "disallow major version changes to dependencies",
+			},
+			&cli.StringSliceFlag{
```
oliverchang, line 113: I think we discussed this, but we need to think of a more extensible semantics for this.

e.g. something like 

```
--disallow-upgrades=PKG-1:major,PKG-2, 
```

Can be a TODO for now.

cuixq (approved):

cmd/osv-scanner/fix/main.go
michaelkedar: Added a TODO

PR #797, opened 2024-02-13
Dont traverse gitignored dirs for gitignore files

robramsaynz:
Updated version of https://github.com/google/osv-scanner/pull/397.

> this seems to be an upstream issue in go-git, and i'll prepare a PR for them soon, but for now this copies in the affected function and fixes it by checking the accumulated patterns while walking the fs looking for gitignore files
> 
> fixes: https://github.com/google/osv-scanner/issues/389

- is located github.com/**ackama**/osv-scanner, rather than github.com/**robotdana**/osv-scanner
- has tests
- has extra changes to handle https://github.com/google/osv-scanner/pull/397#issuecomment-1920529596 (handles mid-tree dirs, handles non-git dirs, handles recursive flag)


robramsaynz:
@another-rex This is ready to look at now. I may add a few more tests to it, but it currently works, and has reasonable test coverage. I'm not sure how to add you as a reviewer so I'm just @-ing you here.

pkg/osvscanner/customgitignore/dir.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	commentPrefix   = "#"
+	gitDir          = ".git"
+	gitignoreFile   = ".gitignore"
+	infoExcludeFile = gitDir + "/info/exclude"
+)
+
+// readIgnoreFile reads a specific git ignore file.
+func readIgnoreFile(fs billy.Filesystem, path []string, ignoreFile string) (ps []gitignore.Pattern, err error) {
```
another-rex, line 23: I'm guessing the path as an slice/array is the case upstream? 

It seems bit weird that all this seems to do is force us to do a split when calling `readIgnoreFile` and `readPattern`, and immediately gets joined back together in these two functions.

pkg/osvscanner/customgitignore/walk_up_to_root.go
```diff
@@ -0,0 +1,234 @@
...
+	// Read parent's dirs up to git-root
+	//
+	if pathAbs != repoRootPath {
+		newPs, err = readIgnoreFilesFromParents(fs, pathRel, repoRootPath)
+		ps = append(ps, newPs...)
+	}
+
+	// Read children's subdirs
+	//
+	switch {
```
another-rex, line 119: nit: I think it'll be clearer to just use a if else here.

another-rex (commented):
Thanks! Added some comments. 

Also, how different is this from upstream? I think ideally we push this change upstream, and avoid having to maintain this. But since it's only a few files, I'm not too concerned if we can't get it upstream.

pkg/osvscanner/customgitignore/walk_up_to_root.go
```diff
@@ -0,0 +1,234 @@
...
+// 		- read all .gitignore files in parent dirs up to and including repo-root
+// 			(ie only dirs that are an ancestor)
+// 		- read .git/info/exclude for repo
+//
+// - parsing a plain dir: .gitignore files are ignored
+//
+// In all cases any dirs matched by a previously read
+// .gitignore are skipped, unless it's the path (ie directly
+// supplied by the user).
+func ParseGitIgnores(path string, recursive bool) (returnPs []gitignore.Pattern, repoRootPath string, err error) {
```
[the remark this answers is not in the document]
G-Rath: pretty sure that lint is disabled, and we're using these in a few places already - overall I think they can have their uses, but that you should explicitly list them in the `return` calls to make it clear (which @robramsaynz is doing), and not use them that often

pkg/osvscanner/customgitignore/dir_test.go
```diff
@@ -0,0 +1,309 @@
+// TODO: check if there's a way to make this: `package customgitignore` instead
+package customgitignore_test
```
G-Rath, line 2:
```suggestion
package customgitignore_test
```

we discussed this IRL - overall it's preferable to use a dedicated `_test` package because it helps enforce you test the _public_ API; it's only when something is really complicated and deeply nested that we want to be testing against private internals

pkg/osvscanner/customgitignore/dir_test.go
```diff
@@ -0,0 +1,309 @@
...
+				"because it's parent dir_b should have been ignored by a rule in repository-root .gitignore")
+		}
+	}
+}
+
+func setupGitRepo(t *testing.T) string {
+	t.Helper()
+
+	// A unique tempdir for local funcs to create a git repo inside
+	gitRepo := filepath.Join(t.TempDir(), "git_repo")
```
G-Rath, line 266: let's use `testutility.CreateTestDir` (once #832 lands) for this, and you want to return the cleanup function to be `defer`'d in each test

pkg/osvscanner/customgitignore/dir_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+	var allPaths string
+
+	allPaths = filepath.Join(gitRepo, filepath.FromSlash(".git/info/"))
+	if err := os.MkdirAll(allPaths, 0755); err != nil {
+		t.Errorf("could not create paths for test: %v", err)
```
G-Rath, line 272: I assume that the tests cannot properly continue if these fail, so they should be `t.Fatalf`

pkg/osvscanner/customgitignore/dir_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+	var hasMatch bool
+
+	// expect ./dir_a/dir_b/.gitignore to be processed
+	//
+	// Normally the `dir_a/dir_b/.gitignore` entry in
+	// REPO_ROOT/.gitignore would stop this .gitignore from being
+	// processed, but it's read because the file has been explicitly
+	// supplied from the command-line.
+	hasMatch = slices.ContainsFunc(patterns, func(p gitignore.Pattern) bool {
```
G-Rath, line 41: I wonder if these might be better handled by using snapshots - you could probably replace all of the checks with just a single `testutility.NewSnapshot().MatchJSON(t, patterns)` (idk what would give nicer output `MatchText` or `MatchJSON` - I'll let you check and decide)

G-Rath (changes requested):
I know you're still working on some improvements to this after our IRL catchup, so I've just done a small sprinkle of initial comments

pkg/osvscanner/customgitignore/dir_test.go
robramsaynz: done

pkg/osvscanner/customgitignore/dir_test.go
robramsaynz: Lets do that. We don't need a `defer`-ed cleanup because that's already done for us by `t.Tempdir()` [LINK](https://pkg.go.dev/testing#T.TempDir)

G-Rath: [`t.Tempdir` is not suitable due our snapshotting](https://github.com/google/osv-scanner/pull/832#discussion_r1508090735), and I think we should be consistent with what we're using - but #832 switches to using `t.Cleanup` which means you still won't have to do the `defer`

pkg/osvscanner/customgitignore/walk_up_to_root.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// path WRT git-root
+	pathRel, err := filepath.Rel(repoRootPath, pathAbs)
+	if err != nil {
+		return ps, "", err
+	}
+
+	fs = osfs.New(repoRootPath)
+
+	// Read parent dirs
```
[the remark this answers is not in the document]
robramsaynz: @another-rex This comment is here as a "code after this is reading the parent dirs" comment, in the same way [this comment](https://github.com/google/osv-scanner/pull/797/files#diff-8b3e805815de31b538fd3c919f4a0d341ad6382fecd193cc1e4dfd38ae4c4331R90) indicates a notable change is the state of the parser.

How would you prefer this is handled?

1. remove this comment
2. remove all "state change" comments
3. change to something like `// --- Reading parent dirs ---` (or other similar style)
4. something else

pkg/osvscanner/customgitignore/walk_up_to_root.go
robramsaynz: done

pkg/osvscanner/customgitignore/dir_test.go
robramsaynz: Sure. Was there any action you wanted now, besides "wait the future"?

G-Rath: Ideally for now I'd pull in the helper like we are elsewhere - that way its consistent pre-#832 and so if this gets merged before that, that PR will get rebased to move this helper and visa versa if #832 get merged first.

Though I expect it'll get landed by Tuesday so probably best to just wait

pkg/osvscanner/customgitignore/dir.go
robramsaynz: Yes. The requirementment comes from upstream. I'm not a fan of thinking about path's this way, but I assume this is related to the fact we're hooking into the internals of a library, and processing dirs this way helps speed/algorthim reasons.

My guess is it makes it easier to work out where you were with respect to the .gitignore that created a pattern.
So a pattern from `REPO_ROOT/dir_a/.gitignore` would only apply to things matching `REPO_ROOT/dir_a/**`.
With this style of `path`, you can efficiently traverse down the tree appending to `path` in every dir.

eg, here's what I got from the tests (note the `domain` field.)

```
&gitignore.pattern{domain:[]string{".", "dir_a", "dir_b"}, pattern:[]string{"DIR_B_GITIGNORE"}, inclusion:false, dirOnly:false, isGlob:false} 

```

pkg/osvscanner/customgitignore/dir_test.go
robramsaynz: I'm guessing not. `p` is a data structure that's handled in a particular way — I've only bothered to learn how the API works, and that returns a bool.

For reference a pattern looks something like this:


```
&gitignore.pattern{domain:[]string{".", "dir_a", "dir_b"}, pattern:[]string{"DIR_B_GITIGNORE"}, inclusion:false, dirOnly:false, isGlob:false} 
```

pkg/osvscanner/customgitignore/walk_up_to_root.go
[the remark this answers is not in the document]
G-Rath: The thing is that you're not just labeling the returns but actually creating variables, which are used when you don't have an explicit return value and the linter will not report that - that's why Go itself recommends their use only for short functions, because it's easy to get them lost (especially as part of refactors) and they can harm readability within the function itself

There is actually a linter that outright bans them which we've disabled because we're agreed they are fine in small doses, but I wouldn't use them extensively and would even go so far as to say the general default convention in Go is to not use them (though happy to be disagreed with on that).

On top of that a lot of the time (including here) you're not actually using the named variable so you're throwing away a small allocation everytime these functions are used.

I don't think you should be using it on this large function, and looking over your smaller functions I don't think any of them really get much benefit either as they can't take advantage of the default variable value

pkg/osvscanner/customgitignore/dir.go
```diff
@@ -0,0 +1,86 @@
+// based on https://github.com/go-git/go-git/blob/v5.7.0/plumbing/format/gitignore/dir.go
+// but modified so that it skips ignored directories while traversing for gitignore files
+
+package customgitignore
```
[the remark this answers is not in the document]
robramsaynz: done

pkg/osvscanner/customgitignore/dir_test.go
robramsaynz: Have added a a TODO for this so it's not lost.

internal/customgitignore/walk_up_to_root.go
```diff
@@ -0,0 +1,247 @@
...
+//
+// 		- read .gitignore in start dir
+// 		- read all .gitignore files in parent dirs up to and including repo-root
+// 			(but only dirs that are an ancestor)
+// 		- read .git/info/exclude for repo
+//
+// NOTE: the dir you're passing in directly could be a dir that is ignored
+// (targeted by a parent's .gitignore or the per-repo exclude file); in this
+// case, the dir's .gitignore file is still processed, but not its sub-dirs.
+
```
another-rex, line 74: I think this also needs a `//` otherwise go will think this is two different comment blocks.
```suggestion
//
```

robramsaynz: good catch. godoc would have made a small, and rather confusing summary for that func.

pkg/osvscanner/customgitignore/walk_up_to_root.go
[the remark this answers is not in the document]
robramsaynz: done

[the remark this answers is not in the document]
another-rex: I don't really have a strong preference on this one, though I think for the purposes of documenting what the return values are we can just put it in the function comment/docstring, avoiding the extra allocation (though I wonder if the go compiler is smart enough to just not do the allocation by default).

pkg/osvscanner/customgitignore/walk_up_to_root.go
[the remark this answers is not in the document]
another-rex: I think what confused me is that immediately underneath you are reading the current directory (which now I see can be interpreted as the "parent" directory of the file passed in). Hmm I think we can just remove this comment since you already comment the individual sections.

robramsaynz:
@another-rex RE:

> Also, how different is this from upstream? I think ideally we push this change upstream, and avoid having to maintain this. But since it's only a few files, I'm not too concerned if we can't get it upstream.

I've moved things round to make them easier to diff, but it shouldn't be massively different. You can see the difference with:

```
I've tried to keep these fairly similar to make diffing easier. You can check the diff with:

$ curl https://raw.githubusercontent.com/go-git/go-git/v5.7.0/plumbing/format/gitignore/dir.go > a.go
$ curl https://raw.githubusercontent.com/ackama/osv-scanner/dont-traverse-gitignored-dirs-for-gitignore-files/pkg/osvscanner/customgitignore/dir.go > b.go
$ diff -u a.go b.go
```

We are messing with a parsing algorithm, so that's risky a little, but also have fairly good tests for this generally. I'm expecting upstream to like the changes. My guess is they wont want `walk_up_to_root.go` given they're a general purpose library and this is the "make it work like ripgrep" file. For instance the existing API of `ReadPatterns()` seems to assume it gets passed the root path of a git repo.

pkg/osvscanner/customgitignore/walk_up_to_root.go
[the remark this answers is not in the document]
robramsaynz: any word on how you want this handled @another-rex ?

internal/customgitignore/dir_test.go
```diff
@@ -0,0 +1,732 @@
...
+		t.Errorf("Expected to find a pattern matching dir_a/dir_b/DIR_B_GITIGNORE from ./dir_a/dir_b/.gitignore")
+	}
+
+	// expect ./.git/info/exclude to be processed, by backtracking up the tree
+	hasMatch = slices.ContainsFunc(patterns, func(p gitignore.Pattern) bool {
+		return p.Match([]string{".", "REPO_EXCLUDE_FILE"}, false) == gitignore.Exclude
+	})
+
+	if !hasMatch {
+		t.Errorf("Expected to find a pattern matching REPO_EXCLUDE_FILE from ./.git/info/exclude ")
```
G-Rath, line 56:
```suggestion
		t.Errorf("Expected to find a pattern matching REPO_EXCLUDE_FILE from ./.git/info/exclude")
```

(please apply to all tests)

internal/customgitignore/dir_test.go
```diff
@@ -0,0 +1,732 @@
...
+	gitRepo := setupGitRepo(t)
+
+	// Read this dir-tree using customgitignore, starting at
+	// the top of the tree at ./dir_a/dir_b
+	start := filepath.Join(gitRepo, "dir_a", "dir_b")
+
+	// Read this dir-tree using customgitignore, starting at the root
+	patterns, _, err := customgitignore.ParseGitIgnores(start, true)
+	if err != nil {
+		t.Errorf("could not read gitignore patterns for test: %v", err)
```
G-Rath, line 31: I assume we cannot continue with any of these tests if this errors:

```suggestion
		t.Fatalf("could not read gitignore patterns for test: %v", err)
```

(please apply to all tests)

internal/customgitignore/dir_test.go
```diff
@@ -0,0 +1,732 @@
...
+		}
+
+		if !d.IsDir() && d.Name() == ".gitignore" {
+			os.Remove(path)
+		}
+
+		return nil
+	})
+	if err != nil {
+		t.Errorf("Could not walk tree removing .gitignores because: %v", err)
```
G-Rath, line 610:
```suggestion
		t.Fatalf("Could not walk tree removing .gitignores because: %v", err)
```

internal/customgitignore/dir_test.go
```diff
@@ -0,0 +1,732 @@
...
+// Hack to test if gip.pattern == []string{"TEST_STRING"}
+//
+// This is necessary because gitignore.Pattern.pattern is
+// non-exported.
+//
+// This matches against the return value from fmt.Sprintf
+// with %#v, which means it may trip up on complicated or
+// user supplied strings.
+//
+// By why can't we just test something like
```
G-Rath, line 718:
```suggestion
// But why can't we just test something like
```

internal/customgitignore/walk_up_to_root.go
```diff
@@ -0,0 +1,254 @@
...
+	// read .gitignore
+	newPs, err := readIgnoreFile(fs, toGoGitPath(pathRel), gitignoreFile)
+	if err != nil {
+		return ps, err
+	}
+	ps = append(ps, newPs...)
+
+	switch {
+	case pathAbs == pathGitRoot:
```
G-Rath, line 172: nit: it feels like it'd be a little simpler to use an `if` here

pkg/osvscanner/customgitignore/walk_up_to_root.go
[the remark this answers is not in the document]
another-rex: Let's just remove this comment

another-rex (approved):
LGTM!

G-Rath (approved):

internal/customgitignore/dir_test.go
```diff
@@ -0,0 +1,732 @@
...
+
+	return gitRepo
+}
+
+func setupPlainDirWithGitignores(t *testing.T) string {
+	t.Helper()
+
+	// A unique tempdir for local funcs to create a git repo inside
+	// TODO: convert `t.TempDir()` to `testutility.CreateTestDir(t)` once PR #832 lands
+	dir := filepath.Join(t.TempDir(), "base_dir")
```
G-Rath, line 647: #832 has been merged so this can be actioned now, though I don't think it should block landing this

Issue #495, opened 2023-08-20
Include CVSS MaxSeverity in JSON output


yene:
feature request

The markdown and table output include a CVSS score, calculated by MaxSeverity().
It would be really handy to have this score in the `results[].packages[].vulnerabilities[]` object for easy scripting against the output.

Note: Since the severity score is so hard to read for humans, maybe max severity should be part of the Vulnerability struct.

thank you

oliverchang:
@another-rex does this seem like something that would fit under `"groups"` ?

yene:
@oliverchang please consider this issue, it would enable scripting in build pipelines, something like
osv-scanner -r -json --format json | jq '.[]....score >= 9'
osv-scanner -r -json --format json | jq '.[]....severity == critical'

thank you for your time.

AppleGamer22:
Hi there.

I'd be happy to jump into this issue and suggest changes to the structs and JSON generation.

Kind regards,
Omri

another-rex:
Happy to take contributions!, 

I think this is a pretty good idea to put the MaxSeverity calculation directly onto the `Group` struct, I just haven't had time to do it yet.

PR #805, opened 2024-02-14
Maximum severity rating for each Group object in JSON output

AppleGamer22:
- Addresses issue #495.
- I'm happy to add tests for the `MaxSeverity` function if this is required for this issue.

### DIFF
--- a/pkg/osvscanner/vulnerability_result.go
+++ b/pkg/osvscanner/vulnerability_result.go
@@ -4,7 +4,7 @@ import (
 	"sort"
 	"strings"
 
-	out "github.com/google/osv-scanner/internal/output"
+	"github.com/google/osv-scanner/internal/output"
 	"github.com/google/osv-scanner/internal/sourceanalysis"
 	"github.com/google/osv-scanner/pkg/grouper"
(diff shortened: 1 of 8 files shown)


pkg/osvscanner/vulnerability_result.go
```diff
@@ -4,6 +4,7 @@ import (
 	"sort"
 	"strings"
 
+	out "github.com/google/osv-scanner/internal/output"
```
G-Rath, line 7: happy for this to be done in a follow up pr sometime, but ideally it would be better to rename the variable

G-Rath (approved):
looks good to me - would be nice to have this change expressed in a test somehow if possible

another-rex (commented):
Thanks! Can you also replace the other places MaxSeverity is being called to use what's now being stored in groups?

I believe it's called in githubannotations.go and table.go

AppleGamer22:
@another-rex @G-Rath thank you both for the thoughtful feedback.

AppleGamer22:
I'll get to writing the `MaxSeverity` test now...

another-rex (approved):
LGTM, thanks!

PR #815, opened 2024-02-21
Automated Updates: add version suggester for Maven

cuixq:
This PR adds the implementation for Maven version suggester:
 - The latest version of the specified package is returned based on the given options.
 - A version range requirement will be replaced to the latest version if this version does not satisfy the constraint.
 - Major updates can be ignored with option `NoMajorUpdates` set to true.

### DIFF
--- a/internal/remediation/suggester/suggest.go
+++ b/internal/remediation/suggester/suggest.go
@@ -1,0 +1,31 @@
+package suggest
+
+import (
+	"context"
+	"fmt"
+
+	"deps.dev/util/resolve"
+)
+
+type SuggestOptions struct {
+	NoMajorUpdates bool
+}
+
+// A VersionSuggester provides an ecosystem-specific method for 'suggesting'
+// the latest version of a dependency based on the given options.
+type VersionSuggester interface {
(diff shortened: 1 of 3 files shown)


internal/remediation/suggester/suggest.go
```diff
@@ -0,0 +1,31 @@
...
+	"deps.dev/util/resolve"
+)
+
+type SuggestOptions struct {
+	NoMajorUpdates bool
+}
+
+// A VersionSuggester provides an ecosystem-specific method for 'suggesting'
+// the latest version of a dependency based on the given options.
+type VersionSuggester interface {
```
michaelkedar, line 16: Thinking of if/how we can combine the Relaxer and Suggester:
Maybe we can have an `Approach` flag in the options that can be `Relax` or `Latest`, and just decide the behaviour based on that?

On the other hand, I'm not actually sure how much reusable code there is between suggesting latest/relaxing, and if keeping them separate makes it easier to tell when it's not implemented.

michaelkedar (approved):
LGTM

another-rex (approved):

oliverchang (approved):

internal/remediation/suggester/suggest.go
cuixq: SGTM!

PR #818, opened 2024-02-22
Align GoVulncheck Go version with go.mod

hogo6002:
GoVulncheck uses local installed GO version to determine vulnerabilities if the env `GOVERSION` is not defined. 
Set the `GOVERSION` value to the one defined in go.mod.

detail: https://github.com/google/osv-scanner/issues/679


pkg/osvscanner/osvscanner.go
```diff
@@ -888,13 +888,14 @@ func patchPackageForRequest(pkg scannedPackage) scannedPackage {
 
 func makeRequest(
 	r reporter.Reporter,
-	packages []scannedPackage,
+	packages *[]scannedPackage,
```
hogo6002, line 891: Updated it to a pointer as we modified this slice (stdlib package version https://github.com/google/osv-scanner/pull/740) in this function.

oliverchang (commented):
Is there a test we can add for this?

internal/sourceanalysis/go.go
```diff
@@ -25,6 +25,15 @@ func goAnalysis(r reporter.Reporter, pkgs []models.PackageVulns, source models.S
...
+	var goVersion string
+	for _, pkg := range pkgs {
+		if pkg.Package.Name == "stdlib" {
+			goVersion = pkg.Package.Version
+		}
+	}
+	os.Setenv("GOVERSION", fmt.Sprintf("go%s", goVersion))
+
 	vulns, vulnsByID := vulnsFromAllPkgs(pkgs)
 	res, err := runGovulncheck(filepath.Dir(source.Path), vulns)
```
spencerschrock, line 38: Can the Go version be passed down to `runGovulncheck` instead? `govulncheck` supports adding env variables to the scan command. https://pkg.go.dev/golang.org/x/vuln/scan#Cmd

`os.Setenv` affects the whole process, so we may get unexpected results when run concurrently. The first thing that comes to mind is tests, either here in osv-scanner, or Scorecard. Some of our e2e tests call osv-scanner concurrently from different Go routines.

pkg/osvscanner/osvscanner.go
```diff
@@ -862,7 +862,7 @@ func filterUnscannablePackages(packages []scannedPackage) []scannedPackage {
 
 // patchPackageForRequest modifies packages before they are sent to osv.dev to
 // account for edge cases.
-func patchPackageForRequest(pkg scannedPackage) scannedPackage {
+func patchPackageForRequest(pkg *scannedPackage) *scannedPackage {
```
another-rex, line 865: This should not be a pointer I believe. The passed in argument should be copied/immutable, the returned value is the patched version.

pkg/osvscanner/osvscanner.go
another-rex: I don't think this needs to be a pointer unless we are adding/removing items from the slice. We can access and directly set the items in the loop, as a slice is already a pointer to the elements within the slice.

pkg/osvscanner/osvscanner.go
```diff
@@ -888,13 +888,14 @@ func patchPackageForRequest(pkg scannedPackage) scannedPackage {
...
+	packages *[]scannedPackage,
 	compareLocally bool,
 	compareOffline bool,
 	localDBPath string) (*osv.HydratedBatchedResponse, error) {
 	// Make OSV queries from the packages.
 	var query osv.BatchedQuery
-	for _, p := range packages {
+	for i := range *packages {
+		p := &(*packages)[i]
 		p = patchPackageForRequest(p)
```
another-rex, line 899:
```suggestion
	for i, p := range packages {
		packages[i] = patchPackageForRequest(p)
```

Something like this should work.

internal/sourceanalysis/go.go
another-rex: +1

another-rex (changes requested):
Thanks! Added some comments

pkg/osvscanner/osvscanner.go
hogo6002: Oh yeah. I misunderstood the purpose of this changes. Reverted changes back. Thanks!

internal/sourceanalysis/go.go
hogo6002: Thanks! updated code

internal/sourceanalysis/go.go
```diff
@@ -25,8 +25,16 @@ func goAnalysis(r reporter.Reporter, pkgs []models.PackageVulns, source models.S
 		return
 	}
 
+	// Set GOVERSION to the Go version in go.mod.
+	var goVersion string
+	for _, pkg := range pkgs {
+		if pkg.Package.Name == "stdlib" {
+			goVersion = pkg.Package.Version
```
another-rex, line 32:
```suggestion
			goVersion = pkg.Package.Version
			break
```

another-rex (approved):
LGTM, thanks!

oliverchang (approved):

PR #823, opened 2024-02-26
Automated Updates: only append dependencies with property to original requirements

cuixq:
Maven original requirements are used to look up if the original requirement contains a property or not, so we don't need all original requirements but only the ones with properties.


internal/resolution/manifest/maven.go
```diff
@@ -24,8 +24,8 @@ const (
 )
 
 type MavenManifestSpecific struct {
-	Properties      []PropertyWithOrigin
-	OriginalImports []resolve.RequirementVersion
+	Properties           []PropertyWithOrigin
+	OriginalRequirements []resolve.RequirementVersion
```
oliverchang, line 28: Should we add a  comment or name this variable a bit better to capture the fact that these are requirements with properties only?

RequirementsWithProperties?

michaelkedar (approved):

internal/resolution/manifest/maven.go
cuixq: done!

oliverchang (approved):

internal/resolution/manifest/maven.go
```diff
@@ -139,15 +131,40 @@ func (m MavenManifestIO) Read(df lockfile.DepFile) (Manifest, error) {
...
-			Properties:      properties,
-			OriginalImports: originalImports,
+			Properties:                 properties,
+			RequirementsWithProperties: reqsWithProps,
 		},
 	}, nil
 }
 
+// ContainsProperty returns whether a string contains property placeholder.
+func ContainsProperty(str maven.String) bool {
```
another-rex, line 144: Add some examples of property placeholders in the docstring (or add a unit test for it.)

another-rex (approved):
LGTM!

PR #824, opened 2024-02-26
Automated Updates: suggest property patches to update for Maven

cuixq:
With this PR, the suggester now can generate property patches for Maven, so that the `update` subcommand does not need to handle ecosystem-specific patch information. 

@oliverchang @michaelkedar let me know how you think about this change!


internal/remediation/suggester/maven.go
```diff
@@ -2,23 +2,124 @@ package suggester
...
+		latest, err := suggestVersion(ctx, client, req, slices.Contains(opts.AvoidMajor, req.Name))
+		if err != nil {
+			return manifest.ManifestPatch{}, fmt.Errorf("suggesting latest version of %s: %w", req.Version, err)
+		}
+		if latest.Version == req.Version {
+			// No need to update
+			continue
+		}
+
+		// TODO: this should go to reporter?
```
oliverchang, line 51: +1. CC  @another-rex can you give some pointers on how to do this?

oliverchang (approved):
I assume most of this is the same as the internally reviewed version?

internal/remediation/suggester/maven.go
michaelkedar: Can this print be moved outside of the `Suggest` function? I don't think this should be printing as a side effect.

internal/remediation/suggester/maven.go
```diff
@@ -2,23 +2,124 @@ package suggester
...
+
+// suggestVersion returns the latest version based on the given Maven requirement version.
 // If there is no newer version available, req will be returned.
 // For a version range requirement,
 //   - the greatest version matching the constraint is assumed when deciding whether the
 //     update is a major update or not.
 //   - if the latest version does not satisfy the constraint, this version is returned;
 //     otherwise, the original version range requirement is returned.
-func (ms *MavenSuggester) Suggest(ctx context.Context, client resolve.Client, req resolve.RequirementVersion, opts SuggestOptions) (resolve.RequirementVersion, error) {
+func suggestVersion(ctx context.Context, client resolve.Client, req resolve.RequirementVersion, noMajorUpdates bool) (resolve.RequirementVersion, error) {
```
michaelkedar, line 122: nit: `suggestVersion` might get confusing when we add other ecosystems to this suggester submodule.

Maybe make this something like `func suggestMavenVersion()` or `func (ms *MavenSuggester) suggestVersion()`.

internal/remediation/suggester/maven.go
cuixq: yes - let me remove this print for now. We can print from the subcommand.

internal/remediation/suggester/maven.go
cuixq: done!

cuixq:
yes, most of this is the same as the internally reviewed version - I just move it from the command to suggester so that `update` subcommand is less ecosystem-specific.

another-rex (approved):

Issue #741, opened 2024-01-10
Investigate why main cli tests are so (relatively) slow


G-Rath:
The test suite has generally been pretty fast as per the nature of Go testing but now it tends to take 30-60 seconds locally with a lot of that time being taken by `github.com/google/osv-scanner/cmd/osv-scanner`.

While that's still relatively fast in the scheme of things, the next slowest package is `semantic` at ~0.200 seconds, _and_ to make it worse the `osv-scanner/cmd/osv-scanner` tests are the most e2e-y, testing the CLI entry function meaning they basically never get to benefit from caching.

I've previously looked into this and I believe it was the result of supporting local databases - in particular, I think one of tests uses the Debian ecosystem which has the latest database by something like a x2 factor; however I didn't explicitly note this anywhere.

Someone should do some digging into what's happening so we have it captured and we can then discuss hopefully how we might be able to make the test(s) faster.

G-Rath:
@billielynch this is for you to pick up

billielynch:
@G-Rath I don't seem to be able to assign myself. Do you want me to be assigned on it now or when I actually have time to look at it?

G-Rath:
@billielynch pretty sure you can only be assigned if you're a collaborator, so for now it'll just have to be unassigned

another-rex:
@billielynch Assigned you to the task now :)

billielynch:
**Confirming the location of the slowness**

Running the tests identifies three files that are taking over 0.1s: 
```
github.com/google/osv-scanner/cmd/osv-scanner   23.991s coverage: 45.4% of statements in ./...
github.com/google/osv-scanner/internal/semantic 0.232s  coverage: 36.8% of statements in ./...
github.com/google/osv-scanner/internal/sourceanalysis   0.960s  coverage: 11.5% of statements in ./...
```
**internal/semantic**
Discussion with @G-Rath indicates that this is just a computationally heavy test group that is doing a lot and fairly optimised already. Going to just pass this one for now.

**cmd/osv-scanner**
[This is by by far the slowest area](https://github.com/google/osv-scanner/blob/main/cmd/osv-scanner/main_test.go). Running each TestRun in `cmd/osv-scanner` we get times: 

- Licenses 2.03s
- LockfileWithExplicitParseAs 1.36s
- GithubActions 1.26s
- SubCommands 0.74s
- LocalDatabases 22.74s

[...]

PR #826, opened 2024-02-27
Added a switch for using cached local db in test to improve speed

billielynch:
regarding : https://github.com/google/osv-scanner/issues/741

As discussed with google team, moving to using cached local db when not running acceptance tests.

### DIFF
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -62,7 +62,7 @@ func normalizeRootDirectory(t *testing.T, str string) string {
 	return strings.ReplaceAll(str, cwd, "<rootdir>")
 }
 
-// normalizeRootDirectory attempts to replace references to the temp directory
+// normalizeTempDirectory attempts to replace references to the temp directory
 // with "<tempdir>", to ensure tests pass across different OSs
 func normalizeTempDirectory(t *testing.T, str string) string {
 	t.Helper()
@@ -508,8 +508,19 @@ func TestRun_LocalDatabases(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			testDir, cleanupTestDir := createTestDir(t)
-			defer cleanupTestDir()
+			var testDir string
+			var cleanupTestDir func()
+
+			if os.Getenv("TEST_ACCEPTANCE") != "true" {
+				testDir = filepath.Join(os.TempDir(), "osv-scanner-test-0")
+				err := os.MkdirAll(testDir, 0777)
+				if err != nil {
+					t.Fatalf("could not create test directory: %v", err)
+				}
+			} else {
+				testDir, cleanupTestDir = createTestDir(t)
+				defer cleanupTestDir()
+			}
 
 			old := tt.args
 
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -523,16 +523,13 @@ func TestRun_LocalDatabases(t *testing.T) {
 		},
 	}
 
-	var testDir string
-	var cleanupTestDir func()
-
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
 			if os.Getenv("TEST_ACCEPTANCE") == "true" {
-				testDir, cleanupTestDir = createTestDir(t)
+				testDir, cleanupTestDir := createTestDir(t)
 				defer cleanupTestDir()
 				old := tt.args
 				tt.args = []string{"", "--experimental-local-db-path", testDir}
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -528,7 +528,7 @@ func TestRun_LocalDatabases(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			if os.Getenv("TEST_ACCEPTANCE") == "true" {
+			if testutility.AcceptanceTestOn() {
 				testDir, cleanupTestDir := createTestDir(t)
 				defer cleanupTestDir()
 				old := tt.args
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -503,16 +503,19 @@ func TestRun_LocalDatabases(t *testing.T) {
 			exit: 0,
 		},
 	}
+
+	var testDir string
+	var cleanupTestDir func()
+	if os.Getenv("TEST_ACCEPTANCE") != "true" {
+		testDir = filepath.Join(os.TempDir(), "osv-scanner-test-0")
+	}
+
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			var testDir string
-			var cleanupTestDir func()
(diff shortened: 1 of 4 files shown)


cmd/osv-scanner/main_test.go
```diff
@@ -508,8 +508,19 @@ func TestRun_LocalDatabases(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			testDir, cleanupTestDir := createTestDir(t)
-			defer cleanupTestDir()
+			var testDir string
+			var cleanupTestDir func()
+
+			if os.Getenv("TEST_ACCEPTANCE") != "true" {
```
G-Rath, line 514: we should use the `testutility.AcceptanceTests` for this

cmd/osv-scanner/main_test.go
```diff
@@ -508,8 +508,19 @@ func TestRun_LocalDatabases(t *testing.T) {
...
+				testDir = filepath.Join(os.TempDir(), "osv-scanner-test-0")
```
G-Rath, line 515: couldn't we just do this outside of the test loop? that should let you reuse `createTestDir` and not need to explicitly name a directory for the normalizer.

i.e. basically do this out of the loop, then here check for `testDir === ""` and if so `testDir, cleanupTestDir = createTestDir(t); defer cleanupTestDir()`

G-Rath (changes requested):

cmd/osv-scanner/main_test.go
billielynch: can do, is there an example of the way it should be used? I didn't think it could be used like that cos I couldn't find any example in the code, might just be looking for the wrong thing though.

G-Rath: https://github.com/ackama/osv-scanner/blob/ff8aee76ad1ea6d56a2e7e8aac88d3deeee58566/internal/sourceanalysis/rust_test.go#L68-L68

billielynch: oh I thought this would just skip the test completely does it not?

billielynch: I think its just skips the tests from messing with it, but I could build something in the utility so that it is only accessing TEST_ACCEPTANCE in one place?

cmd/osv-scanner/main_test.go
```diff
@@ -503,13 +503,27 @@ func TestRun_LocalDatabases(t *testing.T) {
 			exit: 0,
 		},
 	}
+
+	var testDir string
+	var cleanupTestDir func()
+	if os.Getenv("TEST_ACCEPTANCE") != "true" {
```
G-Rath, line 509: we should be able to use `cleanupTestDir`, and we also need to call `defer cleanupTestDir()` to ensure it gets cleaned up

cmd/osv-scanner/main_test.go
```diff
@@ -503,13 +503,27 @@ func TestRun_LocalDatabases(t *testing.T) {
...
+
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			testDir, cleanupTestDir := createTestDir(t)
-			defer cleanupTestDir()
+			if testDir != "" {
+				err := os.MkdirAll(testDir, 0777)
```
G-Rath, line 519: this can go away once we switch to using `cleanupTestDir` at the top level

G-Rath (changes requested):

cmd/osv-scanner/main_test.go
G-Rath: ah yup you're totally right - let's introduce a new `IsRunningAcceptanceTests` helper which we can use in the existing helper plus here

cmd/osv-scanner/main_test.go
```diff
@@ -62,7 +62,25 @@ func normalizeRootDirectory(t *testing.T, str string) string {
 	return strings.ReplaceAll(str, cwd, "<rootdir>")
 }
 
-// normalizeRootDirectory attempts to replace references to the temp directory
+// normalizeUserCacheDirectory attempts to replace references to the current working
+// directory with "<tempdir>", in order to reduce the noise of the cmp diff
```
G-Rath, line 66:
```suggestion
// normalizeUserCacheDirectory attempts to replace references to the users cache
// directory with "<tempdir>", in order to reduce the noise of the cmp diff
```

internal/testutility/utility.go
```diff
@@ -34,11 +34,17 @@ func Skip(t *testing.T, args ...any) {
 	snaps.Skip(t, args...)
 }
 
+// Access to environment variable that toggles acceptance testing execution paths
+// Acceptance testing is "On" only when var set to "true"
+func AcceptanceTestOn() bool {
```
G-Rath, line 39: nit: I feel like it might be nicer to name this something like `IsAcceptanceTesting`

This could just be me though so I'm more mentioning this to see if anyone else feels similar but if not it can stay - I think I'm just not use to "On" being used as the main expression of a function being a conditional 🤷

internal/local/check.go
```diff
@@ -73,10 +73,10 @@ func setupLocalDBDirectory(localDBPath string) (string, error) {
 		}
 	}
 
-	err = os.MkdirAll(path.Join(localDBPath, "osv-scanner"), 0750)
-
+	altPath := path.Join(localDBPath, "osv-scanner")
+	err = os.MkdirAll(altPath, 0750)
 	if err == nil {
-		return path.Join(localDBPath, "osv-scanner"), nil
+		return altPath, nil
```
G-Rath, line 79: nit: I'd prefer this kind of refactoring happen in another PR (if it happens at all) but I won't block over it

cmd/osv-scanner/main_test.go
```diff
@@ -503,18 +522,19 @@ func TestRun_LocalDatabases(t *testing.T) {
...
 			t.Parallel()
 
-			testDir, cleanupTestDir := createTestDir(t)
-			defer cleanupTestDir()
-
-			old := tt.args
-
-			tt.args = []string{"", "--experimental-local-db-path", testDir}
-			tt.args = append(tt.args, old[1:]...)
+			if os.Getenv("TEST_ACCEPTANCE") == "true" {
```
G-Rath, line 531: we should be using our new helper here:

```suggestion
			if testutility.AcceptanceTestOn() {
```

cmd/osv-scanner/main_test.go
G-Rath: note, I think it's fine now but in future we might want to switch to explicitly referring to this as `cachedir` and having a conditional normalize when the `--experimental-local-db` flag is present to replace `cachedir` with `tempdir`, but that seems pretty overkill to me for now

G-Rath (changes requested):

cmd/osv-scanner/main_test.go
billielynch: yeah this is my thoughts too its not pretty but itll do for now

internal/testutility/utility.go
billielynch: I have no strong feelings about this and couldnt see a pattern in the existing things, figured your comment if you cared and had a better suggestionm is is fine for me, will do.

G-Rath (approved):
awesome stuff!

another-rex (approved):
LGTM, thanks!

PR #828, opened 2024-02-27
perf: optimize string formatting and update linting

G-Rath:
I'm kind of on the fence about this - I actually ran the [`perfsprint` benchmarks locally](https://github.com/catenacyber/perfsprint) and these changes are indeed faster, but they're really in the hotpath; at the same time though we're not importing anything new for this and it could one day catch a use on the hotpath so I'm not sure about disabling it either.

Happy to hear what others think

### DIFF
--- a/.github/workflows/lint-action/action.yml
+++ b/.github/workflows/lint-action/action.yml
@@ -22,6 +22,6 @@ runs:
       uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v4.0.0
       with:
         # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-        version: v1.55.2
+        version: v1.56.2
         # https://github.com/golangci/golangci-lint-action/issues/135
         skip-pkg-cache: true
--- a/cmd/osv-scanner/fix/main.go
+++ b/cmd/osv-scanner/fix/main.go
@@ -1,6 +1,7 @@
 package fix
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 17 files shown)


spencerschrock:
> these changes are indeed faster

If you still have the benchmarks around, did it also result in fewer allocs/op? I've seen some comments about avoiding `fmt.Sprintf` when possible, or preferring `errors.New`, so it's nice to see some benchmarks backing it up.

I may consider enabling `perfsprint` on Scorecard, although we have quite a few more violations to fix

G-Rath:
I'll check but tbh it was just a case of checking out the repo and [running the benchmarks](https://github.com/catenacyber/perfsprint/blob/main/analyzer/testdata/src/p/p.go) -

spencerschrock:
> I'll check but tbh it was just a case of checking out the repo and [running the benchmarks]

Ah I thought you were talking about benchmarks for the proposed changes in `osv-scanner`. Thanks for the link and benchmarks.

another-rex:
Hmm I'm not sure about this, it looks like it's splitting up one standard formatting method into a bunch of different functions, but that might just be personal preference. 

The performance improvements are nice, but as you said we're not using it anywhere it'll have big (or any) impact in osv-scanner right now, and the only time it'll have an impact, we'll find it when profiling/benchmarking anyway (and if we can't find it, it's probably not that big of an impact), and can just replace that specific instance. 

So it's really whether we like the look of Sprintf compared to the alternatives.

G-Rath:
@another-rex you've pretty much mirrored my feelings, except:

>  the only time it'll have an impact, we'll find it when profiling/benchmarking anyway

Will you though? afaik there's no automation around that in place unless you've got some internal stuff, so the only way you'd know is if _you're_ using the scanner in a way to hit the performance area and recognize it as such, or if someone is confident enough to report a performance problem vs just saying "well I guess it's that slow"

(I know @spencerschrock you've done some benchmarking before, but I assume that was as part of initial mass adoption and/or in response to noticed areas of slowness, but not thing you're running all the time - let me know if I'm wrong though!)

I don't know if this is worse _enough_ to not do given there is a lint rule that tells you exactly what to do (and that looks like it has an autofix, but I can't trigger it in `golangci-lint`)...?

another-rex:
> Will you though? afaik there's no automation around that in place unless you've got some internal stuff, so the only way you'd know is if you're using the scanner in a way to hit the performance area and recognize it as such, or if someone is confident enough to report a performance problem vs just saying "well I guess it's that slow"

That's a good point, there's no regular benchmarking happening at the moment.

spencerschrock:
> (I know @spencerschrock you've done some benchmarking before, but I assume that was as part of initial mass adoption and/or in response to noticed areas of slowness, but not thing you're running all the time - let me know if I'm wrong though!)

we profile occasionally. On average, our use of osv-scanner is consistently our bottleneck, but on the osv.dev API side of things.

G-Rath:
> I tried --fix with golangci-lint and just assumed it didn't have one

Have opened https://github.com/golangci/golangci-lint/issues/4453

another-rex (approved):

PR #832, opened 2024-02-28
test: make `createTestDir` a general test utility

G-Rath:
We've already got this duplicated in two places, and I have need for it in an upcoming PR and @robramsaynz should be using it in #797 so let's move this into `testutility`

### DIFF
--- a/internal/testutility/utility.go
+++ b/internal/testutility/utility.go
@@ -59,8 +59,9 @@ func ValueIfOnWindows(win, or string) string {
 }
 
 // CreateTestDir makes a temporary directory for use in testing that involves
-// writing and reading files from disk - make sure to call the cleanup function!
-func CreateTestDir(t *testing.T) (string, func()) {
+// writing and reading files from disk, which is automatically cleaned up
+// when testing finishes
+func CreateTestDir(t *testing.T) string {
 	t.Helper()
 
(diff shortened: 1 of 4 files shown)


another-rex (approved):

internal/testutility/utility.go
```diff
@@ -57,3 +57,18 @@ func ValueIfOnWindows(win, or string) string {
 
 	return or
 }
+
+// CreateTestDir makes a temporary directory for use in testing that involves
+// writing and reading files from disk - make sure to call the cleanup function!
+func CreateTestDir(t *testing.T) (string, func()) {
+	t.Helper()
```
spencerschrock, line 64: `t.Cleanup` would be helpful here.
https://pkg.go.dev/testing#T.Cleanup

no need to make caller remember to call cleanup function

G-Rath: oh nice - and yeah I just realised `t.TempDir` is a thing though sadly it doesn't support a prefix which I personally like to easily identify and manually clean up in the case of a failure or debugging.

Oh well, still probably best to just switch to using that

spencerschrock: Sounds reasonable. At least `t.Cleanup` sounds useful in this case.

another-rex (approved):
Thanks!

PR #836, opened 2024-03-04
Replace stereoscope with using go-containerregistry directly

another-rex:
Replace stereoscope with using the lower level go-containerregistry library directly. 

Stereoscope brings in a large list of dependencies to support functions that is not necessary for osv-scanner. This also gives us more low level control of the extraction and scanning. 

It's possible to switch to a new fs.FS lockfile interface without too much work.


internal/image/image.go
```diff
@@ -73,6 +69,172 @@ func ScanImage(r reporter.Reporter, imagePath string) (ScanResults, error) {
 	return scannedLockfiles, err
 }
 
+type Image struct {
+	flattenedFileMaps []FileMap
```
oliverchang, line 73: nit: how about `flattenedLayers`? It's not immediately obvious this is referring to layers from the name otherwise.

internal/image/image.go
```diff
@@ -73,6 +69,172 @@ func ScanImage(r reporter.Reporter, imagePath string) (ScanResults, error) {
...
+	innerImage        *v1.Image
+	tempDir           string
```
oliverchang, line 75: nit: how about extractDir to be more descriptive about what this is?

oliverchang (commented):
nice!

cmd/osv-scanner/main.go
```diff
@@ -118,5 +118,7 @@ func insertDefaultCommand(args []string, commands []*cli.Command, defaultCommand
 }
 
 func main() {
-	os.Exit(run(os.Args, os.Stdout, os.Stderr))
+	exitCode := run(os.Args, os.Stdout, os.Stderr)
```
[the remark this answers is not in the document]
another-rex: It makes it easy to wrap code around the entire program (e.g. for a CPU profiler)

internal/image/image.go
oliverchang: And also per https://google.github.io/styleguide/go/decisions#repetitive-with-type

internal/image/filemap.go
```diff
@@ -0,0 +1,46 @@
...
+
+type FileNode struct {
+	fileType         FileType
+	isWhiteout       bool
+	absoluteDiskPath string
+	virtualPath      string
+}
+
+type FileMap struct {
+	fileNodeTrie *trie.PathTrie
```
oliverchang, line 25: Also, how much memory do we expect to save on average with using a trie structure here?

another-rex: I can check, but I doubt it's a significant amount, this is mainly used to allow us to iterate through directories.

oliverchang: Can you explain a bit more what you mean by "iterate through directories"?

internal/image/image.go
```diff
@@ -73,6 +72,177 @@ func ScanImage(r reporter.Reporter, imagePath string) (ScanResults, error) {
...
+			header, err := tarReader.Next()
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			if err != nil {
+				return Image{}, fmt.Errorf("reading tar: %w", err)
+			}
+			// Some tools prepend everything with "./", so if we don't Clean the
+			// name, we may have duplicate entries, which angers tar-split.
+			cleanedFileName := filepath.Clean(header.Name)
```
oliverchang, line 143: to avoid zip slip, I think we need to make sure the cleaned name does not begin with any '../' ?

internal/image/filemap.go
another-rex: E.g. if we want to scan a node_modules folder, we need to iterate through the contents of the directory. 

It's mainly used for when we want to refactor and implement the fs.FS interface.

oliverchang: got it! this trie dependency is fine for now, but if memory usage differences is negligible I wonder if this can be just simplified to a sorted array of paths that we can binary search on to enable fast lookups and directory listings.

internal/image/image.go
another-rex: Ah yep, added a check, I thought `filepath.Clean` solved that, but it looks like it only resolves internal `../`.

internal/image/image.go
```diff
@@ -73,6 +72,183 @@ func ScanImage(r reporter.Reporter, imagePath string) (ScanResults, error) {
...
+			// Prevent "Zip Slip"
+			if strings.HasPrefix(cleanedFileName, "..") {
```
oliverchang, line 145: "..foo" is a valid filename, so I think we need to add a filepath separator after ".." here.

internal/image/filemap.go
another-rex: Looks like most of the memory usage is from extracting the tar.

internal/image/image.go
```diff
@@ -73,6 +69,172 @@
...
+
+		layerReader, err := layers[i].Uncompressed()
+		if err != nil {
+			return Image{}, err
+		}
+		defer layerReader.Close()
+		tarReader := tar.NewReader(layerReader)
+
+		for {
+			header, err := tarReader.Next()
```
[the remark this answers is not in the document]
another-rex: Fixed with a `filepath.Clean`, with checking if prefix starts with `../`

oliverchang (approved):

internal/sourceanalysis/__snapshots__/integration_test.snap
```diff
@@ -124,7 +124,7 @@
           "position": {
             "filename": "\u003cAny value\u003e",
             "offset": -1,
-            "line": 840,
+            "line": 839,
```
hogo6002, line 127: Do we want to set the line number to a fixed value? so we don't need to update it again when go version changes.

hogo6002 (approved):

internal/sourceanalysis/__snapshots__/integration_test.snap
another-rex: I think it's alright for now, this allows us to check whether the line result in govulncheck is returning correctly. And we shouldn't be updating go that often anyway, so updating it when we do should be fine.

PR #839, opened 2024-03-05
Rename relaxer and suggester

cuixq:
This PR addresses https://github.com/google/osv-scanner/pull/817#issuecomment-1977729821.

Package `relaxer` is renamed to `relax` and package `suggester` is renamed to `suggest`.
The getter and options are also renamed to avoid duplicates in their names.


internal/remediation/relax.go
michaelkedar, line 1: Thoughts on renaming this file to `relock_relax.go` to make it more distinct from `relax/...`?

michaelkedar (approved):
LGTM

G-Rath (approved):

internal/remediation/relax.go
cuixq: SGTM

internal/remediation/relax/relax.go
```diff
@@ -23,7 +23,7 @@ type RequirementRelaxer interface {
 	Relax(ctx context.Context, cl resolve.Client, req resolve.RequirementVersion, allowMajor bool) (resolve.RequirementVersion, bool)
 }
 
-func GetRelaxer(ecosystem resolve.System) (RequirementRelaxer, error) {
+func Get(ecosystem resolve.System) (RequirementRelaxer, error) {
```
another-rex, line 26: Now that it's called relax / suggest, we probably can keep it as GetRelaxer:
```
relaxer := relax.GetRelaxer(eco)
// vs
relaxer := relax.Get(eco)
```

another-rex: But I don't feel strongly about it either way, so up to you, happy to merge it as is.

another-rex (approved):

internal/remediation/relax/relax.go
oliverchang: +1. I think `.Get()` would make sense when the package name is what is being gotten.

cuixq: done!

another-rex (approved):

PR #850, opened 2024-03-08
fix: allow users to override GOVERSION

hogo6002:
fixes https://github.com/google/osv-scanner/issues/848. 
Allow users to specify a `GOVERSION` other than the one in `go.mod` .

Usage: in toml, `GoVersionOverride = "1.21.0"`

### DIFF
--- a/internal/sourceanalysis/go.go
+++ b/internal/sourceanalysis/go.go
@@ -17,8 +17,6 @@ import (
 	"golang.org/x/vuln/scan"
 )
 
-const goVersionEnv = "GOVERSION"
-
 func goAnalysis(r reporter.Reporter, pkgs []models.PackageVulns, source models.SourceInfo) {
 	cmd := exec.Command("go", "version")
 	_, err := cmd.Output()
@@ -140,12 +138,7 @@ func runGovulncheck(moddir string, vulns []models.Vulnerability, goVersion strin
 	var b bytes.Buffer
 	cmd.Stdout = &b
 
-	// Allow GOVERSION to be overridden
-	if val, ok := os.LookupEnv(goVersionEnv); ok {
-		goVersion = val
-	}
-
-	cmd.Env = append(os.Environ(), fmt.Sprintf("%s=%s", goVersionEnv, goVersion))
+	cmd.Env = append(os.Environ(), fmt.Sprintf("GOVERSION=%s", goVersion))
 	if err := cmd.Start(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


internal/sourceanalysis/go.go
```diff
@@ -17,6 +17,8 @@ import (
 	"golang.org/x/vuln/scan"
 )
 
+const goVersionEnv = "GOVERSION"
```
G-Rath, line 20: is this a general env variable i.e. used by Go? or is it worth prefixing this with something like `OSV_SCANNER` to avoid conflict?

another-rex: Yeah, let's change this to OSV_SCANNER_GOVERSION for now, I don't think GOVERSION is actually interpreted by the go compiler in any way (e.g. if you set GOVERSION, it won't change the version that is actually being used, I think that's governed by GOROOT variable), but better to be safe here. 

This will just be a temporary fix, as we probably want this as something we can set in `osv-scanner.toml` eventually.

hogo6002: Not really a general env variable. It's specifically used by govulncheck, and it will be parsed from "go env GOVERSION" if not defined. https://github.com/golang/vuln/blob/563994f0852c21fbdb2424841ebd8de389ff47ad/internal/scan/run.go#L78

hogo6002: It can be changed to `OSV_SCANNER_GOVERSION` for safe when users want to specify a go version, but we still need to pass the `GOVERSION` to govulncheck.

sudo-bmitch (approved):
This works for me.

pkg/osvscanner/osvscanner.go
```diff
@@ -969,3 +971,23 @@ func makeLicensesRequests(packages []scannedPackage) ([][]models.License, error)
...
 }
+
+func overrideGoVersion(r reporter.Reporter, packages []scannedPackage, configManager *config.ConfigManager) {
+	// Overrides Go version
+
+	if len(packages) == 0 {
+		return
+	}
+
+	configToUse := configManager.Get(r, packages[0].Source.Path)
```
another-rex, line 982: Let's have this only get the config when we find stdlib. The config that's being loaded is already being cached, we can just move this Get call into the for loop (line 989)

another-rex (commented):
Thanks! Can you also add a snapshot test for this? (just a folder under cmd/osv-scanner/fixtures filled with a go.mod file and a osv-scanner config file should be enough)

another-rex (approved):
LGTM, thanks!

PR #851, opened 2024-03-09
feat: support scanning `node_modules` generated by NPM in container images

G-Rath:
The actual support was easy because this is really just the NPM extractor applied to a `package-lock.json` with a different name, but this also includes an initial approach at handling management of image fixtures without requiring the actual tars (which can be quite sizable) to be committed to the codebase, thus avoiding bloat.

It's actually turned out pretty well - the main downside is it's not easy to support automatically building if/when the image tests are run specifically but I don't think that's a higher priority vs than just having _something_ to let us avoid committing tars.

~Hopefully it'll work in CI and on Windows + macOS 🤞~

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -52,7 +52,22 @@ jobs:
           check-latest: true
       - name: Run lint action
         uses: ./.github/workflows/lint-action
+  prepare_test_image_fixtures:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          persist-credentials: false
+      - run: scripts/build_test_images.sh
+      - uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
+        with:
+          name: image-fixtures-${{ github.run_number }}-${{ github.run_attempt }}
+          path: internal/image/fixtures/*.tar
+          retention-days: 1
   tests:
+    needs:
+      - prepare_test_image_fixtures
     name: Run unit tests
     strategy:
       fail-fast: false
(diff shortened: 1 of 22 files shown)


another-rex (approved):
LGTM, thanks!

G-Rath:
I just realised I think this might not work on macOS and Windows for the same reason they didn't in CI - those platforms will want OS-specific versions of the images.

It might "just work" on macOS (it didn't in CI because the Docker license actually means GitHub Actions can't include it by default) in which case it's probably good enough to land for now as I assume no one on the core team is using native Windows to develop (personally I'm using WSL), but this'll most likely require follow up - I'm not actually that familiar with docker beyond the basics but I suspect it might be possible to have it build e.g. "linux only" containers regardless of platform (like how you can build other os binaries for Go), which I'll look into next.

Alternatively maybe we should mark these as acceptance tests for now?

another-rex:
Hmm I'm not clear on what the issue with windows docker containers are. I don't have a windows machine to test this at the moment, but wouldn't it already be building Linux containers by default?

G-Rath:
You can switch it to build Linux containers, but I don't know if that's the default - GHA seems to default to Windows containers (though that makes more sense); I think the Windows installer for humans actually asks you to pick - but yes I think you're right that for local development people should be able to use linux containers and get the same result.

another-rex:
Ah I see, I would have guessed that the docker file starting with a Linux base image would change docker to use Linux containers, but maybe that requires installing a completely different runtime, so that's why you choose on install.

Yeah let's mark these as acceptance tests for now.

G-Rath:
> maybe that requires installing a completely different runtime, so that's why you choose on install.

It might have changed, but when I first used Docker on my Windows machine the Windows and Linux containers were completely different - as in you had to active switch between the two, couldn't have both at the same time, and if it didn't wipe your alternative containers it at least cold storaged them somewhere on disk until you switched back...

another-rex:
Looking at this a bit more, it could be good to add a check to rebuild the docker images when the dockerfile changes. Something like checking the modified date of the dockerfile is after the `.tar` file might be enough.

G-Rath:
Yup I've got that down as a follow up since there's a few things in that area we can explore but they should be less critical than unblocking general crafting of tests for images etc

PR #853, opened 2024-03-11
GR: Add npm lockfile read tests

michaelkedar:
Added tests for reading package-lock.json into deps.dev graphs.
I set up the test data by publishing packages to a local private registry, so the lockfiles/graphs aren't massive.

I'm trying to work out a way to mock the npm registry before I do the lockfile writing tests.


cuixq:
I mocked the Maven registry [here](https://github.com/google/osv-scanner/blob/main/internal/resolution/datasource/maven_test.go) - not sure if this helps?

cuixq (approved):

internal/resolution/lockfile/npm_test.go
```diff
@@ -0,0 +1,143 @@
...
+	// This lockfile was generated using a private registry with https://verdaccio.org/
+	// Mock packages were published to it and installed with npm.
+	df, err := lf.OpenLocalDepFile("./fixtures/npm_v2/package-lock.json")
+	if err != nil {
+		t.Fatalf("failed to open file: %v", err)
+	}
+	defer df.Close()
+
+	npmIO := lockfile.NpmLockfileIO{}
+	got, err := npmIO.Read(df)
```
another-rex, line 36: Is this testing if it's reading the graph correctly, or testing all the AddNode/AddEdge functions, or both?

It'll be better for maintainability I think to separate out these two, and use snapshot to test if we read correctly, and a separate snapshot to test whether the Add functions are working correctly.

michaelkedar: This is just testing the graph is being read correctly. The Graph struct/`AddNode/AddEdge` are part of [deps.dev](https://github.com/google/deps.dev/blob/38ffc8dd1d92104fb69d158efdd8cb75778e8b6a/util/resolve/graph.go#L79) - I'm just using it to construct the expected graph in-code.

I have mixed feelings about using a snapshot instead:
- Snapshots need string comparisons, and I'm not sure how complete the string representation of a graph is (it might be comprehensive - I just need to check on it)
- I defined the expected graph based on my understanding of npm dependency resolution (and deps.dev's representations). In my mind, using a snapshot for this feels less... intentional about the expected output. (Though I will admit I'm on the fence about this.)

internal/resolution/lockfile/npm_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"testing"
+
+	"deps.dev/util/resolve"
+	"deps.dev/util/resolve/dep"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/osv-scanner/internal/resolution/lockfile"
+	lf "github.com/google/osv-scanner/pkg/lockfile"
+)
+
+func npmVK(name, version string) resolve.VersionKey {
```
G-Rath, line 13: nit: this should ideally use `t.Helper()` methinks

internal/resolution/lockfile/npm_test.go
another-rex: That makes sense, other concerns I have is how maintainable this is, though I'm not sure what can be done about it. The diff output is not great (I commented out one edge, which seems to completely change the order so the diff gets confused). But that's probably something for the future to figure out, happy with this as is!

another-rex: Eventually I feel like we'll have DSL just to make testing the graphs easier haha.

another-rex (approved):

internal/resolution/lockfile/npm_test.go
michaelkedar: What does `t.Helper()` actually do?
The function just makes a struct so I didn't really want to pass in the `*testing.T` if I didn't have to.

G-Rath: > Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped

While I agree it's a bit annoying, I'd say "that's Go for you 🙃" - to me _because_ this function "just makes a struct" that (unfortunately) means it really should be marked with `t.Helper()` because you want it to be a setup detail of the test rather than the test itself.

There's also the consistency side - we're using it pretty much everywhere else already.

michaelkedar: Ack.
I'll add it here (and also to the similar function in the npm manifest tests)

PR #854, opened 2024-03-11
Find and save the distro version when extracting from debian and alpine

another-rex:
Previously, we ignored the distro version when querying for vulnerabilities. This adds in parsing for distro version. 

To enable this for alpine, which doesn't encode their release version in their `apk-installed` file, I also implemented `Open()` function for image `DepFile`, and for alpine, open the `/etc/alpine-version` file to read the alpine version.

This hid the previous vulnerabilities that were showing up for alpine 3.10, since alpine 3.10 is no longer supported, the advisories weren't being published for those packages, even though those vulnerabilities probably do apply.

---

It is quite difficult to find packages that are actually vulnerable and haven't been updated in alpine, and older packages are removed from the repository, so we can't really test the positive case where a vulnerability is found.


G-Rath:
> It is quite difficult to find packages that are actually vulnerable and haven't been updated in alpine, and older packages are removed from the repository, so we can't really test the positive case where a vulnerability is found.

I've not looked at the changes yet, but I don't follow this - why can't we just craft a alpine lockfile that lists a vulnerable package? even if we can't get alpine to generate one, I would think we know enough about the format to be able to hack a fake one together which'd be better than having no test at all...

another-rex:
Yeah we definitely can, I'll add a crafted test image that shows some vulns. I'm mainly referring to the difficulty of looking for a natural example to make sure we aren't missing anything.

cmd/osv-scanner/__snapshots__/main_test.snap
```diff
@@ -1707,18 +1707,18 @@ No issues found
...
-| https://osv.dev/CVE-2018-25032 | 7.5  | Alpine    | zlib    | 1.2.11-r1 | ./fixtures/oci-image/alpine-tester.tar:/lib/apk/db/installed |
-| https://osv.dev/CVE-2022-37434 | 9.8  | Alpine    | zlib    | 1.2.11-r1 | ./fixtures/oci-image/alpine-tester.tar:/lib/apk/db/installed |
-+--------------------------------+------+-----------+---------+-----------+--------------------------------------------------------------+
+[TestRun_OCIImage/Alpine_3.10_image_tar_with_3.18_version_file - 1]
+Scanning image ../../internal/image/fixtures/test-alpine.tar
++--------------------------------+------+--------------+---------+-----------+---------------------------------------------------------------------+
+| OSV URL                        | CVSS | ECOSYSTEM    | PACKAGE | VERSION   | SOURCE                                                              |
++--------------------------------+------+--------------+---------+-----------+---------------------------------------------------------------------+
+| https://osv.dev/CVE-2018-25032 | 7.5  | Alpine:v3.18 | zlib    | 1.2.11-r1 | ../../internal/image/fixtures/test-alpine.tar:/lib/apk/db/installed |
```
hogo6002, line 1715: Do we want to put the version into ecosystem?

another-rex: We'll ideally want the alpine subversion on all the scanning outputs, but we currently don't get that from the sbom extraction.

oliverchang (approved):
nice!

hogo6002 (approved):

PR #859, opened 2024-03-13
improve error messages in Maven registry client

cuixq:
Follow-up on https://github.com/google/osv-scanner/pull/857

### DIFF
--- a/internal/resolution/datasource/maven_registry.go
+++ b/internal/resolution/datasource/maven_registry.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"deps.dev/util/maven"
+	"github.com/google/osv-scanner/pkg/osvscanner"
 )
 
 const MavenCentral = "https://repo.maven.apache.org/maven2"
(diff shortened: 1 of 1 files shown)


internal/resolution/datasource/maven_registry.go
```diff
@@ -27,17 +28,17 @@ func NewMavenRegistryAPIClient() (*MavenRegistryAPIClient, error) {
...
-		return maven.Project{}, err
+		return maven.Project{}, fmt.Errorf("failed to make new request: %w", err)
 	}
 
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return maven.Project{}, err
+		return maven.Project{}, fmt.Errorf("%w: Maven registry query failed: %w", osvscanner.ErrAPIFailed, err)
 	}
 	defer resp.Body.Close()
```
another-rex, line 44: I think we also need to add ErrAPIFailed to the status code check

another-rex (commented):
Thanks!

another-rex (approved):

PR #860, opened 2024-03-15
Fix/update retry logic in OSV

### LINKED ISSUE #856: upstream request timeout
We run osv-scanner as part of our CI, and have been seeing lots of the following error for the last few days

> scan failed server response error: {"code":504,"message":"upstream request timeout"}

Probably this is related to https://github.com/google/osv.dev/issues/2039

We are scanning a  `Cargo.lock` which is size 202485 and 758 entries in it, so it is a bit large.

AFAICS there is no way to configure the timeout.  I am wondering whether that might be a useful addition, as we wou'd prefer it take longer rather than cause a failure in a large complicated CI workflow.

another-rex:
Fixes #856

Fixes a couple of things with our OSV API retry logic:

- Now also retry non 200 responses, in addition to retrying on network errors. (is there a need to only retry for 500 >= responses? Current logic retrying all non 200)
- Fix batchquery code so that retries are not sending empty data.
- Increase maxRetryAttempts from 3 to 4
- Implement exponential backoff (actually polynomial/quadratic backoff)
- Correctly close the body after reading error message. 
- Implement retries for determineversion API call


G-Rath:
>  (is there a need to only retry for 500 >= responses? Current logic retrying all non 200)

You should not be re-trying 4xx errors as they explicitly mean "client error" and I'm pretty sure they _all_ mean "do not retry without adjustment" but that might not technically be true.

I'd say if there's ever a case the API expects to return a 4xx that should be fine to retry, that's a mistake

(_dashes off to double check his 4xx error codes to make sure he's actually correct)

another-rex:
yeah that's true, but I don't think it's possible to call the osv.go functions with values that could cause a 400 though (other than maybe 429 because you sent too many queries), so I'm not sure if it's worth putting in extra logic there.

G-Rath:
"stick an issue in the backlog" 🤷

cuixq (approved):
Sorry for the late review!

pkg/osv/osv.go
```diff
@@ -319,44 +314,56 @@ func HydrateWithClient(resp *BatchedResponse, client *http.Client) (*HydratedBat
 	return &hydrated, nil
 }
 
+// makeRetryRequest will return an error on both network errors, and if the response is not 200
 func makeRetryRequest(action func() (*http.Response, error)) (*http.Response, error) {
 	var resp *http.Response
 	var err error
 	for i := 0; i < maxRetryAttempts; i++ {
+		time.Sleep(time.Duration(i*i) * time.Second)
```
oliverchang, line 322: should we add some random jitter to this?

oliverchang: nit: might also make a little more logical sense to have the sleep after the actual action. particularly if we add jitter.

pkg/osv/osv.go
```diff
@@ -319,44 +314,56 @@ func HydrateWithClient(resp *BatchedResponse, client *http.Client) (*HydratedBat
...
+		req, err := http.NewRequest(http.MethodPost, DetermineVersionEndpoint, requestBuf)
+		if err != nil {
+			return nil, err
+		}
+		req.Header.Set("Content-Type", "application/json")
+		if RequestUserAgent != "" {
+			req.Header.Set("User-Agent", RequestUserAgent)
+		}
+
+		client := http.DefaultClient
```
oliverchang, line 362: do we need this variable?

oliverchang (approved):
LGTM with some minor comments.

pkg/osv/osv.go
another-rex: added some random jitter. I think it makes more sense at the start than the end, as this avoids having to add a if statement to check if this is the last loop, and avoids having to add `i+1` as well.

PR #866, opened 2024-03-18
test: change how snapshot matchers are called and update example name for consistency

G-Rath:
I was very confused on why _most_ snapshots were all going into `main_test.snap` except for a couple, and it turns out the filename is determined by the call stack (literally `go-snaps` looks for the first frame in a `_test.go` file); I couldn't think of a good way `go-snaps` should enable working around this so instead I decided to refactor our helper a bit.

I think its useful to have all the snaps in their relative files like this vs saving a couple of lines especially given the size and that we're already generating a snapshot for each test.

I've also made a couple of other minor tweaks for consistency while I was at it.


G-Rath:
cc @billielynch @robramsaynz

cuixq (approved):

robramsaynz (approved):

another-rex (approved):

PR #867, opened 2024-03-18
test: regenerate snapshots

G-Rath:
It looks like we've got some old snapshots that weren't cleaned up by `go-snaps` - this reduces the diff of #866


another-rex (approved):
Ah was wondering why there was a lot more deletions in that PR.

PR #869, opened 2024-03-18
Follow up PR851 mark acceptance on image tests

billielynch:
https://github.com/google/osv-scanner/pull/851#issuecomment-1987566176

@G-Rath this the vibe?

Marked the tests in the file edited in previous PR as and acceptance. 

Flagged the additional image build script added in that PR as only called during acceptance testing too (optional can remove)

### DIFF
--- a/internal/image/image_test.go
+++ b/internal/image/image_test.go
@@ -11,6 +11,7 @@ import (
 )
 
 func TestScanImage(t *testing.T) {
+	testutility.SkipIfNotAcceptanceTesting(t, "Not consistent on MacOS/Windows")
 	t.Parallel()
 
 	type args struct {
--- a/internal/image/image_test.go
+++ b/internal/image/image_test.go
@@ -11,8 +11,8 @@ import (
 )
 
 func TestScanImage(t *testing.T) {
-	testutility.SkipIfNotAcceptanceTesting(t, "Not consistent on MacOS/Windows")
 	t.Parallel()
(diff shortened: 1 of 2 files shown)


internal/image/image_test.go
```diff
@@ -11,6 +11,7 @@ import (
 )
 
 func TestScanImage(t *testing.T) {
+	testutility.SkipIfNotAcceptanceTesting(t, "Not consistent on MacOS/Windows")
```
billielynch, line 14: @robramsaynz this is the way to mark tests as run only during acceptance testing. 

It uses [this function](https://github.com/google/osv-scanner/blob/3c2e2ce9cafd6d2e330daa408fee479f6fda2aa0/internal/testutility/utility.go#L45 ) that sets a test property to skip. 

There is also [this function](https://github.com/google/osv-scanner/blob/3c2e2ce9cafd6d2e330daa408fee479f6fda2aa0/internal/testutility/utility.go#L39) for if we just need s switch for some code block rather than to skip the test completely. Which can bee seen used [here](https://github.com/google/osv-scanner/blob/3c2e2ce9cafd6d2e330daa408fee479f6fda2aa0/cmd/osv-scanner/main_test.go#L524) where we create individual local database directories for each test in one case and not in the other.

All of the flow is dependent on the envvar TEST_ACCEPTANCE=true and using the second function as the access point to the envvar.

robramsaynz: Thanks for the info @billielynch 🙏

G-Rath (approved):

another-rex (approved):

PR #872, opened 2024-03-19
ci: harden workflow permissions

G-Rath:
This hardens the workflow permissions so that jobs only have the permissions they need - it's unclear to me from the docs if scorecard currently understands this is the most secure approach (as its docs to me imply it it thinks `permissions: read-all` is best), but that's a scorecard problem.

In doing this, we do harden the permissions of a few jobs that were getting permissions they don't actually need.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -23,11 +23,14 @@ on:
   merge_group:
     branches: [main]
 
-permissions:
-  contents: read # to fetch code (actions/checkout)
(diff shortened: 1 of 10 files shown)


cuixq (commented):
LGTM - wait for Rex to take a look as well.

robramsaynz (approved):

another-rex (approved):

G-Rath:
I don't know for sure this is was entirely because of this change, but we have gone from about a 7.6 to an 8 🎉 

image

PR #873, opened 2024-03-19
Improve code coverage of submodule parsing

robramsaynz:
This is a PR to improve coverage for submodule parsing that's already in the system. Specifically [`func getSubmodules`](https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/osvscanner.go#L507) in `pkg/osvscanner/osvscanner.go`.

This is already covered by a fixture based test in [`pkg/osvscanner/osvscanner_internal_test.go`](https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/osvscanner_internal_test.go#L63) but while have execution inside `getSubmodules`, it doesn't actually cover the working parts of `getSubmodules`.

You can see coverage info by running  `./scripts/generate_coverage_report.sh` and looking in ./coverage.html#file133.

There are a number of states a submodule can be in:

1. a submodule is setup and checked out
2. a submodule is setup and but dir is empty
3. a submodule is setup and and checked out but dir has local changes
4. a submodule is setup and but dir is set to a different commit

I'm planning to test 1 and 2. I think 3 and 4 are nice to haves, and I plan to only tackle them if they're easy once 1 and 2 are done. Possibly submodules have more possible states, but these seem reasonable for coverage.

I see two implementation options: 

- re-use and modify the `func Test_scanGit` in `pkg/osvscanner/osvscanner_internal_test.go`
  but with the fixture including a submodule
- re-use and modify the functions that create git repositories from scratch in
  <https://github.com/google/osv-scanner/pull/797>


robramsaynz:
I've done a bit more poking and managed to create git-repo with submodules that doesn't have a backing github repo, so re-utilising the existing fixture based testing —`func Test_scanGit`— looks like the best option at this stage.

PR #879, opened 2024-03-20
GR: Add tests for npmrc & npm registry api

michaelkedar:
Add some tests for `npmrc` parsing / registry request building and querying of the npm registries.

Currently, the `npmrc` parser can use a few environment variables, which makes it a bit difficult to test properly. I may change it in another PR to make it take them in as an argument so I can sandbox the tests better.


internal/resolution/datasource/npm_registry_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"slices"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/google/osv-scanner/internal/resolution/datasource"
+	"github.com/google/osv-scanner/internal/testutility"
+)
+
+type fakeNpmRegistry struct {
```
michaelkedar, line 19: This is practically the same as the mock http server in the maven tests, and it's probably going to be used in other tests.
I'm going to add a more general mock http server to testutility in another PR

cuixq: that would be nice! I definitely implement this again for Maven parent support....

internal/resolution/datasource/npm_registry_test.go
```diff
@@ -0,0 +1,141 @@
...
+	if _, err := w.Write(resp); err != nil {
+		log.Fatalf("Write: %v", err)
+	}
+}
+
+func TestNpmRegistryClient(t *testing.T) {
+	t.Parallel()
+
+	//nolint:gosec  // "Potential hardcoded credentials" :)
+	const (
```
cuixq, line 67: how are these credentials generated?

cuixq (approved):

internal/resolution/datasource/npm_registry_test.go
michaelkedar: These are just random base64-encoded strings. `auth` is supposed to be the encoded `username:password`, but I'm really just checking these are included in the Authorization header.

These I just randomly went with `cool:auth` and `nice-token` :slightly_smiling_face:

another-rex (approved):

PR #888, opened 2024-03-26
Make MockHTTPServer for tests

michaelkedar:
Factored out the httptest.Server / handler into one struct for easier reuse in tests.


cuixq (approved):
Nice!

PR #890, opened 2024-03-27
Automated Updates: support parents and dependency imports

cuixq:
This PR adds support for parents and dependency imports:
 - fetching and parsing upstream pom.xml files
 - then merging the data into Maven project
 - add `RequirementsFromOtherPOMs` to system-specific data to store requirements that we don't have access

Some of the Maven code should be provided by `deps.dev/util/maven` and `deps.dev/util/resolve` soon...


internal/resolution/manifest/maven.go
```diff
@@ -136,10 +201,53 @@ func (m MavenManifestIO) Read(df lockfile.DepFile) (Manifest, error) {
 		EcosystemSpecific: MavenManifestSpecific{
 			Properties:                 properties,
 			RequirementsWithProperties: reqsWithProps,
+			RequirementsFromOtherPOMs:  otherRequirements,
 		},
 	}, nil
 }
 
+const MaxParent = 100
```
michaelkedar, line 209: is this a limit in Maven itself, or just for us?

michaelkedar (approved):
LGTM, but Maven makes my head hurt

internal/resolution/manifest/maven.go
cuixq: this is just for us

oliverchang (approved):

internal/resolution/manifest/maven.go
another-rex: Can you add a comment here clarifying that?

internal/resolution/manifest/maven.go
```diff
@@ -173,57 +257,6 @@ func mavenOrigin(list ...string) string {
 	return result
 }
 
-func makeMavenDepType(dependency maven.Dependency, origin string) dep.Type {
```
another-rex, line 176: Just to check, this is removed because the logic has been moved to resolve right?

another-rex (approved):
LGTM, added some minor comments.

internal/resolution/manifest/maven.go
cuixq: this is now in `deps.dev/util/resolve`, so yes if this is what you refer to by `resolve`

internal/resolution/manifest/maven.go
cuixq: done

PR #892, opened 2024-03-27
update deps.dev/util/maven

cuixq:
This PR updates `deps.dev/util/maven` to the latest version.


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.21.8
 
 require (
 	deps.dev/api/v3 v3.0.0-20240318003216-81f538734eda
-	deps.dev/util/maven v0.0.0-20240318003216-81f538734eda
+	deps.dev/util/maven v0.0.0-20240322043601-ff53416fec6a
 	deps.dev/util/resolve v0.0.0-20240318003216-81f538734eda
 	deps.dev/util/semver v0.0.0-20240318003216-81f538734eda
```
michaelkedar, line 9: nit: Should we update the psuedo-versions of the other deps.dev modules to be the same?

michaelkedar (approved):

go.mod
cuixq: done

PR #901, opened 2024-04-05
GR: Parse paths in npmrc auth fields correctly

### LINKED ISSUE #899: Guided remediation for npm (`osv-scanner fix`) failed to resolve private dependencies
It looks like the guide doesn't work when my project has private dependencies. After some digging, I found out that the root cause is the guide's reliance on deps.dev, which can't access those private packages.

To reproduce, simply include any private dependencies in `package.json` and run `osv-scanner fix -M package.json`.
The error was:
`cannot find matching versions for ^<version>: package NPM:@<package>: not found`

I've got a couple ideas on how to tackle this:

1. Ignore private dependencies for now. It's not ideal, but at least the guide would work for public packages. This could be an additional CLI boolean flag
2. Switch to using the npm API (that works with local configurations, such as nprmc). It might be slower, but it would let the guide handle both public and private dependencies.

What do you think? I'd love to hear your thoughts on the best way forward. Fixing this would be a big help for folks like me who use private dependencies.

Thanks for all your hard work on OSV-Scanner!

michaelkedar:
Should fix #899 
Changed the npmrc parsing logic when using `--data-source=native` so that specifying registries with paths (e.g. `//my.registry/package/path:_authToken`) will now correctly add the authorization headers.

Used [npm-registry-fetch](https://github.com/npm/npm-registry-fetch/tree/main) as reference.

### DIFF
--- a/internal/resolution/datasource/npm_registry_cache.go
+++ b/internal/resolution/datasource/npm_registry_cache.go
@@ -8,9 +8,9 @@ import (
 )
 
 type npmRegistryCache struct {
-	Timestamp    *time.Time
-	Details      map[string]npmRegistryPackageDetails
-	RegistryURLs map[string]string
+	Timestamp *time.Time
+	Details   map[string]npmRegistryPackageDetails
+	ScopeURLs map[string]string
 }
 
 func (c *NpmRegistryAPIClient) GobEncode() ([]byte, error) {
@@ -23,15 +23,13 @@ func (c *NpmRegistryAPIClient) GobEncode() ([]byte, error) {
 	}
 
 	cache := npmRegistryCache{
-		Timestamp:    c.cacheTimestamp,
-		Details:      c.details,
-		RegistryURLs: make(map[string]string),
+		Timestamp: c.cacheTimestamp,
+		Details:   c.details,
+		ScopeURLs: make(map[string]string),
 	}
 
 	// store the registry URL for each scope (but not the auth info)
-	for scope, reg := range c.registries {
-		cache.RegistryURLs[scope] = reg.URL
-	}
+	cache.ScopeURLs = c.registries.ScopeURLs
 
 	return gobMarshal(&cache)
 }
@@ -58,7 +56,7 @@ func (c *NpmRegistryAPIClient) GobDecode(b []byte) error {
 			scope, _, _ = strings.Cut(pkg, "/")
 		}
 
-		return cache.RegistryURLs[scope] != c.registries[scope].URL
+		return cache.ScopeURLs[scope] != c.registries.ScopeURLs[scope]
 	})
 
 	c.cacheTimestamp = cache.Timestamp
--- a/internal/resolution/datasource/npm_registry_cache.go
+++ b/internal/resolution/datasource/npm_registry_cache.go
@@ -8,9 +8,9 @@ import (
 )
 
 type npmRegistryCache struct {
-	Timestamp *time.Time
-	Details   map[string]npmRegistryPackageDetails
-	ScopeURLs map[string]string
+	Timestamp *time.Time                           // Timestamp of when this cache was made
+	Details   map[string]npmRegistryPackageDetails // For a package name, the versions & their dependencies, and the list of tags
+	ScopeURLs map[string]string                    // The URL of the registry used for a given package @scope. Used to invalidate cache if registry has changed.
 }
 
 func (c *NpmRegistryAPIClient) GobEncode() ([]byte, error) {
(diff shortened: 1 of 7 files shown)


internal/resolution/datasource/npmrc.go
```diff
@@ -150,7 +150,7 @@ func (authInfo npmRegistryAuthInfo) addAuth(header http.Header) {
 		authBytes := []byte(authInfo.username + ":")
 		b, err := base64.StdEncoding.DecodeString(authInfo.password)
 		if err != nil {
-			// TODO: npm seems to actually be quite lenient with invalid encodings
+			// TODO: mimic the behaviour of node's Buffer.from(s, 'base64').toString()
```
oliverchang, line 153: can you add some more details inline on how this differs?

internal/resolution/datasource/npm_registry_cache.go
```diff
@@ -8,9 +8,9 @@ import (
 )
 
 type npmRegistryCache struct {
-	Timestamp    *time.Time
-	Details      map[string]npmRegistryPackageDetails
-	RegistryURLs map[string]string
+	Timestamp *time.Time
+	Details   map[string]npmRegistryPackageDetails
+	ScopeURLs map[string]string
```
oliverchang, line 13: Can you add some comments documenting what these are?

internal/resolution/datasource/npmrc_test.go
```diff
@@ -154,3 +156,141 @@ func TestNpmrcRegistryOverriding(t *testing.T) {
...
+			},
+			requestURL: "https://my.custom.registry/here/",
+			wantAuth:   "Basic " + b64enc("user:pass"),
+		},
+		{
+			name: "token auth",
+			opts: datasource.NpmRegistryAuthOpts{
+				"//my.custom.registry/here/:_authToken": "c0ffee",
+				"//my.custom.registry/here/:token":      "nope",
+				"//my.custom.registry/:_authToken":      "c0ffee",
```
oliverchang, line 188: Should we set different tokens for the two paths here (/here vs / )? To make sure the right one is selected?

oliverchang (commented):
nice! just some minor comments.

internal/resolution/datasource/npmrc_test.go
michaelkedar: Will do.
For some reason npm-registry-fetch has these the same in [this test](https://github.com/npm/npm-registry-fetch/blob/237d33b45396caa00add61e0549cf09fbf9deb4f/test/auth.js#L55-L63):

another-rex (approved):

internal/resolution/datasource/npmrc.go
```diff
@@ -159,37 +161,69 @@ func (authInfo npmRegistryAuthInfo) addAuth(header http.Header) {
...
+// Implementation of npm registry auth matching, adapted from npm-registry-fetch
+// https://github.com/npm/npm-registry-fetch/blob/237d33b45396caa00add61e0549cf09fbf9deb4f/lib/auth.js
+type NpmRegistryAuthOpts map[string]string
 
-// create the http request to the registry api
-// urlComponents should be (package) or (package, version)
-func (info npmRegistryInfo) buildRequest(ctx context.Context, urlComponents ...string) (*http.Request, error) {
-	for i := range urlComponents {
-		urlComponents[i] = url.PathEscape(urlComponents[i])
+var npmAuthFields = [...]string{":_authToken", ":_auth", ":username", ":_password"} // reference of the relevant config key suffixes
```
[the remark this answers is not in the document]
G-Rath: It sets the length of the array based on the number of dots (so in this case 3); my understanding is it's a shortcut `make`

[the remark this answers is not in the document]
another-rex: hmm... but there's 4 elements in the slice?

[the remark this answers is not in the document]
michaelkedar: It means an array with length equal to the number of elements defined:
[From the go spec:](https://go.dev/ref/spec#Composite_literals)
> The length of an array literal is the length specified in the literal type. If fewer elements than the length are provided in the literal, the missing elements are set to the zero value for the array element type. It is an error to provide elements with index values outside the index range of the array. **The notation `...` specifies an array length equal to the maximum element index plus one.**
>```
>buffer := [10]string{}             // len(buffer) == 10
>intSet := [6]int{1, 2, 3, 5}       // len(intSet) == 6
>days := [...]string{"Sat", "Sun"}  // len(days) == 2
>```

oliverchang (approved):

PR #902, opened 2024-04-05
test: cleanup duplicate snapshots

G-Rath:
While working on #889 I found `go-snaps` had a race condition that meant it sometimes duplicates snapshots - this didn't cause any tests to fail, but was confusing and annoying when I was using the snapshots for mass-checking small tweaks to the new output format.

I wrote a small script to clean up these duplicates, but amazingly @gkampitakis has yet again jumped on this quickly and gotten a fix out before I had a chance to mention this to the rest of the team let alone push up my script 😄 

This uses my script to cleanup the duplicates and then updates `go-snaps` to the fixed version so that new test runs won't have this problem.

<details>
<summary>
For those interested, this is the Python script I wrote
</summary>

```
#!/usr/bin/env python


import glob
import re


def deduplicate_snapshots(snapshots):
  by_name = {}

  for snapshot in snapshots:
    if snapshot[1] in by_name:
      are_identical = len(snapshot) == len(by_name[snapshot[1]]) and [
        i for i, j in zip(snapshot, by_name[snapshot[1]]) if i == j
      ]

      print(
        "  removed duplicate of",
        snapshot[1].strip(),
        "(were identical)" if are_identical else "",
      )
      continue
    by_name[snapshot[1]] = snapshot
  return by_name.values()


def sort_snapshots(snapshots):
  return sorted(
    snapshots,
    key=lambda lines: [
      int(x) if x.isdigit() else x for x in re.split(r"(\d+)", lines[1])
    ],
  )


def sort_snapshots_in_file(filename):
  snapshots = []
  snapshot = []

  with open(filename, "r") as f:
    for line in f.readlines():
      if line == "---\n":  # marks the start of a new snapshot
        snapshots.append(snapshot)
        snapshot = []
      else:
        snapshot.append(line)
  print("found", len(snapshots), "snapshots in", filename)

  snapshots = deduplicate_snapshots(snapshots)
  snapshots = sort_snapshots(snapshots)

  if filename.endswith("v_test.snap") or filename.endswith("v_test2.snap"):
    return
  with open(filename, "w") as f:
    for snapshot in snapshots:
      f.writelines(snapshot)
      f.writelines("---\n")


for snapshot_file in glob.iglob("**/__snapshots__/*.snap", recursive=True):
  sort_snapshots_in_file(snapshot_file)
```
</details>

Relates to https://github.com/gkampitakis/go-snaps/issues/96


gkampitakis:
> It seems that my script doesn't give the exact same order as `go-snaps` though I can't see why - it doesn't seem like the order is alphabetical or in test order...?
> 
> This might be a suboptimal thing in `go-snaps` but I wouldn't call it a bug since it seems to be stable across `go-snaps` runs (both with existing and regenerating snapshots) which is the source of truth so I've just completely generated the snapshots with duplicates and moved on 🤷

Indeed it uses natural sort.

G-Rath:
> Indeed it uses natural sort.

Which I went through some annoying conversations with chatgpt to implement in Python without using a package 😅 

But na I think the actual problem is some of the new packages have not actually enabled sorting - I'm doing a new PR to address that

G-Rath:
This should be landed _after_ #904

another-rex (approved):

PR #903, opened 2024-04-05
ci: ensure snapshots are always cleaned up

G-Rath:
While doing #902 I realised we've got a couple of packages using snapshots but not cleaning them up afterwards - this adds a basic script to check for that, based on the existence of the `__snapshots__` directory.

Here's what the annotations look like:

image

Note that because annotations are only shown for _files_, we annotate the first test file in a directory that doesn't contain `testmain_test.go` to ensure it is visible.


G-Rath:
This should be landed _after_ #904

another-rex:
It might be worth it to change the bash script to python to improve readability, the ubuntu-latest image should come with python preinstalled. (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md).

G-Rath:
> ubuntu-latest image should come with python preinstalled.

Yes, but local environments (such as Windows) might not - I'm also not sure if it'll actually be that much more readable given how small the script is (I think it's mainly the `while` and `if` statements that are not immediately obvious); but I'll rewrite it anyway cause we're not expecting this to be a major thing 😄

another-rex (approved):
LGTM, thanks! I think having it in Python actually makes it easier to run on Windows right? As they won't have bash/grep unless they are running on Linux.

G-Rath:
> As they won't have bash/grep unless they are running on Linux.

True though all of our other scripts are written in bash so Windows users will need "bash" installed already (and that doesn't mean they have to use Linux-on-Windows)

another-rex:
How does the snapshot cleanups affect our acceptance tests if we have snapshots in them? Will it "cleanup" snapshots for those skipped tests?

G-Rath:
No, that's why we our skipping is done using `snaps.Skip` rather than `t.Skip` directly as then `go-snaps` still sees the tests we're skipping and doesn't clean them up.

PR #908, opened 2024-04-10
Fix rust call analysis by explicitly disabling stripping of debug info

another-rex:
Latest rust version (1.77) implemented stripping release binaries by default. This makes the call analysis stop working, so explicitly disable the stripping behaviour when building for analysis.

https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#enable-strip-in-release-profiles-by-default


oliverchang (approved):
Is this backwards compatible with older rust tool chains? i.e. this flag being included won't break them?

another-rex:
Yes it should be, the flag has existed for a long time, it's just been defaulted to debuginfo.

oliverchang (approved):

PR #926, opened 2024-04-17
Fix regression for go call analysis in 1.7.0

### LINKED ISSUE #914: Go call analysis regression in 1.7.0
Go call analysis results stop showing uncalled properly for packages in 1.7.0  (still shows for stdlib vulnerabilities). , 1.6.2 still works correctly.

another-rex:
Fixes #914 

govulncheck behaviour changed either in 1.0.3 or 1.0.4.

govulncheck 1.0.2 used to essentially ignore performing call analysis for OSV entries without any symbol information. 
govulncheck 1.0.4 now attempts to perform call analysis assuming that everything all symbols are vulnerable, so any calls to any functions/symbols are considered a vulnerable function called. This is the case for most GHSA entries. 

OSV-Scanner already accounted for OSV entries that doesn't have symbol level information, and assuming that it affects everything, if there are no aliased vulnerability entries that does contain symbol information which is not called. 

the govulncheck change resulted in GHSA entries being incorrectly recorded as having symbol level information, and those symbols have been called, resulting in the following output:

```
"experimentalAnalysis": {
  "GHSA-c3h9-896r-86jm": {
    "called": true
  },
  "GO-2021-0053": {
    "called": false
  }
}
```

When presented with differing called results, the table output defaults to it being called.

Our existing tests did not pick this up, because we didn't have an e2e test, and the unit tests assumed only advisories with symbol information affected the output. 

---

Contents of PR:

- Add extra check when govulncheck returns `called=true` to only pass the result on when advisory actually contains symbol information.
- Add integration/e2e test for call analysis for go, this way any behavior changes will be picked up in the future.
- Edited the fixtures slightly for our existing tests to fail when this new check that's been added is not present.

We could add a GHSA entry without symbol information to the unit tests, but that results in a snapshot >3000 lines long because govulncheck will return traces for everything.


oliverchang (approved):
amazing work debugging this!!

hogo6002 (approved):

PR #930, opened 2024-04-17
Fix paths in test snapshots

another-rex:
Remove local paths from snapshots


michaelkedar (approved):

G-Rath (approved):
nit: ideally `rootdir` would be more consistent with the other tests

another-rex:
Good point, updated!

PR #937, opened 2024-04-23
test: add cases for output functions

G-Rath:
This introduces a set of crafted scanner results that each supported `output` format is run through to showcase how they look across all the different results possible from a scanner run - it originally started life as the tests for #889 but I realised they could base used more generally for testing and reviewing all the outputters, so here we are.

~It looks like this has also revealed the SARIF output is unstable in its ordering, which I'll aim to address in a dedicated PR~


G-Rath:
This is ready for review but #938 should be merged first.

Note that beyond the usual review of the technical implementation, reviewers should also review the test cases to make sure they're correct for the real world and are complete, and also review the snapshots for each outputs for any oddentities that we might want to change (for which dedicated issues should be created - they _won't_ be addressed in this PR)

G-Rath:
fwiw the coverage output for this looks even better after #961 is landed, as it means we've got 100% coverage across most files in the package 😄

another-rex (approved):
LGTM, thanks!

PR #938, opened 2024-04-23
fix: ensure the sarif output has a stable order

G-Rath:
This stabilizes the order in the SARIF output so that it is deterministic, as #937 proved it was not.

### DIFF
--- a/internal/output/result.go
+++ b/internal/output/result.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/google/osv-scanner/pkg/models"
+	"golang.org/x/exp/maps"
 )
 
 type pkgWithSource struct {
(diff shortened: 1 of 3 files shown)


internal/output/result.go
```diff
@@ -15,6 +17,27 @@ type pkgWithSource struct {
...
+// StableKeys returns the pkgWithSource keys in a deterministic order
+func (pss *pkgSourceSet) StableKeys() []pkgWithSource {
+	pkgWithSrcKeys := maps.Keys(*pss)
+
+	slices.SortFunc(pkgWithSrcKeys, func(a, b pkgWithSource) int {
+		for _, fn := range []func() int{
+			func() int { return strings.Compare(a.Source.Path, b.Source.Path) },
+			func() int { return strings.Compare(a.Package.Name, b.Package.Name) },
+			func() int { return strings.Compare(a.Package.Version, b.Package.Version) },
+		} {
```
G-Rath, line 29: I'm super on the fence about this: on the one hand its clever, but on the other maybe it's too clever?

I've got a commit locally that unwinds this into `if` statements which I can push up if folks agree it's too clever

another-rex: I think it's easier to parse than if statements once you understand what it's doing, can you just add a comment above it basically describing it's just sorting by each field in descending priority.

another-rex (approved):
Thanks!

G-Rath:
The coverage drop is fine - we were not covering the original code anyway, and it will be covered by #937, after which just the final `return 0` will be uncovered as I can't seem to hit that :/

internal/output/result.go
G-Rath: sweet done - one other annoying downside though is that we can't cover the `return 0` 😞

another-rex (approved):

internal/output/result.go
G-Rath: Ok turns out there's a `cmp.Or` for exactly this - it's technically less efficient because it's not lazy ([though there's a proposal for that](https://github.com/golang/go/issues/66986)), but these compares should be pretty inexpensive and its nice to remove the unreachable code path

oliverchang (approved):

PR #939, opened 2024-04-23
fix: ensure that npm dependencies retain their "production" grouping

### LINKED ISSUE #924: Scan groups are inconsistent when there are duplicate packages in npm
If the same package version is installed multiple times under different groups in a package-lock.json file (i.e. in both dev and prod), `osv-scanner scan` behaves inconsistently in showing which groups the package belongs to.

e.g. with this [package-lock.json](https://github.com/google/osv-scanner/files/15017137/package-lock.json) (from [package.json](https://github.com/google/osv-scanner/files/15017138/package.json)), the output is randomly:
```
╭─────────────────────────────────────┬──────┬───────────┬─────────┬─────────┬───────────────────╮
│ OSV URL                             │ CVSS │ ECOSYSTEM │ PACKAGE │ VERSION │ SOURCE            │
├─────────────────────────────────────┼──────┼───────────┼─────────┼─────────┼───────────────────┤
│ https://osv.dev/GHSA-v88g-cgmw-v5xw │ 5.6  │ npm       │ ajv     │ 5.5.2   │ package-lock.json │
╰─────────────────────────────────────┴──────┴───────────┴─────────┴─────────┴───────────────────╯
```
or
```
╭─────────────────────────────────────┬──────┬───────────┬───────────┬─────────┬───────────────────╮
│ OSV URL                             │ CVSS │ ECOSYSTEM │ PACKAGE   │ VERSION │ SOURCE            │
├─────────────────────────────────────┼──────┼───────────┼───────────┼─────────┼───────────────────┤
│ https://osv.dev/GHSA-v88g-cgmw-v5xw │ 5.6  │ npm       │ ajv (dev) │ 5.5.2   │ package-lock.json │
╰─────────────────────────────────────┴──────┴───────────┴───────────┴─────────┴───────────────────╯
```

Because ajv@5.5.2 is installed twice, once as a prod dependency, and once as a dev dependency:
```json
    "node_modules/eslint/node_modules/ajv": {
      "version": "5.5.2",
      "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
      "integrity": "sha512-Ajr4IcMXq/2QmMkEmSvxqfLN5zGmJ92gHXAeOXq1OekoH2rfDNsgdDoL2f7QaRCy7G/E6TpxBVdRuNraMztGHw==",
      "dev": true,
      "dependencies": {
        "co": "^4.6.0",
        "fast-deep-equal": "^1.0.0",
        "fast-json-stable-stringify": "^2.0.0",
        "json-schema-traverse": "^0.3.0"
      }
    },
    "node_modules/table/node_modules/ajv": {
      "version": "5.5.2",
      "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
      "integrity": "sha512-Ajr4IcMXq/2QmMkEmSvxqfLN5zGmJ92gHXAeOXq1OekoH2rfDNsgdDoL2f7QaRCy7G/E6TpxBVdRuNraMztGHw==",
      "dependencies": {
        "co": "^4.6.0",
        "fast-deep-equal": "^1.0.0",
        "fast-json-stable-stringify": "^2.0.0",
        "json-schema-traverse": "^0.3.0"
      }
    },
```

Presumably, this would also affect groups in other lockfiles that can support the same package being installed multiple times.

G-Rath:
This resolves an inconsistency in the scanner output for npm packages that appear in the same tree multiple times but in different groups; this happens because the table outputter deduplicates on groups on the assumption that a package can only appear in a single group which is incorrect for the NPM ecosystem.

To address this, I've introduced an internal map type that ensures groups are merged when a package is added, with the twist that if either instance of a package being merged is in no groups then that is the result of the merge because implicitly that means an instance of the package is in the "production" group which takes priority over the other groups.

This is something that should most likely be improved on the future, but right now this fix should be good enough to ship since afaik it doesn't impact any other ecosystem and groups are something of a PoC given we can't resolve that information richly enough across all ecosystems yet.

I think this technically could impact `pnpm` as well, but none of our fixtures gave a good indicator and the latest lockfile version doesn't have that information anymore so frankly I'm not worrying about it at this point.

Resolves #924

### DIFF
--- a/pkg/lockfile/parse-npm-lock-v1_test.go
+++ b/pkg/lockfile/parse-npm-lock-v1_test.go
@@ -445,28 +445,7 @@ func TestParseNpmLock_v1_SamePackageDifferentGroups(t *testing.T) {
 			Version:   "5.5.2",
 			Ecosystem: lockfile.NpmEcosystem,
 			CompareAs: lockfile.NpmEcosystem,
-			DepGroups: nil,
-		},
-		{
-			Name:      "ajv",
-			Version:   "5.5.2",
-			Ecosystem: lockfile.NpmEcosystem,
-			CompareAs: lockfile.NpmEcosystem,
-			DepGroups: []string{"dev"},
-		},
-		{
-			Name:      "ajv",
-			Version:   "5.5.2",
-			Ecosystem: lockfile.NpmEcosystem,
-			CompareAs: lockfile.NpmEcosystem,
-			DepGroups: []string{"optional"},
-		},
-		{
-			Name:      "ajv",
-			Version:   "5.5.2",
-			Ecosystem: lockfile.NpmEcosystem,
-			CompareAs: lockfile.NpmEcosystem,
-			DepGroups: []string{"dev", "optional"},
+			DepGroups: []string{},
 		},
 	})
 }
--- a/pkg/lockfile/parse-npm-lock-v1_test.go
+++ b/pkg/lockfile/parse-npm-lock-v1_test.go
@@ -415,3 +415,58 @@ func TestParseNpmLock_v1_OptionalPackage(t *testing.T) {
 		},
 	})
 }
+
+func TestParseNpmLock_v1_SamePackageDifferentGroups(t *testing.T) {
+	t.Parallel()
+
+	packages, err := lockfile.ParseNpmLock("fixtures/npm/same-package-different-groups.v1.json")
+
+	if err != nil {
+		t.Errorf("Got unexpected error: %v", err)
+	}
+
+	expectPackages(t, packages, []lockfile.PackageDetails{
+		{
+			Name:      "eslint",
+			Version:   "1.2.3",
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: []string{"dev"},
+		},
+		{
+			Name:      "table",
+			Version:   "1.0.0",
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: nil,
+		},
+		{
+			Name:      "ajv",
+			Version:   "5.5.2",
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: nil,
+		},
+		{
+			Name:      "ajv",
+			Version:   "5.5.2",
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: []string{"dev"},
+		},
+		{
+			Name:      "ajv",
+			Version:   "5.5.2",
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: []string{"optional"},
+		},
+		{
+			Name:      "ajv",
+			Version:   "5.5.2",
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: []string{"dev", "optional"},
+		},
+	})
+}
--- a/pkg/lockfile/parse-npm-lock-v1_test.go
+++ b/pkg/lockfile/parse-npm-lock-v1_test.go
@@ -438,7 +438,6 @@ func TestParseNpmLock_v1_SamePackageDifferentGroups(t *testing.T) {
 			Version:   "1.0.0",
 			Ecosystem: lockfile.NpmEcosystem,
 			CompareAs: lockfile.NpmEcosystem,
-			DepGroups: nil,
 		},
 		{
 			Name:      "ajv",
--- a/pkg/lockfile/parse-npm-lock-v1_test.go
+++ b/pkg/lockfile/parse-npm-lock-v1_test.go
@@ -188,6 +188,7 @@ func TestParseNpmLock_v1_NestedDependenciesDup(t *testing.T) {
 		Version:   "6.1.0",
 		Ecosystem: lockfile.NpmEcosystem,
 		CompareAs: lockfile.NpmEcosystem,
+		DepGroups: []string{},
 	})
 
 	expectPackage(t, packages, lockfile.PackageDetails{
@@ -195,6 +196,7 @@ func TestParseNpmLock_v1_NestedDependenciesDup(t *testing.T) {
 		Version:   "5.5.0",
 		Ecosystem: lockfile.NpmEcosystem,
 		CompareAs: lockfile.NpmEcosystem,
+		DepGroups: []string{},
 	})
 
 	expectPackage(t, packages, lockfile.PackageDetails{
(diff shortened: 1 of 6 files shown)


pkg/lockfile/parse-npm-lock.go
```diff
@@ -57,18 +59,43 @@ func pkgDetailsMapToSlice(m map[string]PackageDetails) []PackageDetails {
...
-		details[name] = detail
+type npmPackageDetailsMap map[string]PackageDetails
+
+// mergeNpmDepsGroups handles merging the dependency groups of packages within the
+// NPM ecosystem, since they can appear multiple times in the same dependency tree
+//
+// the merge happens almost as you'd expect, except that if either given packages
+// belong to no groups, then that is the result since it indicates the package
+// is implicitly a production dependency.
+func mergeNpmDepsGroups(a, b PackageDetails) []string {
```
cuixq, line 70: it seems we only need `DepGroups` so can the func be
```
func mergeNpmDepGroups(a, b []string) []string
```
?

pkg/lockfile/parse-npm-lock.go
```diff
@@ -57,18 +59,43 @@ func pkgDetailsMapToSlice(m map[string]PackageDetails) []PackageDetails {
 	return details
 }
 
-func mergePkgDetailsMap(m1 map[string]PackageDetails, m2 map[string]PackageDetails) map[string]PackageDetails {
-	details := map[string]PackageDetails{}
-
-	for name, detail := range m1 {
-		details[name] = detail
+type npmPackageDetailsMap map[string]PackageDetails
```
cuixq, line 62: maybe add a comment stating that the key is name and version

pkg/lockfile/parse-npm-lock.go
G-Rath: It could but that'd mean every caller of this function would have to explicitly access `DepGroups` themselves - so it's cleaner for us to take `PackageDetails` and do that accessing ourselves.

oliverchang (commented):
I'll defer this to @michaelkedar, who is our npm expert. WDYT of this approach?

michaelkedar:
Looking a bit into this, there's another awkward edge case with optional prod dependencies & non-optional dev dependencies (`devOptional`) that we might want to consider.

> dev, optional, devOptional: If the package is strictly part of the `devDependencies` tree, then `dev` will be true. If it is strictly part of the `optionalDependencies` tree, then `optional` will be set. If it is both a `dev` dependency _and_ an `optional` dependency of a non-dev dependency, then `devOptional` will be set. (An `optional` dependency of a `dev` dependency will have both `dev` and `optional` set.)

https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#packages

Our groups parser doesn't seem to get this correct at the moment:
https://github.com/google/osv-scanner/blob/97029011c8adaf14f03fa66519836063204bf474/pkg/lockfile/parse-npm-lock.go#L125-L137

So I guess we ought to differentiate between "optional in prod, required in dev" (`devOptional`) and "not in prod, optional in dev" (`dev && optional`)  somehow, and also apply that to the deduplication logic.


This whole thing kind of comes to some thoughts I've been having on counting vulnerabilities and how or whether we should be deduplicating vulnerabilities in an npm lockfile in the first place. I feel like it'd generally be clearer to have one entry per vulnerability ID, with a complete list of instances of affected packages.
Maybe that's a thing to add to the V2 Wishlist.

G-Rath:
I don't think I disagree with that but frankly it makes my head spin a bit 😅 

Personally I feel like it's probably safest to at least start by just treating groups as part of the uniqueness of a package - I'm pretty sure for most ecosystems that won't change anything because you can only have instance of a package regardless of group, and for ecosystems like NPM it'd mean you would just end up with some more packages in the output.

Then we can work on how to make the output smarter without worrying about the current version of the cli giving false negatives in the meantime

G-Rath:
note to self: consider exploring if an empty string could be useful here - i.e. if a package is not optional or dev, then its `""`

(though thinking about it, I think that could just lead back to this same solution - I think either way, its probably correct more as it needs to be handled within the extractor rather than at the outputter level right now)

pkg/lockfile/parse-npm-lock-v1_test.go
```diff
@@ -415,3 +415,36 @@ func TestParseNpmLock_v1_OptionalPackage(t *testing.T) {
...
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: []string{"dev"},
+		},
+		{
+			Name:      "table",
+			Version:   "1.0.0",
+			Ecosystem: lockfile.NpmEcosystem,
+			CompareAs: lockfile.NpmEcosystem,
+			DepGroups: nil,
```
another-rex, line 441: Is this explicit nil intentional?

another-rex (approved):
I'm happy to leave the case of "optional in prod, required in dev", and (optional in dev, not in prod) as something to consider later.

pkg/lockfile/parse-npm-lock-v1_test.go
G-Rath: Na probably not - I'm on mobile so mind making an inline suggestion removing it so we can get this merged? 😁

another-rex, line 441:
```suggestion
```

another-rex: I don't think I have write permissions, so you'll need to commit it.

G-Rath: Weird, you should - and I apparently can't commit from mobile 😅

I'll deal with it later tonight

another-rex, line 441:
```suggestion
```

another-rex (commented):
Trying again as a review comment

pkg/lockfile/parse-npm-lock-v1_test.go
another-rex: hmm nope still doesn't work

G-Rath: It might just be cause it's on mobile, but are you actually suggesting a multiline edit?

I believe suggestions can't delete the whole selection - they've got to modify, so for deletes you have to suggest modifying the previous and next line as part of the suggestion

another-rex, line 441:
```suggestion
		{
			Name:      "table",
			Version:   "1.0.0",
			Ecosystem: lockfile.NpmEcosystem,
			CompareAs: lockfile.NpmEcosystem,
```

pkg/lockfile/parse-npm-lock-v1_test.go
```diff
@@ -415,3 +415,36 @@ func TestParseNpmLock_v1_OptionalPackage(t *testing.T) {
...
+		},
+		{
+			Name:      "ajv",
+			Version:   "5.5.2",
```
another-rex, line 445:
```suggestion
		},
		{
			Name:      "table",
			Version:   "1.0.0",
			Ecosystem: lockfile.NpmEcosystem,
			CompareAs: lockfile.NpmEcosystem,
		},
		{
			Name:      "ajv",
			Version:   "5.5.2",
```

PR #940, opened 2024-04-23
refactor: use `maps` functions instead of custom implementations

G-Rath:
This replaces some of our custom code for map-based operations with almost-standard-library versions

### DIFF
--- a/pkg/lockfile/parse.go
+++ b/pkg/lockfile/parse.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+
+	"golang.org/x/exp/maps"
 )
 
 func FindParser(pathToLockfile string, parseAs string) (PackageDetailsParser, string) {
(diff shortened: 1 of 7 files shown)


pkg/lockfile/parse.go
```diff
@@ -57,28 +59,14 @@ var ErrParserNotFound = errors.New("could not determine parser")
 
 type Packages []PackageDetails
 
-func toSliceOfEcosystems(ecosystemsMap map[Ecosystem]struct{}) []Ecosystem {
-	ecosystems := make([]Ecosystem, 0, len(ecosystemsMap))
-
-	for ecosystem := range ecosystemsMap {
-		if ecosystem == "" {
```
oliverchang, line 64: Is this logic no longer necessary?

oliverchang (approved):
LGTM with one question. Thanks for the cleanup!

pkg/lockfile/parse.go
G-Rath: I'm not sure that it was ever needed - I think it came over from `osv-detector`, and I think might have been a legacy of how I previously gathered ecosystems but then later refactored.

Either way removing it doesn't cause any test failures for both `osv-detector` and `osv-scanner` to fail, so I think it's best to purposely remove so we can ensure we have a test to cover it if it does turn out to be needed

PR #943, opened 2024-04-26
feat: support parsing `gradle/verification-metadata.xml`

### LINKED ISSUE #915: Support gradle's verification-metadata.xml?
Hi. It looks like `osv-scanner` only checks for `gradle.lockfile` and friends, not `verification-metadata.xml` ([Gradle dependency verification](https://docs.gradle.org/current/userguide/dependency_verification.html)), which is the only "lockfile"-like thing we have for gradle.

We currently use another tool for scanning our gradle dependency tree against NVD, and that tool works with the files we have. We would like to migrate to `osv-scanner` if possible. Is there any reason `osv-scanner` does not parse this file, or is it simply because it has not been added yet?

We currently use the following tool to scan our gradle dependency tree against NVD: https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/index.html.

Current results:
```
$ osv-scanner scan --lockfile android/gradle/verification-metadata.xml
could not determine extractor for /home/foobarbaz/android/gradle/verification-metadata.xml
```

G-Rath:
This adds support for parsing `gradle/verification-metadata.xml` files - since this seems to be like an actual lockfile it's very straightforward: we just parse the file as XML and extract out the name + version of "component".

The interesting part of this is that unlike other project-relative lockfiles this file currently must exist in the `gradle` directory which raises questions about how `--recursive` comes into play  previously we'd not enabled APK and DPKG checking by default but I feel that was more because they were absolute paths and so didn't make sense to do when people were scanning in "project mode".

For now I've just taken the simple route of making the file `gradle/verification-metadata.xml` since that does just work (except for the "find parser" flow which checks against `path.Base` so that has the `gradle` omitted).

Resolves #915


docs/supported_languages_and_lockfiles.md
```diff
@@ -22,19 +22,19 @@ nav_order: 2
...
+| :--------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| C/C++      | `conan.lock`<br>[C/C++ commit scanning](#cc-scanning)                                                                                                       |
+| Dart       | `pubspec.lock`                                                                                                                                              |
+| Elixir     | `mix.lock`                                                                                                                                                  |
+| Go         | `go.mod`                                                                                                                                                    |
+| Java       | `buildscript-gradle.lockfile`<br>`gradle.lockfile`<br>`pom.xml`[\*](https://github.com/google/osv-scanner/issues/35),<br>`gradle/verification-metadata.xml` |
```
cuixq, line 31: should this be alphabetical order?

pkg/lockfile/parse-gradle-verification-metadata_test.go
```diff
@@ -0,0 +1,689 @@
...
+	"testing"
+
+	"github.com/google/osv-scanner/pkg/lockfile"
+)
+
+func TestGradleVerificationMetadataExtractor_ShouldExtract(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
```
cuixq, line 14: delete `name` if it is not used in `t.Run`

docs/supported_languages_and_lockfiles.md
G-Rath: oh yeah good point - I originally thought the `*` was applying to all of these but its actually just `pom.xml` so we could sort it alphabetically.

pkg/lockfile/parse-gradle-verification-metadata.go
```diff
@@ -0,0 +1,56 @@
+package lockfile
+
+import (
+	"encoding/xml"
+	"fmt"
+	"path/filepath"
+)
+
+type GradleVerificationMetadataFile struct {
+	// XMLName             xml.Name              `xml:"project"`
```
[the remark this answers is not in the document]
G-Rath: thanks for the reminder - I was meant to double check what this is actually doing and if we need it

pkg/lockfile/parse-gradle-verification-metadata_test.go
G-Rath: it is being used though? `t.Run(tt.name, func(t *testing.T) {`

cuixq:
I think I reviewed it and pending changes from @G-Rath

G-Rath:
@cuixq if you're requiring changes, could you submit a review requesting changes? The only outstanding change I know of that is needed is [this one](https://github.com/google/osv-scanner/pull/943#discussion_r1585890049) (which I'd completely forgotten about 😅), but that shouldn't hold up others reviewing it.

G-Rath:
Also when we last discussed this, I believe some research was going to be done into the Maven ecosystem to determine how appropriate supporting this file was, which was to be done by @cuixq

cuixq (approved):
I forgot to comment last time with my findings: I think it makes sense to support `vertification-metadata.xml`. I can see this in open source projects - some of these projects may not have `gradle.build` so `verification-metadata.xml` definitely helps with understanding these projects.

Issue #814, opened 2024-02-21
Setting/overwriting license for a dependency


ststroppel:
Hi, 
I've been exploring the new experimental license scanning feature and find it quite impressive. However, I believe there's a crucial functionality that seems to be missing - the ability to specify the license for a dependency.

During my testing of this feature in our npm repository, I encountered several dependencies where the license information could not be detected due to various reasons:

- The license is not properly specified, resulting in the license being resolved as `UNKNOWN`. An example of this is [khroma](https://www.npmjs.com/package/khroma), where the dependency license is declared as a MIT but it was not retrieved correctly. I suspect the necessary information is missing within the package.json.
- The dependency is published under two or more licenses, requiring the user to select the appropriate one. [type-fest](https://www.npmjs.com/package/type-fest) is an example of this scenario.

Therefore, I propose the addition of a feature that allows users to manually set or overwrite the detected license of a dependency, optionally for a specific version or a version range.

oliverchang:
CC  @josieang

arkodg:
also interested in adopting the license scanning feature in Envoy Gateway, and this issue is stopping us from adopting it at this moment, more in https://github.com/envoyproxy/gateway/issues/2917

oliverchang:
Thanks for the feedback all! 

What would be the most convenient way for you to set this information? 

- A config file? i.e. (https://google.github.io/osv-scanner/configuration/)
- A cmdline flag? e.g. `--license-override=PKG,license` or something similar.

ststroppel:
Thanks @oliverchang for coming back to this topic. For me, definitely a config file would be the most convenient way to set it. There are several advantages for this approach:
- The configuration is stored in the repository and versioned
- Additionally, we could specify some version range as the declared license of a package could change with major releases
- The entries could be enriched by additional meta information like `who` approved the license and `why` (e.g. we have a commercial license for the package, that's why license `abc` is the correct one to be used)

oliverchang:
@josieang (who added the license scanning feature initially) would you have some time to help with implementing this? It would also be a great opportunity to look at moving the feature out of experimental.

josieang:
Thanks everyone for feedback on the license scanning feature! Sorry for missing this thread. I'll have time to work on this in early May.

You may have already seen it, but we have a "How are licenses determined?" section at docs.deps.dev/faq.

@ststroppel Just curious:
> Additionally, we could specify some version range as the declared license of a package could change with major releases
[...]

PR #944, opened 2024-04-26
implement filtering by packages through the config

josieang:
I'd like feedback on the config yaml schema, the filter message and it's behaviour if the version is empty (it filters any version of that package).

This is in response to https://github.com/google/osv-scanner/issues/814


pkg/config/config.go
```diff
@@ -34,19 +35,50 @@ type IgnoreEntry struct {
 	Reason      string    `toml:"reason"`
 }
 
+type IgnorePackageVersionEntry struct {
+	Name      string `json:"name"`
+	Version   string `json:"version"` // Q: If empty all versions of the package are ignored?
+	Ecosystem string `json:"ecosystem"`
+	// TODO: Commit? Should we use models.PackageInfo? Or keep this config
```
josieang, line 42: maybe I should also implement Commits too?

pkg/config/config.go
```diff
@@ -34,19 +35,50 @@ type IgnoreEntry struct {
...
+	return shouldIgnoreTimestamp(ignoredLine.IgnoreUntil), ignoredLine
+}
+
+func (c *Config) ShouldIgnorePackageVersion(name, version, ecosystem string) (bool, IgnorePackageVersionEntry) {
+	index := slices.IndexFunc(c.IgnoredPackageVersions, func(elem IgnorePackageVersionEntry) bool {
+		if name != elem.Name {
+			return false
+		}
+
+		return version == elem.Version || elem.Version == ""
```
josieang, line 64: > elem.Version == ""

I think this is helpful behaviour, let me know if you think this would be unexpected for any reason.

pkg/osvscanner/osvscanner.go
```diff
@@ -639,6 +639,19 @@ func filterResults(r reporter.Reporter, results *models.VulnerabilityResults, co
 
 // Filters package-grouped vulnerabilities according to config, preserving ordering. Returns filtered package vulnerabilities.
 func filterPackageVulns(r reporter.Reporter, pkgVulns models.PackageVulns, configToUse config.Config) models.PackageVulns {
+	if ignore, ignoreLine := configToUse.ShouldIgnorePackageVersion(pkgVulns.Package.Name, pkgVulns.Package.Version, pkgVulns.Package.Ecosystem); ignore {
+		pkgString := fmt.Sprintf("%s/%s/%s", pkgVulns.Package.Ecosystem, pkgVulns.Package.Name, pkgVulns.Package.Version)
+		switch len(pkgVulns.Vulnerabilities) {
+		case 1:
+			r.Infof("1 vulnerability for the package %s has been filtered out because: %s\n", pkgString, ignoreLine.Reason)
```
josieang, line 646: maybe I include the vulnerability IDs? like it does for filtering by vuln name?

pkg/config/config.go
```diff
@@ -23,9 +24,11 @@ type ConfigManager struct {
 }
 
 type Config struct {
-	IgnoredVulns      []IgnoreEntry `toml:"IgnoredVulns"`
-	LoadPath          string        `toml:"LoadPath"`
-	GoVersionOverride string        `toml:"GoVersionOverride"`
+	IgnoredVulns                   []IgnoreEntry                        `toml:"IgnoredVulns"`
+	IgnoredPackageVersions         []IgnorePackageVersionEntry          `toml:"IgnoredPackageVersions"`
```
another-rex, line 28: I feel like ignore package versions and override package versions both could have the same "package matching" logic.

Can we consolidate them into the same field, and just have a bool `ignore` which if true ignores the package entirely, otherwise an override field where someone can override the license.

E.g.
```
[[Package]]
name = "pkg-name"
exactVersion = "1.0.0"
ecosystem = "Go"
ignore = false # if true ignores the package entirely
licenseOverride = ["MIT", "0BSD"]
# In the future, if someone requests we can also add a versionOverride field here, or other overrides
reason = "abc"
# major ...
# minor ...
# ...
```

pkg/config/config.go
another-rex: This makes sense to me! If no version is defined, all versions are affected?

pkg/osvscanner/osvscanner.go
another-rex: I don't think we should, since the use case I'm thinking for ignore packages would be if that package is not used at all right (either because it's in fixtures, or it's incorrectly parsed, or it's air gaped or something). In those cases it really doesn't matter what the actual vulnerabilities that affects it are.

pkg/osvscanner/vulnerability_result_internal_test.go
```diff
@@ -400,20 +413,113 @@ func Test_assembleResult(t *testing.T) {
...
+				},
+			},
+		},
+		want: models.VulnerabilityResults{
+			ExperimentalAnalysisConfig: models.ExperimentalAnalysisConfig{
+				Licenses: models.ExperimentalLicenseConfig{
+					Allowlist: []models.License{models.License("MIT"), models.License("0BSD")},
+				},
+			},
+			Results: []models.PackageSource{
```
another-rex, line 449: For maintainability, can we just use the snapshot library here to store/match the results?

pkg/config/config.go
```diff
@@ -34,19 +37,87 @@ type IgnoreEntry struct {
...
+	if index == -1 {
+		return false, IgnorePackageVersionEntry{}
+	}
+	ignoredLine := c.IgnoredPackageVersions[index]
+
+	return shouldIgnoreTimestamp(ignoredLine.IgnoreUntil), ignoredLine
+}
+
+func (c *Config) ShouldOverridePackageVersionLicense(name, version, ecosystem string) (bool, OverridePackageVersionLicenseEntry) {
+	versionParts := strings.Split(version, ".")
```
another-rex, line 86: Not all ecosystems follow semver, not sure how well this will handle non semver versions. Given we have an ecosystem string, can you use the semantic library we have to parse the versions here?

another-rex: Gareth might have some ideas here as well.

another-rex (commented):
Thanks! Main comment is that I think we can consolidate the two overrides into one, so they can share the package matching logic.

pkg/config/config.go
josieang: Ah yes we can, I flagged this as something I would come back to this week: https://github.com/google/osv-scanner/issues/814#issuecomment-2078688416

josieang: Since it's already a dependency, let's use the deps.dev semver package instead of having to import another package.

josieang: Ah sorry I misread - you suggested the `semantic` library. @G-Rath suggested this too in the thread https://github.com/google/osv-scanner/issues/814#issuecomment-2078744502

It supports way more ecosystems than deps.dev's semver library so it might be the better choice actually.

josieang: @G-Rath what do you think is the best way to specify a version prefix range?

I'm thinking an array of [components](https://github.com/G-Rath/osv-detector/blob/e04029c23b143dadd16cdcf0c801773fec58f3d6/pkg/semantic/version.go#L15)?
so [3, 14] means 3.14.x.

But I want to hear other thoughts/suggestions.

josieang: I'm not keen to start making a requirement language for each ecosystem, and I noticed that Components is already an abstraction in the `semantic` library. But I haven't read the library too in depth to figure out if there is a more natural way than specifying an array.

josieang: Rewording what @another-rex suggested offline:

```
[[Package]]
name = "pkg-name"
[[Package.requirement]]
operator = ">"
version = "1.0.0"
[[Package.requirement]]
operator = "<"
version = "1.1.0"
```

since the `semantic` library has a `Compare` function.

josieang: Pros:
> less fields in the config, more flexibility, easier to understand, and probably less implementation as well

G-Rath: fwiw, I personally wonder if it might be too much to be including version ranges in at least this PR - I understand why they could be useful but they don't feel critical to the general feature while requiring you to deal with a bunch of extra complexity.

Having said that, I think your latest proposed config could be a good middle ground since it's hard to argue about supporting more complexities like "1.0.x" for the time being - and frankly while I have a few ideas on how you could do ranges, the more I think about it the more I feel your proposal is good by allowing you to side step a bunch of stuff like differences between version ecosystems.

(part of the reason I say this too is because we're using Go which I've found doesn't combine well with "multiple ecosystems" in this way as everything has to be a single type - that's overall fine, but just means you can't sweep little nounces between ecosystems under the rug for the time being like you could with say TypeScript; and then combine with this being a public cli... 😅)

pkg/config/config.go
```diff
@@ -34,19 +37,87 @@ type IgnoreEntry struct {
...
+	IgnoreUntil time.Time `toml:"ignoreUntil"`
+	Reason      string    `toml:"reason"`
+}
+
+type OverridePackageVersionLicenseEntry struct {
+	Name            string   `toml:"name"`
+	Major           string   `toml:"major"`
+	Minor           string   `toml:"minor"`
+	Patch           string   `toml:"patch"`
+	ExactVersion    string   `toml:"exactVersion"`
```
oliverchang, line 53: Nit: Is there any reason to qualify this with `exact` in `exactVersion` ? IMO `version` on its own is pretty clear.

pkg/config/config.go
oliverchang: +1.

`version` (Which I think we should `exactVersion` to) is optional right? 

Additionally, would it make sense to make `license` a structured field of some sort instead? 

I.e.

```
[[Package]]
[[Package.license]]
override = ["MIT"]
```

pkg/config/config.go
oliverchang: The range syntax makes sense to me, but +1 to @G-Rath's comment to skip this at least in this PR. Is there strong demand for such a feature? 

It adds a bit of complexity to both the config spec and our maintenance.

pkg/config/config.go
josieang: yes that's right.

pkg/config/config.go
josieang: This makes sense to me! done.

pkg/osvscanner/vulnerability_result_internal_test.go
josieang: should be done but check my code just in case I'm not understanding the snapshot library correctly

pkg/config/config.go
josieang: SGTM. So no ranges, you can only specify an exact version (named `version`). If a version is omitted, all versions of the package match.

pkg/config/config.go
josieang: now that we're merging the two, version seems fine to me.

josieang:
Was I supposed to provide a token?

```
Error: Codecov token not found. Please provide Codecov token with -t flag.
Error: Codecov: Failed to properly create commit: The process '/Users/runner/work/_actions/codecov/codecov-action/5ecb98a3c6b[74](https://github.com/google/osv-scanner/actions/runs/8978083643/job/24657957118#step:5:77)7ed38dc09f787459979aebb39be/dist/codecov' failed with exit code 1
```

another-rex:
We seem to be getting these errors every now and then, current workaround is just rerun the tests

pkg/config/config.go
```diff
@@ -24,11 +23,10 @@ type ConfigManager struct {
 }
 
 type Config struct {
-	IgnoredVulns                   []IgnoreEntry                        `toml:"IgnoredVulns"`
-	IgnoredPackageVersions         []IgnorePackageVersionEntry          `toml:"IgnoredPackageVersions"`
-	OverridePackageVersionLicenses []OverridePackageVersionLicenseEntry `toml:"OverridePackageVersionLicenses"`
-	LoadPath                       string                               `toml:"LoadPath"`
-	GoVersionOverride              string                               `toml:"GoVersionOverride"`
+	IgnoredVulns      []IgnoreEntry         `toml:"IgnoredVulns"`
+	PackageVersions   []PackageVersionEntry `toml:"PackageVersions"`
```
another-rex, line 27: Rename this to `PackageOverrides` and `PackageOverrideEntry`, since version is optional, and we are specifying packages for the scanner to override the interpretation of

another-rex (approved):
Minor variable naming nit, otherwise LGTM, thanks!

PR #957, opened 2024-05-01
chore: only extract versions from packages in the generator ecosystem

G-Rath:
Cherry-picked from https://github.com/G-Rath/osv-detector/pull/241

---

Currently the generators assume that all packages in an OSV are for their respective ecosystem which since they download ecosystem-specific databases is _mostly_ true, but there are a few OSVs that are for packages that exist across more than one ecosystem.

This has not been a problem up until now because either the versions in such OSVs happen to be compatible with native ecosystem version parser or we're skipping invalid versions for legacy reasons, but now GHSA-5844-q3fc-56rh exists which has versions that are invalid in Ruby.

### DIFF
--- a/scripts/generators/GenerateMavenVersions.java
+++ b/scripts/generators/GenerateMavenVersions.java
@@ -94,6 +94,10 @@ public class GenerateMavenVersions {
     osvs.forEach(osv -> osv.getJSONArray("affected").forEach(aff -> {
       JSONObject affected = (JSONObject) aff;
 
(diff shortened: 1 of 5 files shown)


G-Rath:
I have a feeling this actually might need to be adjusted to handle suffixing i.e. right now I think for Alpine it might end up skipping a bunch of osvs due to `Alpine:v3.x`; will keep this a draft until I've confirmed

hogo6002 (approved):

another-rex (approved):

PR #958, opened 2024-05-01
ci: setup workflow to run `semantic` tests weekly

G-Rath:
Cherry-pick of https://github.com/G-Rath/osv-detector/pull/182

---

This setups up a dedicated workflow for running the `semantic` test suite using the latest generated fixtures:
  - every sunday
  - whenever a generator or the workflow is changed
  - via a `workflow_dispatch`

This helps ensure high confidence in `semantic` across ecosystems; the workflow also exports the generated files as an artifact which makes it easy to update the committed version of each fixture without requiring the native ecosystem components (i.e. Ruby, Java, R, etc) to be installed.

Note currently this is using v3 of the artifact actions - I'll upgrade to v4 in a follow-up PR, since it will require further work due to them switching to using immutable artifacts.

At some point I'm also keen to explore this for the API side since the logic should be pretty much the same

### DIFF
--- a/.github/workflows/semantic.yml
+++ b/.github/workflows/semantic.yml
@@ -1,0 +1,191 @@
+name: Semantic
+
+on:
+  schedule:
+    - cron: "0 0 * * SUN"
+  workflow_dispatch:
+    inputs:
+      filterResults:
+        description: "What comparator results the generators should print"
+        required: false
+        default: "failures"
+        type: choice
+        options:
+          - all
+          - failures
+          - successes
+  pull_request:
+    paths:
+      - ".github/workflows/semantic.yml"
+      - "scripts/generators/**"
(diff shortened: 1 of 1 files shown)


G-Rath:
Ok so now we have support for Alpine which ideally we should run as part of this, but I found that can take _hours_ when running through docker - for implementing that I did end up compiling a custom binary from the actual C code which was _blazingly_ fast but I'm not sure if there's a way we can easily make that an option?

I'm thinking for now I'll just omit the Alpine generator, and make an issue for adding it + try to explore how we might be able make it faster (such as running the whole generator in a docker container)

G-Rath:
Currently broken due to #1055

another-rex (approved):

PR #909, opened 2024-04-10
GR: resolve tests & mock client

michaelkedar:
Some tests checking dependency resolution works as expected and correctly finds the vulnerable dependency chains.

Includes a new `MockResolutionClient` so I can test guided resolution on toy data without relying on the live OSV / deps.dev APIs.

PR #967, opened 2024-05-07
GR: Add test universe generation script and tests for patch generation

michaelkedar:
Made a go script that creates a universe file for the `MockResolutionClient` (#909) from the current real deps.dev & OSV data used during in-place / relock / relax computation.

This has let me add some real-world in-place & relax patch generation test cases to test for regression without it breaking constantly due to new versions. I've only added one universe because the files are pretty large.

The in-place/relax patch generation tests are basically end-to-end tests - it's not really written in a way that makes testing only the patch computation functionality possible. (This might be a code smell, but I don't think it's practical to structure the code in such a way).


oliverchang (approved):
this is awesome!

scripts/generate_mock_resolution_universe/main.go
```diff
@@ -0,0 +1,297 @@
...
+			reqs, err := cl.Requirements(context.Background(), resolve.VersionKey{
+				PackageKey:  pk,
+				Version:     v,
+				VersionType: resolve.Concrete,
+			})
+			if err != nil {
+				return clienttest.ResolutionUniverse{}, err
+			}
+			for _, r := range reqs {
+				// don't bother writing dev dependencies
```
cuixq, line 189: How about optional dependencies?

cuixq (approved):

scripts/generate_mock_resolution_universe/main.go
michaelkedar: npm tries to install optional dependencies if possible, so I'd like to keep them in the universe

PR #968, opened 2024-05-07
Disable all unimportant vulnerabilities

hogo6002:
osv.dev integrates Debian security tracker data into existing CVE entries, filtering out unimportant vulnerabilities.
image

### DIFF
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -701,6 +701,7 @@ func isUnimportant(affectedPackages []models.Affected) bool {
 			return true
 		}
 	}
+
 	return false
 }
 
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -671,7 +671,8 @@ func filterPackageVulns(r reporter.Reporter, pkgVulns models.PackageVulns, confi
 	var newVulns []models.Vulnerability
 	if len(newGroups) > 0 { // If there are no groups left then there would be no vulnerabilities.
 		for _, vuln := range pkgVulns.Vulnerabilities {
-			if isUnimportant(vuln.Affected) {
+			// Checks if a Debian vulnerability is unimportant. Debian ecosystems may be listed with a version number, such as 'Debian:10'.
+			if strings.HasPrefix(pkgVulns.Package.Ecosystem, string(models.EcosystemDebian)) && isUnimportant(vuln.Affected) {
 				continue
 			}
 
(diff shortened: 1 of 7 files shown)


pkg/osvscanner/osvscanner.go
```diff
@@ -671,7 +671,15 @@ func filterPackageVulns(r reporter.Reporter, pkgVulns models.PackageVulns, confi
 	var newVulns []models.Vulnerability
 	if len(newGroups) > 0 { // If there are no groups left then there would be no vulnerabilities.
 		for _, vuln := range pkgVulns.Vulnerabilities {
-			if _, filtered := ignoredVulns[vuln.ID]; !filtered {
+			unimportant := false
+			for _, affected := range vuln.Affected {
+				if affected.EcosystemSpecific["urgency"] == "unimportant" {
```
oliverchang, line 676: Can we add a check to only do this for Debian? And probably break this out into a separate function.

e.g.

```golang
func isUnimportant(ecosystemSpecific map[string]interface{}) bool {
}
```

pkg/osvscanner/osvscanner.go
```diff
@@ -684,6 +689,18 @@ func filterPackageVulns(r reporter.Reporter, pkgVulns models.PackageVulns, confi
 	return pkgVulns
 }
 
+// isUnimportant returns if a Debian vulnerability is unimportant
+// Details: https://security-team.debian.org/security_tracker.html#severity-levels
+func isUnimportant(affectedPackages []models.Affected) bool {
```
oliverchang, line 694: nit: better for the Debian check to live here too. This way we keep all the related logic in one place.

pkg/osvscanner/osvscanner.go
```diff
@@ -671,6 +671,10 @@ func filterPackageVulns(r reporter.Reporter, pkgVulns models.PackageVulns, confi
 	var newVulns []models.Vulnerability
 	if len(newGroups) > 0 { // If there are no groups left then there would be no vulnerabilities.
 		for _, vuln := range pkgVulns.Vulnerabilities {
+			if isUnimportant(pkgVulns.Package.Ecosystem, vuln.Affected) {
```
another-rex, line 674: Can you aggregate the number of ignored entries and log that number (similar tto how we log ignored entries above).

another-rex: Also, I'm not sure how common this will be, but if there're other items in the vuln group (I.e. aliased or share aliases), should we be ignoring all of them?

another-rex (commented):
Thanks!

oliverchang (approved):

PR #980, opened 2024-05-14
feat: support comparing Alpine versions locally

### LINKED ISSUE #952: Support comparing Alpine versions
Currently `semantic` does not know how to compare versions for Alpine packages, which is required for local/offline mode - this blocks #769

G-Rath:
This introduces support for comparing Alpine versions locally using the same logic as the `apk` package manager, along with a generator for generating fixtures.

There is a bit of fuzziness in the behaviour across different versions of `apk` - the `alpine:3.x` docker images all use `apk` v2.x, which is what the fixture generator uses too and at least `apk` v2.14 (which is used by `alpine:3.19`) and v2.10 pass; however the current latest upcoming version of `apk` technically fails on approximately 30 fixtures which I think is because it has fixed https://gitlab.alpinelinux.org/alpine/abuild/-/issues/10088.

Beyond that I was able to find a handful of other edge cases where the comparison results between these versions was different, but they all seemed to be primarily around the handling of invalid versions which are not expected to be present in OSV data anyway and they look to be the result of bugfixes meaning we'd need special "anti" handling to support in a way that ensures valid versions are still compared correctly, so I think it's good enough to ship.

Resolves #952


internal/semantic/version-alpine.go
```diff
@@ -0,0 +1,338 @@
...
+// parseAlpineNumberComponents parses the given string into AlpineVersion.components
+// and then returns the remainder of the string for continued parsing.
+//
+// Each number component is a sequence of digits (0-9), separated with a ".",
+// and with no limit on the value or amount of number components.
+//
+// This parser must be applied *before* any other parser.
+func parseAlpineNumberComponents(v *AlpineVersion, str string) string {
+	sub := cachedregexp.MustCompile(`^((\d+)\.?)*`).FindString(str)
+	str = strings.TrimPrefix(str, sub)
```
another-rex, line 225: nit: move this down directly to the return statement, it was a bit confusing to me up here as I thought the result was used in the loop.

internal/semantic/version-alpine.go
```diff
@@ -0,0 +1,338 @@
...
+func (v AlpineVersion) compareRemainder(w AlpineVersion) int {
+	if v.remainder == "" && w.remainder != "" {
+		return +1
+	}
+
+	if v.remainder != "" && w.remainder == "" {
+		return -1
+	}
+
+	return 0
```
another-rex, line 183: Should we do a strings.Compare here?

another-rex (commented):
Looks really good! Some minor comments

another-rex (approved):
LGTM, thanks!

PR #981, opened 2024-05-15
move `RequestUserAgent` to `internal/useragent`

cuixq:


internal/useragent/useragent.go
cuixq, line 1: Let me know how you feel about this new location.

spencerschrock: From an external perspective, this would prevent Scorecard from setting a user agent. If that's the direction the team wants to go, I'll close out our corresponding issue.
https://github.com/ossf/scorecard/issues/4029

cuixq: yes and I also noticed that this won't solve our problem of import cycle so I marked this PR to draft

Issue #35, opened 2022-12-12
Better support for transitive deps in Maven/Java (pom.xml) 


oliverchang:
Currently pom.xml support does not support resolving the full dependency graph.

brampat:
I can't find any documentation on the Maven integration for the scanner. Can anyone point me to it?

abhisek:
@oliverchang Can data from deps.dev be used to resolve transitive dependencies from direct dependencies identified from `pom.xml`? If so, no external tool/plugin will be required to get full list of dependencies.

h4sh5:
Could it have something to with an effective pom? 

https://maven.apache.org/plugins/maven-help-plugin/effective-pom-mojo.html

https://medium.com/@harun_ergul/what-is-effective-pom-5c278fe53031

h4sh5:
`mvn dependency:list` might also help (found in https://github.com/opensbom-generator/spdx-sbom-generator/search?q=mvn)

davidrabinowitz:
Also there's a need to configure maven profiles, as some project build different artifacts based on the profile, see https://github.com/GoogleCloudDataproc/spark-bigquery-connector for example

bestbeforetoday:
Based on the suggestion by @h4sh5 above, my current work-around is to run something like this:

```bash
mvn help:effective-pom -Doutput="${TMPDIR}/pom.xml"
osv-scanner --lockfile="${TMPDIR}/pom.xml"
```

oliverchang:
Thanks for the suggestions all! 

I think it may make sense to support a mode to use ecosystem-specifc tooling (e.g. `mvn`) to achieve this, but we're hoping to avoid adding additional dependnencies, and there are additional complexities around profiles/environments as pointed out in @davidrabinowitz. 

We're also exploring approaches using deps.dev as well to resolve this natively without requiring external tooling. We'll keep this bug updated with any new info!

bestbeforetoday:
I found that, since I really care about the dependencies (including transitive dependencies) that will be on the classpath at runtime, and required by any consuming packages, the approach that actually works for me is to generate a Software Bill of Materials (SBOM) and run OSV-Scanner against that SBOM instead of the POM file.

I currently have the [CycloneDX Maven plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin) configured in a profile within my POM file:
```xml
<profile>
  <id>sbom</id>
  <build>
    <plugins>
      <plugin>
        <groupId>org.cyclonedx</groupId>
[...]

PR #982, opened 2024-05-16
Add a new Maven pom.xml extractor

cuixq:
The new Maven lockfile extractor aims to resolve the full Maven dependency graph to provide better transitive support https://github.com/google/osv-scanner/issues/35. This is an experimental feature for now.

This PR uses deps.dev util package to parse Maven pom.xml, also calls deps.dev API for available versions when resolving a range requirement.


pkg/lockfile/parse-maven-lock.go
```diff
@@ -164,3 +168,101 @@ func init() {
...
+		v, _ := semver.Maven.Parse(ver.GetVersionKey().GetVersion())
+		if constraint.MatchVersion(v) && result.Compare(v) < 0 {
+			result = v
+		}
+	}
+
+	return result.String(), nil
+}
+
+func ParseMavenLock2(depsdev depsdevpb.InsightsClient, pathToLockfile string) ([]PackageDetails, error) {
```
cuixq, line 266: not sure if this should be renamed to `ParseMavenLockWithResolver?`

oliverchang: Yes! 

Though, while we're here, we may as well just rename this more accurately, and call this ParseMavenWithResolver. pom.xml isn't a lockfile.

pkg/lockfile/parse-maven-lock.go
```diff
@@ -164,3 +168,101 @@ func init() {
...
+		return "", fmt.Errorf("parsing Maven constraint %s: %w", dep.Version, err)
+	}
+	if constraint.IsSimple() {
+		// Return the constraint if it is a simple version string.
+		return constraint.String(), nil
+	}
+
+	// Otherwise return the greatest version matching the constraint.
+	// TODO: invoke Maven resolver to decide the exact version.
+	resp, err := e.Client.GetPackage(ctx, &depsdevpb.GetPackageRequest{
```
oliverchang, line 245: Question: do you think we'll ever run into cases where we need to parallelise requests to minimize latency?

oliverchang (approved):
LGTM with a question! Also, can you please update your PR title to reflect that this is a version of an extractor that's intended to resolve the full transitive graph? 

Also reference #35 in your PR description.

G-Rath:
fwiw I'm not sure that it makes sense to land this in its current form given that it's not being actually being used and violates the "offline" contract, which I don't think we can necessarily address with the current interface so landing it as-is might prematurely lock us into a public API we immediately want to change.

oliverchang:
> fwiw I'm not sure that it makes sense to land this in its current form given that it's not being actually being used and violates the "offline" contract, which I don't think we can necessarily address with the current interface so landing it as-is might prematurely lock us into a public API we immediately want to change.

That's a good point. Is there an easy way we can keep this interface private, while being able to use it ourselves? 

Alternatively, we can just add some clear documentation to the relevant functions to say: this is experimental and can break.

pkg/lockfile/parse-maven-lock.go
cuixq: I think this will be likely to happen - my plan is to make `pkg/depsdev` handle all requests (parallelised) to deps.dev.

oliverchang:
Following up on this, the right way to do this for now is probably:

- Move this to an "`resolverExtractor` or something similar under `internal`. 
- Modify https://github.com/google/osv-scanner/blob/055ef052bde5a8a259cd563f7d51bee1d5e2f4fb/pkg/osvscanner/osvscanner.go#L344 to prefer this internal resolverExtractor where it exists for a given lockfile? If `--offline` is passed, then we ignore these.

WDYT @cuixq @G-Rath  ?

cuixq:
For now, I would prefer keeping the new extractor in `internal` folder so we can experiment it without breaking anything. 

When it's ready enough, we can think about how to enable this with the `offline` mode, which could be something similar to the second option.

G-Rath (approved):

G-Rath:
fwiw if you want to enable experimental support for it without a lot of work, you should be able to add it as a custom parser [here](https://github.com/google/osv-scanner/blob/055ef052bde5a8a259cd563f7d51bee1d5e2f4fb/pkg/osvscanner/osvscanner.go#L354)

internal/manifest/helpers_test.go
```diff
@@ -0,0 +1,105 @@
...
+	if err == nil {
+		t.Errorf("Expected to get error, but did not")
+	}
+
+	if !errors.Is(err, expected) {
+		t.Errorf("Expected to get \"%v\" error but got \"%v\" instead", expected, err)
+	}
+}
+
+func packageToString(pkg lockfile.PackageDetails) string {
```
another-rex, line 38: nit: Add an example in the function comment of what the output looks like.

internal/manifest/maven.go
```diff
@@ -0,0 +1,118 @@
...
+		pkgDetails := lockfile.PackageDetails{
+			Name:      name,
+			Version:   v,
+			Ecosystem: lockfile.MavenEcosystem,
+			CompareAs: lockfile.MavenEcosystem,
+		}
+		if dep.Scope != "" {
+			pkgDetails.DepGroups = append(pkgDetails.DepGroups, string(dep.Scope))
+		}
+		details[name] = pkgDetails
```
another-rex, line 52: Is the dependency name always unique? e.g. could there be the same dependency at different versions (and recorded twice in the Dependencies array)

I'm guessing it's intentional that this is overwriting so that the last version is always the one resolved, can you add a comment a comment here clarifying this.

internal/manifest/maven.go
```diff
@@ -0,0 +1,118 @@
...
```
another-rex, line 71: Comment here as well

internal/manifest/maven.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return maps.Values(details), nil
+}
+
+func (e MavenResolverExtractor) resolveVersion(ctx context.Context, dep maven.Dependency) (string, error) {
+	constraint, err := semver.Maven.ParseConstraint(string(dep.Version))
+	if err != nil {
+		return "", fmt.Errorf("parsing Maven constraint %s: %w", dep.Version, err)
```
another-rex, line 80:
```suggestion
		return "", fmt.Errorf("failed parsing Maven constraint %s: %w", dep.Version, err)
```

internal/manifest/maven.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if constraint.IsSimple() {
+		// Return the constraint if it is a simple version string.
+		return constraint.String(), nil
+	}
+
+	// Otherwise return the greatest version matching the constraint.
+	// TODO: invoke Maven resolver to decide the exact version.
```
another-rex, line 88: Can you clarify the difference between what it's currently doing (querying deps.dev) vs invoking the Maven resolver?

another-rex (commented):
Thanks! mostly LGTM, added some comments

internal/manifest/maven.go
cuixq: I would say don't worry about this too much for now since the logic will be overwritten soon (including how managed dependencies would work).

I copied these code from the current Maven extractor is sort of PoC that deps.dev Maven parser should return the same results if we use the same logic.

another-rex (approved):

PR #986, opened 2024-05-20
Update snapshots

cuixq:
Currently tests are broken due to a new vulnerability in test fixtures. This PR aims to fix these failures by updating the snapshots

### DIFF
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -176,7 +176,7 @@ func TestRun(t *testing.T) {
 		{
 			name: "Scan locks-many",
 			args: []string{"", "./fixtures/locks-many"},
-			exit: 0,
+			exit: 1,
 		},
 		// all supported lockfiles in the directory should be checked
 		{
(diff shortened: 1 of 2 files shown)


cmd/osv-scanner/main_test.go
```diff
@@ -176,7 +176,7 @@ func TestRun(t *testing.T) {
 		{
 			name: "Scan locks-many",
 			args: []string{"", "./fixtures/locks-many"},
-			exit: 0,
+			exit: 1,
```
cuixq, line 179: @another-rex we may need to update the fixtures to make sure no vulnerabilities. will you be happy to help on this?

cmd/osv-scanner/main_test.go
```diff
@@ -576,12 +576,12 @@ func TestRun_Licenses(t *testing.T) {
 		{
 			name: "No vulnerabilities with license summary",
 			args: []string{"", "--experimental-licenses-summary", "./fixtures/locks-many"},
-			exit: 0,
+			exit: 1,
```
oliverchang, line 579: Should we just bump the version in the relevant lock files in this dir instead? otherwise we're no longer testing the same thing.

cuixq: `busybox` version `1.35.0-r29` has a vulnerability in the test snapshot which I am not sure if we want to keep this version for some testing purpose. 

also bumping the version `alpine.cdx.xml` includes changing `cpe`, `purl` and `properties`, so I mentioned @another-rex to see if he could help with preparing the new fixture.

cuixq:
this is addressed in https://github.com/google/osv-scanner/pull/990

PR #990, opened 2024-05-23
Fix snapshots and alpine version

another-rex:
This updates busybox to 1.36.1-r27 to resolve all current vulnerabilities, and then updates the snapshots to match.

This is a bit odd as 1.36.1-r27 doesn't actually exist on the distro this SBOM is created for (alpine 3.17) , where the highest version is 1.35.0-r30. However, as 3.17 is now out of support, no more fixes are being backported for 1.35.0. 

The *ideal(?)* behavior would not show the 3.19/3.20 vulnerabilities on 1.36.1 when scanning Alpine 3.17, but because of distro in purls still being undefined, all alpine advisories are returned. When this is eventually implemented, we should revert this PR.

### DIFF
--- a/cmd/osv-scanner/fixtures/sbom-insecure/alpine.cdx.xml
+++ b/cmd/osv-scanner/fixtures/sbom-insecure/alpine.cdx.xml
@@ -182,18 +182,18 @@
         <property name="syft:location:0:path">/bin/busybox</property>
       </properties>
     </component>
-    <component bom-ref="pkg:apk/alpine/busybox@1.35.0-r29?arch=x86_64&amp;upstream=busybox&amp;distro=alpine-3.17.2&amp;package-id=623d53216342d45e" type="library">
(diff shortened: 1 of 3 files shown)


oliverchang (approved):

cmd/osv-scanner/fixtures/sbom-insecure/alpine.cdx.xml
```diff
@@ -601,4 +601,4 @@
       </properties>
     </component>
   </components>
-</bom>
```
cuixq, line 604: is this intentional?

cuixq (approved):

cmd/osv-scanner/fixtures/sbom-insecure/alpine.cdx.xml
another-rex: I think the new line got added because of vscode formatting.

PR #996, opened 2024-05-27
Now that we have updated to go1.21.10, we can remove the ignore line from osv-scanner.toml

another-rex:
Now that we have updated to go1.21.10, we can remove the ignore line from osv-scanner.toml which was ignoring a vulnerability in go1.21.8


michaelkedar (approved):

cuixq (approved):
Can we remove this file?

another-rex:
I think we should keep the file to make adding rules in the future easier, and give an example of an osv-scanner.toml file. Though it's pretty fast to add the file when we need it as well.

PR #1000, opened 2024-05-28
Maven standard dependencies should take precedence over managed dependencies

cuixq:
Managed dependencies are not real dependencies so they should not take precedence over standard dependencies.

Dependency management is used to control the versions of artifacts used in transitive dependencies. https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management

Also, version requirements in managed dependencies are only referred when the requirement is not defined for that dependency in standard dependencies section.


G-Rath:
Can you just check this against what I believe is the section of the docs I referred to when I did this?

> * a and c both are declared as dependencies of the project so version 1.0 is used due to dependency mediation. Both also have runtime scope since it is directly specified.
> * b is defined in B's parent's dependency management section and since **dependency management takes precedence over dependency mediation for transitive dependencies**, version 1.0 will be selected should it be referenced in a or c's POM. b will also have compile scope.
> * Finally, since d is specified in B's dependency management section, should d be a dependency (or transitive dependency) of a or c, version 1.0 will be chosen - again **because dependency management takes precedence over dependency mediation** and also because the current POM's declaration takes precedence over its parent's declaration.

I assume my mistake is in not understanding what's being referred to is triggered in specific cases which are not true (most of the time at least) for the scanner when it's reading a pom, but would like to have that confirmed.

cuixq:
I think the dependency management in **root** pom.xml takes precedence over the dependencies in pom.xml of **transitive** dependencies.

However, since we are only scanning the root pom.xml with direct dependencies only, dependencies take precedence over the managed dependencies, and this is also what `mvn` does.

G-Rath (approved):

another-rex (approved):

PR #1002, opened 2024-05-29
Transitive dependency support for Maven pom.xml

cuixq:
Issue https://github.com/google/osv-scanner/issues/35

In this PR, the new Maven extractor invokes Maven resolver to compute the transitive dependencies of a Maven pom.xml.

Since managed dependencies are not actually being depended on, they are not in the resolved dependency graph, and thus they are not included in the scan results.


internal/manifest/fixtures/maven/with-scope.xml
```diff
@@ -1,10 +1,14 @@
...
+  <groupId>com.mycompany.app</groupId>
+  <artifactId>my-app</artifactId>
+  <version>1.0</version>
+
   <dependencies>
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
-      <scope>test</scope>
```
cuixq, line 7: @michaelkedar I think we should fix the resolver not to ignore `test` dependencies. `junit:junit` is not in the resolve graph. :(

another-rex (approved):
LGTM!

michaelkedar (approved):

PR #1004, opened 2024-05-29
✨ Adding a flag to stop execution after dependency parsing

marcwieserdev:
## About this PR

This PR adds a simple experimental flag to stops the execution of osv-scanner after having extracted dependencies from the lockfiles.

Adding this will let us use osv-scanner to extract dependencies from lockfiles, but continue to use our internal tools to detect vulnerabilties from them.

It is the first step for us to export packages in the CycloneDX format.

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -1593,6 +1593,42 @@ Warning: `scan` exists as both a subcommand of OSV-Scanner and as a file on the 
 
 ---
 
+[TestRun_SubCommands/scan_with_only_package_being_returned - 1]
+{
+  "results": [
+    {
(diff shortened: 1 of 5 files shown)


another-rex:
Thank you! I generally want to be careful about adding additional flags though. 

I think the `--experimental-offline` flag essentially does what you are proposing here (especially if you don't already have a local db downloaded (via `--experimental-local-db`)). The only problem is it will spam the stderr with database not found errors, which I opened  #1005 to fix.

Can you check if the offline flag fulfils your requirements?

marcwieserdev:
Thanks a lot for your insight ! I have tested it and it seems to do the job we want when we combine it with `--experimental-all-packages` and `--no-call-analysis=all`

I'm closing this PR then

Issue #769, opened 2024-01-25
Missing vulnerabilities for debian purls for --experimental-local-db


likhithn:
I tried running --format json --sbom bom.json --experimental-local-db ~/database and it missed all the purls that has 
pkg:deb/debian but I get hits for --format json --sbom bom.json

another-rex:
Good catch! It looks like our server side purl matching is case insensitive (seems to be correct according to the spec: https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst), but the local-db is not, which is a bug. 

@G-Rath This looks like another good starter issue for your team!

G-Rath:
@another-rex I think the issue is maybe a bit more complicated, and something I've be aware of for a while but not brought up since it didn't seem to be a high priority.

The OSV spec says for the Alpine ecosystem:

> The Alpine package ecosystem; the name is the name of the source package. The ecosystem string must have a :v<RELEASE-NUMBER> suffix to scope the package to a particular Alpine release branch (the v prefix is required). E.g. v3.16.

My interpretation of that is the ecosystem string _must always_ have a suffix, which seems to be the case for `ecosystem` fields but not the purl fields; though noting it could be that this is meant as an "if" kind of condition (i.e. "The ecosystem string must have a ... suffix _if_ the package is to be scoped to ...") - that doesn't fix the problem outright but would change what/where the fix could be.

Assuming my current interpretation is correct if we use CVE-2022-48174 as an example, which is returned by the API when scanning `fixtures/sbom-insecure/alpine.cdx.xml`, the [entry](https://api.osv.dev/v1/vulns/CVE-2022-48174) has the following `affected[0].package`:

```
"package": {
  "name": "busybox",
  "ecosystem": "Alpine:v3.18",
  "purl": "pkg:apk/alpine/busybox?arch=source"
},
```

and when we scan our test SBOM, we get the following PURL for the `busybox` package:

```
pkg:apk/alpine/busybox@1.35.0-r29?arch=x86_64&upstream=busybox&distro=alpine-3.17.2
```

When we're preparing the osv "query" for comparison, we [build package details](https://github.com/google/osv-scanner/blob/main/internal/local/check.go#L22-L45) which favors the `PURL` if present; and in that case we use whatever `ecosystem` we're given from `models.PURLToPackage`, which in this case is just returning `Alpine`.

[...]

PR #1006, opened 2024-05-30
Remove feature from changelog as it's still blocked on #769

another-rex:
Remove feature from changelog as it's still blocked on #769

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ### Features:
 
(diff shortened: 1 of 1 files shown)


michaelkedar (approved):

G-Rath:
strictly speaking I'm not sure it is blocked, it's just the scanner doesn't find alpine as easily - but e.g. if you manually pass in a JSON file with the alpine ecosystem I'm pretty sure it'll now work whereas before it'd error

PR #1007, opened 2024-05-30
fix: only care about ecosystem suffix if present in both ecosystems when determining equality

### LINKED ISSUE #769: Missing vulnerabilities for debian purls for --experimental-local-db

G-Rath:
This changes how we compare ecosystems so that the suffix is only considered when it is present for _both_ ecosystems being compared, since we can't reliably extract that.

Resolves #769

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -143,6 +143,36 @@ Loaded filter from: <rootdir>/fixtures/go-project/osv-scanner.toml
 
 ---
 
+[TestRun/PURL_SBOM_case_sensitivity_(api) - 1]
+Scanning dir ./fixtures/sbom-insecure/alpine.cdx.xml
+Scanned <rootdir>/fixtures/sbom-insecure/alpine.cdx.xml as CycloneDX SBOM and found 14 packages
+Loaded filter from: <rootdir>/fixtures/sbom-insecure/osv-scanner.toml
+CVE-2022-37434 has been filtered out because: This is a intentionally vulnerable test sbom
+Filtered 1 vulnerability from output
++--------------------------------+------+-----------+---------+-----------+---------------------------------------+
+| OSV URL                        | CVSS | ECOSYSTEM | PACKAGE | VERSION   | SOURCE                                |
++--------------------------------+------+-----------+---------+-----------+---------------------------------------+
+| https://osv.dev/CVE-2018-25032 | 7.5  | Alpine    | zlib    | 1.2.10-r0 | fixtures/sbom-insecure/alpine.cdx.xml |
++--------------------------------+------+-----------+---------+-----------+---------------------------------------+
+
+---
+
+[TestRun/PURL_SBOM_case_sensitivity_(api) - 2]
+
+---
+
+[TestRun/PURL_SBOM_case_sensitivity_(local) - 1]
+Scanning dir ./fixtures/sbom-insecure/alpine.cdx.xml
+Scanned <rootdir>/fixtures/sbom-insecure/alpine.cdx.xml as CycloneDX SBOM and found 14 packages
+Loaded Alpine local db from <tempdir>/osv-scanner/Alpine/all.zip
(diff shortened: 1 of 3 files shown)


cmd/osv-scanner/__snapshots__/main_test.snap
```diff
@@ -141,6 +141,43 @@ Loaded filter from: <rootdir>/fixtures/go-project/osv-scanner.toml
 
 ---
 
+[TestRun/PURL_SBOM_case_sensitivity_(api) - 1]
```
G-Rath, line 144: the fact that the offline output has vulnerabilities whereas the online one doesn't indicates there's a potential bug in the api's Alpine comparator 🤔

another-rex: Yes there is a bug in alpine version enumeration, just fixed this week in: https://github.com/google/osv.dev/pull/2241 but not in production yet.

G-Rath: Cool - I'd seen a few of those go around this/last week, so figured it might have been known 😅

another-rex (commented):
LGTM, thanks!. Going to wait till the next OSV.dev release (with the alpine enumeration fix) before merging this, so we'll have the correct snapshots.

G-Rath:
Turns out we're not accurately sorting the table output at least for local - I'll tackle that somewhere...

hogo6002 (approved):
LGTM

cuixq (approved):

G-Rath:
I cannot for the life of me replicate that sorting difference locally 😕

hogo6002:
> I cannot for the life of me replicate that sorting difference locally 😕

I tried running  `update_snaps` for this branch locally, and it did change the order, but it's still a little different from the [Ubuntu results](https://github.com/google/osv-scanner/actions/runs/9488640750/job/26148082254?pr=1007). 
image

G-Rath:
Yeah, it seems that every couple of hours the order changes - I'm not sure if its coinciding with something on the osv.dev side like work being landed or if it's just some clock-shift based randomness, but it seems to be "hours" based rather than "seconds" based...

hogo6002:
> Yeah, it seems that every couple of hours the order changes - I'm not sure if its coinciding with something on the osv.dev side like work being landed or if it's just some clock-shift based randomness, but it seems to be "hours" based rather than "seconds" based...

Yeah I see what you mean. I just tested again and the result shows different.

hogo6002:
I had a look with @michaelkedar  on this issue. We noticed that the result changes after the export cron job publishes a new `all.zip`. We think this is because the export runs in parallel (https://github.com/google/osv.dev/pull/2167), which may cause the order to be slightly different each time. To solve this, I guess we need to either sort the `all.zip` on the OSV.dev side or order all vulnerabilities alphabetically on the OSV-Scanner side. What do you think @another-rex

G-Rath:
oohh good find - fwiw locally I've been playing around with sorting the table rows within packages by their url though that got blocked by it being an `interface{}`; I've not gotten back to it yet so might be an easy fix

another-rex:
Hmm... is this behaviour new? All.zip is built concurrently, but that has always been the case if I understand correctly.

hogo6002:
> Hmm... is this behaviour new? All.zip is built concurrently, but that has always been the case if I understand correctly.

I don't think a new behaviour. Even though all.zip is built concurrently, the files are usually modified in alphabetical order, just a few are out of place. That's probably why most of our tests are passing. But this PR has a bunch of vulnerabilities with adjacency IDs (like 9840, 9841, 9842, and 9843), which is probably why it's failing. But I am not 100% sure if this is the issue as the modified order doesn't really align with the scanner output order.
image

another-rex:
I think this just needs a flag change in the test to match the new download-offline-db flag and should be good to merge.

PR #1008, opened 2024-05-30
Fix some Maven manifest & resolver issues

michaelkedar:
Some fixes for a few Maven resolver issues I've come across:
1. Requirement origins weren't being tracked correctly if a package key was set by a property. Fixed by checking dependency keys before and after interpolation, and updating the map if they changed. (I've modified one of the tests to check for this case)
2. To work around the resolver not resolving test or optional dependencies, made it so the pom.xml parser removes the test scope and optional flags.
3. `resolve.PackageKey` was not sufficient to uniquely key the requirements for the `Groups` map in the `Manifest`. Made a new `RequirementKey` type with ecosystem-specific information for both npm and maven to solve this.


internal/remediation/suggest/maven_test.go
```diff
@@ -98,7 +119,7 @@ func TestSuggest(t *testing.T) {
 					VersionType: resolve.Requirement,
 					Version:     "4.12",
 				},
-				Type: dep.NewType(dep.Test),
+				// Type: dep.NewType(dep.Test),
```
cuixq, line 122: maybe add a comment?

internal/resolution/manifest/maven.go
```diff
@@ -76,6 +91,56 @@ func (m MavenManifestIO) Read(df lockfile.DepFile) (Manifest, error) {
...
+		}
+
+		prevKeys := allKeys()
+		if err := project.Interpolate(); err != nil {
+			return err
+		}
+
+		newKeys := allKeys()
+		if len(prevKeys) != len(newKeys) {
+			return errors.New("number of dependencies changed after interpolation")
```
cuixq, line 131: the number of dependencies may get changed after interpolation if we are not able to resolve any placeholder, but this should be very rare...

internal/resolution/manifest/maven.go
```diff
@@ -238,6 +309,12 @@ func (m MavenManifestIO) MergeParents(ctx context.Context, result *maven.Project
 //   - identifier to locate the profile/plugin which is profile ID or plugin name
 //   - (optional) suffix indicates if this is a dependency management
 func makeRequirementVersion(dep maven.Dependency, origin string) resolve.RequirementVersion {
+	// Treat test & optional dependencies as regular dependencies to force the resolver to resolve them.
```
cuixq, line 312: I think this should be fine for now, but we may "fix" the resolver to handle `test` and `optional` dependencies if it is not that much work, and this would benefit transitive support when scanning pom.xml.

cuixq (approved):

internal/resolution/manifest/maven.go
michaelkedar: It's kind of hard to know which dependencies are which in those cases, so I'll leave it as an error for now.

Added a comment here.

PR #1011, opened 2024-05-31
Add go binary scanning

another-rex:
Add go binary scanning extractor, and use it in image scanning. 

This shows quite a few false positives that can be resolved with call analysis, which will be implemented in a followup PR.


pkg/lockfile/go-binary.go
```diff
@@ -0,0 +1,76 @@
...
+		// Assume if a file has an extension (that's not exe), it is not a go binary
+		// This also filters out hidden files on Unix
+		// This is a heuristic to improve performance and can result in false negatives
+		// TODO(another-rex): When we have access to the full FS interface, we can open and check
+		// magic bytes to be more accurate
+		return false
+	}
+
+	// Any other path can be a go binary
+	return true
```
oliverchang, line 32: Does this incur much performance penalty when scanning a container filesystem?

oliverchang (approved):

pkg/lockfile/go-binary.go
another-rex: This sort of depends on whether the DepFile passed into this extractor has the ReaderAt interface. If it does, then the performance cost is relatively minimal, as it checks the headers to see if it's a relevant file (e.g. has ELF header, various magic bytes, has go debug info...etc) and returns early if it isn't.

oliverchang: Ack, I think it's reasonable to expect ReaderAt in most cases.

PR #1014, opened 2024-05-31
✨  Adding CycloneDX 1.4 and 1.5 reporter

marcwieserdev:
## About this PR

This PR aims to add the CycloneDX 1.4 and 1.5 format. The idea behind it is to be able to export everything in a commonly used format to later reuse the result to enrich it. (ex: re-running a vulnerability matching process on the packages to find emerging vulnerabilities).

This PR adds 2 versions of CycloneDX even though at the moment they report the exact same thing. 

In future PR, we intend to extract where packages has been found during the extraction process (file, line, column), which is an information only supported by CycloneDX 1.5.

We need a CycloneDX 1.4 format for backward compatibility reasons.

## Additional informations

You may want to review this PR commit by commit, as it should be easier to read


internal/utility/purl/composer_extractor.go
```diff
@@ -0,0 +1,21 @@
+package purl
+
+import (
+	"log"
+	"strings"
+
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+func ExtractPURLFromComposer(packageInfo models.PackageInfo) (namespace string, name string, ok bool) {
```
oliverchang, line 10: nit: `Extract` is a bit overloaded for us -- we already use it in the context of "extracting" dependencies from e.g. a lockfile. 

Could we just omit the `extract` bit in this directory? And further following [Go conventions to avoid repetition](https://google.github.io/styleguide/go/best-practices#avoid-repetition), could we just call these functions `FromComposer`, `FromMaven` etc? Then users will just be calling `purl.FromComposer`, `purl.FromMaven`.

internal/utility/purl/composer_extractor.go
```diff
@@ -0,0 +1,21 @@
...
+	nameParts := strings.Split(packageInfo.Name, "/")
+	if len(nameParts) != 2 {
+		log.Printf("invalid packagist package_name=%s", packageInfo.Name)
```
oliverchang, line 13: can we return an error instead for all of these? we can replace the `ok bool` with an `err error`.

internal/utility/purl/golang_extractor.go
```diff
@@ -0,0 +1,25 @@
...
+	"log"
+	"strings"
+
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+func ExtractPURLFromGolang(packageInfo models.PackageInfo) (namespace string, name string, ok bool) {
+	nameParts := strings.Split(packageInfo.Name, "/")
+	if len(nameParts) == 0 || len(packageInfo.Name) == 0 {
+		log.Printf("invalid golang package_name=%s", packageInfo.Name)
```
oliverchang, line 13: same comment here generally: can we return errors instead of just printing and exiting?

internal/utility/purl/golang_extractor.go
```diff
@@ -0,0 +1,25 @@
+package purl
+
+import (
+	"log"
+	"strings"
+
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+func ExtractPURLFromGolang(packageInfo models.PackageInfo) (namespace string, name string, ok bool) {
```
oliverchang, line 10: nit: can we just call this `FromGo`?

pkg/grouper/grouper_test.go
```diff
@@ -1,83 +1,82 @@
-package grouper_test
+package grouper
```
oliverchang, line 1: Are these changes necessary for this PR?

pkg/grouper/package_grouper.go
```diff
@@ -0,0 +1,45 @@
+package grouper
+
+import (
+	"github.com/google/osv-scanner/internal/utility/purl"
+	"github.com/google/osv-scanner/pkg/models"
+	"slices"
+)
+
+func GroupByPURL(packageSources []models.PackageSource) map[string]models.PackageVulns {
```
oliverchang, line 9: Can you add some function level comments explaining what this function is doing?

pkg/osvscanner/vulnerability_result.go
```diff
@@ -110,5 +110,6 @@ func buildVulnerabilityResults(
 		results.ExperimentalAnalysisConfig.Licenses.Allowlist = allowlist
 	}
 
+	results.ResultsByPURL = grouper.GroupByPURL(results.Results)
```
oliverchang, line 113: Can we avoid storing this, and just compute these on demand when this is needed?

pkg/grouper/package_grouper.go
oliverchang: Also side comment: `grouper` is likely not the right package for this.. @another-rex WDYT?

oliverchang (commented):
thank you very much for contributing this! we'd love to support this feature in OSV-Scanner. I just have some comments.

pkg/grouper/package_grouper.go
marcwieserdev: Thanks for the review ! :) 
What package would you recommend ? (I thought of the `internal/utility/purl` package and then renaming the method to `Group`)

oliverchang: thanks for addressing the comments! yep `internal/utility/purl` and renaming the method to `Group` sounds great to me!

marcwieserdev: I'm on holiday right know, I will do it right when I come back (later this week)

go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/package-url/packageurl-go v0.1.3
 	github.com/pandatix/go-cvss v0.6.2
 	github.com/spdx/tools-golang v0.5.4
+	github.com/stretchr/testify v1.9.0
```
another-rex, line 29: These 3 new dependencies seem to be only used for one assert, can it be done manually instead?

another-rex (commented):
Thanks! I haven't had time to look into this recently, but I'll take a deeper look this week.

pkg/grouper/package_grouper.go
marcwieserdev: I moved the package there, I'm resolving the conversation, feel free to open it back if anything looks off ;)

pkg/reporter/sbom/models.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+type CycloneDXVersion int
+
+const (
+	CycloneDXVersion14 CycloneDXVersion = iota
+	CycloneDXVersion15
+)
+
+var SpecVersionToBomCreator = map[CycloneDXVersion]BomCreator{
```
another-rex, line 16: We will wish to add SPDX support in the future as well, can you rename this to CycloneDXBomCreator?

another-rex (commented):
Some minor comments, otherwise LGTM!

another-rex (approved):
Thanks! Will aim to get this merged in next week.

Issue #531, opened 2023-09-12
Better support for project inheritance in Maven/Java (pom.xml)


cuixq:
Currently pom.xml support does not support merging data from other projects.
However, there might be dependencies and properties defined in parent projects. 

Context: https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Inheritance

oliverchang:
Related: #35

PR #1019, opened 2024-06-03
Merge parent projects for Maven pom.xml

cuixq:
https://github.com/google/osv-scanner/issues/531

This PR merges parent pom.xml by paring locally or fetching from upstream (Maven Central for now).

When merging a parent pom.xml, only `pom` packaging is allowed. Once we fetch a parent from upstream, parsing from local is no longer allowed.

The project is also interpolated to get rid of properties, and dependencies are also processed (dedup and import).

### DIFF
--- a/internal/manifest/maven.go
+++ b/internal/manifest/maven.go
@@ -3,7 +3,9 @@ package manifest
 import (
 	"context"
 	"encoding/xml"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"deps.dev/util/maven"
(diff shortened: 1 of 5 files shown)


michaelkedar (approved):

internal/manifest/maven.go
```diff
@@ -95,12 +113,73 @@ func (e MavenResolverExtractor) Extract(f lockfile.DepFile) ([]lockfile.PackageD
...
+// start indicates the index of the current parent project, which is used to
+// check if the packaging has to be `pom`.
+// allowLocal indicates whether parsing local parent pom.xml is allowed.
+// path holds the path to the current pom.xml, which is used to compute the
+// relative path of parent.
+func (e MavenResolverExtractor) mergeParents(ctx context.Context, result *maven.Project, current maven.Parent, start int, allowLocal bool, path string) error {
+	currentPath := path
+	visited := make(map[maven.ProjectKey]bool, MaxParent)
+	for n := start; n < MaxParent; n++ {
+		if current.GroupID == "" || current.ArtifactID == "" || current.Version == "" {
```
another-rex, line 132: I'm guessing this will only happen when parent is empty?

another-rex (approved):
LGTM!

internal/manifest/maven.go
cuixq: yes, but we observed "bad" pom.xml that one or two of the three fields are empty and we should avoid fetching for these fields.

PR #1025, opened 2024-06-05
Start on override strategy for maven guided remediation

michaelkedar:
Very much a WIP.

Start on a new 'override' remediation strategy for guided remediation for maven, intending to remediate by forcing the use of non-vulnerable versions by `dependencyManagement`.

This is just the logic to identify possible override patches by searching for the first non-vulnerable version of an affected package.

Things I still need to do/think about:
- Generate a proper Maven ManifestPatch, depending on how the patch needs to be applied (e.g. override direct dependency, override existing dependencyManagement, create new dependencyManagement, update properties?)
- Re-resolving with the proposed changes, to check if new vulns are introduced (and try also fix those)
- How to structure program to use override for maven and relock for npm.

### DIFF
--- a/internal/remediation/override.go
+++ b/internal/remediation/override.go
@@ -1,0 +1,187 @@
+package remediation
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+	"slices"
+	"strings"
+
+	"deps.dev/util/resolve"
+	"deps.dev/util/semver"
+	"github.com/google/osv-scanner/internal/resolution"
+	"github.com/google/osv-scanner/internal/resolution/client"
+	"github.com/google/osv-scanner/internal/resolution/util"
+	"github.com/google/osv-scanner/internal/utility/vulns"
+)
+
+// TODO: need to make a ManifestPatch with ecosystem-specific fields
+type OverridePatch struct {
+	resolve.PackageKey
+	OrigVersion   string
+	NewVersion    string
+	ResolvedVulns []resolution.ResolutionVuln
+}
+
+func (p OverridePatch) String() string {
+	vulns := make([]string, len(p.ResolvedVulns))
+	for i, v := range p.ResolvedVulns {
+		vulns[i] = v.Vulnerability.ID
+	}
+
+	return fmt.Sprintf("%s@%s -> %s %v", p.Name, p.OrigVersion, p.NewVersion, vulns)
+}
+
+type OverrideUnfixable struct {
+	resolve.VersionKey
+	resolution.ResolutionVuln
+}
+
+func (u OverrideUnfixable) String() string {
+	return fmt.Sprintf("%s@%s [%s]", u.Name, u.Version, u.Vulnerability.ID)
+}
+
+type OverrideResult struct {
+	Patches   []OverridePatch
+	Unfixable []OverrideUnfixable
+}
+
+func (r OverrideResult) String() string {
+	s := &strings.Builder{}
+	fmt.Fprintln(s, "PATCHES:")
+	for _, p := range r.Patches {
+		fmt.Fprintln(s, p)
+	}
+
+	fmt.Fprintln(s, "UNFIXABLE:")
+	for _, unf := range r.Unfixable {
+		fmt.Fprintln(s, unf)
+	}
+
+	return s.String()
+}
+
+func ComputeOverridePatches(ctx context.Context, cl client.ResolutionClient, result *resolution.ResolutionResult, opts RemediationOptions) (OverrideResult, error) {
+	// Filter the original result just in case it hasn't been already
+	result.FilterVulns(opts.MatchVuln)
+
+	// Find the vulns affecting each version key to count vulns as the scan action does.
+	// TODO: Make ResolutionResult do this kind of thing.
+	vkVulns := make(map[resolve.VersionKey][]*resolution.ResolutionVuln)
+	for i, v := range result.Vulns {
+		seenVks := make(map[resolve.VersionKey]struct{})
+		for _, c := range v.ProblemChains {
+			vk, _ := c.End()
+			if _, seen := seenVks[vk]; !seen {
+				vkVulns[vk] = append(vkVulns[vk], &result.Vulns[i])
+				seenVks[vk] = struct{}{}
+			}
+		}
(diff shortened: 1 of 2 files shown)


internal/remediation/override.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/google/osv-scanner/internal/utility/vulns"
+)
+
+type overridePatch struct {
+	resolve.PackageKey
+	OrigVersion string
+	NewVersion  string
+}
+
+func ComputeOverridePatches(ctx context.Context, cl client.ResolutionClient, result *resolution.ResolutionResult, opts RemediationOptions) ([]resolution.ResolutionDiff, error) {
```
oliverchang, line 24: Can you add some documentation to all these functions?

internal/remediation/override.go
```diff
@@ -0,0 +1,279 @@
...
+	// returns errOverrideImpossible if 0 vulns are patchable.
+	var effectivePatches []overridePatch
+	for {
+		// Find the relevant vulns affecting each version key.
+		vkVulns := make(map[resolve.VersionKey][]*resolution.ResolutionVuln)
+		for i, v := range result.Vulns {
+			if !slices.Contains(vulnIDs, v.Vulnerability.ID) {
+				continue
+			}
+			seenVks := make(map[resolve.VersionKey]struct{})
```
oliverchang, line 123: nit: seenVKs

internal/remediation/override.go
```diff
@@ -0,0 +1,279 @@
...
+			for _, c := range v.ProblemChains {
```
oliverchang, line 124: Can you add some high level comments on the logic here?

internal/remediation/override.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	if len(effectivePatches) == 0 {
+		return nil, nil, errOverrideImpossible
+	}
+
+	return result, effectivePatches, nil
+}
+
+func patchManifest(patches []overridePatch, m manifest.Manifest) (manifest.Manifest, error) {
```
oliverchang, line 238: @cuixq is probably best for reviewing this.

internal/remediation/override.go
```diff
@@ -0,0 +1,279 @@
...
+		for vk, vulnerabilities := range vkVulns {
+			// Consider vulns affecting packages we don't want to change unfixable
+			if slices.Contains(opts.AvoidPkgs, vk.Name) {
+				continue
+			}
+
+			sys := vk.Semver()
+			// Get & sort all the valid versions of this package
+			// TODO: (Maven) skip unlisted versions and versions on other registries
+			versions, err := cl.Versions(ctx, vk.PackageKey)
```
oliverchang, line 156: Can we break up this function into smaller functions? e.g. getSortedPackageVersions()

Having a lot of this logic inline makes the logic a bit hard to follow.

oliverchang (commented):
looks like a good start overall, just some general comments.

internal/remediation/override.go
michaelkedar: Split `getVersionsGreater()` out, still considering other parts to split out.

internal/resolution/manifest/maven.go
```diff
@@ -189,17 +189,20 @@ func (m MavenManifestIO) Read(df lockfile.DepFile) (Manifest, error) {
...
 				otherRequirements = append(otherRequirements, reqVer)
-			} else {
-				requirements = append(requirements, reqVer)
 			}
-			if dep.Scope != "" {
+			origin, ok := reqVer.Type.GetAttr(dep.MavenDependencyOrigin)
+			if ok && (strings.HasPrefix(origin, "parent@") && strings.HasSuffix(origin, "@management")) {
+				reqVer.Type.AddAttr(dep.MavenDependencyOrigin, "management")
+			}
+			requirements = append(requirements, reqVer)
```
michaelkedar, line 202: This is a work-around to get the parent's `dependencyManagement` to be resolved properly.
I'd rather re-think the Maven manifest representation a bit instead of doing this.

scripts/generate_mock_resolution_universe/main.go
```diff
@@ -183,34 +214,17 @@ func makeUniverse(cl *client.DepsDevClient) (clienttest.ResolutionUniverse, erro
 				VersionType: resolve.Concrete,
 			})
 			if err != nil {
-				return clienttest.ResolutionUniverse{}, err
+				continue
 			}
 			for _, r := range reqs {
 				// don't bother writing dev dependencies
```
cuixq, line 220: also test dependencies

internal/remediation/override.go
```diff
@@ -0,0 +1,310 @@
...
+	return versions[offset:], nil
+}
+
+// patchManifest applies the overridePatches to the manifest in-memory. Returns a copy of the manifest that has been patched.
+func patchManifest(patches []overridePatch, m manifest.Manifest) (manifest.Manifest, error) {
+	if m.System() != resolve.Maven {
+		return manifest.Manifest{}, errors.New("unsupported ecosystem")
+	}
+
+	// TODO: may need special handling for the artifact's type and classifier
```
cuixq, line 274: what is this TODO about?

internal/remediation/override.go
```diff
@@ -0,0 +1,310 @@
...
+		found := false
+		i := 0
+		for _, r := range patched.Requirements {
+			if r.PackageKey != p.PackageKey {
+				patched.Requirements[i] = r
+				i++
+
+				continue
+			}
+			if origin, hasOrigin := r.Type.GetAttr(dep.MavenDependencyOrigin); !hasOrigin || origin == "management" {
```
cuixq, line 288: use `manifest.OriginManagement`?

cuixq (commented):
LGTM!

internal/remediation/override.go
michaelkedar: Updated the comment.
Basically, this currently doesn't look at or set the `type` or `classifier` fields, so it fails to actually create overrides for packages that have them

Thinking about it now, this might have the possibility of creating an infinite loop - I'll fix this in another PR.

cuixq: sounds good

cuixq (approved):

PR #1027, opened 2024-06-05
Dependency imports should always be fetched from upstream

cuixq:
There are two places that we call `mergeParents()`:
 - Merging data from parent pom.xml files
 - Importing dependency management from another project
 
In `mergeParents()`, we first check if `relativePath` is defined to know if we can parse parent locally.
However, this only applies for the first case but not for importing dependency management.
Also, once we start fetching parent pom.xml from upstream, we should no longer parse locally. 

This PR adds `allowLocal` to `mergeParents()` to specify if we allow parsing local parent pom.xml, and once a parent is fetched from upstream, `allowLocal` is set to false.

### DIFF
--- a/internal/resolution/manifest/maven.go
+++ b/internal/resolution/manifest/maven.go
@@ -215,7 +215,7 @@ func (m MavenManifestIO) mergeParents(ctx context.Context, result *maven.Project
 			}
 		} else {
 			// Once we fetch a parent pom.xml from upstream, we should not allow
-			// paring parent pom.xml locally anymore.
+			// parsing parent pom.xml locally anymore.
 			allowLocal = false
 
 			var err error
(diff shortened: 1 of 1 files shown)


internal/resolution/manifest/maven.go
```diff
@@ -214,6 +214,10 @@ func (m MavenManifestIO) MergeParents(ctx context.Context, result *maven.Project
 				return fmt.Errorf("failed to unmarshal project: %w", err)
 			}
 		} else {
+			// Once we fetch a parent pom.xml from upstream, we should not allow
+			// paring parent pom.xml locally anymore.
```
another-rex, line 218:
```suggestion
			// parsing parent pom.xml locally anymore.
```

another-rex (approved):
LGTM

PR #1028, opened 2024-06-05
Invoke `MavenResolverExtrator` when scanning pom.xml

cuixq:
https://github.com/google/osv-scanner/issues/35

In this PR, `MavenResolverExtrator` is invoked when scanning pom.xml to report vulnerabilities in transitive dependencies.
However, the default Maven extractor is still being used with offline mode.


internal/resolution/datasource/maven_registry.go
cuixq, line 1: this is to avoid import cycle :(

oliverchang: I think this makes sense also. the OSV-Scanner ErrAPIFailed seems like it should be kept just to OSV-specific API calls.

pkg/osvscanner/osvscanner.go
```diff
@@ -399,6 +406,33 @@ func scanLockfile(r reporter.Reporter, path string, parseAs string) ([]scannedPa
...
+		return lockfile.Lockfile{}, err
+	}
+	extractor := manifest.MavenResolverExtractor{
+		DependencyClient: depClient,
+	}
+	packages, err := extractor.Extract(f)
+	if err != nil {
+		err = fmt.Errorf("failed extracting %s: %w", f.Path(), err)
+	}
+	sort.Slice(packages, func(i, j int) bool {
```
oliverchang, line 421: I assume this sorting is just for test consistency? If so, let's add a comment saying so.

pkg/osvscanner/osvscanner.go
```diff
@@ -790,7 +824,7 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
 			r.Errorf("Failed to resolved path with error %s\n", err)
 			return models.VulnerabilityResults{}, err
 		}
-		pkgs, err := scanLockfile(r, lockfilePath, parseAs)
+		pkgs, err := scanLockfile(r, lockfilePath, parseAs, !actions.CompareLocally)
```
oliverchang, line 827: nit: it's a bit confusing that scanLockfile takes the inverse of actions.CompareLocally, while scanDir doesn't.

For consistency, it might be better for both to just accept actions.CompareLocally directly, and handle the logic of inverting that within scanLockfile?

oliverchang (commented):
AWESOME!

pkg/osvscanner/osvscanner.go
```diff
@@ -103,7 +106,7 @@ const (
 //   - Any lockfiles with scanLockfile
 //   - Any SBOM files with scanSBOMFile
 //   - Any git repositories with scanGit
-func scanDir(r reporter.Reporter, dir string, skipGit bool, recursive bool, useGitIgnore bool, compareOffline bool) ([]scannedPackage, error) {
+func scanDir(r reporter.Reporter, dir string, skipGit bool, recursive bool, useGitIgnore bool, compareLocally bool) ([]scannedPackage, error) {
```
G-Rath, line 109: I don't think this new name is accurate - comparing offline is a very different concern from comparing locally; while that different doesn't matter too much in this function, I think for the sake of consistency it's better to keep this as `compareOffline`

pkg/osvscanner/osvscanner.go
```diff
@@ -341,7 +344,7 @@ func scanImage(r reporter.Reporter, path string) ([]scannedPackage, error) {
 
 // scanLockfile will load, identify, and parse the lockfile path passed in, and add the dependencies specified
 // within to `query`
-func scanLockfile(r reporter.Reporter, path string, parseAs string) ([]scannedPackage, error) {
+func scanLockfile(r reporter.Reporter, path string, parseAs string, allowResolver bool) ([]scannedPackage, error) {
```
G-Rath, line 347: as above, I think this should be more "offline" themed as that is the important constraint - the current name reflects an implementation; people don't care about if resolvers are allowed, they care if any online actions will be taken

pkg/osvscanner/osvscanner.go
G-Rath: This also should be checking `actions.CompareOffline` as that's different from comparing locally, and I don't think there's a reason why the new resolver couldn't be used if you're comparing locally.

pkg/osvscanner/osvscanner.go
```diff
@@ -815,7 +849,7 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
 
 	for _, dir := range actions.DirectoryPaths {
 		r.Infof("Scanning dir %s\n", dir)
-		pkgs, err := scanDir(r, dir, actions.SkipGit, actions.Recursive, !actions.NoIgnore, actions.CompareOffline)
+		pkgs, err := scanDir(r, dir, actions.SkipGit, actions.Recursive, !actions.NoIgnore, actions.CompareLocally)
```
G-Rath, line 852: This should not be changed, as comparing locally is different from comparing offline - only the latter requires the former

G-Rath (changes requested):

G-Rath:
sorry, after posting my review and seeing all my comments together it's very clear they're all just a single comment:

> comparing locally is different from comparing offline - only the latter requires the former; we should continue checking if we're offline, not if we're comparing locally

pkg/osvscanner/osvscanner.go
cuixq: transitive dependency resolution requires network access for two purposes:
 - to generate effective pom.xml: this may require fetching other pom.xml files from upstream
 - to call deps.dev resolver: the resolver sends requests to deps.dev API for requirements 

so I don't think you can use the resolver locally or offline.

pkg/osvscanner/osvscanner.go
cuixq: could you explain more about the difference?

I had a discussion with @another-rex and we agree that both indicates users not willing to send package information to API (which is required by dependency resolution). 

if `CompareOffline` is true, `CompareLocally` is also true, but not vice versa.

If we check `CompareLocally`, and it is false, both `CompareOffline` and `CompareLocally` are false. 
If we only check `CompareOffline`, when `CompareOffline` is false there might be chance that `CompareLocally` still true, but package information will still be sent to deps.dev API and I think that is not what a user wants.

I have to say I also felt confused, so that's why I use `allowResolver` first and check if both `CompareOffline` and `CompareLocally` false.

pkg/osvscanner/osvscanner.go
cuixq: replied in another comment.

pkg/osvscanner/osvscanner.go
another-rex: +1 to Xueqin's comment, that's also my understanding. I believe the only usage of compareLocally here in scanDir is for the vendored directories scanning, which should be turned off when compareLocally is on for 2 reasons:

- It sends granular information (hashes of all files)
- The results are very likely to be affected git commits, which we don't scan in local mode anyway, so there's no point to the scan.

G-Rath: I don't disagree with the thinking, but imo it doesn't match with the name - really we should change it to be "compare offline" and "download offline databases", because that's why we have the two flags, and that'd resolve this confusion

cuixq:
@another-rex @G-Rath PTAL- thanks!

oliverchang (approved):
LGTM, but are your concerns addressed @G-Rath ?

G-Rath (approved):

another-rex (approved):
LGTM, though the test packages seem to have a lot of vulnerabilities though, can we pick one with only a few vulns?

cuixq:
I updated the fixtures with the example in the blog post.

PR #1036, opened 2024-06-11
test: update snapshots and exit codes

G-Rath:
Someone that knows these tests better should double check that these changes in exit codes mean they're no longer covering what they were meant to

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -226,7 +226,12 @@ Scanned <rootdir>/fixtures/locks-many/yarn.lock file and found 1 package
 Loaded filter from: <rootdir>/fixtures/locks-many/osv-scanner.toml
 GHSA-whgm-jr23-g3j9 and 1 alias have been filtered out because: Test manifest file
 Filtered 1 vulnerability from output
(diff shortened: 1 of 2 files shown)


cuixq:
I think we should update [cmd/osv-scanner/fixtures/locks-many/alpine.cdx.xml](https://github.com/google/osv-scanner/blob/main/cmd/osv-scanner/fixtures/locks-many/alpine.cdx.xml)

G-Rath:
Is that something you're ok to pick up? I don't actually know how to generate an sbom 😅 (everytime I try, the CLIs confuse me)

cuixq:
@another-rex could you help to update the fixture?

cuixq:
ideally, I would like to run the tests against some mock database, so we don't need to update the fixtures all the time.

G-Rath:
eh that would be nice but then require a lot of different work - something I've been thinking of exploring which could be a useful middleground is creating a workflow that runs say daily to automatically open a PR with snapshot updates.

Another alternative is the osv schema itself could define some kind of "test" ecosystem and then the api could have a bunch of osvs using that ecosystem, with the idea being you can officially represent advisories of different structures for the purposes of testing so they're a lot more controlled

PR #1039, opened 2024-06-12
Add `experimental-download-offline-databases` flag

cuixq:
Currently flags `experimental-offline` and `experimental-local-db` are confusing sometimes.

This PR renames `experimental-local-db` to `experimental-download-database` to make it more explicit whether to download the database or not.

For now, `experimental-download-database` only works when `experimental-offline` is set. 

`internal/local` is also modified to reflect the change in the naming and meaning of this flag.


docs/offline-mode.md
```diff
@@ -59,17 +59,17 @@ The offline database flag `--experimental-offline` causes OSV-Scanner to scan yo
...
 
-## Local database option
+## Download database option
 
-The local database flag `--experimental-local-db` causes OSV-Scanner to download or update your local database and then scan your project against it.
+The download database flag `--experimental-download-database` causes OSV-Scanner to download or update your local database and then scan your project against it. This option only works when you also set the offline flag.
 
 ```bash
-osv-scanner --experimental-local-db ./path/to/your/dir
+osv-scanner --experimental-offline --experimental-local-db ./path/to/your/dir
```
andrewpollock, line 67: Should the `--experimental-download-database` flag be getting included here?

andrewpollock (commented):
This is cool! 

I'm often downloading one or a few ecosystems for offline validation or other examination purposes, and have this shell script I had Gemini cook up for me to make this streamlined.

Is the intention for this feature to be used independently of a scanning run, or as part of one? 

Would it be possible to include another flag that accepted a subset of ecosystems to download, in the interests of saving time, bandwidth and disk space for this purpose?

cuixq:
For now, this flag is only valid when `experimental-offline` is set. The motivation is that `experimental-offline` and `experiment-local-db` is confusing. I think ideally, we would like to see some subcommand for example `download` to handle downloading the database, and I like the idea to specify the systems to download in the command!

internal/local/zip.go
```diff
@@ -27,8 +27,8 @@ type ZipDB struct {
 	Name string
 	// the url that the zip archive was downloaded from
 	ArchiveURL string
-	// whether this database should make any network requests
-	Offline bool
```
G-Rath, line 31: fwiw I think it would be better to keep the existing structure within the database logical, as it's technically possible they could be implemented to make network requests before downloading and (more practically) it means a smaller diff + aligns better with everything else (e.g. the error is `ErrOfflineDatabaseNotFound`, the tests say "if the db is offline", etc).

The fact that we're exposing this to the user as "download databases" is something of an implementation detail of the CLI.

Having said that, I won't block if you can't be bothered reverting 🙂

pkg/osvscanner/osvscanner.go
```diff
@@ -753,17 +753,17 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
-	if actions.CompareLocally {
 		actions.SkipGit = true
 
 		if len(actions.ScanLicensesAllowlist) > 0 || actions.ScanLicensesSummary {
 			return models.VulnerabilityResults{}, errors.New("cannot retrieve licenses locally")
 		}
 	}
 
+	if !actions.CompareOffline && actions.DownloadDatabase {
+		return models.VulnerabilityResults{}, errors.New("database is only downloaded under offline mode")
```
G-Rath, line 764: I think something like this reads better:

```suggestion
		return models.VulnerabilityResults{}, errors.New("databases can only be downloaded when running in offline mode")
```

G-Rath (changes requested):
This looks good, but the flag should be plural because there are many databases and really the number of them is an implementation detail (right now its 1:1 with ecosystems, but maybe in future they'll be sharded differently 🤷)

I also wonder if the flag should be something like `--download-offline-databases`, to tie it more to the `--offline` flag 🤔

G-Rath:
> Is the intention for this feature to be used independently of a scanning run, or as part of one?
> Would it be possible to include another flag that accepted a subset of ecosystems to download

@andrewpollock I proposed basically this exact thing in subcommand form at our weekly catchup 😄

My thinking is especially now that we have both subcommands and interactive-ness, we could provide a very nice interface for managing these databases and their location without overloading `scan` with a lot of flags

Part of me now wonders if it could be worth having a dedicated `offline` subcommand instead of a flag? but maybe that's just trading a (flag) swing for a (subcommand) roundabout?

internal/local/zip.go
cuixq: I just feel it's weird to use `downloadDB` to indicate `offline` so would like to keep consistent in the names, but I don't mind too much keeping `offline` in `ZipDB`.

another-rex:
I personally think `download-offline-databases` makes more sense than `offline-download-databases`.

another-rex (approved):
LGTM, just think we should swap offline and download around in the flag.

cuixq:
@G-Rath do you mind taking another look? Thanks.

docs/offline-mode.md
```diff
@@ -59,17 +59,17 @@ The offline database flag `--experimental-offline` causes OSV-Scanner to scan yo
 osv-scanner --experimental-offline ./path/to/your/dir
 ```
 
-## Local database option
+## Download offline databases option
 
-The local database flag `--experimental-local-db` causes OSV-Scanner to download or update your local database and then scan your project against it.
+The download offline databases flag `--experimental-download-offline-databases` causes OSV-Scanner to download or update your local database and then scan your project against it. This option only works when you also set the offline flag.
```
G-Rath, line 64: nit: I would phrase this more as "allowed" rather than "causes", as I think that aligns more with the theme - we're trying to be as offline as possible, but sometimes we can't, this flag _allows_ the scanner to be a little less offline in those cases

you could frame it something like

```suggestion
The download offline databases flag `--experimental-download-offline-databases` allows OSV-Scanner to download or update your local database when running in offline mode, to make it easier to get started. This option only works when you also set the offline flag.
```

G-Rath (approved):
looks good! got a minor comment about the doc wording, but it's optional :)

PR #1049, opened 2024-06-19
Package tracing PoC

another-rex:
Package/vulnerability tracing capability, used to determine which layer introduced a vulnerability.

This required some minor refactoring of the experimental image scanning types. 

TODOs:

- [x] Perform error handling when attempting to parse older layer files.
- [x] Determine the best way to pass back up the layer information.
- [x] Add tests


pkg/osvscanner/vulnerability_result.go
```diff
@@ -49,9 +49,10 @@ func buildVulnerabilityResults(
 
 		if rawPkg.Version != "" && rawPkg.Ecosystem != "" {
 			pkg.Package = models.PackageInfo{
-				Name:      rawPkg.Name,
-				Version:   rawPkg.Version,
-				Ecosystem: string(rawPkg.Ecosystem),
+				Name:          rawPkg.Name,
+				Version:       rawPkg.Version,
+				Ecosystem:     string(rawPkg.Ecosystem),
+				OriginLayerID: rawPkg.OriginLayerID,
```
hogo6002, line 55: I noticed that the final vulnerability JSON output only includes the layer ID, but not the specific command associated with that layer. Is there any specific reason for that?
image

pkg/lockfile/types.go
```diff
@@ -1,12 +1,14 @@
 package lockfile
 
+// TODO(v2): These fields do not need JSON tags I believe
```
cuixq, line 3: make a v2 wishlist bug?

cuixq (commented):
LGTM - I will let people with full knowledge to approve.

pkg/osvscanner/vulnerability_result.go
another-rex: Updated now to return:
- originLayerID
- Creation command
- Whether it's likely in a base image or not.

internal/image/image.go
```diff
@@ -30,61 +31,112 @@ type ScanResults struct {
...
+// of the corresponding layer
+func (img *Image) LayerIDToCommand(id string) string {
+	file, err := (*img.innerImage).ConfigFile()
+	if err != nil {
+		log.Panicln("Failed to get inner image config file")
+	}
+	idxCount := img.layerIDToIndex[id]
+	var i int
+	// Match history to layer IDX by skipping empty layer history entries
+	for i = 0; idxCount >= 0; i++ {
```
oliverchang, line 57: why is this `>= 0` as opposed to `> 0` ? 

If `idxCount == 0`, that means we've found `img.layerIDToIndex[id]` non-empty layers right? at that point `file.History[i]` should contain what we want?

internal/image/image.go
```diff
@@ -30,61 +31,112 @@ type ScanResults struct {
...
+	innerImage     *v1.Image
+	extractDir     string
+	baseImageIndex int
+	configFile     *v1.ConfigFile
+	layerIDToIndex map[string]int
 }
 
-func (img *Image) LastLayer() fileMap {
-	return img.flattenedLayers[len(img.flattenedLayers)-1]
+func (img *Image) LayerIDToIdx(id string) int {
```
oliverchang, line 43: nit: instead of `Idx`, let's just use the full `Index` for this function name. e.g. https://pkg.go.dev/strings#Index

oliverchang (commented):
nice!

internal/image/guess_base_image.go
```diff
@@ -0,0 +1,80 @@
+package image
+
+import (
+	"strings"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+// Originally from https://github.com/aquasecurity/trivy/blob/1f5f34895823fae81bf521fc939bee743a50e304/pkg/fanal/image/image.go#L111
```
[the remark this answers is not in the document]
another-rex: I'm not sure, they have the same Apache 2.0 license as our code. We might need to add a NOTICE file?

internal/image/image.go
another-rex: I believe `>=` is correct. because the layer index is 0 based. Imagine a simple image with just 1 non-empty layer:
```
  | 0
  | 1
0 | 2
  | 3

i = 0
idxCount = 0

i = 1
idxCount = 0

i = 2
idxCount = 0

i = 3
idxCount = -1
```

so the return layer is (`i - 1 = 2`).

Not sure what's the best way to clarify this in the comments, should I put the example above in?

oliverchang: Ah my bad, yeah `idxCount` is 0-indexed in the first place. Ignore this comment then!

internal/image/image.go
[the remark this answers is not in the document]
another-rex: Turns out this is not used/needed now that we are returning the history in the initial extraction step. Removed now.

internal/image/image.go
```diff
@@ -30,61 +31,112 @@ type ScanResults struct {
...
+	return img.layerIDToIndex[id]
+}
+
+// LayerIDToCommand takes in a layer id and returns the history CreatedBy field
+// of the corresponding layer
+func (img *Image) LayerIDToCommand(id string) string {
```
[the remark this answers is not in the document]
another-rex: Added an error for this to return when no history is available or no config file is available.

Also made this function private, as it doesn't need to be public anymore now that we are returning history in the initial extraction step.

internal/image/image.go
```diff
@@ -30,61 +31,112 @@ type ScanResults struct {
...
+	file, err := (*img.innerImage).ConfigFile()
+	if err != nil {
+		log.Panicln("Failed to get inner image config file")
```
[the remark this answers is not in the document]
another-rex: Switched this to use the preloaded config file. That config file will never be nil as it will return an error when initially loading the image.

internal/image/layer.go
```diff
@@ -0,0 +1,78 @@
...
+
+// imgLayer represents all the files on a layer
+type imgLayer struct {
+	fileNodeTrie *trie.PathTrie
+	id           string
+	rootImage    *Image
+	// TODO: Use hashmap to speed up path lookups
+}
+
+func (filemap imgLayer) GetFileNode(path string) (fileNode, error) {
```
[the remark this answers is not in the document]
another-rex: made this private

internal/image/image.go
another-rex: Function has been removed entirely as it's not needed anymore

internal/image/guess_base_image.go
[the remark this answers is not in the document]
oliverchang: OK, let's keep this as is for now then.

oliverchang (approved):

hogo6002 (approved):

PR #1055, opened 2024-06-21
fix: ensure that `package` exists in `affected` property

G-Rath:
This has always been allowed by the spec but now there's at least one real-world advisory in the Debian database like this which causes it to error.


G-Rath:
This unbreaks #958

hogo6002 (approved):

another-rex (approved):

cuixq (approved):

PR #1061, opened 2024-06-21
ci: don't fail if codecov errors when uploading

G-Rath:
Codecov has been just too flakey to have as a requirement for landing changes, so for the time being we're going to ignore if uploading fails


G-Rath:
It looks like the underlying issues might have been resolved in the last couple of hours via https://github.com/codecov/codecov-action/issues/1487#issuecomment-2183079349, so I'm going to optimistically close this for now

PR #1066, opened 2024-06-24
Add new line to filtered out vulns

another-rex:
These logs weren't adding a new line, causing all of them to be printed in one big line.

I wonder if we can add a lint for this, seems to happen pretty often to everyone using the reporter.


hogo6002 (approved):
Thanks Rex! I didn't notice that.

hogo6002:
Replaced by: https://github.com/google/osv-scanner/pull/1067

PR #1076, opened 2024-06-27
Changed min and max to inbuilt functions

neilnaveen:
- Changed min and max to inbuilt functions.


G-Rath (approved):
I have no idea how I missed that these got added in v1.21, but I love it! 🎉

hogo6002 (approved):
LGTM. Thanks Neil!

cuixq (approved):

PR #1083, opened 2024-07-01
Refactoring `manifest.Read()` for Maven

cuixq:
The current design of Maven manifest reading has some problems:
 - requirements that are not useful for resolution are also included
 - requirement origins mess the dependency resolution as well

To address the issues above, in this PR:
 - requirements returned are only those ones needed for dependency resolution
 - passing Maven dependencies with origins so we can look up their original requirement
 - passing requirements that are needed for updates only individually

However there are still known issues for this implementation:
 - profiles are not yet fully supported (pending deps.dev changes)
 - original requirements may not be found for dependencies with placeholders in their names

### DIFF
--- a/internal/remediation/suggest/maven.go
+++ b/internal/remediation/suggest/maven.go
@@ -52,22 +52,20 @@ func (ms *MavenSuggester) Suggest(ctx context.Context, client resolve.Client, mf
 			Type:       req.Type,
 			NewRequire: latest.Version,
 		}
-		origReq, depOrigin := originalRequirement(patch, specific.OriginalRequirements)
-		if origReq == "" {
-			// Cannot find the original requirement indicates the requirement not in base project
+		origDep := originalDependency(patch, specific.OriginalRequirements)
+		if origDep.Name() == "" {
+			// An empty name indicates the dependency is not found, so the original dependency is not in the base project.
 			continue
 		}
 
-		dep, _, _ := resolve.MavenDepTypeToDependency(patch.Type)
-		patch.Type = resolve.MavenDepType(dep, depOrigin)
-
-		if !origReq.ContainsProperty() {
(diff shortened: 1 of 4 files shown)


internal/resolution/manifest/maven.go
```diff
@@ -51,186 +51,105 @@ func NewMavenManifestIO() MavenManifestIO {
 }
 
 type MavenManifestSpecific struct {
-	Properties                 []PropertyWithOrigin
-	RequirementsWithProperties []resolve.RequirementVersion
-	RequirementsFromOtherPOMs  []resolve.RequirementVersion // Requirements that we cannot modify directly
+	Properties             []PropertyWithOrigin
```
michaelkedar, line 54: From what I understand, these properties don't include properties from parents, right?
Maybe add a comment to note this?

internal/remediation/suggest/maven.go
```diff
@@ -52,23 +52,29 @@ func (ms *MavenSuggester) Suggest(ctx context.Context, client resolve.Client, mf
...
+		origReq, depOrigin := originalRequirement(patch, specific.OriginalRequirements)
+		if origReq == "" {
+			// Cannot find the original requirement indicates the requirement not in base project
+			continue
+		}
 
-		origReq := originalRequirement(req, specific.RequirementsWithProperties)
-		if !requirementHasProperty(origReq) {
+		dep, _, _ := resolve.MavenDepTypeToDependency(patch.Type)
+		patch.Type = resolve.MavenDepType(dep, depOrigin)
```
michaelkedar, line 62: nit: maybe you can avoid calling `resolve.MavenDepTypeToDependency` twice if `originalRequirement()` just returns the `manifest.DependencyWithOrigin`?

michaelkedar (commented):
Nice! a few comments:

michaelkedar (approved):
LGTM

PR #1087, opened 2024-07-02
refactor: just disable color output rather than tracking terminal width

G-Rath:
This lets us drop some conditionals - weirdly it doesn't seem to work if we move the conditional `text.DisableColors` call into `reporter.New` even though that does work for the vertical reporter...

### DIFF
--- a/internal/output/markdowntable.go
+++ b/internal/output/markdowntable.go
@@ -4,15 +4,17 @@ import (
 	"io"
 
 	"github.com/google/osv-scanner/pkg/models"
-
 	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
 )
 
 // PrintTableResults prints the osv scan results into a human friendly table.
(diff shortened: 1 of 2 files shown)


another-rex (approved):
Thanks!

another-rex:
> This lets us drop some conditionals - weirdly it doesn't seem to work if we move the conditional `text.DisableColors` call into `reporter.New` even though that does work for the vertical reporter...

Maybe it's because there multiple reporter.New() calls every run, and the disable color function is a single global variable.

G-Rath:
yeahhhhh but then why would it work for the vertical output...? something odd is def going on but I think it's probably best we deal with that once we've landed the vertical report since we plan to shuffle things around a bit re table anyway right

G-Rath:
Weirdly too it seems like it's a very stable global? I've run the test suite a ton and it never seems to give different results like I'd expect if this is an unstable use of a global...?

(fwiw I vote we land this for now - we can easily revert if it turns out to be flakey)

hogo6002 (approved):

PR #1093, opened 2024-07-02
test: add output fixtures for call analysis

G-Rath:
These fixtures were based on the output from the `cmd/osv-scanner/fixtures/call-analysis-go-project` fixture, though from the struct I think it's pretty complete; the most notable thing is it looks like call analysis does play favorites unlike most of the scanner: only one advisory of a vulnerability will be marked as called/uncalled regardless of how many aliases that vulnerability has.


G-Rath:
This unblocks support call analysis in #889

another-rex (approved):
Thanks!

PR #1104, opened 2024-07-08
Move sbom to internal, and add standard output tests

another-rex:
- Move SBOM versions enum into models folder. 
  - This change is done to proactively avoid future circular dependency issues
- Move CycloneDX logic into internal/output/sbom
  - Avoid making a public library yet for something that we might wish to change in the near future. 
- By having CycloneDX in output with the rest of the output formats, we can add the standard output tests as well.


another-rex:
Depending on the order #1093 is merged, either this PR or #1093 needs to be updated.

internal/output/cyclonedx.go
```diff
@@ -0,0 +1,25 @@
...
+	"io"
+
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/google/osv-scanner/internal/output/sbom"
+	"github.com/google/osv-scanner/internal/utility/purl"
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+// PrintJSONResults writes results to the provided writer in JSON format
+func PrintCycloneDXResults(vulnResult *models.VulnerabilityResults, cycloneDXVersion models.CycloneDXVersion, outputWriter io.Writer) error {
```
G-Rath, line 14:
```suggestion
// PrintCycloneDXResults writes results to the provided writer in CycloneDX format
func PrintCycloneDXResults(vulnResult *models.VulnerabilityResults, cycloneDXVersion models.CycloneDXVersion, outputWriter io.Writer) error {
```

pkg/models/cyclonedx.go
```diff
@@ -0,0 +1,8 @@
+package models
+
+const (
+	CycloneDXVersion14 CycloneDXVersion = iota
+	CycloneDXVersion15
+)
+
+type CycloneDXVersion int
```
michaelkedar, line 8: nit: can you move the `type` definition above the `const`?

G-Rath (approved):

michaelkedar (approved):

cuixq (approved):
LGTM - maybe explain a bit more in the description why we make this change?

Issue #989, opened 2024-05-21
Ability to add new sections to pom.xml in Maven updater


cuixq:
Currently [Maven updater](https://github.com/google/osv-scanner/blob/main/internal/resolution/manifest/maven.go) can only substitute elements in pom.xml, however, we may want to add new section (for example `<dependencyManagment/>`) to pom.xml for guided remediation.

PR #1105, opened 2024-07-09
Add a dependency to pom.xml if it is not from the base project

cuixq:
https://github.com/google/osv-scanner/issues/989

For the purpose of overriding dependency requirements, we may need to add new dependencies or even dependency management section to pom.xml. This PR:
 - moves the patch generation from suggester to updater
 - before writing patches, prepare dependency and property patches
 - when updating dependencies, if a dependency is not from the base project, we should add a new section
 - when updating a project, if dependency management is not updated and there are patches from dependency management, add a new dependency management section
 - marks the parent dependency patch to have type "pom"

TODOs:
 - need to find a way to differentiate patches for guided remediation or general updates, since we should not add new sections for general updates (using `OrigRequire` seems reasonable)
 - make the indentation for the new sections configuration, now defaults to a white space


internal/resolution/manifest/maven.go
```diff
@@ -329,6 +330,196 @@ func mavenOrigin(list ...string) string {
...
+	for _, patch := range patches {
+		origDep := originalDependency(patch, specific.OriginalRequirements)
+		if origDep.Name() == "" {
+			// An empty name indicates the dependency is not found, so the original dependency is not in the base project.
+			// TODO: enable for guided remeidation
+			// if err := depPatches.addPatch(patch, false); err != nil {
+			//	return MavenDependencyPatches{}, MavenPropertyPatches{}, err
+			// }
+			continue
+		}
```
michaelkedar, line 403: Is there a particular reason to have this disabled?

cuixq: currently patches from both guided remediation and general updates do not set `OrigRequire` which makes it hard to differentiate them (we don't want to add patches unfound in base project if they are for general updates) and there is also a TODO for that purpose.

PR #1107, opened 2024-07-10
Fix npm grouping

another-rex:
Fix bug in the npm grouping code, the current maps.Copy method doesn't use the merging applied in details.add, causing inconsistent test failures.


hogo6002 (approved):

G-Rath (approved):
I swear I tested this when I found out about the function to confirm it was 1:1, but oh well...

PR #1108, opened 2024-07-10
Activate profiles before merging parent

cuixq:
We should merge active profiles before merging parent data. Currently, we only merge default profiles by passing empty JDK and OS information (this could be a TODO for the future).

Some known issue or improvement:
 - Inactive profiles may not be interpolated but we probably want them to be updated as well
 - Now default profiles are merged but we may consider leveraging JDK and OS information to activate accurate profiles


michaelkedar (approved):
LGTM

cuixq:
There is going to be an update on `deps.dev/util/semver` and I will wait a bit to include it.

PR #1113, opened 2024-07-11
Sort dependencies before writing to pom.xml

cuixq:
Currently, the write test is a bit flaky because the dependencies to be added are not sorted and their order in pom.xml is not guaranteed. 

This PR adds sorting of dependencies before they are going to be written.


internal/resolution/manifest/maven.go
```diff
@@ -843,7 +854,6 @@ func updateDependency(w io.Writer, enc *xml.Encoder, raw string, patches map[Mav
 					// A Maven dependency key consists of Type and Classifier together with GroupID and ArtifactID.
 					if patch.DependencyKey == rawDep.Key() {
 						req = patch.NewRequire
-						patches[patch] = true
```
cuixq, line 846: this is no longer needed since we pre-compute which patches are to be added in `buildPatches`.

internal/resolution/manifest/maven.go
```diff
@@ -585,22 +629,7 @@ func write(buf *bytes.Buffer, w io.Writer, depPatches MavenDependencyPatches, pr
...
-						}
-						if di.Type != dj.Type {
-							return di.Type < dj.Type
-						}
-						if di.Classifier != dj.Classifier {
-							return di.Classifier < dj.Classifier
-						}
-
-						return di.Version < dj.Version
+						return compareDependency(dm.Dependencies[i], dm.Dependencies[j])
```
michaelkedar, line 632: Not something to change, but something I've wondered about:
Is there a reason to use `sort.Slice` over [`slices.SortFunc`](https://pkg.go.dev/slices#SortFunc), or is it just a preference thing?

I guess technically `slices.SortFunc` would make this one line instead of two:
```golang
slices.SortFunc(dm.Dependencies, compareDependency)
```

michaelkedar (approved):
LGTM

internal/resolution/manifest/maven.go
G-Rath: fwiw the docs technically recommend you do change it:

> Note: in many situations, the newer [slices.SortFunc](https://pkg.go.dev/slices#SortFunc) function is more ergonomic and runs faster.

though I think when I looked into it, it felt like their comment was more based on what types you were using that lead to different compare functions rather than "we've spent more time optimizing `slices.SortFunc` than this one" kind of thing?

also kind of related, here's [some prior art](https://github.com/google/osv-scanner/blob/main/internal/output/result.go#L23-L42) on how I implemented a chain of compares similar to what you've done with `compareDependency`.

Also also, note that there's [`cmp.Or`](https://pkg.go.dev/cmp@master#Or) which _might_ be useful here, but also might not be... 🤷

cuixq: Nice to know both `slice.SortFunc` and `com.Or`!

internal/resolution/manifest/maven.go
```diff
@@ -542,6 +543,38 @@ func projectStartElement(raw string) string {
...
+	}
+	if patch.Type != "" && patch.Type != "jar" {
+		d.Type = string(patch.Type)
+	}
+
+	return d
+}
+
+func compareDependency(d1, d2 dependency) int {
+	return cmp.Or(cmp.Compare(d1.GroupID, d2.GroupID), cmp.Compare(d1.ArtifactID, d2.ArtifactID),
```
G-Rath, line 574: Nit: I think it'd be more readable if you put each condition on a newline

internal/resolution/manifest/maven.go
cuixq: sadly `cmp.Or` only for Go 1.22 :(

G-Rath: Ek sorry I learned that the same hard way when I first discovered it 😬

PR #1126, opened 2024-07-17
Bump goreleaser build version to 1.22. 

### LINKED ISSUE #1123: issue when running osv-scanner on project running with golang 1.22
Hello, I have some issues with my project which uses golang v1.22.5. 
On the CI, I'm pulling the latest version of OSV-scanner from ghcr.io. 

```
osv-scanner:
    runs-on: xxx
    container: ghcr.io/google/osv-scanner:v1.8.2
        ....
```

I got this error when running the osv-scanner, does it because the osv-scanner was built using go v1.21?
```
This application uses version go1.21 of the source-processing packages but runs version go1.22 of 'go list'. It may fail to process source files that rely on newer language features. If so, rebuild the application using a newer version of Go.
```

another-rex:
Remove the patch version as we always want the latest patch version.

Fixes #1123 

We just need to keep in mind that the library code should be compatible with the previous go release.


cuixq (approved):

hogo6002:
Should we also update our docs: https://github.com/search?q=repo%3Agoogle%2Fosv-scanner%201.21.11&type=code

PR #1030, opened 2024-06-07
Update Go patch version

hogo6002:
Updates/Adds Go patch version to docs

PR #1133, opened 2024-07-21
Reflect Go 1.21.12 change more broadly

andrewpollock:
Replicate #1109 more like #1030 to avoid Renovate failures like https://github.com/google/osv-scanner/pull/1120#issuecomment-2227395239

The Docker files are already at 1.22.5, not sure if we want to update to this globally?


another-rex (approved):

another-rex:
We probably should update to 1.22.5 in most places (but still keep the library compatible with 1.21.12), https://github.com/google/osv-scanner/issues/1123#issuecomment-2237343942 is also worth looking at

PR #1136, opened 2024-07-23
feat(guided remediation): add non-interactive Maven remediation by override

michaelkedar:
#1141 
Adds `--strategy=override` for `osv-scanner fix --non-interactive` for `pom.xml` manifest files. For now interactive mode will print an error telling you to use non-interactive mode if you try `pom.xml`.

Also, made non-interactive mode decide which strategy to use if not explicitly specified based on the provided manifests / lockfiles.


michaelkedar:
I need to fix [the possible infinite loop](https://github.com/google/osv-scanner/pull/1025#discussion_r1671418639) before this gets released - I've run into it on a couple of random projects I've tested.

internal/remediation/override.go
```diff
@@ -66,8 +66,8 @@ func ComputeOverridePatches(ctx context.Context, cl client.ResolutionClient, res
 		}
 
 		if res.err != nil {
-			// TODO: stop goroutines
-			return nil, res.err
+			// Resolution errors aren't uncommon - just skip if we encounter one.
```
oliverchang, line 69: Can you please expand on this? And why this isn't a concern?

 Should we at least log the error?

cmd/osv-scanner/fix/noninteractive.go
```diff
@@ -247,6 +247,113 @@ func relockUnfixableVulns(diffs []resolution.ResolutionDiff) []*resolution.Resol
...
+	pkgChanged := make(map[resolve.PackageKey]bool) // dependencies we've already applied a patch to
+	fixedVulns := make(map[string]bool)             // vulns that have already been fixed by a patch
+	for _, diff := range diffs {
+		// If this patch is incompatible with existing patches, skip adding it to the patch list.
+
+		// A patch is incompatible if any of its changed packages have already been changed by an existing patch.
+		if slices.ContainsFunc(diff.Deps, func(dp manifest.DependencyPatch) bool { return pkgChanged[dp.Pkg] }) {
+			continue
+		}
+		// A patch is also incompatible if any fixed vulnerability has already been fixed by another patch.
```
oliverchang, line 334: can you explain a bit more why this makes patches incompatible?

oliverchang (approved):

internal/remediation/override.go
michaelkedar: Expanded briefly.

Currently I don't have a good way to log in these functions (and in a way that will play nicely with interactive mode), so I've just left a TODO.

cmd/osv-scanner/fix/noninteractive.go
michaelkedar: Expanded more with an example.

oliverchang (approved):
thanks!

cuixq (approved):

cmd/osv-scanner/fix/main.go
```diff
@@ -232,11 +230,29 @@ func action(ctx *cli.Context, stdout, stderr io.Writer) (reporter.Reporter, erro
 	r := reporter.NewTableReporter(stdout, stderr, reporter.InfoLevel, false, 0)
 	maxUpgrades := ctx.Int("apply-top")
 
-	switch ctx.String("strategy") {
+	strategy := ctx.String("strategy")
+
+	if !ctx.IsSet("strategy") {
+		// Choose a default strategy based on the manifest/lockfile provided.
+		switch {
```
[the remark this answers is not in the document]
michaelkedar: Yeah, I thought relocking/relaxing is less potentially breaking than override which is why I've made it preferred.

cmd/osv-scanner/fix/interactive.go
```diff
@@ -13,6 +13,10 @@ import (
 
 func interactiveMode(ctx context.Context, opts osvFixOptions) error {
 	if !remediation.SupportsRelax(opts.ManifestRW) && !remediation.SupportsInPlace(opts.LockfileRW) {
+		if remediation.SupportsOverride(opts.ManifestRW) {
```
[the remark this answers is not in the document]
michaelkedar: It's inside to check if override is the only strategy supported. If something supports override and relax then we can still launch interactive mode.

Not that it matters anyway - I'm not implementing override for npm / relax for maven any time soon :)

PR #1139, opened 2024-07-24
fix: handle parsing versions in `composer.lock` files that are numbers rather than strings

### LINKED ISSUE #1138: Composer version not always a string
While analyzer a `composer.lock` file we have the following error:

```shell
(extracting as composer.lock) could not extract from /app/composer.lock: json: cannot unmarshal number into Go struct field ComposerPackage.packages.version of type string
```

The version in this case is not a string:
```shell
➜  /tmp grep version composer.lock
            "version": "5.9.1",
            "version": "1.99",
            "version": "0.1.1",
            "version": "1.2.8",
            "version": "v1.9.0",
            "version": "4.6.2",
            "version": "v2.11.0",
            "version": "5.5.3",
            "version": "5.5.3",
            "version": "2.0.0",
            "version": "v1.8.0",
            "version": "v0.8.1",
            "version": "1.8.2",
            "version": "2.4.20.3",
            "version": "2.6.1",
            "version": "3.7.2",
            "version": "1.4.3",
            "version": "20190220",
            "version": "1.11.5",
            "version": "3.1",
            "version": "1.5",
            "version": "0.5.7",
```

G-Rath:
It seems it's technically possible for `composer.lock` files to have a number for `version` in packages - I can't actually reproduce this with `composer` itself, but if I manually edit a lockfile to have a number instead of a string it doesn't complain or change the value.

It would be good to understand more about how this could happen in the wild, but it's easy enough to support either way.

Resolves #1138

### DIFF
--- a/pkg/lockfile/fixtures/composer/with-bad-version.json
+++ b/pkg/lockfile/fixtures/composer/with-bad-version.json
@@ -1,0 +1,183 @@
+{
+  "_readme": [
+    "This file locks the dependencies of your project to a known state",
+    "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+    "This file is @generated automatically"
+  ],
+  "content-hash": "035277326fc4da8f74a4f00bb10a888a",
+  "packages": [
+    {
+      "name": "composer/installers",
+      "version": "v2.3.0",
+      "source": {
+        "type": "git",
+        "url": "https://github.com/composer/installers.git",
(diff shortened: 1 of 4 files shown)


another-rex:
I think we should try to figure out why this is generated first before merging this fix. I'm not sure if we can confidently support this before figuring out the root cause. (e.g. will the number ever have a decimal point in it?)

cuixq:
It seems `composer` does not complain about strings and valid numbers (for example `"version":20190220` or `"version":1.2`) but complains about invalid numbers (for example `"version":1.2.3`).

PR #1140, opened 2024-07-25
Add Maven dependency management to override client

cuixq:
Currently, Maven dependency management is not added to the override client so they are not considered when computing Maven dependency graph. 

This PR adds all direct dependency management to override client so that transitive dependencies are resolved correctly.


michaelkedar (approved):
LGTM

PR #1142, opened 2024-07-26
OSV scalibr interface

another-rex:
This PR begins the work of migrating lockfile extractors to use the new interface. 

To aid in the review, I copied all of `pkg/lockfile` into `internal/lockfilescalibr`, and pushed that to `osv-scanner-v2`, which this is now merging in.

Some notes on the migration:

- This is still in a really draft state
- None of the fixtures has been touched with the exception of alpine, which had some minor directory moving to support testing OS version extraction.
- Still missing a few extractors:
  - dpkg (scalibr already has dpkg)
  - osv-scanner config format
  - csv (intentionally left out as I'm not sure if we need it)
  - node_modules (scalibr already has a better version, that uses package.jsons)
  - Maven + transition dependency resolution (might have to wait till we get a unified client)
- Moved to tabular tests, 
  - Updated tests to test Inventory output
  - Tests now all use similar helper functions to make future refactoring a lot easier, and to make writing tests easier
- Kept PackageDetails for some packages and added a simple conversion function
  to inventory before returning (currently the only one still using PackageDetails is npm extractor, as adding the merging code is a bit convoluted odd with inventory)
- Copied the new interface into lockfilescalibr/extractor.go. This is temporary until the move into osv-scalibr, which will contain both
- All ToPURL functions need to be looked at to see if we should be adding additional qualifiers/namespaces..etc.
- We need to add tests for ToPurl() and Ecosystem() functions
- Because scalibr uses a virtual FS to walk over files, all paths are absolute from ScanRoot, but will not start with /

---

Update:

- Moved all of the osv-scalibr native types to packages with the same name as in osv-scalibr, to make migrating easier.
- Moved test helper functions into their own package to allow it to be shared with separate package extractors. (`sharedtesthelpers`)
- Moved most extractors into their own package, the remaining ones are still in progress to be moved. This follows a similar folder structure as osv-scalibr.
- Not sure where to put metadata, so separated it out into the `othermetadata` package


internal/lockfilescalibr/go-binary.go
```diff
@@ -94,7 +96,7 @@ func (e GoBinaryExtractor) ToCPEs(i *extractor.Inventory) ([]string, error) { re
 func (e GoBinaryExtractor) Ecosystem(i *extractor.Inventory) (string, error) {
 	switch i.Extractor.(type) {
 	case GoBinaryExtractor:
-		return "go", nil
+		return string(GoEcosystem), nil
```
oliverchang, line 99: do we need the string() here since `GoEcosystem` is already a string?

internal/lockfilescalibr/language/dotnet/nugetpackagelock/extractor-v1_test.go
```diff
@@ -1,113 +1,113 @@
-package lockfilescalibr_test
+package nugetpackagelock_test
```
oliverchang, line 1: Thinking more, is the convention of "<language>/<pkg manager>" something we definitely need to keep?

There are some edge cases. e.g. https://en.wikipedia.org/wiki/Conda_(package_manager), which is a package manager for R **and** python

PR #1147, opened 2024-07-29
Improve Maven manifest updater

cuixq:
This PR includes some improvements on Maven manifest updater:
 - rename functions from `update` to `write`
 - trimming white space when writing a new string value
 - for dependency with an empty version, keep looking for non-empty requirement


internal/resolution/manifest/maven.go
```diff
@@ -894,18 +895,19 @@ func updateString(enc *xml.Encoder, raw string, values map[string]string) error
...
-					return err
-				}
-				if err := enc.EncodeElement(value, tt); err != nil {
-					return err
-				}
-
-				continue
+				str = value
+			}
+			if err := enc.EncodeElement(strings.TrimSpace(str), tt); err != nil {
```
cuixq, line 906: I spent some time trying to keep the white space but could not find a good way to do this, so I decided to trim out the white space. This is not ideal but good enough, @michaelkedar let me know how you feel about this.

internal/resolution/manifest/maven_test.go
```diff
@@ -236,6 +257,7 @@ func TestMavenRead(t *testing.T) {
 				{Property: maven.Property{Name: "maven.compiler.source", Value: "1.7"}},
 				{Property: maven.Property{Name: "maven.compiler.target", Value: "1.7"}},
 				{Property: maven.Property{Name: "junit.version", Value: "4.12"}},
+				{Property: maven.Property{Name: "zeppelin.daemon.package.base", Value: "\n\t    ../bin\n    "}},
```
cuixq, line 260: `deps.dev/util/maven` probably needs to be updated to trim the white space when parsing properties.

internal/resolution/manifest/maven.go
```diff
@@ -490,7 +490,8 @@ func originalDependency(patch DependencyPatch, origDeps []DependencyWithOrigin)
 	dependency.ArtifactID = maven.String(IDs[1])
 
 	for _, d := range origDeps {
-		if d.Key() == dependency.Key() {
+		if d.Key() == dependency.Key() && d.Version != "" {
```
cuixq, line 493: this should solve the current problem of not being able to update dependency without a version since we agreed we would like to update dependency management section for this case. @michaelkedar

internal/resolution/manifest/maven.go
michaelkedar: I'm fine with this if it's overly difficult to preserve the white space.
Though nit: maybe add a comment here explaining it.

michaelkedar (approved):

PR #1149, opened 2024-07-31
Handle Maven parent relative path

cuixq:
When Maven looks for the parent POM, it first looks up the specified relative path, then look for the default relative path which is `../pom.xml`, and lastly in the remote repository. If only a directory is specified in relative path, `pom.xml` will be looked automatically.

Reference: https://maven.apache.org/ref/3.9.8/maven-model/maven.html#parent

Currently, OSV-Scanner only do some steps above, this PR corrects this.

Also, considering both `internal/manifest` and `internal/resolution/manifest` require basically the same logic for merging parent POM, I would like to refactor this in a following PR.


internal/resolution/manifest/maven.go
```diff
@@ -301,6 +309,28 @@ func (m MavenManifestIO) mergeParents(ctx context.Context, result *maven.Project
 	return result.Interpolate()
 }
 
+// Maven looks for the parent POM first in 'relativePath',
+// then the local repository '../pom.xml',
+// and lastly in the remote repo.
+func MavenParentPOMPath(currentPath, relativePath string) string {
+	if relativePath == "" {
```
cuixq, line 316: there is some caveat here: we are not able to tell whether the relative path is not set (should default to `../pom.xml`) or set to an empty string (parent should be fetched from upstream).
but we still check if the parent pom.xml exists and its key matches what we want, so I think this might be fine...

michaelkedar: Maybe there's a change that can be made in the deps.dev repo to set a default value of `relativePath` to `../pom.xml` if it's unspecified?

internal/resolution/manifest/maven.go
```diff
@@ -263,20 +264,23 @@ func (m MavenManifestIO) mergeParents(ctx context.Context, result *maven.Project
...
 			}
-		} else {
+			if proj.ProjectKey != current.ProjectKey || proj.Packaging != "pom" {
+				// The identifiers or packaging in parent does not match what we want,
+				// mark proj as empty so parent can be fetched from upstream.
+				proj = maven.Project{}
+			}
+		}
+		// proj being empty indicates that we are not able to find parent pom.xml locally.
+		if reflect.DeepEqual(proj, maven.Project{}) {
```
michaelkedar, line 283: nit: Not sure if I like this `reflect.DeepEqual()` here.
Could maybe instead use a pointer and check `proj == nil` or add a `parentFound bool` and check that?

Though, I'm fine to keep it as-is if you would prefer.

internal/resolution/manifest/maven.go
```diff
@@ -301,6 +309,28 @@ func (m MavenManifestIO) mergeParents(ctx context.Context, result *maven.Project
...
+// and lastly in the remote repo.
+func MavenParentPOMPath(currentPath, relativePath string) string {
+	if relativePath == "" {
+		relativePath = "../pom.xml"
+	}
+	path := filepath.Join(filepath.Dir(currentPath), relativePath)
+	if info, err := os.Stat(path); err == nil {
+		if !info.IsDir() {
+			return path
+		}
```
michaelkedar, line 323: Wondering what maven does if the `relativePath` points to a file that is not named `pom.xml`

michaelkedar (approved):
LGTM

internal/resolution/manifest/maven.go
cuixq: this is acceptable if the file exists and is a valid parent pom.xml, otherwise, maven tries to download from upstream.

internal/resolution/manifest/maven.go
cuixq: this can be investigated - we need to make sure we are able to differentiate the two situations.

internal/resolution/manifest/maven.go
cuixq: actually I considered `parentFound` so just changed the PR to use it

PR #1151, opened 2024-07-31
Do not attempt to remediate vulnerabilities in Maven artifacts that have defined `<classifier>` or `<type>`

michaelkedar:
There is currently no way in the deps.dev API to determine the Maven classifier/type that exist for a given version of a package. Not knowing this, the override strategy can end up bumping to a version that does not exist, which would cause a compilation failure that can't be detected in-code.

I'm avoiding the bigger issue here by just explicitly preventing remediation for those artifacts...

This will also prevent the infinite loop I mentioned in https://github.com/google/osv-scanner/pull/1025#discussion_r1671418639


michaelkedar:
@cuixq the issue with these would also affect the Maven updater - not sure what we want to do about that...

cuixq (approved):

internal/remediation/override.go
```diff
@@ -128,13 +129,24 @@ func overridePatchVulns(ctx context.Context, cl client.ResolutionClient, result
 			seenVKs := make(map[resolve.VersionKey]struct{})
 			// Use the DependencyChains to find all the affected nodes.
 			for _, c := range v.ProblemChains {
+				// Currently, there is no way to know if a specific classifier or type exists for a given version with deps.dev.
```
oliverchang, line 132: Can you add a TODO here to better handle this by checking against the Maven registry ?

oliverchang (approved):

Issue #1127, opened 2024-07-18
Support snapshot artifacts when fetching Maven parent POMs


Malayke:
Hello,

I encountered an issue while performing a security scan on one of my Java projects using `osv-scanner`. The scan failed, and the error output is as follows:

```
❯ ./osv-scanner -skip-git -r $HOME/gitlab/abcd/broker/my-kyc-center
Scanning dir /Users/user/gitlab/abcd/broker/my-kyc-center
Attempted to scan lockfile but failed: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-api/pom.xml
Attempted to scan lockfile but failed: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-common/pom.xml
Attempted to scan lockfile but failed: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-job/pom.xml
Attempted to scan lockfile but failed: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-persist/pom.xml
Attempted to scan lockfile but failed: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-service/pom.xml
Scanning directory for vendored libs: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-service/src/main/java/com/my/kyc/service/vendor
Scanning potential vendored dir: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-service/src/main/java/com/my/kyc/service/vendor/model
Scanning potential vendored dir: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-service/src/main/java/com/my/kyc/service/vendor/security
Attempted to scan lockfile but failed: /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-web/pom.xml
Attempted to scan lockfile but failed: /Users/user/gitlab/abcd/broker/my-kyc-center/pom.xml
No package sources found, --help for usage information.
```

After debugging the source code, I obtained more detailed error information:

```
"failed extracting /Users/user/gitlab/abcd/broker/my-kyc-center/kyc-api/pom.xml: failed to merge parents: failed to get Maven project com.my.kyc:kyc-center:2.0.0-SNAPSHOT: API query failed: Maven registry query status: 404 Not Found"
```

The directory `/Users/user/gitlab/abcd/broker/my-kyc-center` contains a multi-module Maven project. Through debugging, I think I discovered the following two points causing the scan failure:

1. **Hardcoded Maven Central URL:**
[...]

PR #1160, opened 2024-08-02
feat: support fetching snapshot versions from a Maven registry

cuixq:
https://github.com/google/osv-scanner/issues/1127

Maven snapshot versions cannot be requested directly. Instead, we should request version level metadata first to get the exact version value and then request the pom.xml.

This PR supports fetching snapshot versions by:
 - adding functions to fetch artifact-level and version-level metadata
 - when fetching a project, first check if it's a snapshot version, if yes, request the metadata to get the exact version value
 - add a test case for fetching snapshot versions


internal/resolution/datasource/maven_registry.go
```diff
@@ -24,31 +24,95 @@ func NewMavenRegistryAPIClient(registry string) *MavenRegistryAPIClient {
...
+	// Fetch version metadata for snapshot versions.
+	metadata, err := m.getVersionMetadata(ctx, groupID, artifactID, version)
+	if err != nil {
+		return maven.Project{}, err
+	}
+
+	snapshot := ""
+	for _, sv := range metadata.Versioning.SnapshotVersions {
+		if sv.Extension == "pom" {
+			// We only look for pom.xml for project metadata.
```
michaelkedar, line 43: Do you know what happens if the `pom` and `jar` versions are mismatched?

e.g. `foo-1.0.0-SNAPSHOT` has `jar` = `1.0.0-1` and `pom` = `1.0.0-2`.
Presumably, maven would download `foo-1.0.0-1.jar` for my dependency, but would it look for `foo-1.0.0-1.pom` or `foo-1.0.0-2.pom` for its dependencies?

internal/resolution/datasource/maven_registry.go
```diff
@@ -24,31 +24,95 @@ func NewMavenRegistryAPIClient(registry string) *MavenRegistryAPIClient {
 
 var errAPIFailed = errors.New("API query failed")
 
+// GetProject fetches a pom.xml specified by groupID, artifactID and version and parses it to maven.Project.
+// For a snapshot version, version level metadata is used to find the extact version value.
```
michaelkedar, line 28: Could you add some more information on how snapshot versions behave in maven (is there a doc you can link to?)

It took me a while (& looking at your tests & some maven repos) to try understand how it works.
From what I understand, when you depend on `pkg@1.2.3-SNAPSHOT`, it looks up in the registry `pkg/1.2.3-SNAPSHOT/maven-metadata.xml` to determine which file to actually download, and downloads it from `pkg/1.2.3-SNAPSHOT/1.2.3-realversion.pom`?

michaelkedar (approved):
LGTM - I've thought of a few corner cases that can be left as TODOs.

internal/resolution/datasource/maven_registry.go
cuixq: Yes correct.

The closest official documentation I find is https://maven.apache.org/ref/3.9.9/maven-repository-metadata/ and  https://maven.apache.org/repositories/metadata.html but the link to V-level metadata is not found...

I added both links to the comment.

internal/resolution/datasource/maven_registry.go
```diff
@@ -24,31 +24,95 @@ func NewMavenRegistryAPIClient(registry string) *MavenRegistryAPIClient {
...
-	u, err := url.JoinPath(m.registry, strings.ReplaceAll(groupID, ".", "/"), artifactID, version, fmt.Sprintf("%s-%s.pom", artifactID, version))
+	if !strings.HasSuffix(version, "SNAPSHOT") {
+		return m.getProject(ctx, groupID, artifactID, version, "")
+	}
+
+	// Fetch version metadata for snapshot versions.
+	metadata, err := m.getVersionMetadata(ctx, groupID, artifactID, version)
+	if err != nil {
+		return maven.Project{}, err
+	}
```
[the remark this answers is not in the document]
cuixq: I am not sure if it is valid for a snapshot version to not have a metadata file.

Based on the link in the comment above, a non-snapshot version should not have a metadata file with `snapshotVersion`.

internal/resolution/datasource/maven_registry.go
cuixq: I don't know exact answer to this question. A snapshot version is the version in development before release and the actual version string is the timestamp + build number - I would expect them to be the same for the same snapshot version.

ifuatgucluer:
good job

PR #1168, opened 2024-08-07
fix: only trim XML elements with no inner elements

cuixq:
This PR fixes two issues when reading/writing Maven manifest:
 - when merging parents, if `groupId` or `version` is empty, inherit these values from the parent;
 - only trim elements with no inner elements, otherwise inner elements will be wiped.

### DIFF
--- a/internal/resolution/manifest/maven.go
+++ b/internal/resolution/manifest/maven.go
@@ -646,9 +646,7 @@ func writeProject(w io.Writer, enc *xml.Encoder, raw, prefix, id string, patches
 						req = k.NewRequire
 					}
 				}
-				if err := writeString(enc, "<parent>"+rawParent.InnerXML+"</parent>", func(s string) bool {
-					return s != "parent"
-				}, map[string]string{"version": req}); err != nil {
+				if err := writeString(enc, "<parent>"+rawParent.InnerXML+"</parent>", []string{"parent"}, map[string]string{"version": req}); err != nil {
 					return fmt.Errorf("updating parent: %w", err)
 				}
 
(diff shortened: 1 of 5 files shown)


internal/resolution/manifest/maven.go
```diff
@@ -827,7 +833,8 @@ func writeDependency(w io.Writer, enc *xml.Encoder, raw string, patches map[Mave
 	return enc.Flush()
 }
 
-func writeString(enc *xml.Encoder, raw, tag string, values map[string]string) error {
+// noInnder indicates whether the element of the given XML name only contains string but no more inner element.
```
oliverchang, line 836: nit: noInner is not spelled correctly. Also, since we are adding a comment here, can we document the function fully as well? 

i.e.

```
// writeString writes ..... 
```

internal/resolution/manifest/maven.go
```diff
@@ -837,7 +844,7 @@ func writeString(enc *xml.Encoder, raw, tag string, values map[string]string) er
 		if err != nil {
 			return err
 		}
-		if tt, ok := token.(xml.StartElement); ok && tt.Name.Local != tag {
+		if tt, ok := token.(xml.StartElement); ok && noInner(tt.Name.Local) {
```
oliverchang, line 847: All the uses of `noInner` above seems to be checking a string against a blocklist. Can we simplify `noInner` to just be a slice instead? Then inside `writeString`, we just do `slices.Contains`.

oliverchang: Alternatively, since this seems a bit brittle (we need to know all the cases where we don't want to trim out, is this something we can more reliably trim out by tracking nesting levels? 

e.g.. only replace values 1 level down.

cuixq: considering Maven POM models is quite well defined so we have confidence in knowing all the tag names and their levels except for property names (they can be any text) so I think it should be fine.

cuixq: now I rethink this - maybe this should be an improvement here (property names may be texts that we do not allow for trimming), though I need to investigate and understand more XML...

michaelkedar:
> - when merging parents, does not check parent.groupId since Maven allows this to happen;

I don't think this is quite the correct behaviour. e.g. if you say your parent's groupId is `org.foo` but the parent pom says `org.bar` Maven throws an error.

The problem we are seeing is that the groupId is inherited from the parent if it is missing. So we'd need to add some logic somewhere like
```golang
if proj.GroupID == "" {
  proj.GroupID = parent.GroupID
}
  ```

cuixq:
thanks for checking that @michaelkedar !

oliverchang (approved):

PR #1175, opened 2024-08-12
fix(workflow): correct permission name

andrewpollock:
Is there a way to catch errors like this, pre-merge?

### DIFF
--- a/.github/workflows/staleness.yml
+++ b/.github/workflows/staleness.yml
@@ -1,6 +1,6 @@
 name: "Close stale issues and PRs"
 
-permissions: all-read
+permissions: read-all
 
 on:
   schedule:


G-Rath (approved):
I think https://github.com/rhysd/actionlint would catch this

andrewpollock:
> I think https://github.com/rhysd/actionlint would catch this

Great minds think alike :-)
I just independently tracked this down, and yes:

```
$ ~/go/bin/actionlint | grep staleness.yml
.github/workflows/staleness.yml:3:14: "all-read" is invalid for permission for all the scopes. available values are "read-all" and "write-all" [permissions]
```

another-rex (approved):

PR #1178, opened 2024-08-13
chore: add PR title check to follow Git commit convention

hogo6002:
Enforces PR naming to follow with the Git [commit convention](https://www.conventionalcommits.org/en/v1.0.0/)
- Uses https://github.com/amannn/action-semantic-pull-request to check
- Fails to comply will result in the PR title check failing:
image

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -1,0 +1,34 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: OSV PR format check
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+
+permissions:
+  pull-requests: read
+
+jobs:
+  title-check:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


cuixq (approved):

.github/workflows/pr-check.yml
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: OSV PR format check
+
+on:
+  pull_request:
```
another-rex, line 18: Hmm looking at the documentation https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#event-triggers, does this need to be `pull_request_target` instead?

hogo6002: Maybe I'm wrong, but I think `pull_request` is sufficient and better for us. Considering it only provides [read-only permissions on forked repos](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/), so it's safer (and we only need read permissions).

I think that GitHub Actions has features for editing pull requests, so they recommend using `pull_request_target `to get full permissions (including writing).

(My this PR is merging from a forked repo and has successfully passed the title check action, so I think it should be fine.)

another-rex: Ah that makes sense, can you add a comment here saying pull_request_target is only necessary for editing just to remind us in the future? Otherwise LGTM

another-rex (approved):

.github/workflows/pr-check.yml
hogo6002: Updated, thanks!

.github/workflows/pr-check.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  # `pull_request_target` is only required when editing PRs from forks.
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+
+permissions:
```
andrewpollock, line 25: I think based on my recent faffing around, you can move this out of the top level and into the specific job that requires it for more precise definition.

Then explicitly set the permissions at this level to `{}`

hogo6002: I do see our other Git action yaml file set the permission at each job level. I thought this is just because there are multiple jobs with different permissions required. 

Is there any benefit of setting the PR `read` permission at the job level instead of the top level for PR check workflow? (I am assuming all jobs within pr_check.yml would require PR read permission)

oliverchang (approved):

Issue #1184, opened 2024-08-16
Unexpected diff when writing XML


cuixq:
Texts are escaped to `&#x9;` when we write XML in Maven updater.
Go does escaping in `EncodeToken`: https://github.com/golang/go/blob/master/src/encoding/xml/marshal.go#L223

cuixq:
Not only tabs are escaped but also other characters: https://github.com/golang/go/blob/master/src/encoding/xml/xml.go#L1916

cuixq:
Besides the escaped texts, due to issue https://github.com/golang/go/issues/21399 self-closing tags are not encoded as expected. 

We may consider refactoring how to write XMLs: instead of calling `xml.Encode()`, write the content directly to avoid unexpected behaviour.

oliverchang:
Are there remaining issues to track here, or can we close this @cuixq  ?

cuixq:
I noticed one issue of multiple-line self-closing tag encoded into one line:
```
<a 
something
something />
```
is now
```
<a something something />
```

this seems low priority though.

cuixq:
Going to track this in separate issues.

PR #1190, opened 2024-08-19
fix: unescape tabs before writing to pom.xml

cuixq:

When encoding tokens to pom.xml, tabs are escaped and this is not what we want.
In this PR, before writing to pom.xml, we replace all escaped tabs with unescaped characters.

### DIFF
--- a/internal/resolution/manifest/__snapshots__/maven_test.snap
+++ b/internal/resolution/manifest/__snapshots__/maven_test.snap
@@ -1,4 +1,148 @@
 
+[TestMavenReadWrite - 1]
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.mycompany.app</groupId>
+  <artifactId>my-app</artifactId>
+  <version>1.0</version>
+
+  <name>my-app</name>
+  <!-- FIXME change it to the project's website -->
+  <url>http://www.example.com</url>
+
+  <parent>
+    <groupId>org.parent</groupId>
+    <artifactId>parent-pom</artifactId>
+    <version>1.1.1</version>
+    <relativePath>../parent/pom.xml</relativePath>
+  </parent>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.source>1.7</maven.compiler.source>
+    <maven.compiler.target>1.7</maven.compiler.target>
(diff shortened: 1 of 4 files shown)


oliverchang (commented):
nice! just have some questions:
- will this result in unrelated diffs in the result? 
- is there a bug in the Go encoding/xml package, or some more context you can give on why this occurs?

cuixq:
> nice! just have some questions:
> 
> * will this result in unrelated diffs in the result?

Yes, but I think this is definitely better than having `&#x9;` in the result.

> * is there a bug in the Go encoding/xml package, or some more context you can give on why this occurs?

I believe this is due to `EncodeToken` escaping the text: https://github.com/golang/go/blob/master/src/encoding/xml/marshal.go#L222

Also from https://github.com/golang/go/issues/47464, I don't think Go refers this as a bug.

oliverchang:
> > nice! just have some questions:
> > 
> > * will this result in unrelated diffs in the result?
> 
> Yes, but I think this is definitely better than having `&#x9;` in the result.

This is a bit of a hack, but is it possible to replace this with a sentinel value. e.g. "REPLACEWITHTAB", and then replace this ourselves when we're about to write it to the output?

> 
> > * is there a bug in the Go encoding/xml package, or some more context you can give on why this occurs?
> 
> I believe this is due to `EncodeToken` escaping the text: https://github.com/golang/go/blob/master/src/encoding/xml/marshal.go#L222
> 
> Also from [golang/go#47464](https://github.com/golang/go/issues/47464), I don't think Go refers this as a bug.

Got it, so it's correct but non-ideal behaviour.

cuixq:
I think replacing with some sentinel value may not work since this will make the pom.xml no longer valid. However this hints me that maybe we could let the texts be escaped during writing and replace all escaped `&#x9;` by `\t` at the end. I will investigate the possibility.

oliverchang:
> I think replacing with some sentinel value may not work since this will make the pom.xml no longer valid. However this hints me that maybe we could let the texts be escaped during writing and replace all escaped `&#x9;` by `\t` at the end. I will investigate the possibility.

yeah, great point. we can just unescape if we are able to hook into that part of the output.

internal/resolution/manifest/__snapshots__/maven_test.snap
```diff
@@ -26,7 +170,9 @@
     <maven.compiler.source>1.7</maven.compiler.source>
     <maven.compiler.target>1.7</maven.compiler.target>
     <junit.version>4.13.2</junit.version>
-    <zeppelin.daemon.package.base>../bin</zeppelin.daemon.package.base>
+&#x9;  <zeppelin.daemon.package.base>
+&#x9;    ../bin
+&#x9;  </zeppelin.daemon.package.base>
```
michaelkedar, line 175: There seems to be escaped tabs in this test snapshot?

cuixq: this is expected - this is result from `write()` and then we unescape all the tabs before writing to the actual file in `Write()`

oliverchang (approved):

michaelkedar (approved):

PR #1191, opened 2024-08-20
feat(guided remediation): add `--upgrade-config` flag

### LINKED ISSUE #1177: Allow control over allowed individual package upgrade versions in non-interactive guided remediation
Currently, `osv-scanner fix` has two flags:
- `--disallow-major-upgrades` (bool), which prevents major updates in all packages
- `--disallow-package-upgrades` (list of string), which prevents all updates to the listed packages.

We should replace these flags with per-package allowed update levels e.g. `--allowed-upgrades=foo=none,bar=major,baz=minor`.

Syntax might be a bit tricky and needs some thought:
- Should this be `allowed-upgrades` or `disallowed-upgrades`?
- Is `pkg1=type1,pkg2=type2` okay syntax for all ecosystems? (i.e. does any ecosystem allow for '=' in a package name?)
- How to apply rules to all packages (e.g. `all=major` could refer to the package named `all`)
- What can be allowed.? e.g. `major`/`minor`/`patch`/`none`
  - Consider what this would mean for non-semver ecosystems
  - Maybe allow for some regex rules to deal with e.g. [com.google.guava:guava](https://deps.dev/maven/com.google.guava%3Aguava) having versions that end with `-jre` and `-android`

michaelkedar:
closes #1177

- Adds `--upgrade-config` flag for configuring allowed upgrades on a per-package basis.
- Hide & deprecate previous `--disallow-major-upgrades` and `--disallow-package-upgrades` flags.
- Update docs & example script to use new flag.

### DIFF
--- a/docs/guided-remediation.md
+++ b/docs/guided-remediation.md
@@ -235,7 +235,9 @@ The following flag may be used to limit the patches allowed for your dependencie
   - `patch` allows for updates `>=1.2.3, <1.3.0`
   - `none` disallows any updates
 
-  If `package-name:` is omitted, `level` is applied to all packages. Default is `--upgrade-config=major`.
+  If `package-name:` is omitted, `level` is applied to all packages. The specific `package-name:level` will take precedence over the general `level` (e.g. specifying both `minor` `pkg:none` will use `none` as the allowed level for `pkg`).
+
+  Default behaviour is `--upgrade-config=major`.
 
   Example usage:
 
--- a/docs/guided-remediation.md
+++ b/docs/guided-remediation.md
@@ -227,16 +227,18 @@ A vulnerability is only considered if it satisfies all the conditions set by the
 The following flag may be used to limit the patches allowed for your dependencies:
 
 - `--upgrade-config=<[package-name:]level>` Sets the maximum upgrade level allowed for a package. Can be repeated for multiple packages.
-  
-  `level` is the SemVer component to allow updates to, can be one of `major`, `minor`, `patch`, or `none`. e.g. If a package was at version `1.2.3` 
+
+  `level` is the SemVer component to allow updates to, can be one of `major`, `minor`, `patch`, or `none`. e.g. If a package was at version `1.2.3`
+
   - `major` allows for updates to any version `>=1.2.3`
   - `minor` allows for updates `>=1.2.3, <2.0.0`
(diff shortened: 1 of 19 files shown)


docs/guided-remediation.md
```diff
@@ -224,10 +224,24 @@ A vulnerability is only considered if it satisfies all the conditions set by the
...
+- `--upgrade-config=<[package-name:]level>` Sets the maximum upgrade level allowed for a package. Can be repeated for multiple packages.
+
+  `level` is the SemVer component to allow updates to, can be one of `major`, `minor`, `patch`, or `none`. e.g. If a package was at version `1.2.3`
+
+  - `major` allows for updates to any version `>=1.2.3`
+  - `minor` allows for updates `>=1.2.3, <2.0.0`
+  - `patch` allows for updates `>=1.2.3, <1.3.0`
+  - `none` disallows any updates
+
+  If `package-name:` is omitted, `level` is applied to all packages. Default is `--upgrade-config=major`.
```
oliverchang, line 238: Can we add some detail to describe precedence? 

i.e. our behavior is to take the most specific setting right? if a specific package is specified in a config, that takes precedence over a global config.

oliverchang (approved):
nice!

docs/guided-remediation.md
cuixq: +1 - it is not clear to me whether the specific settings overwrite the global ones until I see the example usage.

internal/remediation/upgrade/config_test.go
cuixq, line 1: maybe add some test for Maven names like `foo:bar` considering the presence of `:` in the name

cuixq (commented):
LGTM!

docs/guided-remediation.md
michaelkedar: Added some explanation w.r.t precedence here

internal/remediation/upgrade/config_test.go
michaelkedar: The logic for parsing the `pkg:level` strings is in `parseUpgradeConfig()` in `fix/main.go` - I need to work out a test for that function

michaelkedar: This as a test case in `fix/main_test.go`

cuixq (approved):

PR #1194, opened 2024-08-21
fix: stop finding more parent pom if the path is empty

cuixq:
If we don't break when the parent path is empty, the current parent key keeps the same so an error of parent cycle is returned.


internal/resolution/manifest/maven.go
```diff
@@ -303,27 +303,31 @@ func (MavenManifestIO) Write(df lockfile.DepFile, w io.Writer, patch ManifestPat
...
-			if mavenutil.ProjectKey(proj) != parent.ProjectKey || proj.Packaging != "pom" {
-				// This is not the project that we are looking for, we should fetch from upstream
-				// that we don't have write access so we give up here.
-				break
-			}
-			origin := mavenOrigin(mavenutil.OriginParent, parentPath)
-			specific.OriginalRequirements = append(specific.OriginalRequirements, buildOriginalRequirements(proj, origin)...)
-			specific.Properties = append(specific.Properties, buildPropertiesWithOrigins(proj, origin)...)
-			parent = proj.Parent
+		f, err := os.Open(currentPath)
```
oliverchang, line 312: do we need to close this somewhere?

oliverchang (approved):

michaelkedar (approved):

PR #1201, opened 2024-08-22
fix: correct for breaking change in glamour v0.8.0

andrewpollock:
The styles have been moved into their own package


andrewpollock:
Once this is merged and https://github.com/google/osv-scanner/pull/1187 gets rebased, I think that PR should pass muster...

cuixq (approved):

PR #1206, opened 2024-08-25
perf: ignored packages should be filtered out before scanning

giovanni-bozzano:
Currently, vulnerabilities in packages that are ignored in the config file are filtered out after the scanning process.
To optimize the scanning process, these packages should be preemptively ignored.


another-rex (approved):
LGTM, thanks!

G-Rath (approved):
🚀

cmd/osv-scanner/__snapshots__/main_test.snap
```diff
@@ -198,6 +202,7 @@ Scanned <rootdir>/fixtures/sbom-insecure/alpine.cdx.xml as CycloneDX SBOM and fo
 [TestRun/PURL_SBOM_case_sensitivity_(local) - 1]
 Scanning dir ./fixtures/sbom-insecure/alpine.cdx.xml
 Scanned <rootdir>/fixtures/sbom-insecure/alpine.cdx.xml as CycloneDX SBOM and found 14 packages
+Loaded filter from: <rootdir>/fixtures/sbom-insecure/osv-scanner.toml
```
another-rex, line 205:
```suggestion
```

Issue #1124, opened 2024-07-16
Support ignoring a package only from license scanning


shahar-h:
Today `[[PackageOverrides]]` can be configured to ignore a package entirely from both vulnerability and license scanning using `ignore` field:
```toml
[[PackageOverrides]]
# The package name, version, and ecosystem to match against
name = "lib"
# If version is not set or empty, it will match every version
version = "1.0.0"
ecosystem = "Go"
# Ignore this package entirely, including license scanning
ignore = true
```

However, there are cases where we only want to ignore a package from license scanning, for example when the package is only used in tests we may still want to detect vulnerabilities but ignore license since this package is not compiled in our application.

G-Rath:
You should be able to do this using `license.override` to override the license to one you consider allowed, e.g.

```toml
[[PackageOverrides]]
# The package name, version, and ecosystem to match against
name = "lib"
# If version is not set or empty, it will match every version
version = "1.0.0"
ecosystem = "Go"
# Ignore the license of this package
license.override = ["IGNORED"]
``

and then call `osv-scanner` with `--experimental-licenses IGNORED`.

While a little messier than being to do say `license.ignore = true`, I don't know if there's enough of an advantage as I'd expect we would still want to at least output that ignoring like we do when a license is overridden so you'd still have the same amount of output..

@another-rex @cuixq what do you think?

shahar-h:
> You should be able to do this using `license.override` to override the license to one you consider allowed, e.g.
> 
> ```toml
> [[PackageOverrides]]
> # The package name, version, and ecosystem to match against
> name = "lib"
> # If version is not set or empty, it will match every version
> version = "1.0.0"
> ecosystem = "Go"
> # Ignore the license of this package
> license.override = ["IGNORED"]
> ```
> 
> and then call `osv-scanner` with `--experimental-licenses IGNORED`.
> 
> While a little messier than being to do say `license.ignore = true`, I don't know if there's enough of an advantage as I'd expect we would still want to at least output that ignoring like we do when a license is overridden so you'd still have the same amount of output..
> 
> @another-rex @cuixq what do you think?

Currently I do something similar - I override the package license to an allowed one like `MIT`, but having an explicit `license.ignore` flag would make it more readable.

PR #1214, opened 2024-08-29
feat: support composite-based package overrides

### LINKED ISSUE #1155: Is there a way to ignore all vulnerabilities for a given lockfile?
I'm currently having an `osv-scanner.toml` file next to the `Cargo.lock` file for which I don't care about vulnerabilities:

```toml
# This crate is used for benchmarks only. Those alerts can't be fixed because one of the comparison
# point, namely wasm3, is dead although it is the best benchmark reference so far.
IgnoredVulns = [
  { id = "GHSA-crf8-h2wq-2h9x" },
  { id = "GHSA-g98v-hv3f-hcfr" },
  { id = "GHSA-gq4p-4hxv-5rg9" },
  { id = "GHSA-r7qv-8r2h-pg27" },
  { id = "RUSTSEC-2021-0139" },
]
```

Ideally, I would like something like:

```toml
# This crate is used for benchmarks only. Those alerts can't be fixed because one of the comparison
# point, namely wasm3, is dead although it is the best benchmark reference so far.
ignore = true
```

This is more maintenable and describes the intent better.

I tried to find the schema of the `osv-scanner.toml` file, but only found the few examples in https://google.github.io/osv-scanner/configuration/.

Thanks!

### LINKED ISSUE #1211: Add dependency groups filter for PackageOverrides config
Hi,

I think it would be nice if the `PackageOverrides` config would support matches on the package dependency groups (`dev`, `optional`, etc). The simplest solution would be a simple exact match like it is used for `name` or `ecosystem`.

Examples:

# Ignore a dependency group entirely

This may be useful for projects that don’t care about any issues in a certain dependency group.

```toml
[[PackageOverrides]]
dependency_group = "dev"
ignore = true
```

# Ignore a package only if it is in a certain dependency group

This is useful if vulnerabilities for a package should only be ignored as long as the package is part of a certain dependency group. Example: One might want to ignore a package if it is a `dev` dependency but it should raise errors if it is a production dependency.

```toml
[[PackageOverrides]]
name = "axios"
ecosystem = "npm"
dependency_group = "dev"
ignore = true
```

Thanks for your work on osv-scanner!

Cheers

G-Rath:
This rewrites the package overrides logic to be composition based, granting a lot more flexibility:

```
# ignore everything
[[PackageOverrides]]
ignore = true

# ignore everything in this group
[[PackageOverrides]]
group = "dev"
ignore = true

# ignore everything in this ecosystem
[[PackageOverrides]]
ecosystem = "go"
ignore = true

# ignore all packages named "axios" regardless of ecosystem or group
[[PackageOverrides]]
name = "axios"
ignore = true

# ignore all packages named "axios" in the npm ecosystem that are in the dev group
[[PackageOverrides]]
name = "axios"
ecosystem = "npm"
group = "dev"
ignore = true

# ... and so on
```

While some of these might seem a bit extreme, ultimately I think this is probably the way to go as the logic itself is very straightforward and it gives a lot more power to the people.

Since `config` is a public package, I've had to deprecated the related existing public methods and there's a bit of naming & structural yuck but I figure that's not a big deal since v2 is right around the corner and again the logic itself is very straightforward.

Resolves #1211
Resolves #1155

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -49,7 +49,7 @@ effectiveUntil = 2022-11-09 # Optional exception expiry date, after which the ov
 reason = "abc" # Optional reason for the override, to explain why it was added
 ```
 
-Each override entry will be checked against every package being scanned, and applied if all the configured fields match. This enables both very broad and very specific overrides based on your needs:
+Overrides are applied if all the configured fields match, enabling you to create very broad or very specific overrides based on your needs:
 
 ```toml
 # ignore everything
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -52,26 +52,26 @@ reason = "abc" # Optional reason for the override, to explain why it was added
 Overrides are applied if all the configured fields match, enabling you to create very broad or very specific overrides based on your needs:
 
 ```toml
-# ignore everything
+# ignore everything in the current directory
 [[PackageOverrides]]
 ignore = true
 
-# ignore everything in a particular group
+# ignore a particular group
 [[PackageOverrides]]
 group = "dev"
 ignore = true
 
-# ignore everything in a particular ecosystem
+# ignore a particular ecosystem
 [[PackageOverrides]]
 ecosystem = "go"
 ignore = true
 
-# ignore all packages named "axios" regardless of ecosystem or group
+# ignore packages named "axios" regardless of ecosystem or group
 [[PackageOverrides]]
 name = "axios"
 ignore = true
 
-# ignore all packages named "axios" in the npm ecosystem that are in the dev group
+# ignore packages named "axios" in the npm ecosystem that are in the dev group
 [[PackageOverrides]]
 name = "axios"
 ecosystem = "npm"
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -29,22 +29,54 @@ reason = "No external http servers are written in Go lang."
 
 Ignoring a vulnerability will also ignore vulnerabilities that are considered aliases of that vulnerability.
 
-## Override specific package
+## Override packages
 
-To ignore a specific a package, or manually set its license, enter the package name and ecosystem under the `PackageOverrides` key.
+You can specify overrides for particular packages to have them either ignored entirely or to set their license using the `PackageOverrides` key:
 
 ```toml
 [[PackageOverrides]]
-# The package name, version, and ecosystem to match against
+# One or more fields to match each package against:
 name = "lib"
-# If version is not set or empty, it will match every version
 version = "1.0.0"
 ecosystem = "Go"
-# Ignore this package entirely, including license scanning
+group = "dev"
+
+# Actions to take for matching packages:
+ignore = true # Ignore this package entirely, including license scanning
+license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not ignored
+
+effectiveUntil = 2022-11-09 # Optional exception expiry date, after which the override will no longer apply
+reason = "abc" # Optional reason for the override, to explain why it was added
+```
+
+Each override entry will be checked against every package being scanned, and applied if all the configured fields match. This enables both very broad and very specific overrides based on your needs:
+
+```toml
+# ignore everything
+[[PackageOverrides]]
 ignore = true
-# Override the license of the package
-# This is not used if ignore = true
-license.override = ["MIT", "0BSD"]
-# effectiveUntil = 2022-11-09 # Optional exception expiry date
-reason = "abc"
+
+# ignore everything in a particular group
+[[PackageOverrides]]
+group = "dev"
+ignore = true
+
+# ignore everything in a particular ecosystem
+[[PackageOverrides]]
+ecosystem = "go"
+ignore = true
+
+# ignore all packages named "axios" regardless of ecosystem or group
+[[PackageOverrides]]
+name = "axios"
+ignore = true
+
+# ignore all packages named "axios" in the npm ecosystem that are in the dev group
+[[PackageOverrides]]
+name = "axios"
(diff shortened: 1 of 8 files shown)


G-Rath:
(note that the docs should probably be updated, but I want to get confirmation that we're happy with this general change first)

G-Rath:
Note that this should also be usable to resolve #1124 - I think technically you could do that right now by e.g. "consider all packages in the `test` group as having a license we're happy with" but it might be better to have a special "skip" type value that does actually mean the overridden package is completely omitted from license checking code paths...

cuixq (approved):
LGTM!

pkg/osvscanner/osvscanner.go
```diff
@@ -967,7 +967,17 @@ func filterIgnoredPackages(r reporter.Reporter, packages []scannedPackage, confi
 	out := make([]scannedPackage, 0, len(packages))
 	for _, p := range packages {
 		configToUse := configManager.Get(r, p.Source.Path)
-		if ignore, ignoreLine := configToUse.ShouldIgnorePackageVersion(p.Name, p.Version, string(p.Ecosystem)); ignore {
+		if ignore, ignoreLine := configToUse.ShouldIgnorePackage(
```
another-rex, line 970: nit: Separate this out from the if statement as it is quite long, and just have it on it's own line.

another-rex (approved):
LGTM! good idea making it more general. I'm not sure about using the models.PackageVulns type here, but that's probably something to consider in the v2 rework of the types, (we have too many types that roughly represent the same thing currently), so happy to merge as is.

docs/configuration.md
```diff
@@ -29,22 +29,54 @@ reason = "No external http servers are written in Go lang."
...
+
+# Actions to take for matching packages:
+ignore = true # Ignore this package entirely, including license scanning
+license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not ignored
+
+effectiveUntil = 2022-11-09 # Optional exception expiry date, after which the override will no longer apply
+reason = "abc" # Optional reason for the override, to explain why it was added
+```
+
+Each override entry will be checked against every package being scanned, and applied if all the configured fields match. This enables both very broad and very specific overrides based on your needs:
```
another-rex, line 52: When ignoring vulnerabilities, we only apply the config to the lockfiles that is in the same directory, is this the same behavior here? Or is it applying to everything?

docs/configuration.md
```diff
@@ -29,22 +29,54 @@ reason = "No external http servers are written in Go lang."
...
+
+```toml
+# ignore everything
+[[PackageOverrides]]
+ignore = true
```
another-rex, line 57: Is the use case for this just scanning for licenses? In that case can we specify that in the comment above.

docs/configuration.md
G-Rath: I've not touched the logic that handles deciding if it should check if an override applies for a package (none of the code I touched dealt with paths, among other things) - so that behaviour should remain unchanged, and also wasn't mentioned in the original docs.

In saying that, I assume what probably triggered this was the "will be checked against every package being scanned" which I agree could make people think it applies to... well _every_ package being scanned 😅 

I'll reword this to remove that ambiguity

docs/configuration.md
G-Rath: I'm not quite sure what you mean - I've only changed the logic around when an override should be applied to a particular package, not the supported "actions"; so the behaviour should be whatever `ignore` did previously...?

G-Rath:
@another-rex 

> I'm not sure about using the models.PackageVulns type here, but that's probably something to consider in the v2 rework of the types, (we have too many types that roughly represent the same thing currently), so happy to merge as is.

I think we're on exactly the same page - overall, my view here is naming wise this suffers because of already known issues which are a major epic to be addressed in v2 and that the more important thing is this change introduces a showcase of how the configuration could be improved that is very useful to have landed _before_ v2 so it can be part of shaping v2 itself.

A good example of that imo is with files: it could be very reasonable to add the ability to ignore specific files but that's hard because `models.PackageVulns` doesn't have that information - so if the new interface(s) for v2 included that (which iirc I think it will?), we'd be able to very easily support that as a field here.

docs/configuration.md
G-Rath: fwiw, my expectation is that this enables #1155

docs/configuration.md
another-rex: Just had another look at the logic because I wasn't too sure what the current behavior is, but yes you're right we do only affect lockfiles in the same directory as the ignore file.

(we probably can add some tests specifically for nested lockfiles+configs, though it'll conflict with our osv-scanner-action and scorecard..., I'll add an issue for this).

docs/configuration.md
another-rex: I meant why someone would want to ignore everything. This was before I remembered that it only applies to the file in the same directory, so this will be pretty useful. Can you change the comment above this to specify that this is ignoring everything in the same directory?

another-rex: (we should probably also change ignore to ignoreVulns, and deprecate ignore, as it is ambiguous as to whether we are still doing license scanning (we are), we can do that in a separate PR though. I'll make an issue for this.)

pkg/osvscanner/osvscanner.go
G-Rath: done - I went with keeping the function in the if and using a variable for the struct as I think that's a bit more correct: in the long-run, we expect to be passing a whole "thing" to this function (like what we're already doing in `vulnerability_result.go`), so it's the `models.PackageVuln` part that's a temp workaround

another-rex (approved):

Issue #1215, opened 2024-08-30
Indentation when writing new dependencies to Maven pom.xml


cuixq:
New dependencies should have more indentation as shown below.

```
<dependencyManagement>
    <dependencies>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-codec-http</artifactId>
      <version>4.1.108.Final</version>
    </dependency>
      <dependency>
          ...
      </dependency>
    </dependencies>
</dependencyManagement>
```

PR #1216, opened 2024-08-30
fix(internal): encode XML tokens without escaping

cuixq:

Currently, texts are escaped in `xml.EncodeToken()` and this generates unexpected diff when writing updates to XML.

This PR adds `func encodeToken()` that writes `CharData` directly to avoid escaping, and all buffered XML is flushed after encoding.

This PR also fixes https://github.com/google/osv-scanner/issues/1215 by adding two more space to indentation.


internal/resolution/manifest/maven.go
```diff
@@ -716,17 +704,30 @@ func write(raw string, w io.Writer, patches MavenPatches) error {
...
 			return err
 		}
 	}
 
 	return nil
 }
 
+// encodeToken encodes a XML token without escaping text.
+// For a CharData token, writew the token directly to avoid escaping.
+// For all other token types, usew xml.EncodeToken().
```
michaelkedar, line 717: typos in this comment

michaelkedar (approved):
LGTM

PR #1219, opened 2024-09-02
fix(guided remediation): Add special handling for specific Maven packages

### LINKED ISSUE #1183: Handle special versioning of `com.google.guava:guava`
[com.google.guava:guava](https://central.sonatype.com/artifact/com.google.guava/guava/versions) has two variants: `jre` and `andriod`.

Currently we do not consider variant when updating. A more correct behaviour is only updating to a newer version of the same variant.

michaelkedar:
Add some special cases for overriding versions of Maven packages to prevent undesirable outcomes from following the correct versioning rules:
- `com.google.guava:guava`: Preserve the version's 'flavor' (`-android` or `-jre`) when updating (Closes #1183)
- Apache `commons-*`: Sort the ancient CalVer versions (e.g.`20021128.082114`) before the newer versions (`1.0`).

The logic from this workaround should probably also be added to the updater, and also the other remediation strategies might want to make use of similar workarounds.


internal/remediation/override.go
```diff
@@ -279,6 +279,58 @@ func getVersionsGreater(ctx context.Context, cl client.DependencyClient, vk reso
 	return versions[offset:], nil
 }
 
+// comparisonFunctionWithWorkarounds returns a version comparison function with special behaviour for specific packages,
+// producing more desirable ordering using non-standard comparison.
+func comparisonFunctionWithWorkarounds(vk resolve.VersionKey) func(resolve.Version, resolve.Version) int {
```
cuixq, line 284: can we make this re-usable (for example `osv-scanner update` may need this as well)?

internal/remediation/override.go
```diff
@@ -279,6 +279,58 @@ func getVersionsGreater(ctx context.Context, cl client.DependencyClient, vk reso
...
+			return -1
+		}
+	}
+
+	if vk.System == resolve.Maven && strings.HasPrefix(vk.Name, "commons-") {
+		// Old versions of apache commons- libraries (commons-io:commons-io, commons-math:commons-math, etc.)
+		// used date-based versions (e.g. 20040118.003354), which naturally sort after the more recent semver versions.
+		// We manually force the date versions to come before the others to prevent downgrades.
+		return func(a, b resolve.Version) int {
+			aCal := strings.HasPrefix(a.Version, "200")
```
cuixq, line 316: I think it's better to check if the first 8 numbers represent a date. https://pkg.go.dev/time#Parse

michaelkedar: It seems [pretty inefficient](https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/time/format.go;l=1044) to keep calling `Time.Parse("20060102", a.Version[:8])` repeatedly, especially since we don't actually need to use the parsed year/month/day.

I've taken a quick look at the `commons-*` packages on Maven central, and the date versions seem to only have been used from the years 2002-2005 (which seems to be up to the release of Maven version 2).
I don't think we'll see any of these packages versioned like this post 2010, and I also doubt we'd ever see the major versions of these packages reach up to `200.0.0`.

internal/remediation/override.go
michaelkedar: Added as a TODO, it'll probably need to be moved to avoid circular import problems.

internal/remediation/override.go
```diff
@@ -279,6 +279,59 @@ func getVersionsGreater(ctx context.Context, cl client.DependencyClient, vk reso
...
+
+			if aIsAndroid == wantAndroid {
+				return 1
+			}
+
+			return -1
+		}
+	}
+
+	if vk.System == resolve.Maven && strings.HasPrefix(vk.Name, "commons-") {
```
oliverchang, line 312: Are these versions in maven-metadata.xml as well ? 

e.g. I see the problematic  [20030203.000550](https://deps.dev/maven/commons-io%3Acommons-io/20030203.000550) for `commons-io:commons-io`, but it doesn't exist in [maven-metadata.xml](https://repo.maven.apache.org/maven2/commons-io/commons-io/maven-metadata.xml)

Is checking maven-metadata.xml the proper fix for this here? If so, let's add a TODO.

oliverchang (approved):
LGTM with a question

internal/remediation/override.go
cuixq: I think we can filter out most of the versions by checking if the length is greater than 8, or maybe even implement our own simplified checker. if all the versions are from year 2002-2005, add a comment about this specialty.

internal/remediation/override.go
michaelkedar: > Are these versions in maven-metadata.xml as well ?

In nearly all cases, no. I did find [`commons-primitives:commons-primitives`](https://repo1.maven.org/maven2/commons-primitives/commons-primitives/maven-metadata.xml) and [`commons-sql:commons-sql`](https://repo1.maven.org/maven2/commons-sql/commons-sql/maven-metadata.xml) with both formats, and [`commons-compress:commons-compress`](https://repo1.maven.org/maven2/commons-compress/commons-compress/maven-metadata.xml) has only date-based versions. So it's not guaranteed.

> Is checking maven-metadata.xml the proper fix for this here?

Checking the metadata file (which is a necessity for the native registry client) will usually exclude those versions from the list, but we don't do that for the deps.dev API.
If a `pom.xml` was already using e.g. `commons-io:commons-io` version `20030203.000550`, we still would probably want to upgrade it through `0.1` -> `1.4` -> `2.16.1`.

That said, I don't think I've seen any OSV/GHSA records that correctly list those versions as vulnerable :shrug:

internal/remediation/override.go
michaelkedar: I've added a comment to the code justify checking only the first 3 characters.

Checking it properly would ultimately come down to doing a similar check, and I'd rather not over-engineer it for a case that doesn't exist.

cuixq (approved):

internal/remediation/override.go
oliverchang: Ack, so we can't reliably rely on this in any case. Also great point re still wanting to support the proper upgrade path.

PR #1223, opened 2024-09-04
refactor: use forked xml package for writing

cuixq:

We would like to modify Go's implementation on self-closing tags and this PR is the first step:
 - add `encoding/xml` to `internal/thirdparty` (also ignore this directory for lint)
 - make Maven manifest writing to use the internal version of `xml`
 - modify `xml` to not escape char data so we can call `xml.EncodeToken()`


internal/thirdparty/xml/xml.go
```diff
@@ -0,0 +1,2075 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package xml implements a simple XML 1.0 parser that
+// understands XML name spaces.
+package xml
```
oliverchang, line 7: can you add some comments here to explain why we forked this?

oliverchang (approved):

michaelkedar (approved):

PR #1224, opened 2024-09-04
refactor(guided remediation): Take `PreFetch` out of `DependencyClient`  interface and prevent repeated datasource network calls

michaelkedar:
What I mentioned in https://github.com/google/osv-scanner/pull/1207#discussion_r1733737395

Make `PreFetch` a standalone function that takes in a client that uses every `DependencyClient` method call.
Since the underlying datasources tend to use the same request for multiple methods, I've made a `requestCache` type that uses logic based on the [singleflight](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.8.0:singleflight/singleflight.go;l=91) package to prevent the same requests being made multiple times. I've simplified it a bit by skipping the bespoke handling of panics / `runtime.Goexit`.


cuixq (approved):
nice!

internal/resolution/client/client.go
```diff
@@ -2,28 +2,119 @@ package client
...
+			pbvk := node.GetVersionKey()
+			vk := resolve.VersionKey{
+				PackageKey: resolve.PackageKey{
+					System: resolve.System(pbvk.GetSystem()),
+					Name:   pbvk.GetName(),
+				},
+				Version:     pbvk.GetVersion(),
+				VersionType: resolve.Concrete,
+			}
+			go c.Requirements(ctx, vk)        //nolint:errcheck
```
oliverchang, line 99: Do we still need a TODO somewhere here to limit the number of goroutines we create?

oliverchang (approved):
nice!

PR #1228, opened 2024-09-06
fix: govulncheck calls on C code

### LINKED ISSUE #1220: [GitHub Action] scan fails on go projects that import C code
When using osv-scanner GitHub action to scan a go project that imports C code - the scan fails.
In example: Scanning the root `go.mod` file in https://github.com/envoyproxy/gateway fails with:
```log
Failed to run code analysis (govulncheck) on '/app/go.mod' because govulncheck: loading packages: 
There are errors with the provided package patterns:

-: # github.com/containers/storage/pkg/unshare
unshare.c:21:10: fatal error: linux/limits.h: No such file or directory
   21 | #include <linux/limits.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
/root/go/pkg/mod/github.com/containers/storage@v1.55.0/pkg/unshare/getenv_linux_cgo.go:14:8: could not import C (no metadata for C)
/root/go/pkg/mod/github.com/containers/storage@v1.55.0/drivers/btrfs/btrfs.go:19:8: could not import C (no metadata for C)
-: # github.com/containers/storage/drivers/quota
/root/go/pkg/mod/github.com/containers/storage@v1.55.0/drivers/quota/projectquota_supported.go:18:10: fatal error: linux/fs.h: No such file or directory
   18 | #include <linux/fs.h>
      |          ^~~~~~~~~~~~
compilation terminated.
/root/go/pkg/mod/github.com/containers/storage@v1.55.0/drivers/quota/projectquota_supported.go:42:8: could not import C (no metadata for C)
-: # github.com/containers/storage/pkg/loopback
/root/go/pkg/mod/github.com/containers/storage@v1.55.0/pkg/loopback/loop_wrapper.go:7:10: fatal error: linux/loop.h: No such file or directory
    7 | #include <linux/loop.h> // FIXME: present only for defines, maybe we can remove it?
      |          ^~~~~~~~~~~~~~
compilation terminated.
/root/go/pkg/mod/github.com/containers/storage@v1.55.0/pkg/loopback/loop_wrapper.go:18:8: could not import C (no metadata for C)
/root/go/pkg/mod/github.com/containers/storage@v1.55.0/drivers/copy/copy_linux.go:13:8: could not import C (no metadata for C)
```

In order to resolve this issue `CGO` needs to be disabled in GitHub action [Dockerfile](https://github.com/google/osv-scanner/blob/main/goreleaser-action.dockerfile) by adding `ENV CGO_ENABLED=0`.
See more info [here](https://semaphoreci.com/blog/govulncheck) under 'Govulncheck quirks' section.

another-rex:
Fix CGO and govulncheck not working well together. govulncheck does not work on C code anyway, so disabling CGO doesn't change any capabilities.

Fixes #1220


another-rex:
Actually, a better solution would be to set CGO_ENABLED=0 just before calling govulncheck itself in the code. I'll make that change.

oliverchang (approved):

PR #1235, opened 2024-09-10
feat(guided remediation): remediate unresolved dependency management vulns

michaelkedar:
Adds functionality to allow guided remediation to fix vulns in `dependencyManagement` dependencies that do not appear in the resolved dependency graph of the POM - useful for 'remediating' POMs without any actual dependencies.

I've accomplished this by checking if each of the original management dependencies (*excluding* those inherited from parents) appear in the graph after the initial resolution. If they're missing, I add them to the graph as direct dependencies (not resolving their transitive dependencies).

This behaviour is disabled by default, and I've added a `--maven-fix-management` flag to enable it. I was going to try combine this and `--ignore-dev` into a `--groups` flag but it seemed like it would be a bit too complicated.

### DIFF
--- a/cmd/osv-scanner/fix/main.go
+++ b/cmd/osv-scanner/fix/main.go
@@ -163,8 +163,8 @@ func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 			},
 			&cli.BoolFlag{
 				Category: vulnCategory,
-				Name:     "maven-ignore-management",
-				Usage:    "(pom.xml) ignore vulnerabilities in dependencyManagement packages that do not appear in the resolved dependency graph",
+				Name:     "maven-fix-management",
+				Usage:    "(pom.xml) also remediate vulnerabilities in dependencyManagement packages that do not appear in the resolved dependency graph",
 			},
 		},
 		Action: func(ctx *cli.Context) error {
@@ -240,7 +240,7 @@ func action(ctx *cli.Context, stdout, stderr io.Writer) (reporter.Reporter, erro
 	opts := osvFixOptions{
 		RemediationOptions: remediation.RemediationOptions{
 			ResolveOpts: resolution.ResolveOpts{
-				MavenManagement: !ctx.Bool("maven-ignore-management"),
+				MavenManagement: ctx.Bool("maven-fix-management"),
 			},
 			IgnoreVulns:   ctx.StringSlice("ignore-vulns"),
 			ExplicitVulns: ctx.StringSlice("vulns"),
--- a/cmd/osv-scanner/fix/main.go
+++ b/cmd/osv-scanner/fix/main.go
@@ -11,6 +11,7 @@ import (
 	"deps.dev/util/resolve"
 	"github.com/google/osv-scanner/internal/remediation"
 	"github.com/google/osv-scanner/internal/remediation/upgrade"
+	"github.com/google/osv-scanner/internal/resolution"
 	"github.com/google/osv-scanner/internal/resolution/client"
 	"github.com/google/osv-scanner/internal/resolution/datasource"
 	"github.com/google/osv-scanner/internal/resolution/lockfile"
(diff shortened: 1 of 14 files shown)


cmd/osv-scanner/fix/main.go
```diff
@@ -160,6 +161,11 @@ func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 				Name:     "ignore-dev",
 				Usage:    "ignore vulnerabilities affecting only development dependencies",
 			},
+			&cli.BoolFlag{
+				Category: vulnCategory,
+				Name:     "maven-ignore-management",
```
oliverchang, line 166: should we invert this to `maven-fix-management` ? and make the default ignore these?

oliverchang: Alternatively, should we generalize this to `--groups` instead as we discussed previously?

internal/resolution/resolve.go
```diff
@@ -104,6 +114,92 @@ func Resolve(ctx context.Context, cl client.ResolutionClient, m manifest.Manifes
...
+	if m.System() == resolve.Maven && opts.MavenManagement {
+		// Add a node & edge for each dependency in dependencyManagement that doesn't already appear in the resolved graph
+		manifestSpecific, ok := m.EcosystemSpecific.(manifest.MavenManifestSpecific)
+		if !ok {
+			return graph, errors.New("invalid MavenManifestSpecific data")
+		}
+
+		// Search through OriginalRequirements management dependencies in this pom only (not parents).
+		for _, req := range manifestSpecific.OriginalRequirements {
+			if req.Origin != mavenutil.OriginManagement {
```
cuixq, line 127: how about dependency management in profiles?

internal/resolution/resolve.go
```diff
@@ -104,6 +114,92 @@ func Resolve(ctx context.Context, cl client.ResolutionClient, m manifest.Manifes
...
+func resolvePostProcess(ctx context.Context, cl client.ResolutionClient, m manifest.Manifest, opts ResolveOpts, graph *resolve.Graph) (*resolve.Graph, error) {
+	if m.System() == resolve.Maven && opts.MavenManagement {
+		// Add a node & edge for each dependency in dependencyManagement that doesn't already appear in the resolved graph
+		manifestSpecific, ok := m.EcosystemSpecific.(manifest.MavenManifestSpecific)
+		if !ok {
+			return graph, errors.New("invalid MavenManifestSpecific data")
+		}
+
+		// Search through OriginalRequirements management dependencies in this pom only (not parents).
+		for _, req := range manifestSpecific.OriginalRequirements {
```
cuixq, line 126: maybe we can refactor `manifestSpecific.RequirementsForUpdates` for this purpose?

internal/resolution/resolve.go
```diff
@@ -104,6 +114,92 @@ func Resolve(ctx context.Context, cl client.ResolutionClient, m manifest.Manifes
 	return result, nil
 }
 
+func resolvePostProcess(ctx context.Context, cl client.ResolutionClient, m manifest.Manifest, opts ResolveOpts, graph *resolve.Graph) (*resolve.Graph, error) {
+	if m.System() == resolve.Maven && opts.MavenManagement {
```
cuixq, line 118: I prefer returning first if it does not require further processing (just my preference).

cmd/osv-scanner/fix/main.go
oliverchang: discussed offline re `--group` -- keeping this flag (but inverted) SGTM.

internal/resolution/resolve.go
michaelkedar: I was thinking in the future there may be some other post-processing steps (e.g. possibly dealing with npm `peerDependencies` here), so we wouldn't want to immediately return in that case.

internal/resolution/resolve.go
michaelkedar: I'm going to put this as a TODO, since writing to profiles is a bit iffy - https://github.com/google/osv-scanner/issues/1239#issuecomment-2345028438

internal/resolution/resolve.go
michaelkedar: Possibly, but it seems like a bigger change and I don't know how it would impact the maven suggester

cuixq: fair - maybe a TODO to explore later (also I don't know if we still want to continue working on `osv-scanner update`)

internal/resolution/resolve.go
cuixq: TODO may also include the dependencies in non-activated profiles? not only the dependency management

cuixq (approved):

PR #1241, opened 2024-09-11
fix: don't assume there's always a reason for a package being filtered out

G-Rath:
Currently if you don't provide a reason for an override the scanner will just output "because: " which looks weird and results in a trailing space in our snapshots which IDEs try to trim.

Since we want to encourage people to provide a reason, I've changed the scanner to explicitly state "(no reason provided)" in this situation

### DIFF
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -707,18 +707,18 @@ func filterPackageVulns(r reporter.Reporter, pkgVulns models.PackageVulns, confi
 
 				reason := ignoreLine.Reason
 
-				if reason != "" {
-					reason = " because: " + reason
+				if reason == "" {
+					reason = "(no reason given)"
 				}
 
 				// NB: This only prints the first reason encountered in all the aliases.
 				switch len(group.Aliases) {
 				case 1:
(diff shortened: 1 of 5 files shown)


cuixq (approved):
LGTM

pkg/osvscanner/osvscanner.go
```diff
@@ -704,14 +704,21 @@ func filterPackageVulns(r reporter.Reporter, pkgVulns models.PackageVulns, confi
 				for _, id := range group.Aliases {
 					ignoredVulns[id] = struct{}{}
 				}
+
+				reason := ignoreLine.Reason
+
+				if reason != "" {
+					reason = " because: " + reason
+				}
```
another-rex, line 712: Let's go with the "because of an unknown reason" here to encourage people adding their the reasoning in!

G-Rath: done - I've gone with "(no reason given)" as I think that makes more grammatical sense, and that the "(...)" should make it stand out a little more vs other stuff

another-rex (approved):

hogo6002 (approved):

PR #1242, opened 2024-09-11
fix: announce when a config file is invalid and exit with a non-zero code

G-Rath:
When looking into some other config stuff I realised the current loading logic just assumes that an error is because a config doesn't exist and silently falls back to the default one, when really it could be that there is a config but it's invalid


pkg/config/config.go
```diff
@@ -174,6 +175,9 @@ func (c *ConfigManager) Get(r reporter.Reporter, targetPath string) Config {
 	if configErr == nil {
 		r.Infof("Loaded filter from: %s\n", config.LoadPath)
 	} else {
+		if !errors.Is(configErr, errConfigFileNotFound) {
```
cuixq, line 178: why not check if the error is errConfigFileInvalid?

G-Rath: because it doesn't really matter much since there's only two errors but I felt it was technically a bit better to check like this because this is the scenario we've got logic to handle - so inversely, _any_ other error we're not expecting, even though right now the only other error could be `errConfigFileInvalid` 😅 

(we could actually omit one of those errors entirely, but I figured it doesn't have a big cost and probably a good practice to be following)

cuixq: my understanding is that the error message is about "invalid config file" so we should be checking the invalid config error.

G-Rath: I disagree per my previous comment - checking for that error explicitly would mean we would silently eat other errors that could come along; while very unlikely (as it'd probably be the result of a refactor which would have a high-chance of changing this code anyway), I still think it's better to do it this way given there no other significant cost i.e. its not faster or results in a smaller binary if we check for this error than the other one.

(tbh, really we should actually be checking `os.Open` returned an `os.ErrNotExist` in the first place 🤷 )

cuixq: I think a comment is needed to explain why we are check "not found" for "invalid config" error, and I agree we should check not found when opening the file instead of return it at the end.

G-Rath: I've refactored this to check for `os.ErrNotExist`, which I realised we could do in a simpler way

cuixq (approved):

another-rex (approved):

Issue #1226, opened 2024-09-05
Add vulnerabilities.ignore flag to just ignore vulnerabilties.


another-rex:
~~the `PackageOveride.ignore` field is ambiguous as to whether license scanning is still performed. Solution could be to rename `ignore` to `ignoreVulns`.~~

Ignore flag already ignores both vulnerabilities and license violations. We should have a second flag for just vulnerabilities. The opposite of #1124

G-Rath:
I thought that `ignore` had it ignored from both license scanning and vulnerabilities?

fwiw, I'd recommend doing something like `vulnerabilities.ignore` to mirror `license.overrides`

another-rex:
Huh you are right, I am just very confused at the behavior of our configs. I'll update this issue to be to introduce a just ignore vulns option.

PR #1243, opened 2024-09-11
feat: allow explicitly ignoring the license of a package in config

### LINKED ISSUE #1124: Support ignoring a package only from license scanning

G-Rath:
This allows you to configure the scanner to completely ignore the license of a package in a way that is explicit, as oppose to configuring `license.overrides` to set the package license to an allowed one.

Resolves #1124

### DIFF
--- a/cmd/osv-scanner/fixtures/osv-scanner-complex-licenses-config.toml
+++ b/cmd/osv-scanner/fixtures/osv-scanner-complex-licenses-config.toml
@@ -9,13 +9,18 @@ license.override = ["0BSD"]
 
 [[PackageOverrides]]
 ecosystem = "Alpine"
-name = "apk-tools"
-license.ignore = true
-license.override = ["UNKNOWN"] # this takes priority over license.ignore
+name = "musl"
+license.override = ["UNKNOWN"]
 
 [[PackageOverrides]]
 ecosystem = "Alpine"
-name = "musl"
+name = "musl-utils"
+license.ignore = true # this takes priority over license.override
+license.override = ["UNKNOWN"]
+
+[[PackageOverrides]]
+ecosystem = "Alpine"
+name = "apk-tools"
 license.ignore = false
 
 [[PackageOverrides]]
--- a/cmd/osv-scanner/fixtures/osv-scanner-complex-licenses-config.toml
+++ b/cmd/osv-scanner/fixtures/osv-scanner-complex-licenses-config.toml
@@ -1,0 +1,23 @@
+[[PackageOverrides]]
+ecosystem = "npm"
+ignore = true
+license.override = ["0BSD"]
+
+[[PackageOverrides]]
+ecosystem = "Packagist"
+license.override = ["0BSD"]
+
+[[PackageOverrides]]
+ecosystem = "Alpine"
+name = "apk-tools"
+license.ignore = true
+license.override = ["UNKNOWN"] # this takes priority over license.ignore
+
+[[PackageOverrides]]
+ecosystem = "Alpine"
+name = "musl"
+license.ignore = false
+
+[[PackageOverrides]]
+ecosystem = "Alpine"
+license.ignore = true
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -42,8 +42,9 @@ ecosystem = "Go"
 group = "dev"
 
 # Actions to take for matching packages:
-ignore = true # Ignore this package entirely, including license scanning
-license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not ignored
+ignore = true # Ignore this package completely, including license scanning
+license.ignore = true # Ignore the license of the package, if it is not already completely ignored at the top level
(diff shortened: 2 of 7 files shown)


cmd/osv-scanner/fixtures/osv-scanner-complex-licenses-config.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[[PackageOverrides]]
+ecosystem = "Packagist"
+license.override = ["0BSD"]
+
+[[PackageOverrides]]
+ecosystem = "Alpine"
+name = "apk-tools"
+license.ignore = true
+license.override = ["UNKNOWN"] # this takes priority over license.ignore
```
G-Rath, line 14: note: I realised while doing this that the new composition logic combines with the order to allow you to selectively override specific packages under a particular condition while having related packages have different logic, though only if the logic is truthy

For now I think it's worth keeping as an undocumented feature given you cannot do e.g. `ignore = false` (as that's a negative condition), but it's cool and might be worth exploring further in future

docs/configuration.md
```diff
@@ -42,8 +42,9 @@ ecosystem = "Go"
 group = "dev"
 
 # Actions to take for matching packages:
-ignore = true # Ignore this package entirely, including license scanning
-license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not ignored
+ignore = true # Ignore this package completely, including license scanning
+license.ignore = true # Ignore the license of the package, if it is not already completely ignored at the top level
+license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not completely ignored at the top level
```
G-Rath, line 47: note: I've purposely kept the documentation here light as I think it'll be easier to document once #1226 is landed

cuixq:
@G-Rath I am a bit confused about what this PR does - could you update the description with more details?

G-Rath:
@cuixq I've added a short description, though I think the linked issue gives context too - let me know if there's anything specific you're still confused on that you'd like me to clarify

pkg/osvscanner/vulnerability_result.go
```diff
@@ -73,7 +73,12 @@ func buildVulnerabilityResults(
 				for j, license := range entry.License.Override {
 					overrideLicenses[j] = models.License(license)
 				}
-				r.Infof("overriding license for package %s/%s/%s with %s\n", pkg.Package.Ecosystem, pkg.Package.Name, pkg.Package.Version, strings.Join(entry.License.Override, ","))
+				// reaching here without any override licenses means that the license should be ignored completely
+				if len(overrideLicenses) == 0 {
+					r.Infof("ignoring license for package %s/%s/%s\n", pkg.Package.Ecosystem, pkg.Package.Name, pkg.Package.Version)
```
cuixq, line 78: Upper case?

pkg/osvscanner/vulnerability_result.go
```diff
@@ -73,7 +73,12 @@ func buildVulnerabilityResults(
...
+				} else {
+					r.Infof("overriding license for package %s/%s/%s with %s\n", pkg.Package.Ecosystem, pkg.Package.Name, pkg.Package.Version, strings.Join(entry.License.Override, ","))
```
cuixq, line 80: Upper case?

cmd/osv-scanner/fixtures/osv-scanner-complex-licenses-config.toml
cuixq: I do think `license.ignore` should take the priority - is this hard to implement?

pkg/osvscanner/vulnerability_result.go
G-Rath: I think that would be best changed in a dedicated pull request since it'll change unrelated test snapshots - I can do that if you want, and then once it's landed it'll make sense to change the `ignoring` to be upper case here

another-rex: Let's keep this as it is for now.

cmd/osv-scanner/fixtures/osv-scanner-complex-licenses-config.toml
G-Rath: done

cuixq (approved):

docs/configuration.md
another-rex, line 47:
```suggestion
license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not completely ignored
```
I think this should be updated now that ignore takes priority.

another-rex (approved):

PR #1248, opened 2024-09-15
refactor: simplify and reuse `tryLoadConfig`

G-Rath:
`toml.DecodeFile` does literally what we're doing except without the custom error messages, which I don't think should matter as we don't intentionally expose those as an actual API; there also doesn't seem to be anything special about what `UseOverride` is doing to prevent it from using `tryLoadConfig`.

This will mean error output will probably have changed slightly, but it should still be accurate and we're not purposely providing custom or dedicated errors in this package so I don't think it should be considered breaking.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -205,8 +205,8 @@ func normalizeConfigLoadPath(target string) (string, error) {
 	return configPath, nil
 }
 
-// tryLoadConfig tries to load config in `target` (or it's containing directory)
-// `target` will be the key for the entry in configMap
+// tryLoadConfig attempts to parse the config file at the given path as TOML,
+// returning the Config object if successful or otherwise the error
 func tryLoadConfig(configPath string) (Config, error) {
 	config := Config{}
 	_, err := toml.DecodeFile(configPath, &config)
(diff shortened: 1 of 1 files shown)


G-Rath:
Noting that this should already be compatible with #1242

pkg/config/config.go
```diff
@@ -201,22 +199,11 @@ func normalizeConfigLoadPath(target string) (string, error) {
...
-		_, err := toml.NewDecoder(file).Decode(&config)
-		if err != nil {
-			return Config{}, fmt.Errorf("failed to parse config file: %w", err)
-		}
-		config.LoadPath = configPath
-
-		return config, nil
-	}
+	config := Config{LoadPath: configPath}
+	_, err := toml.DecodeFile(configPath, &config)
```
G-Rath, line 206: note: I've realised this actually makes it possible to "configure" the `LoadPath`

I will look back into this shortly

G-Rath: I ended up just refactoring to only set the `LoadPath` if `err` is `nil` - while it makes the function a bit bigger, it means we maintain the existing public behaviour without costing us a lot, and (importantly) lets us defer a bikeshed 😅

I did consider that it might be useful to have the `LoadPath` set in all cases since in the event of an error you could use that to programmatically know what file it was erroring on, but again that'd mean doing some bikeshedding and frankly I don't think it's useful enough to do it for now when we can easily change the behaviour later 🤷

cuixq (approved):

another-rex (approved):

PR #1250, opened 2024-09-16
test: restructure internal `config` cases and fixtures

G-Rath:
These tests are so old that they predate the repository itself and are not giving the best coverage of what they say they're testing so I've refactored them so that:
  - the fixtures now live in the package, since that's the only place they're used
    - (we could probably improve this a bit further, but I've not done anything extra for now as I don't think it's as important)
  - the tests are now laid out as tables and run in parallel
  - methods are being tested independently, and with improved comparators
  - include a few missing cases such as when the file is missing


pkg/config/fixtures/testdatainner/osv-scanner-load-path.toml
```diff
@@ -0,0 +1 @@
+LoadPath = "a/b/c"
```
G-Rath, line 1: note: technically due to the tags on `Config` this will get read, but we always override it so isn't not possible for this to work (which is what we want)

_however_ if someone were to refactor `tryLoadConfig` (like what I've done in #1248), then this could start getting loaded - hence, I've added this test to safeguard, with the intent to address the actual bug in a follow-up

pkg/config/config_internal_test.go
```diff
@@ -1,102 +1,185 @@
...
+			args: args{
+				target: "./fixtures/testdatainner/does-not-exist",
 			},
-			{
-				ID: "GO-2022-1059",
+			want:    "",
+			wantErr: true,
+		},
+		{
+			name: "target is file in directory with config",
```
cuixq, line 44: maybe I miss something - should this be `directory without config`? there isn't `osv-scanner.toml` in `innerFolder`

G-Rath: 🤦 I had it written without that the first time, but then got confused with the `tryLoadConfig` tests and added it - thanks for catching this!

cuixq: so now there are two `target is file in directory` tests - any difference between two tests?

G-Rath: there were always two tests like this - I've just ported them across. I think technically they're different because one has a config file in them and the other doesn't, but the original test didn't make it clear if that was a part of the test or just a coincidence.

I wasn't that fussed because these tests are very fast to run so I didn't have a huge issue with retaining a potentially duplicated test here (given it's not a _total_ duplicate from a file system POV even if that's not super revertant to the implementation)

cuixq: I think it's better to state the difference in the test names - maybe @another-rex knows why we have these tests.

G-Rath: fow now I've just adjusted the names to be unique, as I'd like to get this landed ASAP so I can rebase #1248 and #1249

cuixq (approved):

PR #1252, opened 2024-09-16
fix: don't allow `LoadPath` to be set via config file

G-Rath:
This is a pseudo bug I introduced way back in #190 - it doesn't actually happen as our implementation always explicitly sets the property after we've successfully parsed the file into the struct, but it is possible for refactors like #1248 to reveal this and ultimately we don't want this behaviour so we might as well fix it.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -31,7 +31,7 @@ type ConfigManager struct {
 type Config struct {
 	IgnoredVulns      []IgnoreEntry          `toml:"IgnoredVulns"`
 	PackageOverrides  []PackageOverrideEntry `toml:"PackageOverrides"`
-	LoadPath          string                 `toml:"LoadPath"`
+	LoadPath          string                 `toml:"-"`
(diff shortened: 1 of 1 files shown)


cuixq (approved):

pkg/config/config.go
```diff
@@ -27,7 +27,7 @@ type ConfigManager struct {
 type Config struct {
 	IgnoredVulns      []IgnoreEntry          `toml:"IgnoredVulns"`
 	PackageOverrides  []PackageOverrideEntry `toml:"PackageOverrides"`
-	LoadPath          string                 `toml:"LoadPath"`
+	LoadPath          string                 `toml:"-"`
```
cuixq, line 30: Could you add a comment for this?

hogo6002 (approved):

pkg/config/config.go
G-Rath: I've added a comment and moved the field to the end of struct to make it clearer that this is not a parsed field

another-rex (approved):

PR #1258, opened 2024-09-18
feat(output): add HTML output format

hogo6002:
Implement https://github.com/google/osv-scanner/issues/1274
Add an HTML output format to display results more effectively, particularly for container scanning.
This format retains all the information from the existing table output, including called/uncalled vulnerability details.

New features:
* Provides a package summary page
* Groups vulnerabilities by source package and ecosystem. Vulnerabilities from OS images are displayed last by default.
* Provides a total count of vulnerabilities (excluding uncalled vulnerabilities by default).
* Surfaces container scanning metrics, such as layer commands.
* Shows fix version information.
image
image
image


internal/output/html/vuln_table_template.html
```diff
@@ -0,0 +1,34 @@
...
+  {{ $index := uniqueID}}
+  <tr class='vuln-tr' id="vuln-tr-{{ $element.Summary.ID }}-{{ $rowIndex }}-{{ $index }}"
+    onclick="showVulnDetails('{{ $element.Summary.ID }}-{{ $rowIndex }}-{{ $index }}')">
+    <td><a href="https://osv.dev/{{ $element.Summary.ID }}">{{ $element.Summary.ID }}</a></td>
+    <td>{{ $element.Summary.PackageName }}</td>
+    <td>{{ $element.Summary.InstalledVersion }}</td>
+    <td {{ if eq $element.Summary.FixedVersion "No fix available" }} class="warning" {{ end }}>{{
+      $element.Summary.FixedVersion }}</td>
+    <td>{{ $element.Summary.SeverityRating }} {{ if ne $element.Summary.SeverityRating "UNKNOWN" }}({{
+      $element.Summary.SeverityScore }}){{ end }}</td>
```
another-rex, line 19: nit: Fix the line breaks here so that {{ and }} are in the same line.

internal/output/html/report_template.html
```diff
@@ -0,0 +1,272 @@
...
+
+      <div id="vuln-tab" class="view-tab">
+        {{ range .EcosystemResults }}
+        <div class="ecosystem">
+          <h2>{{ .Ecosystem }}</h2>
+          <div class="source-section">
+            {{ range .Sources }}
+            <h3>Source: {{ .Ecosystem }}: {{ .Source }}</h3>
+            <p>
+              Found {{ index .PackageCount 0 }} package(s) with issues, {{
```
another-rex, line 261: nit: Fix line breaks

internal/output/html/report_template.html
```diff
@@ -0,0 +1,272 @@
...
+        <img
+          src="https://google.github.io/osv.dev/assets/logo.png"
+          alt="OSV-Scanner Logo"
+          class="logo"
+        />
+        <h1>Vulnerability Scan Report</h1>
+      </header>
+      <h2>Scanning Results</h2>
+      <p>
+        Total {{ .HTMLVulnCount.Called }} vulnerabilities found ({{
```
another-rex, line 227: nit: Fix line breaks.

internal/output/html/package_table_template.html
```diff
@@ -0,0 +1,38 @@
...
+    <td {{ if eq $element.FixedVersion "No fix available" }} class="warning" {{ end }}>{{ $element.FixedVersion }}</td>
+    <td>{{ printSeverityCountShort $element.HTMLVulnCount }}</td>
+  </tr>
+  <tr>
+    <td colspan="100%">
+      <div id="vuln-details-{{ $element.Name }}-{{ $rowIndex }}" class="vuln-details">
+        <p><strong>Source: {{ $element.Source }}</strong></p>
+        <p>Vulnerabilities:</p>
+        {{ range $index, $vuln := $element.CalledVulns }}
+          {{ if eq $vuln.Summary.FixedVersion "No fix available" }}
```
another-rex, line 28: nit: Keep indentation here on the same level

internal/output/html.go
```diff
@@ -0,0 +1,486 @@
...
+	CalledVulns      []HTMLVulnResult
+	UncalledVulns    []HTMLVulnResult
+	InstalledVersion string
+	FixedVersion     string
+	HTMLVulnCount    HTMLVulnCount
+}
+
+type HTMLVulnResult struct {
+	Summary HTMLVulnResultSummary
+	Detail  map[string]string
```
another-rex, line 50: As discussed offline, rather than looping through the fields, just list them out explicitly here.

internal/output/html.go
```diff
@@ -0,0 +1,486 @@
...
+// populating the called and uncalled maps.
+func processVulnerabilityGroups(groupIDs map[string]models.GroupInfo, vulnPkg models.PackageVulns, uncalledVulnIDs map[string]bool, calledPackages map[string]bool, uncalledPackages map[string]bool) {
+	for _, group := range vulnPkg.Groups {
+		slices.SortFunc(group.IDs, identifiers.IDSortFunc)
+		representID := group.IDs[0]
+		groupIDs[representID] = group
+
+		if !group.IsCalled() {
+			uncalledVulnIDs[representID] = true
+			uncalledPackages[vulnPkg.Package.Name] = true
```
another-rex, line 258: I'm not sure if this is correct? 

If the package has one vuln that is called, one that is not called, this will count the package as an uncalled package.

internal/output/html.go
```diff
@@ -0,0 +1,486 @@
...
+		packageResults[index].Ecosystem = ecosystemName
+		packageResults[index].Source = packageSource.Source.Path
+		updateCount(&count, &packageResult.HTMLVulnCount)
+	}
+
+	return &SourceResult{
+		Source:         packageSource.Source.String(),
+		Ecosystem:      ecosystemName,
+		PackageResults: packageResults,
+		PackageCount:   [2]int{len(calledPackages), len(uncalledPackages)},
```
another-rex, line 152: Split this out into two variables, as it is a lot easier to keep track of which one is called and which is uncalled by explicitly naming them.

internal/output/html.go
```diff
@@ -0,0 +1,486 @@
...
+
+type SourceResult struct {
+	Source         string
+	Ecosystem      string
+	PackageResults []PackageResult
+	PackageCount   [2]int // called and uncalled package count
+	HTMLVulnCount  HTMLVulnCount
+}
+
+type PackageResult struct {
```
another-rex, line 37: Name this HTMLPackageResult

internal/output/html.go
```diff
@@ -0,0 +1,486 @@
...
+// HTML templates directory
+const TEMPLATEDIR = "html/*"
+
+// supportedBaseImages lists the supported OS base images for container scanning.
+var baseImages = []string{"Debian", "Alpine", "Ubuntu"}
+
+//go:embed html/*
+var templates embed.FS
+
+var vulnIndex = 0
```
another-rex, line 85: Suggest using an atomic integer here just in case we multi thread this in the future.

another-rex: And add a comment describing what this variable is for.

internal/output/html.go
```diff
@@ -0,0 +1,486 @@
...
+	Medium   int
+	Low      int
+	Unknown  int
+	Called   int
+	Uncalled int
+	Fixed    int
+	UnFixed  int
+}
+
+const UNFIXED = "No fix available"
```
another-rex, line 74: Rename to UnfixedDescription

another-rex (commented):
Looking good! Did a quick first pass at reviewing this.

internal/output/html.go
hogo6002: I think this would count that package as both uncalled and called. We record the called vulnerability information by each vulnerability, and this loop through each vuln and update the package count.

internal/output/html/vuln_table_template.html
hogo6002: Updated, I feel it got messed up by the reformat script

internal/output/html/report_template.html
```diff
@@ -0,0 +1,272 @@
...
+<html>
+  <head>
+    <title>Vulnerability Scan Report</title>
+    <link
+      rel="icon"
+      href="https://google.github.io/osv.dev/assets/icon.png"
+      type="image/x-icon"
+    />
+    <style>
+      body {
```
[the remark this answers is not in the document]
hogo6002: I wanted to keep my CSS in a separate `.css` file, but VS Code kept complaining about `<style>{{template "style.css"}}</style>` even though it worked fine.  To get rid of that annoying error, I just put the CSS in an HTML file and used `{{template "style.html"}}` to include it.

internal/output/html/package_table_template.html
hogo6002: Updated

internal/output/html.go
hogo6002: Updaetd with a `HTMLVulnResultDetail` struct

internal/output/html.go
hogo6002: Updated all naming to start with HTML

internal/output/html.go
hogo6002: Removed this variable and replaced with random method

hogo6002:
Added a tooltip for layer command file name and enabled UT

internal/output/html.go
another-rex: Hmm... I think if a package has any called vulns, it should be categorised as called package.

another-rex: Discussed offline

pkg/reporter/format.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/google/osv-scanner/pkg/models"
 )
 
-var format = []string{"table", "vertical", "json", "markdown", "sarif", "gh-annotations", "cyclonedx-1-4", "cyclonedx-1-5"}
+var format = []string{"html", "table", "vertical", "json", "markdown", "sarif", "gh-annotations", "cyclonedx-1-4", "cyclonedx-1-5"}
```
cuixq, line 10: Is there any particular order for format here? if not, shall we order alphabetically?

internal/output/html.go
```diff
@@ -0,0 +1,475 @@
...
+	Fixed    int
+	UnFixed  int
+}
+
+const UnfixedDescription = "No fix available"
+
+// HTML templates directory
+const TemplateDir = "html/*"
+
+// supportedBaseImages lists the supported OS base images for container scanning.
```
cuixq, line 91: baseImages

internal/output/html.go
```diff
@@ -0,0 +1,475 @@
...
+	return func() int {
+		*index += 1
+		return *index
+	}
+}
+
+// getFixVersion returns the lowest fixed version for a given package and
+// its current installed version, considering the affected ranges. If no fix is
+// available, it returns "No fix available".
+func getFixVersion(allAffected []models.Affected, installedVersion string, installedPackage string, ecosystem models.Ecosystem) string {
```
cuixq, line 362: Personally I would prefer to return a bool to indicate whether a fix is available plus the string, however I also understand we want to pass the fix version to the string field so we have to make the return value a string...

cuixq (approved):
LGTM

pkg/reporter/format.go
hogo6002: Removed html from help command for now.

internal/output/html.go
hogo6002: Good catch, thanks!

internal/output/html.go
hogo6002: I think we will add a centralized fix version function in the future. So I'll leave the current implementation as is for now and apply this suggestion when creating the new function.

internal/output/html.go
```diff
@@ -0,0 +1,475 @@
...
+	ecosystemPrefix := models.Ecosystem(strings.Split(string(ecosystem), ":")[0])
+	vp := semantic.MustParse(installedVersion, ecosystemPrefix)
+	minFixVersion := UnfixedDescription
+	for _, affected := range allAffected {
+		if affected.Package.Name != installedPackage || affected.Package.Ecosystem != ecosystem {
+			continue
+		}
+		for _, affectedRange := range affected.Ranges {
+			for _, affectedEvent := range affectedRange.Events {
+				if affectedEvent.Fixed == "" || vp.CompareStr(affectedEvent.Fixed) > 0 {
```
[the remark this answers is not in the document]
hogo6002: Updated

internal/output/html.go
```diff
@@ -0,0 +1,475 @@
...
+				}
+			}
+		}
+	}
+
+	return minFixVersion
+}
+
+func getMaxFixedVersion(ecosystem string, allVulns []HTMLVulnResult) string {
+	ecosystemPrefix := models.Ecosystem(strings.Split(string(ecosystem), ":")[0]) //nolint:all
```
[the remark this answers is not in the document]
hogo6002: Right, updated

internal/output/html.go
```diff
@@ -0,0 +1,475 @@
...
+	var allVulns []HTMLVulnResult
+	var calledPackages = make(map[string]bool)
+	var uncalledPackages = make(map[string]bool)
+	var uncalledVulnIDs = make(map[string]bool)
+	var groupIDs = make(map[string]models.GroupInfo)
+	ecosystemName := ""
+
+	for _, vulnPkg := range packageSource.Packages {
+		if ecosystemName == "" {
+			ecosystemName = vulnPkg.Package.Ecosystem
```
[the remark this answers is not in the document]
hogo6002: Discussed offline

internal/output/html.go
```diff
@@ -0,0 +1,494 @@
...
+	for _, vuln := range allVulns {
+		groupInfo, isIndex := groupIDs[vuln.Summary.ID]
+		if !isIndex {
+			// We only display one vulnerability from one group
+			continue
+		}
+
+		// Add group IDs info
+		if len(groupInfo.IDs) > 1 {
+			vuln.Detail.GroupIDs = strings.Join(groupInfo.IDs[1:], ", ")
```
oliverchang, line 183: nit: in general it's better to leave all formatting decisions to the template itself, to enforce more separation between the logic (i.e. go) and the formatting (the template)

e.g. something similar to this example here: https://pkg.go.dev/text/template#example-Template-Block

```go
func main() {
	const (
		master  = `Names:{{block "list" .}}{{"\n"}}{{range .}}{{println "-" .}}{{end}}{{end}}`
		overlay = `{{define "list"}} {{join . ", "}}{{end}} `
	)
	var (
		funcs     = template.FuncMap{"join": strings.Join}
		guardians = []string{"Gamora", "Groot", "Nebula", "Rocket", "Star-Lord"}
	)
	masterTmpl, err := template.New("master").Funcs(funcs).Parse(master)
```

Is that doable here?

oliverchang (commented):
this looks awesome!!!

internal/output/html.go
```diff
@@ -0,0 +1,494 @@
...
+var templates embed.FS
+
+// BuildHTMLResults builds HTML results from vulnerability results.
+func BuildHTMLResults(vulnResult *models.VulnerabilityResults) HTMLResult {
+	var ecosystemMap = make(map[string][]HTMLSourceResult)
+	var resultCount HTMLVulnCount
+
+	for _, packageSource := range vulnResult.Results {
+		sourceName := packageSource.Source.String()
+		if strings.Contains(sourceName, "/usr/lib/") {
```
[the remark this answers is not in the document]
hogo6002: Updated. This is a temporary solution, discussed with Rex previously. This should be filtered out during the container scanning stage.

internal/output/html.go
```diff
@@ -0,0 +1,494 @@
...
+	"strings"
+
+	"github.com/google/osv-scanner/internal/cachedregexp"
+	"github.com/google/osv-scanner/internal/identifiers"
+	"github.com/google/osv-scanner/internal/semantic"
+	"github.com/google/osv-scanner/internal/utility/severity"
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+type HTMLResult struct {
```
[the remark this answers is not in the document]
hogo6002: Good point! Added for all structs

internal/output/html.go
hogo6002: Yeah, updated!

internal/output/html.go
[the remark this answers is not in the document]
another-rex: Can you also describe what/why we have this workaround? E.g. it is a heruestic to ignore installed packages which are already covered by OS specific vulnerabilties.

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+				FixedVersion:     fixedVersion,
+			},
+			Detail: vulnDetails,
+		}
+	}
+
+	return vulnResults
+}
+
+// processVulnerabilityGroups processes vulnerability groups and IDs,
```
another-rex, line 278: Can you list out exactly what this function does in the docstring with each of the arguments?
E.g. 
- groupIDs: Adds sorted group IDs in vulnPkg to groupIDs
- uncalledVulnIDs, ... : Marks vulnID in the relevent maps.

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+// populating the called and uncalled maps.
+func processVulnerabilityGroups(groupIDs map[string]models.GroupInfo, vulnPkg models.PackageVulns, uncalledVulnIDs map[string]bool, calledPackages map[string]bool, uncalledPackages map[string]bool) {
```
another-rex, line 280: nit: groupIDs should go after vulnPkg. Logic is vulnPkg contains the information you are trying to distribute to the arguments after it.

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+	// Sort packageResults to ensure consistent output
+	sort.Slice(results, func(i, j int) bool {
+		return results[i].Name < results[j].Name
+	})
+
+	return results
+}
+
+// processVulnerabilities processes vulnerabilities for a package
+// and returns a slice of HTMLVulnResult.
```
another-rex, line 247:
```suggestion
// processVulnerabilities converts each vulnerability for a package
// to an HTMLVulnResult
```

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+func processVulnerabilities(vulnPkg models.PackageVulns) []HTMLVulnResult {
+	vulnResults := make([]HTMLVulnResult, len(vulnPkg.Vulnerabilities))
+	for i, vuln := range vulnPkg.Vulnerabilities {
+		aliases := vuln.Aliases
```
another-rex, line 251: Why is this copied into this variable?

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+	for _, ecosystemResult := range ecosystemResults {
+		for _, sourceResult := range ecosystemResult.Sources {
+			results = append(results, sourceResult.PackageResults...)
+		}
+	}
+
+	return results
+}
+
+func formatLayerCommand(command string) (string, string) {
```
another-rex, line 466: Add a small comment here describing the return values.

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+		Source:               packageSource.Source.String(),
+		Ecosystem:            ecosystemName,
+		PackageResults:       packageResults,
+		CalledPackageCount:   len(calledPackages),
+		UncalledPackageCount: len(uncalledPackages),
+		HTMLVulnCount:        count,
+	}
+}
+
+// processPackageResults splits the given vulnerabilities into called and uncalled
```
another-rex, line 182: I think the docstring is incorrect here?

internal/output/html.go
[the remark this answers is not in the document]
hogo6002: Updated

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+		if len(result.CalledVulns) > 0 {
+			result.InstalledVersion = result.CalledVulns[0].Summary.InstalledVersion
+			result.FixedVersion = getMaxFixedVersion(ecosystemPrefix, result.CalledVulns)
+		}
+
+		results = append(results, *result)
+	}
+
+	// Sort packageResults to ensure consistent output
+	sort.Slice(results, func(i, j int) bool {
```
another-rex, line 239: Hmm does this also need to sort on version? E.g. if a package is repeated with 2 different versions installed.

another-rex (commented):
Last review!

internal/output/html.go
```diff
@@ -0,0 +1,505 @@
...
+			continue
+		}
+
+		// Process vulnerabilities for each source
+		sourceResult := processSource(packageSource)
+		if sourceResult == nil {
+			continue
+		}
+
+		sourceList, ok := ecosystemMap[sourceResult.Ecosystem]
```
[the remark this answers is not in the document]
hogo6002: Updated

internal/output/html.go
hogo6002: Added

internal/output/html.go
hogo6002: Good catch, updated

internal/output/html.go
```diff
@@ -234,21 +246,24 @@ func processPackageResults(allVulns []HTMLVulnResult, groupIDs map[string]models
 	}
 
 	// Sort packageResults to ensure consistent output
-	sort.Slice(results, func(i, j int) bool {
-		return results[i].Name < results[j].Name
+	slices.SortFunc(results, func(a, b HTMLPackageResult) int {
+		if c := cmp.Compare(a.Name, b.Name); c != 0 {
```
another-rex, line 250: a trick you can use here is the cmp.Or() function

another-rex (approved):
LGTM, one minor comment

internal/output/html.go
hogo6002: Good hint, updated!

.prettierignore
```diff
@@ -1,3 +1,4 @@
 **/fixtures/**
 **/fixtures-go/**
 /docs/vendor/**
+/internal/output/html/*
```
G-Rath, line 4: fwiw it looks like you should be able to format at least the css and js files with Prettier as they've not got any templating, which I think would be good for consistency

You've also not added any final newlines which is something Prettier would catch if we can run it on these files 😅

(that is defined in our `.editorconfig` - if you're using vscode might be worth checking you've got https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig installed and enabled)

hogo6002: Thanks Gareth! I actually didn't install editorConfig. Updated HTML with newlines and only disable linting on HTML templates

hogo6002: @G-Rath I'm merging this PR now to gather internal feedback. Feel free to try out the new HTML output and share any issues or ideas you have here: https://github.com/google/osv-scanner/issues/1274

Issue #1257, opened 2024-09-18
Reenable `revive` golangci-lints


another-rex:
It looks like by default, golangci has some preset for revive. By disabling a single rule (increment-decrement), this somehow disables all revive rules.

We should re-enable revive, as it is the replacement for golint. This is not trival, as there are quite a few changes we need to make for it to be compliant with revive again.

G-Rath:
fwiw I did realise and look into this a while ago, and found only https://github.com/google/osv-scanner/pull/997 to be worth enabling at the time.

While I didn't write up any detailed notes, I believe my overall findings were:
  - a lot of what's being linted is already been covered by other enabled lints (or in fact clash with them)
  - some would require breaking changes, and didn't seem like something we cared strongly about based on existing code (e.g. stuttering and `unexported-return`)
  - there didn't seem to be any complaints, common mistakes, dislikes, etc slipping through to make it worth chasing these

I'm not against working towards re-enabling `revive`, but I think it would be best to start by capturing a list or table of all the rules that should be (re)enabled so that we can easily work through them and also coordinate with the upcoming v2. 

I also think ideally we should be trying to enable the rules we're addressing for as we go to ensure we don't regress - that could be done either by a file-by-file, package-by-package, or by enabling across the whole codebase and explicitly disabling the rule inline for existing violations (happy to do a PR or two to showcase what I mean if that's not clear).

michaelkedar:
> some would require breaking changes, and didn't seem like something we cared strongly about based on existing code (e.g. stuttering and unexported-return)

fwiw, running `revive` with defaults, the only things I see that are not in `internal`  and would be breaking changes are:
```
pkg/config/config.go:18:6: exported: type name will be used as config.ConfigManager by other packages, and that stutters; consider calling this Manager (revive)
type ConfigManager struct {
     ^
pkg/osvscanner/osvscanner.go:67:5: error-naming: error var NoPackagesFoundErr should have name of the form ErrFoo (revive)
var NoPackagesFoundErr = errors.New("no packages found in scan")
    ^
pkg/osvscanner/osvscanner.go:73:5: error-naming: error var VulnerabilitiesFoundErr should have name of the form ErrFoo (revive)
var VulnerabilitiesFoundErr = errors.New("vulnerabilities found")
    ^
pkg/osvscanner/osvscanner.go:78:5: error-naming: error var OnlyUncalledVulnerabilitiesFoundErr should have name of the form ErrFoo (revive)
[...]

PR #1259, opened 2024-09-19
chore: make guided remediation follow revive's default lint rules

michaelkedar:
Helping with #1257
Changed the guided remediation code (`cmd/osv-scanner/fix`, and `internal/resolution`, `remediation` and `tui`) to fix lint errors found when using `revive`'s default settings. All of this is internal, so there's no API breakages.

It was mostly `unexported-return` and stuttering complaints (e.g. `resolution.ResolutionResult` -> `resolution.Result`), so a bunch of structs have been renamed.

### DIFF
--- a/cmd/osv-scanner/fix/main.go
+++ b/cmd/osv-scanner/fix/main.go
@@ -32,9 +32,9 @@ type osvFixOptions struct {
 	remediation.Options
 	Client     client.ResolutionClient
 	Manifest   string
-	ManifestRW manifest.IO
+	ManifestRW manifest.ReadWriter
 	Lockfile   string
-	LockfileRW lockfile.IO
+	LockfileRW lockfile.ReadWriter
 	RelockCmd  string
 }
 
@@ -260,7 +260,7 @@ func action(ctx *cli.Context, stdout, stderr io.Writer) (reporter.Reporter, erro
 	system := resolve.UnknownSystem
 
 	if opts.Lockfile != "" {
-		rw, err := lockfile.GetLockfileIO(opts.Lockfile)
+		rw, err := lockfile.GetReadWriter(opts.Lockfile)
 		if err != nil {
 			return nil, err
 		}
@@ -269,7 +269,7 @@ func action(ctx *cli.Context, stdout, stderr io.Writer) (reporter.Reporter, erro
 	}
 
 	if opts.Manifest != "" {
-		rw, err := manifest.GetManifestIO(opts.Manifest)
+		rw, err := manifest.GetReadWriter(opts.Manifest)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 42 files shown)


cmd/osv-scanner/fix/main.go
```diff
@@ -29,12 +29,12 @@ const (
...
 type osvFixOptions struct {
-	remediation.RemediationOptions
+	remediation.Options
 	Client     client.ResolutionClient
 	Manifest   string
-	ManifestRW manifest.ManifestIO
+	ManifestRW manifest.IO
 	Lockfile   string
-	LockfileRW lockfile.LockfileIO
+	LockfileRW lockfile.IO
```
michaelkedar, line 37: I've had the tendency to name these variables `rw` (for read-write).
It might make sense to rename the structs `manifest.ReadWriter` and `lockfile.ReadWriter`.
Any thougts/opinions?

G-Rath:
fwiw I've written up [my thoughts on this on the issue](https://github.com/google/osv-scanner/issues/1257#issuecomment-2359758232) - overall my main comment here is that I think it would be best to see if you can get the now-not-violated `revive` rules enabled for these files/packages to prevent future regressions and to let us know what in these files still needs actioning via explicitly inline disables.

internal/resolution/resolve.go
```diff
@@ -17,38 +17,38 @@ import (
 	"github.com/google/osv-scanner/pkg/models"
 )
 
-type ResolutionVuln struct {
-	Vulnerability models.Vulnerability
-	DevOnly       bool
+type Vulnerability struct {
+	OSV     models.Vulnerability
```
cuixq, line 21: I feel we can have a better name.. but I don't have a good suggestion neither...

cuixq (approved):
Thanks Michael!

cmd/osv-scanner/fix/main.go
cuixq: I don't have preference between the two options, both SGTM.

oliverchang (approved):

michaelkedar:
> overall my main comment here is that I think it would be best to see if you can get the now-not-violated `revive` rules enabled for these files/packages to prevent future regressions and to let us know what in these files still needs actioning via explicitly inline disables.

I can't seem to work out good a way to only enable revive rules for specific files (apparently you can add an `exclude:` do individual rules, but I can't get that to work.

Best I can do is add something like
```yaml
issues:
  exclude-rules:
    - path-except: internal/(resolution|remediation|tui)/
      linters:
        - revive
```

But that's not great, and it'll disable the `indent-error-flow` lint that we currently have enabled for everything else...

cmd/osv-scanner/fix/main.go
michaelkedar: I've gone and renamed them to `ReadWriter`s

internal/resolution/resolve.go
michaelkedar: Yeah, it's not great. I settled on `OSV` because it is the format of the structure.

PR #1262, opened 2024-09-19
feat: group DSA and its CVEs together

hogo6002:
For container scanning, we should only display DSA results and hide their related CVEs.
- Add DSA's related CVEs to the same group.
- Sort the group IDs with the DSA first. (DSA is more important than CVE, and one DSA contains multiple CVEs)
- Don't print CVE records when there is a DSA (this reduces noise in container scanning results).

### DIFF
--- a/pkg/grouper/grouper.go
+++ b/pkg/grouper/grouper.go
@@ -3,10 +3,10 @@ package grouper
 import (
 	"slices"
 	"sort"
-	"strings"
 
 	"golang.org/x/exp/maps"
 
+	"github.com/google/osv-scanner/internal/identifiers"
 	"github.com/google/osv-scanner/pkg/models"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/grouper/grouper_models.go
```diff
@@ -14,6 +18,13 @@ func ConvertVulnerabilityToIDAliases(c []models.Vulnerability) []IDAliases {
 			ID:      v.ID,
 			Aliases: v.Aliases,
 		}
+
+		// For Debian Security Advisory data,
+		// all related CVEs should be bundled together, as they are part of this DSA.
+		if strings.Split(v.ID, "-")[0] == "DSA" {
```
another-rex, line 24: This feels like it could be inaccurate as it makes the DSA related transitive as well. I think it's fine for now but can you add a TODO here to revisit this implementation later.

another-rex: I.e. we probably want to add Related field to IDAliases, and move this logic into Grouper.

pkg/grouper/grouper.go
```diff
@@ -20,6 +21,26 @@ func hasAliasIntersection(v1, v2 IDAliases) bool {
 	return slices.Contains(v1.Aliases, v2.ID) || slices.Contains(v2.Aliases, v1.ID)
 }
 
+func sortIDs(groupIDs []string) []string {
```
another-rex, line 24: Please use the existing `internal/output/identifiers.go` sort functions.  I think it's a good idea to sort it in the grouper directly like you have it here, so can you try moving the sort functions out into it's own internal package, and make it public?

another-rex (commented):
Thanks! Added some comments.

pkg/grouper/grouper_models.go
hogo6002: Yeah, added a todo to provide a general way to handle all linux security advisories,

pkg/grouper/grouper.go
hogo6002: Updated.

another-rex (approved):
LGTM!

PR #1268, opened 2024-09-23
feat: support `vulnerabilities.ignore` in package overrides

### LINKED ISSUE #1226: Add vulnerabilities.ignore flag to just ignore vulnerabilties.

G-Rath:
This implements the ability to ignore vulnerabilities in a matching group of packages while still reporting license violations, as the inverse to `license.ignore`.

Resolves #1226

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -368,6 +368,7 @@ Attempted to scan lockfile but failed: <rootdir>/fixtures/locks-many-with-invali
 ---
 
 [TestRun/config_file_can_be_broad - 1]
+Scanned <rootdir>/fixtures/locks-insecure/osv-scanner-flutter-deps.json file as a osv-scanner and found 3 packages
 Scanning dir ./fixtures/locks-many
 Scanned <rootdir>/fixtures/locks-many/Gemfile.lock file and found 1 package
 Scanned <rootdir>/fixtures/locks-many/alpine.cdx.xml as CycloneDX SBOM and found 14 packages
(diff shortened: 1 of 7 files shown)


cmd/osv-scanner/__snapshots__/main_test.snap
```diff
@@ -364,13 +365,16 @@ overriding license for package Packagist/league/flysystem/1.0.8 with 0BSD
...
-+-------------------+-----------+------------------+---------+---------------------------------------+
++-------------------+-----------+------------------------------------------------+---------+-------------------------------------------------------+
+| LICENSE VIOLATION | ECOSYSTEM | PACKAGE                                        | VERSION | SOURCE                                                |
++-------------------+-----------+------------------------------------------------+---------+-------------------------------------------------------+
+| 0BSD              | Packagist | league/flysystem                               | 1.0.8   | fixtures/locks-insecure/composer.lock                 |
+| UNKNOWN           |           | https://github.com/flutter/buildroot.git       |         | fixtures/locks-insecure/osv-scanner-flutter-deps.json |
+| UNKNOWN           |           | https://github.com/brendan-duncan/archive.git  |         | fixtures/locks-insecure/osv-scanner-flutter-deps.json |
+| UNKNOWN           |           | https://chromium.googlesource.com/chromium/src |         | fixtures/locks-insecure/osv-scanner-flutter-deps.json |
```
G-Rath, line 374: note: I realised as part of this that git-based dependencies are a bit of an edge case for the more composition-based config override logic

internally they don't have an ecosystem which is reflected in the license violation table but the vulnerability table uses `GIT` instead, but then `ecosystem = "GIT"` does not match these while `ecosystem = ""` just ignores _everything_.

I don't know if its worth fixing but we could account for this by mapping a value of `GIT` as a special case when checking for overrides; alternatively (and maybe the better short-term change) we could change the vulnerabilities table output to not have this logic so it at least matches the licenses table...

docs/configuration.md
```diff
@@ -42,9 +42,10 @@ ecosystem = "Go"
 group = "dev"
 
 # Actions to take for matching packages:
-ignore = true # Ignore this package completely, including license scanning
-license.ignore = true # Ignore the license of the package, if it is not already completely ignored at the top level
-license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not ignored
+ignore = true # Ignore this package completely, including both reporting vulnerabilities and license violations
+vulnerability.ignore = true # Ignore vulnerabilities for this package, while still checking the license (if not also ignored)
+license.ignore = true # Ignore the license of the package, while still checking for vulnerabilities (if not also ignored)
+license.override = ["MIT", "0BSD"] # Override the license of the package, if it is not ignored from license scanning completely
```
G-Rath, line 48: note: I think overall this reads better using the term "reporting" and "checking"

imo that reduces the expectation that these options mean the underlying actions for the matched packages don't happen at all, as that's not the case (i.e we will always include the package in the API calls to determine if there are vulnerabilities, it's just if there are we ignore them)

pkg/config/config.go
```diff
@@ -118,6 +123,15 @@ func (c *Config) ShouldIgnorePackageVersion(name, version, ecosystem string) (bo
...
 }
 
+// ShouldIgnorePackageVulnerabilities determines if the given package should have its vulnerabilities ignored based on override entries in the config
+func (c *Config) ShouldIgnorePackageVulnerabilities(pkg models.PackageVulns) bool {
+	overrides, _ := c.filterPackageVersionEntries(pkg, func(e PackageOverrideEntry) bool {
+		return e.Vulnerability.Ignore
+	})
+
+	return overrides
+}
```
G-Rath, line 133: note: the names here don't really make much sense but I think that's a bigger issue that should be tackled at the package level

I also didn't keep the interface matching the other functions because I couldn't think of _any_ unsurprising way of leveraging "overrides for vulnerabilities" like with licenses, and I feel like having this be as specific as possible should make it easier to do non-breaking changes in future

i.e. my hope is that if in future we do think up more uses for this struct, we'll be able to introduce `ShouldOverridePackageVulnerabilities` and co, and just deprecate this method

Given this is a public method, I am happy to switch this to returning the override entry to be on the safe side anyway given there's no major downside to doing that straight away even if we never use it

pkg/osvscanner/vulnerability_result.go
```diff
@@ -59,11 +59,14 @@ func buildVulnerabilityResults(
 		pkg.DepGroups = rawPkg.DepGroups
 
 		if len(vulnsResp.Results[i].Vulns) > 0 {
-			includePackage = true
-			pkg.Vulnerabilities = vulnsResp.Results[i].Vulns
-			pkg.Groups = grouper.Group(grouper.ConvertVulnerabilityToIDAliases(pkg.Vulnerabilities))
-			for i, group := range pkg.Groups {
-				pkg.Groups[i].MaxSeverity = output.MaxSeverity(group, pkg)
+			configToUse := configManager.Get(r, rawPkg.Source.Path)
```
G-Rath, line 62: note: it might make sense to pull this out of the condition since we do the exact same thing for license scanning below

but I've purposely not done that for now as I think the performance impact will be pretty minimal, it should be easily doable in a follow-up PR, and there might be a reason why we don't want to do that...

cuixq (approved):
LGTM

docs/configuration.md
another-rex: Huh I thought https://github.com/google/osv-scanner/pull/1206 ignored packages from being sent to the API in the first place.

G-Rath: oh yeah you're right I think, though I still think my wording here is fine because really that's an optimization - our public contract in this sense I think is still (at least right now) "we won't report" not "we don't check at all"

another-rex (approved):

PR #1271, opened 2024-09-24
feat: assume `txt` files with "requirements" in their name are `requirements.txt` files

### LINKED ISSUE #67: Relax filename requirements for requirements.txt parsing.

### LINKED ISSUE #370: Support parsing requirements.txt files that aren't specifically named "requirements.txt"
It's somewhat common in the Python ecosystem to have a requirements.txt file with a different name, e.g. `ci-requirements.txt` or `requirements/docs.txt`.

Having a way to parse these would be super valuable. I imagine either the filename support could be expanded, or an additional CLI arg to specify the lockfile format could work.

### LINKED ISSUE #1266: Accept other names for requirements.txt files or provide a way to specify an extractor
I would expect the following command to work, extracting a Python `requirements.txt` type file:

```
$ osv-scanner scan --lockfile requirements-dev.txt
```

Instead it says "could not determine extractor".

G-Rath:
This relaxes our logic for deciding if we can parse a file as a `requirements.txt`: we now assume any file that has "requirements" in its name and ends with `.txt` is probably a `requirements.txt`, which matches logic found elsewhere in the wild (e.g. dependabot).

Resolves #1266
Resolves #370
Resolves #67


cuixq (approved):

another-rex (commented):
LGTM! 

(I thought we had to merge this change with the scalibr extractor, but turns out they already do this)

oliverchang (approved):
thanks!

PR #1279, opened 2024-09-25
fix(offline): report all ecosystems without local databases in one single line

### LINKED ISSUE #1005: Errors spamming the stderr output
Some errors should only be displayed once, but spams the output because of its location in the code flow.

Currently the only example of this I have is :

https://github.com/google/osv-scanner/blob/main/internal/local/check.go/#L147-L148

Solution proposed:
Have some global hashmap that checks if this error has already been logged and don't repeat the log.

G-Rath:
This reduces the amount of noise generated when using `--offline` if a database is not available for a particular ecosystem. While it was suggested that we could have some form of general error cache, for now I've just optimized for this specific case as really it's the one we expect the most and (in part) I think the nature of Go makes the more generic improvement a bit too faffy vs the gain.

Resolves #1005

### DIFF
--- a/cmd/osv-scanner/main_test.go
+++ b/cmd/osv-scanner/main_test.go
@@ -641,6 +641,35 @@ func TestRun_LocalDatabases(t *testing.T) {
 	}
 }
 
+func TestRun_LocalDatabases_AlwaysOffline(t *testing.T) {
+	t.Parallel()
+
+	tests := []cliTestCase{
+		// a bunch of different lockfiles and ecosystem
+		{
+			name: "",
+			args: []string{"", "--config=./fixtures/osv-scanner-empty-config.toml", "--experimental-offline", "./fixtures/locks-requirements", "./fixtures/locks-many"},
+			exit: 127,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			testDir := testutility.CreateTestDir(t)
+			old := tt.args
+			tt.args = []string{"", "--experimental-local-db-path", testDir}
+			tt.args = append(tt.args, old[1:]...)
+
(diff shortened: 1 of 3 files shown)


cmd/osv-scanner/main_test.go
```diff
@@ -641,6 +641,35 @@ func TestRun_LocalDatabases(t *testing.T) {
 	}
 }
 
+func TestRun_LocalDatabases_AlwaysOffline(t *testing.T) {
```
G-Rath, line 644: note: I've had to put this in a dedicated test because our other local db tests try to reuse the database as an optimization when run locally, which is exactly what we don't want here

internal/local/check.go
```diff
@@ -143,8 +148,13 @@ func MakeRequest(r reporter.Reporter, query osv.BatchedQuery, offline bool, loca
 		db, err := loadDBFromCache(pkg.Ecosystem)
 
 		if err != nil {
-			// currently, this will actually only error if the PURL cannot be parses
-			r.Errorf("could not load db for %s ecosystem: %v\n", pkg.Ecosystem, err)
+			if errors.Is(err, ErrOfflineDatabaseNotFound) {
+				missingDbs = append(missingDbs, string(pkg.Ecosystem))
+			} else {
+				// the most likely error at this point is that the PURL could not be parsed
+				r.Errorf("could not load db for %s ecosystem: %v\n", pkg.Ecosystem, err)
```
G-Rath, line 155: note: this no longer has coverage, but that's fine

while my comment mentions a possible error being a PURL could not be parsed, we've not got a test for that and I'm not sure how easy it'll be - I've made a note to look into that as a separate thing, but we shouldn't let it block this as I think we want to be playing it safe i.e. by logging the error rather than assume the only possible error could be that the database cannot be found

cuixq (approved):

Issue #1045, opened 2024-06-17
Support private registries for Maven


cuixq:
Now by default, [deps.dev API](https://docs.deps.dev/api/v3/index.html) is queried for package versions and requirements when resolving dependencies. 

We should also support fetching these data from private registries as well.

jenspopp:
That would be a highly appreciated feature ;-) Right now it already fails for my most simple projects with:

failed extracting pom.xml: failed to merge parents: failed to get Maven project my.maven:super-pom:1.7.2: API query failed: Maven registry query status: 4
04 Not Found

cuixq:
Some updates regarding this issue:
 - There is a [MavenRegistryAPIClient](https://github.com/google/osv-scanner/blob/main/internal/resolution/datasource/maven_registry.go) to talk to a Maven registry directly for metadata. Currently the URL is hard-coded to Maven Central, and we would like to make this configurable.
 - There is also [MavenRegistryClient](https://github.com/google/osv-scanner/blob/main/internal/resolution/client/maven_registry_client.go) which can be used as a DependencyClient in dependency resolution. More work is needed to make it work with multiple-registry resolution which may involve working on deps.dev's Maven resolver.

another-rex:
Should this issue be closed with #1286 ?

cuixq:
There is one optimisation PR in progress and this can be closed after that.

Issue #1269, opened 2024-09-24
Read Maven configurations from `settings.xml`


cuixq:
`settings.xml` contains the information about:
 - registry: local repository, remote repository servers, and authentication information
 - project: profiles and properties

To give a better support for Maven dependency resolution, configurations should be read from `settings.xml`.

oliverchang:
@michaelkedar this is part of the auth support right?

michaelkedar:
`settings.xml` parsing for auth is complete.

We haven't parsed them for profiles, however.

cuixq:
Considering we are moving the implementation of Maven client to scalibr and there is already an issue about `settings.xml`, I will close this bug and and track the progress there in https://github.com/google/osv-scalibr/issues/409.

PR #1286, opened 2024-09-30
feat: fetch Maven metadata from specified repositories

cuixq:
https://github.com/google/osv-scanner/issues/1045

There are [repositories](https://maven.apache.org/pom.html#Repositories) defined in a Maven pom.xml. When looking for an artifact, these repositories are searched one by one until the artifact is found. Maven Central is the default registry to try at the last.

To support this behaviour, this PR:
 - makes `MavenRegistryAPIClient` host a list of registries besides the default registry
 - adds `UpdateRegistries` to `DependencyClient` to update the registries
 - adds a new flag to specify the default maven registry for `fix`
 - add new experimental options to `scan` to align with what we have for `fix`

TODO: 
 - still need to update documentation for new options/flags
 - update deps.dev Maven resolver for mutil-registry resolution
 - record not found requests to optimize performance


internal/resolution/datasource/maven_registry.go
```diff
@@ -27,50 +33,98 @@ type MavenRegistryAPIClient struct {
...
+		projects:        NewRequestCache[string, maven.Project](),
+		metadata:        NewRequestCache[string, maven.Metadata](),
+	}, nil
+}
+
+// Add adds the given registry to the list of registries if it has not been added.
+func (m *MavenRegistryAPIClient) Add(registry string) {
+	if !slices.Contains(m.registries, registry) {
+		if _, err := url.Parse(registry); err == nil {
+			m.registries = append(m.registries, registry)
```
michaelkedar, line 54: I don't think this should silently fail - can we return this error?

internal/resolution/client/maven_registry_client.go
```diff
@@ -106,6 +106,9 @@ func (c *MavenRegistryClient) Requirements(ctx context.Context, vk resolve.Versi
 	if err := proj.MergeProfiles("", maven.ActivationOS{}); err != nil {
 		return nil, err
 	}
+	for _, repo := range proj.Repositories {
+		c.api.Add(string(repo.URL))
+	}
```
michaelkedar, line 111: Should this be adding new repositories in `Requirements`? The seems like you'd inherit the repositories of your dependencies?

michaelkedar (commented):
A few initial comments

internal/resolution/client/maven_registry_client.go
cuixq: hmmmm does this happen only for dependency resolution? I assume you may not know what repositories defined in dependencies if only preparing effective POM?

cuixq: discussed offline - a project does not inherit repositories of your dependencies so we shouldn't add new repositories here

cmd/osv-scanner/fix/noninteractive.go
```diff
@@ -295,6 +295,19 @@ func autoOverride(ctx context.Context, r reporter.Reporter, opts osvFixOptions,
 		return err
 	}
 
+	if opts.ManifestRW.System() == resolve.Maven {
+		// Update Maven registries based on the pom.xml
+		specific, ok := manif.EcosystemSpecific.(manifest.MavenManifestSpecific)
```
oliverchang, line 300: are there other places where we should be checking for the registries to use beyond the pom.xml? i.e. per https://maven.apache.org/guides/mini/guide-multiple-repositories.html#repository-order

```
effective settings:
Global settings.xml
User settings.xml
local effective build POM:
Local pom.xml
Parent POMs, recursively
Super POM
effective POMs from dependency path to the artifact.
```

We need to check `settings.xml` in various places, as well as the parent POMs? Do we need to add a TODO here?

cmd/osv-scanner/fixtures/maven-transitive/parent.xml
```diff
@@ -0,0 +1,30 @@
...
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java-util</artifactId>
+      <version>4.28.0</version>
+    </dependency>
+  </dependencies>
+
+  <repositories>
+    <repository>
+      <id>google-andriod</id>
```
oliverchang, line 25: nit: not a big deal in this context but there's a typo :)

internal/manifest/maven.go
```diff
@@ -52,6 +61,15 @@ func (e MavenResolverExtractor) Extract(f lockfile.DepFile) ([]lockfile.PackageD
 		return result.DependencyManagement, nil
 	})
 
+	if registries := e.MavenRegistryAPIClient.GetRegistries(); len(registries) > 0 {
+		cleintRegs := make([]client.Registry, len(registries))
```
oliverchang, line 65: nit: wrong spelling for client

internal/resolution/client/maven_registry_client.go
```diff
@@ -151,6 +151,16 @@ func (c *MavenRegistryClient) MatchingVersions(ctx context.Context, vk resolve.V
 	return resolve.MatchRequirement(vk, versions), nil
 }
 
+func (c *MavenRegistryClient) UpdateRegistries(registries []Registry) error {
+	for _, reg := range registries {
+		if err := c.api.AddRegistry(reg.URL); err != nil {
```
oliverchang, line 156: do we need to clear any existing registries? `Update` implies we will overwrite all existing ones?

internal/resolution/datasource/maven_registry.go
```diff
@@ -27,50 +33,101 @@ type MavenRegistryAPIClient struct {
...
-		metadata: NewRequestCache[string, maven.Metadata](),
+		defaultRegistry: registry,
+		projects:        NewRequestCache[string, maven.Project](),
+		metadata:        NewRequestCache[string, maven.Metadata](),
+	}, nil
+}
+
+// Add adds the given registry to the list of registries if it has not been added.
+func (m *MavenRegistryAPIClient) AddRegistry(registry string) error {
+	if !slices.Contains(m.registries, registry) {
```
oliverchang, line 52: nit: it seems better to invert this condition and return nil instead, to avoid more indentation on the common case.

oliverchang (commented):
Looks mostly good with some nits/questions.

cmd/osv-scanner/fix/noninteractive.go
cuixq: repositories defined in parent POMs will be merged when we merge other data from parents.

I made an [issue](https://github.com/google/osv-scanner/issues/1269) for reading settings.xml, but I think it's a good idea to refer to that issue here.

internal/resolution/client/maven_registry_client.go
cuixq: We don't want to clear the existing registries - maybe this should be renamed to `AddRegistries` and have another `SetRegistries` to clear and set the registries (though not sure if this has any use case)

cuixq:
I mark this PR to draft - the registries should not be added when importing dependencies, but that is now done by `MergeParents` for all cases.

oliverchang (approved):

michaelkedar (approved):
LGTM - just a couple of small things

PR #1289, opened 2024-09-30
fix: warn about and ignore duplicate entries in SBOMs

### LINKED ISSUE #330: Duplicate results
When parsing a hierarchical CycloneDX SBOM, there is the possibility of duplicate references to some components. These are evaluated (?) and reported multiple times, without a way to differentiate them (or trace the to the originating component).

The result represented by `osv-scanner` should:
- [ ] not contain duplicate entries, and
- [ ] allow to see, where the faulty component originates (ie. present a component tree like `npm audit` or `cargo audit`)

G-Rath:
While from what I understand duplicates should not be possible in a valid SBOM, apparently they happen and it's useful for us to report + skip them.

Since doing this efficiently requires use of a map we in turn have to sort the packages to ensure a consistent output order, leading to me discovering that we're not already sorting the packages - I've opened #1288 to land that change first.

Resolves #330

### DIFF
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -443,7 +443,7 @@ func extractMavenDeps(f lockfile.DepFile) (lockfile.Lockfile, error) {
 // within to `query`
 func scanSBOMFile(r reporter.Reporter, path string, fromFSScan bool) ([]scannedPackage, error) {
 	var errs []error
-	var packages []scannedPackage
+	packages := map[string]scannedPackage{}
 	for _, provider := range sbom.Providers {
 		if fromFSScan && !provider.MatchesRecognizedFileNames(path) {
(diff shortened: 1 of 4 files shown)


pkg/osvscanner/osvscanner.go
```diff
@@ -515,11 +520,17 @@ func scanSBOMFile(r reporter.Reporter, path string, fromFSScan bool) ([]scannedP
 				}
 			}
 
-			slices.SortFunc(packages, func(i, j scannedPackage) int {
+			packages2 := make([]scannedPackage, 0, len(packages))
```
cuixq, line 523: rename `packages` and `packages2` to something indicating one is a map and one is a slice.

cuixq (approved):

another-rex (approved):

PR #1290, opened 2024-10-01
fix: bump osv max concurrent requests

hogo6002:
The current `maxConcurrentRequests` setting is insufficient for container scanning, where one scan may report more than a thousand vulnerabilities. Bump it up to 1000 to fix it.

### DIFF
--- a/pkg/osv/osv.go
+++ b/pkg/osv/osv.go
@@ -28,7 +28,7 @@ const (
 	// maxQueriesPerRequest splits up querybatch into multiple requests if
 	// number of queries exceed this number
 	maxQueriesPerRequest  = 1000
-	maxConcurrentRequests = 25
+	maxConcurrentRequests = 1000
 	maxRetryAttempts      = 4
 	// jitterMultiplier is multiplied to the retry delay multiplied by rand(0, 1.0)
 	jitterMultiplier = 2


andrewpollock:
Trying to understand the impact on an end user in a bandwidth-challenged environment...

Does this result in up to 1,000 concurrent separate API requests for individual vulnerabilities? 
Or a single batch query for up to 1,000 vulnerabilities?

pkg/osv/osv.go
```diff
@@ -28,7 +28,7 @@ const (
 	// maxQueriesPerRequest splits up querybatch into multiple requests if
 	// number of queries exceed this number
 	maxQueriesPerRequest  = 1000
-	maxConcurrentRequests = 25
+	maxConcurrentRequests = 1000
```
cuixq, line 31: I feel 1000 might be too much. how long to run with maxConcurrentRequests set to 100?

another-rex:
> Does this result in up to 1,000 concurrent separate API requests for individual vulnerabilities?
Or a single batch query for up to 1,000 vulnerabilities?

The former, 1000 concurrent separate API requests.

pkg/osv/osv.go
hogo6002: Discussed offline, 1000 should be ok to use and doesn't have any disadvantages.

hogo6002:
> Does this result in up to 1,000 concurrent separate API requests for individual vulnerabilities? Or a single batch query for up to 1,000 vulnerabilities?

It's for the individual vulnerability queries.

michaelkedar (approved):

another-rex (approved):

Issue #1295, opened 2024-10-04
Support local database for guided remediation


oliverchang:
i.e. respect https://google.github.io/osv-scanner/experimental/offline-mode/ for scanning.

another-rex:
Is this resolved with #1306?

Issue #1048, opened 2024-06-18
Add a client to fetch vulnerabilities from osv.dev


cuixq:
Something like [internal/resolution/client/osv_client.go](https://github.com/google/osv-scanner/blob/main/internal/resolution/client/osv_client.go)

This will be helpful for testing (mocking the client instead of fetching directly from OSV.dev API).

cuixq:
I believe this is done in https://github.com/google/osv-scanner/pull/1464

PR #1306, opened 2024-10-08
feat(guided remediation): support offline database in fix subcommand

michaelkedar:
#1295
Support the `--experimental-offline` and `--experimental-download-offline-databases` flags in `osv-scanner fix`.

This is pretty inefficient since `local.MakeRequest()` reads and parses the zip file every time it's called. I didn't want to make too many changes to make it more efficient since we'll probably end up refactoring it in v2.

Also fixed the links to headings on the offline docs page.

### DIFF
--- a/internal/resolution/client/osv_offline_client.go
+++ b/internal/resolution/client/osv_offline_client.go
@@ -4,27 +4,42 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"sync"
 
 	"deps.dev/util/resolve"
 	"github.com/google/osv-scanner/internal/local"
 	"github.com/google/osv-scanner/internal/resolution/util"
 	"github.com/google/osv-scanner/pkg/models"
 	"github.com/google/osv-scanner/pkg/osv"
+	"github.com/google/osv-scanner/pkg/reporter"
 )
 
 type OSVOfflineClient struct {
 	// TODO: OSV-Scanner v2 plans to make vulnerability clients that can be used here.
-	offline     bool
 	localDBPath string
-	mu          sync.Mutex
 }
 
-func NewOSVOfflineClient(downloadDBs bool, localDBPath string) *OSVOfflineClient {
-	return &OSVOfflineClient{
-		offline:     !downloadDBs,
-		localDBPath: localDBPath,
+func NewOSVOfflineClient(r reporter.Reporter, system resolve.System, downloadDBs bool, localDBPath string) (*OSVOfflineClient, error) {
+	if system == resolve.UnknownSystem {
+		return nil, errors.New("osv offline client created with unknown ecosystem")
 	}
+	// Make a dummy request to the local client to log and make sure the database is downloaded without error.
+	q := osv.BatchedQuery{Queries: []*osv.Query{{
+		Package: osv.Package{
+			Name:      "foo",
+			Ecosystem: string(util.OSVEcosystem[system]),
+		},
+		Version: "1.0.0",
+	}}}
+	_, err := local.MakeRequest(r, q, !downloadDBs, localDBPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.HasErrored() {
+		return nil, errors.New("error creating osv offline client")
+	}
+
+	return &OSVOfflineClient{localDBPath: localDBPath}, nil
 }
 
 func (c *OSVOfflineClient) FindVulns(g *resolve.Graph) ([]models.Vulnerabilities, error) {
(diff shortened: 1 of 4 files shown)


internal/resolution/client/osv_offline_client.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// local.MakeRequest only logs an error if the database cannot be found.
+	// For guided remediation this should be a fatal error, since there's only the one ecosystem.
+	// Set up a reporter to capture error logs and return the logs as an error.
+	r := &errorReporter{}
+
+	// Not entirely sure if the local database is thread safe.
+	// Chucking it in a mutex just in case.
+	c.mu.Lock()
```
G-Rath, line 50: this seems like something we should confirm, and ideally make it thread-safe if not already

I'm happy to do the leg work but it's been a while since I've had to actively deal with thread safety, so would appreciate a refresher if possible 🙂 (ideally, if we can get a test crafted that proves it is (eventually) thread-safe, that'd be the ultimate help I think)

My current thinking is that `local` has three primary parts: 1. fetching the database, 2. load the database, 3. doing the queries - the latter should be thread-safe for sure because it's just comparing to loaded content, so it's pure read (i.e. we don't have any counter-like writes or nothing).

In theory 1. should be fine I think so long as the underlying writing of the file is atomic - otherwise the rest of the logic should work from any point and the worse case is we do some work twice (i.e. in one thread we might check if the file exists, decide its not, and download it - even though there's another thread that is in the middle of doing that whose file we'll end up overriding).

In theory 2. should also be fine, I think because the loading is done into a local variable - we do cache the load, but that's done as an assignment to a map _after_ we're loaded into that local variable, which should make it atomic too?

Also fwiw the whole test suite for `local` is run in parallel which from what I understand of how that works should help prove it's thread-safe (though I don't think it's a perfect test because we use test directories, but some of the code should overlap thread-wise)

Happy for this to be punted to a dedicated issue/pr, but I'd appreciate joining you on the journey as a chance to refresh my memory on threading

internal/resolution/client/osv_offline_client.go
```diff
@@ -0,0 +1,94 @@
...
+				Ecosystem: string(util.OSVEcosystem[node.Version.System]),
+			},
+			Version: node.Version.Version,
+		}
+	}
+
+	// local.MakeRequest only logs an error if the database cannot be found.
+	// For guided remediation this should be a fatal error, since there's only the one ecosystem.
+	// Set up a reporter to capture error logs and return the logs as an error.
+	r := &errorReporter{}
```
G-Rath, line 46: I think we should at least try to preserve or replicate the "Loaded xyz local db from path/to/db" call, as knowing that path can be useful especially for offline mode

internal/resolution/client/osv_offline_client.go
```diff
@@ -0,0 +1,94 @@
...
+	"deps.dev/util/resolve"
+	"github.com/google/osv-scanner/internal/local"
+	"github.com/google/osv-scanner/internal/resolution/util"
+	"github.com/google/osv-scanner/pkg/models"
+	"github.com/google/osv-scanner/pkg/osv"
+)
+
+type OSVOfflineClient struct {
+	// TODO: OSV-Scanner v2 plans to make vulnerability clients that can be used here.
+	offline     bool
```
cuixq, line 18: `offline` is a bit confusing here - sounds like this indicates if we query the local database for vulnerabilities, but this is already an offline client. any reason that we don't use `downloadDBs` here?

internal/resolution/client/osv_offline_client.go
michaelkedar: I think I agree with your assessment (though, I'm not sure if file writes are atomic).
I've changed it to make a single request when created to possibly download the database, then let the future concurrent requests be made without downloading and without the mutex.

As-is, I'm pretty sure this usage is safe - each call to `local.MakeRequest` re-reads the zip file, so nothing should end up being used across goroutines. This is quite inefficient, and we probably want to have another look at this with v2.

internal/resolution/client/osv_offline_client.go
michaelkedar: Propagated the log from the first request in `NewOSVOfflineClient`. 
Keeping all of them would be too noisy since the database ends up being loaded after every dependency resolution we make.

internal/resolution/client/osv_offline_client.go
michaelkedar: Ended up removing this variable.

internal/resolution/client/osv_offline_client.go
G-Rath: > each call to local.MakeRequest re-reads the zip file, so nothing should end up being used across goroutines. This is quite inefficient, and we probably want to have another look at this with v2.

I mean technically you're not wrong, but I wouldn't say it's _quite_ inefficient as the whole design is that it takes a batch of queries - so you're meant to be passing it everything in one so that it can then handle doing things in parallel.

This mirrors `osv.MakeRequest` though as that's just an HTTP client it's implementation is lighter.

fwiw since this is a private package we can address it right now, but I don't think there's anything to actually address? i.e. we only ever make one call currently to `local.MakeRequest` which is the complete batch, and from what you've commented in this PR it doesn't sound like guided remediation would benefit from this straight away because it's only ever using a single ecosystem

michaelkedar: Guided remediation (at least with the `relax`/`override` strategies) ends up doing something like:
1. Resolve the dependency graph
2. Find vulns in graph
3. For each vulnerability, find patch that should fix the vuln
4. Re-resolve for the the patched graph
5. Check the new graph for the vulnerability and any newly introduced vulns
6. Possibly repeat 3-5 to until vuln is remediated and no new vulns are introduced.

So it'll end up calling `local.MakeRequest` at least once per discovered vulnerability (in parallel).

I worked around this in the guided remediation HTTP client by querying packages without versions and caching the responses to match locally. If it ends up being notably slow I might also do that here in a later PR.

Could also change `local` to work better with this, but it might be a decent amount of work for something that might end up changing anyway in v2 #1048

oliverchang (approved):

cuixq (approved):

PR #1307, opened 2024-10-08
fix: address a number of typos

G-Rath:
I found these while working on other stuff

### DIFF
--- a/exit_code_redirect.sh
+++ b/exit_code_redirect.sh
@@ -28,7 +28,7 @@ last_arg="${!total_args}"
 # Remove the last argument from the list
 args=${@:1:$((total_args - 1))}
 
-# () inteprets spaces as spearate entries in an array
+# () interprets spaces as separate entries in an array
 # tr replaces newlines with spaces
 split_args=($(echo "$last_arg" | tr '\n' ' '))
 
--- a/internal/customgitignore/dir_test.go
+++ b/internal/customgitignore/dir_test.go
@@ -721,7 +721,7 @@ func hasPatternContaining(gips []gitignore.Pattern, test string) bool {
 //
 // ... because the changes in customgitignore adjust the
 // implementation details of the upstream package so that
-// it doesn't read .gitingore files from ignored dirs.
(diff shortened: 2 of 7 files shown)


another-rex (approved):

cmd/osv-scanner/fix/main.go
```diff
@@ -274,7 +274,7 @@ func action(ctx *cli.Context, stdout, stderr io.Writer) (reporter.Reporter, erro
 			return nil, err
 		}
 		opts.ManifestRW = rw
-		// Prefer the manifest's system over the lockfile's.
```
cuixq, line 277: I think this is correct?

internal/thirdparty/xml/read.go
cuixq, line 1: this is forked from Go so I would prefer to have it not changed, so we are able to diff it in the future.

internal/thirdparty/xml/typeinfo.go
cuixq, line 1: same here

hogo6002 (approved):

cuixq (approved):

cuixq:
@G-Rath can you fix the lint issue so I can merge this PR?

G-Rath:
@cuixq the lint failure is a very weird fluke - it doesn't happen locally and I didn't change anything related to that file; hopefully it'll pass on this rerun

G-Rath:
@cuixq looks like this is just a general failure on `main` - I can look into why that's happening shortly, but in the meantime you should be fine to merge this:

image

cuixq:
yes I also noticed that - maybe this is an issue of the linter itself

G-Rath:
@cuixq CI is green now so technically this is good to merge - I couldn't actually find an answer to why linting was failing but I'm hoping it's a very rare fluke; I suspect it could be caching related.

cuixq:
@G-Rath thank you! I am also thinking to upgrade the linter...

G-Rath:
ha yah I saw your PR - I was thinking of doing the exact same thing, but thankfully saw your PR first 😄

PR #1309, opened 2024-10-08
feat: deprecate axillary public packages in favor of private versions

G-Rath:
As part of the upcoming v2 release we're planning on making as many of our currently-public packages that we can private, in order to make it easier for us to maintain them since most are "axillary" packages that only expect to be used by the scanner, and that it's not a breaking change to make particular packages public again in future if that does prove useful.

While we cannot remove the existing public packages until v2 is released, we can still create the private versions and switch over to using them - this means that we can start making what'd be breaking changes to their interfaces, and also gives people a chance to report their usage of a particular package due to them now being marked as deprecated.

Of our currently public packages, these are being made private and marked as deprecated:
- `config`
- `depsdev`
- `grouper`
- `spdx`

Of the others:
  - `lockfile` will be replaced with https://github.com/google/osv-scalibr, but we won't be marking it as deprecated until @another-rex has finished migrating the implementations here over to that library
  - `models` is going to be broken up, with at least the `constants` being moved to a dedicated package in https://github.com/ossf/osv-schema and at least some of the other content moved "elsewhere" (though the exact fate of everything is yet to be determined)
  - `osvscanner` is purposely public and will remain that way, which also necessitates `reporter` being public since it's a parameter to the API functions so it'd be burdensome for consumers to have it private
  - `osv` is used by `osvscanner`, but technically might not need to public; but that is still I think something we're interested in keeping public at some point - I believe there are talks of oneday maybe housing it in https://github.com/google/osv.dev, but for now it can remain public too

Note this is focused on just "moving" the existing packages - there are already a few functions in e.g. `config` that are marked as deprecated, which I'll remove in a follow-up pull request to make it easier to review.


G-Rath:
I've confirmed this is valid locally by deleting the public versions of the packages being made private, and everything still builds and passes etc

cuixq (commented):
LGTM, just one question - we have a v2 branch which I think is for v2 stuff, especially for this sort of breaking change. @another-rex wdyt?

G-Rath:
This isn't a breaking change, which is the point - it's breaking to _remove_ the packages, but not for us to stop using them; once this has been landed, then you can rebase `osv-scanner-v2` onto `main` and then actually delete the public packages

cuixq (approved):

another-rex (approved):

internal/grouper/grouper.go
```diff
@@ -0,0 +1,73 @@
...
+	return slices.Contains(v1.Aliases, v2.ID) || slices.Contains(v2.Aliases, v1.ID)
+}
+
+// Group groups vulnerabilities by aliases.
+func Group(vulns []IDAliases) []models.GroupInfo {
+	// Mapping of `vulns` index to a group ID. A group ID is just another index in the `vulns` slice.
+	groups := make([]int, len(vulns))
+
+	// Initially make every vulnerability its own group.
+	for i := range len(vulns) {
```
another-rex, line 30:
```suggestion
	for i := range vulns {
```

internal/grouper/grouper.go
```diff
@@ -0,0 +1,73 @@
...
+		groups[i] = i
+	}
+
+	// Do a pair-wise (n^2) comparison and merge all intersecting vulns.
+	for i := range len(vulns) {
```
another-rex, line 35:
```suggestion
	for i := range vulns {
```

PR #1310, opened 2024-10-08
fix: update spdx license ids

G-Rath:
While working on #1309 I learned that this list is out of date, so this regenerates it - I don't think it's a big deal, but I've also made a local note to look into potentially automating this more (it probably ideally should happen before every release)


cuixq (approved):
LGTM - shall we set up some regular process to regenerate the file regularly?

G-Rath:
> shall we set up some regular process to regenerate the file regularly?

Yup I plan(ned) to look into this as a follow-up - I don't think it's super critical given how few new licenses there are on the whole and that no one has opened an issue (that I know of) about a license being missing, but I ended up doing #1312 which I think is a good starting improvement.

Since I don't do the releases, I'd have to defer to others on what would be sufficient enough beyond that workflow change - but I'm happy to do the work once someone tells me what should be done 🙂

cuixq:
I think it's a good idea to add an extra step to our release process to generate all the files. maybe make a script so that we can run it to regenerate everything?

PR #1311, opened 2024-10-08
feat: replace "skip-git" with "include-git", making git repository scanning not the default

### LINKED ISSUE #1277: Consider making "skip git" the default in v2
(I've been meaning to raise this to discuss for a few months and we're getting closer to v2 so I'm making a public issue to force my own hand 😄)

Currently when I run the scanner without any config it will include scanning the commit if it sees its in a git repository which generally it _always_ will be for me as everything I do that involves code and dependencies uses git for source control, however _most of the time_ these git repositories will not be public because they're proprietary applications meaning there will never a result of scanning the commit and arguably that's a bit of private data being sent to the API (which unlike e.g. the dependency tree, will never give a positive result).

My understanding is that generally there are two main criteria for git scanning to be useful:
  1. the codebase is public (meaning its commits can be indexed, and whatnot)
  2. the codebase is using a very "commit driven" language like C++ (i.e. while possible for other ecosystems to have commits related to vulnerabilities, it's far more likely in just about every ecosystem that version numbers/semver will be being used instead)

Given the number of ecosystems the scanner supports and frankly just how much better/easier non-git based vuln info is to handle right now, I strongly suspect more than half of the uses of the scanner are in a context where at least one of those conditions are not true meaning this current default is not useful for most people and arguably a little negative (though I admit it's not a huge privacy concern).

Personally I'd prefer if this was opt-in (i.e. `--check-git` or `--include-git`), though maybe there's another way to try make this more automatic - for example, maybe it would make sense for the scanner to fallback to checking git if it doesn't find any other lockfiles, or to make this managed through a config property so codebases could opt-in (i.e. open source repos could create an `osv-scanner.toml` with a marker effective telling the scanner "hey this is a public repo so feel free to be more 'aggressive' in what you check")

G-Rath:
BREAKING CHANGE: don't scan git repositories by default, replacing "--skip-git" with "--include-git"

I've gone with renaming the property so it aligns with being false by default, though we could keep the existing property if we really wanted; likewise, I've also changed this for both the API and CLI as I think it makes sense to have them aligned even though we could technically have them different.

Resolves #1277

(also, I think the fact that we use `--skip-git` in our workflows is another good sign it should be disabled by default 😅)


another-rex:
Can you update the target to the v2 branch?

G-Rath:
@oliverchang given we don't have any tests for this area of the scanner and the conflicts with the v2 branch, it might be better if someone else on the team whose more familiar with the C/C++ stuff take this up, if that's possible

PR #1320, opened 2024-10-11
fix(guided remediation): update deps.dev Maven resolver

### LINKED ISSUE #1240: Maven resolution fails with native data source when `maven-metadata.xml` is missing from repository
Currently, the Maven resolver always tries to get the list of versions of a package, even when it shouldn't need to (in the case of soft requirements). The native Maven dependency client gets these versions from the `maven-metadata.xml` file, which may be missing, in which case the resolution fails entirely.

e.g. [`org.w3c.css:sac` on the Maven central repository](https://repo.maven.apache.org/maven2/org/w3c/css/sac/) does not have a `maven-metadata.xml`, but it is possible and valid to depend on `org.w3c.css:sac@1.3`. Running `osv-scanner fix --data-source=native` on a pom.xml that brings in this dependency exits with the error `API query failed: Maven registry query status: 404 Not Found`

I'm documenting this issue here, but the fix for this probably involves changes to the resolver logic in [the deps.dev repo](https://github.com/google/deps.dev/tree/main/util/resolve/maven).

michaelkedar:
Update deps.dev module to https://github.com/google/deps.dev/commit/b3ba03369df10d7148cd6ae39e8147aa21b6454e to pick up resolver changes.
Should fix #1240


cuixq (approved):
should this be chore?

michaelkedar:
> should this be chore?

yeah, probably

G-Rath:
No it shouldn't because it _fixes_ a bug, so it's a fix

michaelkedar:
:shrug:

PR #1329, opened 2024-10-18
feat: support evaluating spdx license expressions

### LINKED ISSUE #1299: license checker problems with SPDX expressions AND / OR
These should be excluded
```
> osv-scanner --experimental-licenses MIT,ISC,Apache-2.0,BSD-3-Clause,BSD-2-Clause,CC0-1.0,BlueOak-1.0.0,CC-BY-3.0 --lockfile pnpm-lock.yaml
╭─────────────────────────┬───────────┬──────────────────────────────────┬──────────────┬────────────────╮
│ LICENSE VIOLATION       │ ECOSYSTEM │ PACKAGE                          │ VERSION      │ SOURCE         │
├─────────────────────────┼───────────┼──────────────────────────────────┼──────────────┼────────────────┤

│ Apache-2.0 AND MIT      │ npm       │ @swc/core-darwin-arm64           │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-darwin-x64             │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-linux-arm64-gnu        │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-linux-arm64-musl       │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-linux-x64-gnu          │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-linux-x64-musl         │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-win32-arm64-msvc       │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-win32-ia32-msvc        │ 1.7.26       │ pnpm-lock.yaml │
│ Apache-2.0 AND MIT      │ npm       │ @swc/core-win32-x64-msvc         │ 1.7.26       │ pnpm-lock.yaml │
...
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 0.13.1       │ pnpm-lock.yaml │
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 0.20.2       │ pnpm-lock.yaml │
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 0.21.3       │ pnpm-lock.yaml │
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 1.4.0        │ pnpm-lock.yaml │
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 2.13.0       │ pnpm-lock.yaml │
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 2.19.0       │ pnpm-lock.yaml │
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 3.13.1       │ pnpm-lock.yaml │
│ CC0-1.0 OR MIT          │ npm       │ type-fest                        │ 4.26.1       │ pnpm-lock.yaml │
```
And this workaround doesnt work
```
> osv-scanner --experimental-licenses "MIT,ISC,Apache-2.0,BSD-3-Clause,BSD-2-Clause,CC0-1.0,BlueOak-1.0.0,CC-BY-3.0,Apache-2.0 AND MIT,CC0-1.0 OR MIT" --lockfile pnpm-lock.yaml
--experimental-licenses requires comma-separated spdx licenses. The following license(s) are not recognized as spdx: Apache-2.0 AND MIT,CC0-1.0 OR MIT
```

G-Rath:
This implements a parser for SPDX license expressions in accordance with [annex D of the v2 spec](https://spdx.github.io/spdx-spec/v2-draft/SPDX-license-expressions/), to allow the scanner to properly determine if packages with such expressive licenses are permitted based on the licenses allowed by the user.

To do this I've implemented a two-phase parser which starts by tokenizing the license string and then turns it into an AST of nodes that can be walked to determine if the full expression is satisfied; for no particularly good reason I have used a string for the base token type rather than a `struct`, meaning the tokens value is also it's type - the tradeoff with this is while it means we don't have to do as much referencing or work in the tokenizer, we do have to do some extra work when walking the tree to resolve the "simple expression" tokens.

I'm proposing landing the current implementation as I don't think using a `struct` would be strictly better, though in hindsight it probably would have been a bit quicker to implement and so I plan to explore how much simpler (or complex) it might be as a follow up.

Currently this is focused on `AND` and `OR` support, as I believe those are the two primary operators that are relevant to the scanner, though we still might want to have richer handling for the `WITH` and `+` operators; currently both of those just get treated as being part of the license expression (though it's not actually possible right now to allow a license with `WITH` as the CLI expects license values to not have any spaces - this too will be a follow-up for me).

Finally, I've purposely not put any caching in place even though that should be easy, due to wanting to get this landed and as I don't expect it to actually have a significant impact on the scanner performance (ultimately most complex expressions in the real-world will be made up of a single operator, and chopping+looping over strings in memory is extremely fast)

Resolves #1299


jayvdb:
Hi , I'd love to see this progress.  We are already using osv-scanner in one small private repo replacing a fork of https://github.com/jslicense/licensee.js that adds pnpm support, and would like to continue the transition but this PR is needed to avoid adding a ridiculous number of ignores in our `osv-scanner.yaml` files.

another-rex:
Thanks for the interest! We are in the process of getting this feature in, but are a bit blocked on some internal issues. We'll try to get this feature merged in a week or so.

another-rex (approved):
Since this works already let's merge this in first, and then use the deps.dev library later.

PR #1330, opened 2024-10-18
feat: Use lockfile scalibr interface

another-rex:
This PR contains all the code required to move to osv-scalibr while making the existing code compile and pass all tests (container tests not passing because of a bug in the scalibr alpine extractor). 

Changes not mentioned in the following list will be split off in separate PRs which should land before this PR.

Those are:
- [x] #1337 
- [x] #1331 
- [x] #1338 
- [x] #1341
- [x] #1345

Changes in this PR:
- Fixture changes:
  - Scalibr Python requirements.txt extractor currently doesn't support packages without versions, so added some version strings to the test files
- Image package required quite a bit of reworking to successfully update.
  - Add the ability to iterate through a directory via the pathtree library
  - Support scalibr FS interface for Layers
  - Add conversion code to convert inventories from osv-scalibr back to v1's lockfile and Inventory
    - This is done to minimize snapshot changes. Followup PRs should remove this conversion
- Add `internal/lockfilescalibr` package:
  - `errors.go` adds common extraction errors we want to translate.
  - `translation.go` adds helper functions and translation logic between osv-scanner v1 extractor names, and osv-scalibr extractor names.

Changes in followup PRs:
- Delete lockfiles package and migrate everything to use osv-scalibr extractors
- Remove conversion code in image


another-rex:
This PR is ready for review, but the snapshots tests will be updated to the updated version to make the test pass once https://github.com/google/osv-scalibr/pull/241 is merged. 

We also currently use a pseudo version of osv-scalibr until a release is made with all the latest changes.

pkg/osvscanner/osvscanner.go
```diff
@@ -356,27 +366,30 @@ func scanImage(r reporter.Reporter, path string) ([]scannedPackage, error) {
...
-				parsedLockfile, err = lockfile.ExtractDeps(f, parseAs)
+
+	var inventories []*extractor.Inventory
+
+	// special case for the APK and DPKG parsers because they have a very generic name while
+	// living at a specific location, so they are not included in the map of parsers
+	// used by lockfile.Parse to avoid false-positives when scanning projects
+	switch parseAs {
+	case "apk-installed":
+		// inventories, err := apkinstalled.Extractor{}.Extract(context.Background(), &si)
```
oliverchang, line 377: any reason why we keep this commented line around? Can you add a comment / link to a relevant bug?

internal/image/layer.go
```diff
@@ -47,7 +111,43 @@ type Layer struct {
...
+
+	return output, nil
+}
+
+var _ fs.FS = Layer{}
+var _ fs.StatFS = Layer{}
+var _ fs.ReadDirFS = Layer{}
+
 func (filemap Layer) getFileNode(path string) (*FileNode, error) {
+	if !filepath.IsAbs(path) {
```
oliverchang, line 147: Can you add a comment here explaining why we need to handle this? 

As opposed to say, enforcing this precondition that paths passed have to be absolute paths?

internal/image/scan.go
```diff
@@ -22,33 +29,100 @@ func ScanImage(r reporter.Reporter, imagePath string) (ScanResults, error) {
...
 		if file.fileType != RegularFile {
 			continue
 		}
-		parsedLockfile, err := extractArtifactDeps(file.virtualPath, img.LastLayer())
+
+		// TODO: Currently osv-scalibr does not correctly annotate OS packages
+		// causing artifact extractors to double extract elements here.
+		// So let's skip all these directories for now.
+		// See (b/364536788)
+		if strings.HasPrefix(file.virtualPath, "/usr/local/") {
```
oliverchang, line 47: why skip scanning if  /usr/local/ ? 

/usr/local generally has stuff not installed by OS package managers right? it's everything else under /usr that's problematic wrt double scanning?

internal/lockfilescalibr/translation.go
```diff
@@ -0,0 +1,199 @@
...
+//
+// If no extractors are found, then ErrNoExtractorsFound is returned.
+func Extract(ctx context.Context, localPath string, extractAs string) ([]*extractor.Inventory, error) {
+	info, err := os.Stat(localPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if extractAs != "" {
+		for _, ext := range lockfileExtractors {
```
oliverchang, line 123: nit: for readability might be worth putting this in a helper private function

internal/lockfilescalibr/translation.go
```diff
@@ -0,0 +1,199 @@
...
+
+		return nil, fmt.Errorf("%w, requested %s", ErrExtractorNotFound, extractAs)
+	}
+
+	output := []*extractor.Inventory{}
+	extractorFound := false
+
+	for _, ext := range lockfileExtractors {
+		if ext.FileRequired(localPath, info) {
+			extractorFound = true
```
oliverchang, line 151: a lot of this looks very similar to lines 125-139. Could we factor this out into a common function?

pkg/osvscanner/osvscanner.go
another-rex: Ah that comment should not be there, removed.

internal/image/scan.go
another-rex: I think we found a lot of OS packages get installed here. @hogo6002 IIRC you found a lot of go compiler binaries here?

hogo6002: Yes, we found a lot of go compiler binaries there

hogo6002: Example: 
`GO-2023-2375` has been found in the actual project binary.
image
But it also shows 20 times in /usr/lib
image
image

another-rex: Ah gotcha, I'll update this path. I think I found some npm packages in `/usr/local` though. I'll double check

internal/lockfilescalibr/translation.go
another-rex: Done

internal/lockfilescalibr/translation.go
another-rex: Done

internal/image/layer.go
another-rex: Added a comment:
```
	// We expect all paths queried to be absolute paths rooted at the container root
	// However, scalibr uses paths without a prepending /, because the paths are relative to ScanRoot
	// ScanRoot will always be '/' for container scanning, so prepend with / if necessary.
```

internal/image/scan.go
another-rex: Checking https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

It looks like we should ignore the entire /usr directory. I think everything in there is written by os package managers, since it is suppose to be read only.

oliverchang (approved):

internal/image/image.go
```diff
@@ -268,6 +268,7 @@ func inWhiteoutDir(fileMap Layer, filePath string) bool {
 		if filePath == dirname {
 			break
 		}
+
 		node := fileMap.fileNodeTrie.Get(dirname)
```
G-Rath, line 272: nit: I actually agree with this change, but since it's the only one in this file, removing it would take a while file out of the diff 😅 

```suggestion
		}
		node := fileMap.fileNodeTrie.Get(dirname)
```

pkg/osvscanner/osvscanner.go
```diff
@@ -11,15 +13,21 @@ import (
 	"path"
 	"path/filepath"
 	"slices"
-	"sort"
 	"strings"
 
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/extractor/filesystem/os/apk"
+	"github.com/google/osv-scalibr/extractor/filesystem/os/dpkg"
+	scalibrosv "github.com/google/osv-scalibr/extractor/filesystem/osv"
```
G-Rath, line 21: _bro_ 🤜

G-Rath (approved):
awesome! I won't pretend that I understand all of it, but it's exciting to see 😄

> Scalibr Python requirements.txt extractor currently doesn't support packages without versions, so added some version strings to the test files

that defeats the point of those test fixtures, ~and makes this a breaking change~ (oh wait, we're already doing one of those 🙈)

I'm guessing it's not an easy fix, but it'd nice if we could note somewhere that ideally these should be removed once support is added

---

I'm not sure if there's anything else like that - if there is, maybe would it be useful if we created like a "post v2 follow-up" issue where we could just chuck this stuff into so we don't forget, and then we can come back after the majors out to properly split/link things with proper issues? (probably not worth it though if this is the only one)

internal/image/scan.go
```diff
@@ -22,33 +29,105 @@ func ScanImage(r reporter.Reporter, imagePath string) (ScanResults, error) {
...
 	for _, file := range allFiles {
 		if file.fileType != RegularFile {
 			continue
 		}
-		parsedLockfile, err := extractArtifactDeps(file.virtualPath, img.LastLayer())
+
+		// TODO: Currently osv-scalibr does not correctly annotate OS packages
+		// causing artifact extractors to double extract elements here.
+		// So let's skip all these directories for now.
+		// See (b/364536788)
```
cuixq, line 46: I think this bug number should be removed?

another-rex: I think this is just reference for us to keep track.

PR #1331, opened 2024-10-18
feat: Migrate pomxml extractor which also performs transitive dependency resolution

another-rex:
Part of #1330 

No functional change is made compared to the version in `internal/manifest`, just switched to use the osv-scalibr interface.

Extractors moved to lockfilescalibr as a temporary staging ground before moving to osv-scalibr.


cuixq:
There is a PR about Maven registry support https://github.com/google/osv-scanner/pull/1286, I would recommend to merge this PR afterwards (refactoring is needed I think).

another-rex:
@cuixq This should be ready to review now!

cuixq (approved):

PR #1338, opened 2024-10-22
feat: Add osvscanner.json extractor

another-rex:
Add/migrate osv-scanner.json extractor to use the scalibr format. This is intended to replace the current extractor located here: pkg/lockfile/osv-vuln-results.go

This is extractor is intended to stay in osv-scanner and not moved to osv-scalibr as it is very osv-scanner specific.


cuixq:
I know this is done in another CL but I think it's better to state that `lockfilescalibr` is temporary.
also is this extractor something new? or we are just copying from existing codebase?

cuixq (approved):

G-Rath (approved):

oliverchang (approved):

internal/lockfilescalibr/language/osv/osvscannerjson/extractor.go
```diff
@@ -0,0 +1,87 @@
+// Package osvscannerjson extracts osv-scanner's json output.
```
another-rex, line 1:
```suggestion
// Package osvscannerjson extracts osv-scanner's json output (--format json).
```

internal/lockfilescalibr/language/osv/osvscannerjson/extractor.go
```diff
@@ -0,0 +1,87 @@
+// Package osvscannerjson extracts osv-scanner's json output (--format json).
```
another-rex, line 1:
```suggestion
// Package osvscannerjson extracts osv-scanner's json output.
```

PR #1181, opened 2024-08-14
docs(guided remediation): document override strategy for Maven remediation

michaelkedar:
Update guided remediation docs page for Maven override strategy. 
I've tried to make sure it covers everything - let me know if something's missing.

Currently at rendered at https://michaelkedar.github.io/osv-scanner/experimental/guided-remediation/

PR #1340, opened 2024-10-23
docs: update documentation about Maven registry support

cuixq:
https://github.com/google/osv-scanner/pull/1286 adds support for Maven registry during resolution.

As a follow up, this PR updates the documentation for transitive scanning about specifying data source during resolution as well as specifying Maven registry.

This PR also corrects the deps.dev API version we are using.

We also need to update the documentation in https://github.com/google/osv-scanner/pull/1181.


docs/supported_languages_and_lockfiles.md
```diff
@@ -81,6 +81,14 @@ After the dependency resolution, the OSV database is queried for the vulnerabili
 {: .note }
 Test dependencies are not supported yet in the computed dependency graph for Maven pom.xml.
 
+### Data source
+
+By default, we use the [deps.dev API](https://docs.deps.dev/api/v3/) to find version and dependency information of packages during transitive scanning.
+
+If instead you'd like to fetch data from [Maven Central](https://repo.maven.apache.org/maven2/), you can use the `--experimental-resolution-data-source=native` flag.
+
+If your project uses mirrored or private registries, in addition to setting `--experimental-resolution-data-source=native`, you will need to use the `--experimental-maven-registry` flag to specify the registry.
```
another-rex, line 90:
```suggestion
If your project uses mirrored or private registries, in addition to setting `--experimental-resolution-data-source=native`, you will need to use the `--experimental-maven-registry=<full-registry-url>` flag to specify the registry.
```

Is this what you mean? Can you also add an example to clarify details like whether you need to include (https://) for example.

cuixq: yes - example added

michaelkedar (approved):

docs/supported_languages_and_lockfiles.md
```diff
@@ -81,6 +81,14 @@ After the dependency resolution, the OSV database is queried for the vulnerabili
 {: .note }
 Test dependencies are not supported yet in the computed dependency graph for Maven pom.xml.
 
+### Data source
+
+By default, we use the [deps.dev API](https://docs.deps.dev/api/v3/) to find version and dependency information of packages during transitive scanning.
+
+If instead you'd like to fetch data from [Maven Central](https://repo.maven.apache.org/maven2/), you can use the `--experimental-resolution-data-source=native` flag.
+
+If your project uses mirrored or private registries, in addition to setting `--experimental-resolution-data-source=native`, you will need to use the `--experimental-maven-registry=<full-registry-url>` flag to specify the registry (e.g. `--experimental-maven-registry=https://repo.maven.apache.org/maven2/`).
```
oliverchang, line 90: Unrelated to documentation, but should we name the flag `--resolution-registry` to be more generic? Or is it very specific to Maven?

cuixq: I have thought of this considering we will soon support transitive scanning for Python requirements.txt - ideally I would like to have something `--resolution-registry=Maven:[URL],PyPI:[URL]`, but I don't have a very clear picture about how does Python private registry work so I don't want to put too much effort on finalizing this flag now, and this is also why I mark this flag as experimental. Also for the ecosystem string to be used in the flag, this should follow the spec that Gareth is working on which is also not finallized.

oliverchang: SGTM

another-rex (approved):

PR #1341, opened 2024-10-23
feat: Use custom pathtree library instead of third party trie

another-rex:
Turns out we don't need most of the functionality of the third party tire library. Built a simple and focused library pathtree which is just a tree with insert and get functions to allow us to implement a virtual filesystem easily.


internal/image/image.go
```diff
@@ -11,9 +11,9 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/dghubble/trie"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/osv-scanner/internal/image/internal/pathtree"
```
cuixq, line 16: is there a way to only have one internal in the path?
like `google/osv-scanner/internal/pathtree` or `google/osv-scanner/internal/image/pathtree`?

cuixq (approved):

internal/image/image.go
oliverchang: +1 is there a reason to have a nested `internal` here?

internal/image/internal/pathtree/pathtree.go
```diff
@@ -0,0 +1,133 @@
...
+			return nil
+		}
+		cursor = next
+	}
+
+	return cursor.value
+}
+
+// Get retrieves all the direct children of this given path
+func (node *Node[V]) GetChildren(path string) []*V {
```
oliverchang, line 80: are there any callers of this besides tests?

oliverchang (approved):

internal/image/internal/pathtree/pathtree.go
another-rex: Yes, this is needed for scalibr filesystem to implement ReadDIr, which is part of the parent PR.

internal/image/image.go
another-rex: I'll remove the second layer of internal, it is originally here since pathtree features are tied quite closely to image scanning.

PR #1342, opened 2024-10-23
feat: add `--experimental-offline-vulnerabilities` and `--experimental-no-resolve` flags

### LINKED ISSUE #1121: Support a --no-resolve flag to avoid resolving transitive dependencies during scanning
Transitive dependencies cannot be fixed. If I want to use only the direct dependencies of the Maven project pom without using transitive dependencies, how can I pass parameters?

### LINKED ISSUE #1339: Allow offline vulnerability database and transitive resolution to be enabled/disabled separately.
Currently, enabling [offline mode (`--experimental-offline`)](https://google.github.io/osv-scanner/experimental/offline-mode/) in scanning makes the scanner use a local vulnerability database *and* disables transitive dependency resolution for Maven pom.xml files.

But there are use cases for having one and not the other, i.e.
- Disable transitive resolution, but still use the OSV.dev vulnerability API, as in #1121 
- Use a local vuln db, but still do transitive scanning in environments with access to the maven registry but not osv.dev.

With #1286 adding the `--experimental-resolution-data-source` flag, we could allow setting this to `none` or `disable` to disable transitive resolution, and make the `--experimental-offline` flag affect only the vulnerability database. (But that would mean 'offline mode' may still attempt to make network requests...)

michaelkedar:
Closes #1339 and closes #1121 
Adds flags to use offline mode for vulnerabilities (`--experimental-offline-vulnerabilities`) and transitive resolution separately (`--experimental-no-resolve`)

The original `--experimental-offline` flag retains the same behaviour by functionally setting both of these flags.

### DIFF
--- a/cmd/osv-scanner/scan/main.go
+++ b/cmd/osv-scanner/scan/main.go
@@ -109,6 +109,10 @@ func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 			},
 			&cli.BoolFlag{
 				Name:  "experimental-offline",
+				Usage: "run in offline mode",
+			},
+			&cli.BoolFlag{
+				Name:  "experimental-offline-vulnerabilities",
 				Usage: "checks for vulnerabilities using local databases that are already cached",
 			},
 			&cli.BoolFlag{
@@ -137,6 +141,10 @@ func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 				Usage:     "scan an exported *docker* container image archive (exported using `docker save` command) file",
 				TakesFile: true,
 				Hidden:    true,
+			},
+			&cli.BoolFlag{
+				Name:  "experimental-no-resolve",
+				Usage: "disable transitive dependency resolution of manifest files",
 			},
 			&cli.StringFlag{
 				Name:  "experimental-resolution-data-source",
@@ -221,6 +229,18 @@ func action(context *cli.Context, stdout, stderr io.Writer) (reporter.Reporter, 
 		callAnalysisStates = createCallAnalysisStates(context.StringSlice("call-analysis"), context.StringSlice("no-call-analysis"))
 	}
 
+	offlineVulns := context.Bool("experimental-offline-vulnerabilities")
+	disableTransitive := context.Bool("experimental-no-resolve")
+	if context.Bool("experimental-offline") {
+		// Set the corresponding offline flags, unless they were already set explicitly.
+		if !context.IsSet("experimental-offline-vulnerabilities") {
+			offlineVulns = true
+		}
+		if !context.IsSet("experimental-no-resolve") {
+			disableTransitive = true
+		}
+	}
+
 	vulnResult, err := osvscanner.DoScan(osvscanner.ScannerActions{
 		LockfilePaths:        context.StringSlice("lockfile"),
 		SBOMPaths:            context.StringSlice("sbom"),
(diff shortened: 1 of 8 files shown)


cmd/osv-scanner/scan/main.go
```diff
@@ -221,6 +229,18 @@ func action(context *cli.Context, stdout, stderr io.Writer) (reporter.Reporter,
 		callAnalysisStates = createCallAnalysisStates(context.StringSlice("call-analysis"), context.StringSlice("no-call-analysis"))
 	}
 
+	offlineVulns := context.Bool("experimental-offline-vulnerabilities")
+	disableTransitive := context.Bool("experimental-no-resolve")
+	if context.Bool("experimental-offline") {
+		// Set the corresponding offline flags, unless they were already set explicitly.
+		if !context.IsSet("experimental-offline-vulnerabilities") {
+			offlineVulns = true
+		}
```
michaelkedar, line 238: The `CompareOffline` field that `experimental-offline-vulnerabilities` sets also affects the `SkipGit` and license scanning  in `osvscanner.DoScan`.
I think it might be a bit clearer to instead set them here similarly, but that could create a subtle change in behaviour in `DoScan`.

cmd/osv-scanner/scan/main.go
```diff
@@ -221,6 +229,18 @@ func action(context *cli.Context, stdout, stderr io.Writer) (reporter.Reporter,
 		callAnalysisStates = createCallAnalysisStates(context.StringSlice("call-analysis"), context.StringSlice("no-call-analysis"))
 	}
 
+	offlineVulns := context.Bool("experimental-offline-vulnerabilities")
+	disableTransitive := context.Bool("experimental-no-resolve")
+	if context.Bool("experimental-offline") {
+		// Set the corresponding offline flags, unless they were already set explicitly.
+		if !context.IsSet("experimental-offline-vulnerabilities") {
```
oliverchang, line 236: Why do we need to check if they were already set? if someone is setting `--experimental-offline`, then IMO that should just always disable everything regardless of what other flags are set?

cuixq: I think I may understand the use case a bit here: someone wants to disable all network requests but not the ones for vulnerabilities. 
Also as Michael mentioned below, `CompareOffline` also controls others things for example license scanning, maybe we should change it to a struct with a list of booleans? Each boolean field corresponds to one type of request requiring network.
Even further, can we change `experimental-offline` to be a list of strings taking `all` or `vulnerability,license` and `all` takes precedence over all other.

michaelkedar: NB: I moved this logic to the `Action` of the offline flag in [aa0a734](https://github.com/google/osv-scanner/pull/1342/commits/aa0a73405e98ae8d7b009e640456e4f571c73d21), but I've not addressed this discussion.

oliverchang: IMO let's keep this simple for now and go with @michaelkedar 's approach. 

`--experimental-offline` should just only be set when  somebody wants to turn everything off, and can't be overriden. There's already individual flags to control behaviour (after this PR) for the independent bits. 

Re taking a slice of strings instead, we went with that for `--call-analysis` to take in a list of languages, but I think full list of languages would be much larger than the list of functionality where we can control online/offline. (i.e. we'd have --call-analysis-rust, --call-analysis-go, --call-analysis-python, --call-analysis-{language}, ...)

And since this is all experimental, we can change this later :)

oliverchang (approved):

cuixq (approved):

PR #1355, opened 2024-10-29
feat: add support for comparing Red Hat versions locally

### LINKED ISSUE #1336: Add Red Hat version comparison support
Following osv.dev's recent addition of support for [RPM versions](https://github.com/google/osv.dev/blob/master/osv/third_party/univers/rpm.py), osv-scanner should also add a `version-rmp.go` file in [internal/semantic](https://github.com/google/osv-scanner/tree/main/internal/semantic) to support Red Hat version comparison.

G-Rath:
This adds support for comparing versions in the Red Hat ecosystem locally, in the same manner as the native `rpm` tool does.

Huge shout out to @jantman whose fantastically concise [blog post](https://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/#how-rpm-compares-version-parts) on how the comparator works provided just about everything I needed to implement this without involvement from other sources - the only thing it was missing is the caret special character, which I assume was added (probably long) after that blog post was written.

I did have a peek at his PR for Puppet and some of the original sources for the manual test fixtures, but I appreciated not having to actively dig into the C code like I had to with Alpine 😅

Note that for the first time such a pull request for `semantic` does _not_ include a fixture generator because there are so many advisories/packages/versions in the osv.dev database that the file came out as over 1mb, so I'm going to do a dedicated pull request for the generator which will avoid committing the actual file but instead rely on the weekly `semantic` runs.

Also, I believe some of the other ecosystems use this format under the hood (AlmaLinux and SUSE were mentioned by @hogo6002 , and I feel like I heard somewhere that Rocky Linux also might use `rpm`?); I have purposely not looked into those since I'm not familiar with this part of the digital world and ultimately support for them would be best landed in a dedicated pull request.

Resolves #1336


hogo6002 (approved):
Thanks Gareth! yeah Rocky Linux also uses rpm

internal/semantic/version-redhat.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+
+		var ac, bc string
+
+		for _, c := range a[vi:] {
+			if !iser(c) {
+				break
+			}
+
+			ac += string(c)
```
another-rex, line 128: Not sure how much this will matter performance wise, but we might want to use stringbuilder here instead.

internal/semantic/version-redhat.go
```diff
@@ -0,0 +1,235 @@
...
+		if bc == "" {
+			if isDigit {
+				return +1
+			}
+
+			return -1
+		}
+
+		// 9. If the leading segments were both numeric, discard any leading zeros and whichever one is longer wins. If `a` is longer than `b` (without leading zeroes), return 1, and vice versa. If they’re of the same length, continue on.
+		if isDigit {
```
another-rex, line 151: Hmm... would this be clearer to just parse and compare numerically?

internal/semantic/version-redhat.go
```diff
@@ -0,0 +1,235 @@
...
+			if len(ac) > len(bc) {
+				return +1
+			}
+			if len(ac) < len(bc) {
+				return -1
+			}
+		}
+
+		// 10. Compare the leading segments with strcmp() (or <=> in Ruby). If that returns a non-zero value, then return that value. Else continue to the next iteration of the loop.
+		if diff := strings.Compare(ac, bc); diff != 0 {
```
another-rex, line 164: If this does == 0, is the idea to check the next run of characters/numbers? 

E.g. This is for comparing 1234abcd5678 vs 1234abcd5600

internal/semantic/version-redhat.go
```diff
@@ -0,0 +1,235 @@
...
+		return -1
+	}
+	if a != "" && b == "" {
+		return +1
+	}
+
+	var vi, wi int
+
+	for {
+		// 1. Trim anything that’s not [A-Za-z0-9] or tilde (~) from the front of both strings.
```
another-rex, line 45: Update comment to mention ^ as well

another-rex (commented):
Thanks! Added some comments

internal/semantic/version-redhat.go
```diff
@@ -0,0 +1,235 @@
...
+// manner as rpmvercmp(8) does.
+func compareRedHatComponents(a, b string) int {
+	if a == "" && b != "" {
+		return -1
+	}
+	if a != "" && b == "" {
+		return +1
+	}
+
+	var vi, wi int
```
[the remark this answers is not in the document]
G-Rath: Ah yeah good point - this was named back when it was a receiver method and the args were `v` and `w`

internal/semantic/version-redhat.go
G-Rath: I don't think so - I tried it previously when doing some performance testing, and it actually broke the tests so I think at best it'll require the same amount of code 🤷‍♂️

internal/semantic/version-redhat.go
G-Rath: correct - effectively the version boils down to segments which are either a run of letters or digits, + some special characters (namely, `~` and `^`).

So everything that is not one of those gets discarded, then the first not-discarded-not-special character sets the "type" of the run (either digits or letters), which we collect and compare; then we move onto the next segment, and continue until one of the strings runs out or a special rule is met

So in your example, we end up with the following segment comparisons:
  - 1234 vs 1234
  - abcd vs abcd
  - 5678 vs 5600

internal/semantic/version-redhat.go
G-Rath: from my testing, using `strings.Builder` makes no difference at all to the performance, and seems to require a couple more lines while looking pretty much the same 🤷 

I've not actually used `strings.Builder` though so maybe I've missed a trick:
```
var sba strings.Builder
for _, c := range a[ai:] {
	if !iser(c) {
		break
	}

	sba.WriteRune(c)
	ai++
}
as := sba.String()

var sbb strings.Builder
for _, c := range b[bi:] {
	if !iser(c) {
		break
	}

	sbb.WriteRune(c)
	bi++
}
bs := sbb.String()
```

<details>
<summary>fwiw, this is my benchmark function</summary>

```
func Benchmark_compareRedHatComponents(b *testing.B) {
	// this is just `runAgainstEcosystemFixture` without the assertion part
	lines := loadEcosystemFixture(b, "redhat-versions-generated.txt")
	versions := make([][]RedHatVersion, 0, len(lines))

	for _, line := range lines {
		pieces := strings.Split(line, " ")

		if len(pieces) != 3 {
			b.Fatalf(`incorrect number of peices in fixture "%s" (got %d)`, line, len(pieces))
		}

		versions = append(versions, []RedHatVersion{
			parseRedHatVersion(pieces[0]),
			parseRedHatVersion(pieces[1]),
		})
	}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		for _, v := range versions {
			compareRedHatComponents(v[0].version, v[1].version)
		}
	}
}
```
</details>

internal/semantic/version-redhat.go
```diff
@@ -0,0 +1,235 @@
...
+
+		// 6. End the loop if either string has reached zero length.
+		if vi == len(a) || wi == len(b) {
+			break
+		}
+
+		// 7. If the first character of `a` is a digit, pop the leading chunk of continuous digits from each string (which may be "" for `b` if only one `a` starts with digits). If `a` begins with a letter, do the same for leading letters.
+		isDigit := isASCIIDigit(rune(a[vi]))
+
+		var iser func(r rune) bool
```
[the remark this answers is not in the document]
another-rex: hmm `isExpectedRunType`?

another-rex (approved):

jantman:
Thanks for the shout-out @G-Rath - crazy to me that a decade later that post is still relevant to some people. And sorry that yet another person has to implement this crazy logic...

PR #1356, opened 2024-10-29
ci: add versions fixture generator for Red Hat

G-Rath:
This adds a `semantic` fixture generator for the Red Hat ecosystem - as this ends up being just under 2mb, I've not committed the file like the others, instead making the test suite only use it if it's actually present.

It works by leveraging [`rpm`s built-in Lua interpreter](https://rpm-software-management.github.io/rpm/manual/lua.html), which includes exposing the `vercmp` function meaning we can have `rpm` tell us what the result of comparing two versions are so long as there's a version of that available.

In terms of actually getting a version of `rpm`, that turns out to be a lot harder with Docker I assume just because CentOS & co is a less open-source-y distro vs the others; there are images, but they're all generally very big. In an interesting twist though it turns out Debian actually ships a version of `rpm` for building and the like - it has had it's package management abilities neutered, but that's not something we need so a simple `apt install rpm` gives us the dependency we need.

### DIFF
--- a/.github/workflows/semantic.yml
+++ b/.github/workflows/semantic.yml
@@ -59,6 +59,39 @@ jobs:
           path: /tmp/debian-versions-generator-cache.csv
           key: ${{ runner.os }}-${{ hashFiles('debian-db.zip') }}
 
+  generate-redhat-versions:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+
+      - uses: actions/cache/restore@3624ceb22c1c5a301c8db4169662070a689d9ea8 # v4.1.1
+        with:
+          path: /tmp/redhat-versions-generator-cache.csv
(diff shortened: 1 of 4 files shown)


G-Rath:
> Semantic / generate-redhat-versions (pull_request) Successful in 15m

nice, that's a very reasonable "worst case with no cache" time given the size

(as a reminder, Debian is also pretty big typically taking 5+ minutes, but with caching it only takes ~20 seconds 😄)

.github/workflows/semantic.yml
```diff
@@ -59,6 +59,39 @@ jobs:
...
+          path: /tmp/redhat-versions-generator-cache.csv
+          key: ${{ runner.os }}-
+
+      - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
+        with:
+          python-version: "3.10"
+      - run: sudo apt install rpm
```
oliverchang, line 80: TIL you could do this.

oliverchang (approved):

PR #1359, opened 2024-10-30
chore: update Go to version 1.23.2

slonka:
we're getting a lot of errors when scanning 1.23.2 [kuma codebase](https://github.com/kumahq/kuma/):

```
Scanned ./projects/kuma/go.mod file and found 222 packages
2024/10/30 08:38:53 internal error: error "package requires newer Go version go1.23" (*errors.errorString) without position
2024/10/30 08:38:53 internal error: error "package requires newer Go version go1.23" (*errors.errorString) without position
...
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/google/osv-scanner
 
-go 1.22.7
+go 1.23.2
 
 require (
 	deps.dev/api/v3 v3.0.0-20241010035105-b3ba03369df1


cuixq:
There are also [renovate.json](https://github.com/google/osv-scanner/blob/main/renovate.json#L30) and [docs/installation.md](https://github.com/google/osv-scanner/blob/main/docs/installation.md?plain=1#L80) that need to be updated.

another-rex:
We already build with the latest go version, v1.9.0 was built with (go1.23.1), and the just released v1.9.1 is now built with go1.23.2

The go mod version is for API support, so osv library users can still import and use OSV with go 1.22

PR #1364, opened 2024-10-30
test(semantic): include ecosystems not supported by `lockfile`

G-Rath:
Notably this adds coverage over the Ubuntu entry, since that reuses Debian so we don't have any explicit tests for that in the comparison tests.

We also no longer need to be manually adding `CRAN` since `lockfile` supports it

### DIFF
--- a/internal/semantic/parse_test.go
+++ b/internal/semantic/parse_test.go
@@ -14,8 +14,7 @@ func TestParse(t *testing.T) {
 
 	ecosystems := lockfile.KnownEcosystems()
 
-	// todo: remove once CRAN is supported by lockfile
-	ecosystems = append(ecosystems, "CRAN")
+	ecosystems = append(ecosystems, "Alpine", "Debian", "Ubuntu")
 
 	for _, ecosystem := range ecosystems {
 		_, err := semantic.Parse("", models.Ecosystem(ecosystem))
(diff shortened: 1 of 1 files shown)


G-Rath:
ugh so it turns out that ironically the Alpine parser does not support an empty version - I've opened #1365 to fix that, which I think is the right move given all the other version parsers don't panic on this

internal/semantic/parse_test.go
```diff
@@ -14,8 +14,7 @@ func TestParse(t *testing.T) {
 
 	ecosystems := lockfile.KnownEcosystems()
 
-	// todo: remove once CRAN is supported by lockfile
-	ecosystems = append(ecosystems, "CRAN")
+	ecosystems = append(ecosystems, "Alpine", "Debian", "Ubuntu")
```
cuixq, line 17: should these be consts defined in `pkg/lockfile`?

G-Rath: Given a big part of v2 is about sorting our "ecosystem" type, I don't think we should be making any constants right now 😅

(I believe in the long-run these should eventually be sourced from the upcoming `osvschema` package)

cuixq: I think we have the constants already defined for example https://github.com/google/osv-scanner/blob/main/pkg/lockfile/apk-installed.go#L11?

G-Rath: but it doesn't have one for `Ubuntu`, and `lockfile` won't be available when this is moved to `scalibr` - I can change if you really want, I just don't think it's a big deal for these two lines to be using the string format given the related upcoming movement that'll likely end up changing this anyway 🤷

cuixq: fair - I think the point is to avoid having free-text ecosystem strings and the ecosystems should be from [osv-schema](https://github.com/ossf/osv-schema/blob/main/bindings/go/osvschema/constants.go).

cuixq (approved):

PR #1371, opened 2024-10-31
test: use internal utility for creating tmp directory

G-Rath:
This isn't major, but it means we're more consistent across the codebase in how we're doing tmp directories in tests


cuixq (approved):

another-rex (approved):

G-Rath:
the linting check failure is a timeout, not an actual failure - running it should get it passing

PR #1382, opened 2024-11-05
test: rewrite `models` to use snapshots and tables

G-Rath:
This should make it easier to maintain since we won't need to edit raw JSON - I wouldn't say this is strictly better as some of the tests were small enough that I'd argue the initial overhead of using snapshots makes it less of a win, but ultimately these don't change often so I think it's still better to be consistent with the rest of the codebase

(I've also got a fix or two coming up in this section which'll be made easier by this)


G-Rath:
This will conflict with #1381 - that pull request would probably be the better one to land first

another-rex (approved):

PR #1383, opened 2024-11-06
feat(output): update HTML output to a new design

hogo6002:
HTML output added https://github.com/google/osv-scanner/pull/1258
New design:
- Changed the color scheme to match osv.dev.
- Merged the package view and vulnerability view into a single nested table.
- Added a layer view.
- Added a vulnerability type filter (project, OS, and uncalled).
- Added a vulnerability search bar.
- Refined the vulnerability severity count design.
- Integrated the osv.dev vulnerability page into results (allowing users to click to view details).
- Improved the UI.

image
image
image
image


internal/output/html/package_table_template.html
```diff
@@ -1,53 +1,62 @@
...
-    <td>{{ $element.InstalledVersion }}</td>
-    {{ if ne $element.HTMLVulnCount.UnFixed 0 }}
-    {{ if eq $element.HTMLVulnCount.Fixed 0}}
-    <td class="warning">
-      {{ $element.FixedVersion }}
+  {{ $index := uniqueID }}
+  <tr class="table-tr package-tr {{if $element.HTMLPackageDetail }}has-layer-info"
+    data-layer="{{ $element.HTMLPackageDetail.LayerID }}{{ end }}" id="table-tr-{{ $index }}">
+    <td class="icon-td">
+      <div class="expand-icon" onclick="showPackageDetails('{{ $index }}')">
```
oliverchang, line 14: Would it make sense to have the details expand if we click anywhere in this entire row? 

Right now, we can only expand if we click on the rather small triangle, which also feels like it's not ideal for accessibility.

internal/output/html/package_table_template.html
```diff
@@ -1,53 +1,62 @@
...
-        {{ range $index, $vuln := $element.CalledVulns }}
-        {{ if eq $vuln.Summary.FixedVersion "No fix available" }}
-        <p><a href="https://osv.dev/{{ $vuln.Summary.ID }}">{{ $vuln.Summary.ID }}</a>
-          ({{$vuln.Summary.SeverityRating}}) has no fix available</p>
-        {{ else }}
-        <p><a href="https://osv.dev/{{ $vuln.Summary.ID }}">{{ $vuln.Summary.ID }}</a>
-          ({{$vuln.Summary.SeverityRating}}) has a fixed version at {{$vuln.Summary.FixedVersion}}</p>
-        {{ end }}
+      <div id="table-tr-{{ $index }}-details" class="package-details">
+        {{ if and $element.HTMLPackageDetail (not (eq $element.HTMLPackageDetail.LayerID "")) }}
```
oliverchang, line 44: Is this "Layer introduced in"? We should be clearer about exactly what Layer means here.

oliverchang (commented):
this is awesome!!!

internal/output/html.go
```diff
@@ -335,11 +381,89 @@ func buildHTMLResult(ecosystemMap map[string][]HTMLSourceResult, resultCount HTM
...
+	all := 0
+
+	for _, ecosystem := range result {
+		for _, source := range ecosystem.Sources {
+			if ecosystem.IsOS {
+				osType += source.HTMLVulnCount.Called
+			} else {
+				projectType += source.HTMLVulnCount.Called
+			}
+			all += source.HTMLVulnCount.Called
```
G-Rath, line 412: question: is this actually "all called"?

suggestion: you could just calculate the value when returning the struct by doing `osType + projectType`

internal/output/html.go
```diff
@@ -335,11 +381,89 @@ func buildHTMLResult(ecosystemMap map[string][]HTMLSourceResult, resultCount HTM
...
-		HTMLVulnCount:    resultCount,
+		EcosystemResults:    ecosystemResults,
+		HTMLVulnCount:       resultCount,
+		IsContainerScanning: isContainerScanning,
+		AllLayers:           layers,
+		VulnTypeCount:       vulnTypeCount,
+	}
+}
+
+func getVulnTypeCount(result []HTMLEcosystemResult) VulnTypeCount {
```
G-Rath, line 399: nit: you should be able to take advantage of default values here I think, and so just do a `var vtc VulnTypCount` to get all the "variables" initialized to zero

internal/output/html.go
```diff
@@ -335,11 +381,89 @@ func buildHTMLResult(ecosystemMap map[string][]HTMLSourceResult, resultCount HTM
...
+	for _, ecosystem := range result {
+		for _, source := range ecosystem.Sources {
+			for _, packageInfo := range source.PackageResults {
+				layerID := packageInfo.HTMLPackageDetail.LayerID
+				layerCommand := packageInfo.HTMLPackageDetail.LayerCommand
+
+				// Check if this layer ID and command combination is already in the map
+				if _, ok := layerMap[layerID]; !ok {
+					var resultCount HTMLVulnCount
+					updateCount(&resultCount, &packageInfo.HTMLVulnCount)
```
G-Rath, line 439: nit: probably worth exploring at some point if it would be cleaner to make this a method on `HTMLVulnCount`

(pointers always confuse me since I don't work with them day-to-day, so I won't say off the top of my head that it would definitely be worth it 😅)

internal/output/html.go
```diff
@@ -335,11 +381,89 @@ func buildHTMLResult(ecosystemMap map[string][]HTMLSourceResult, resultCount HTM
...
+					layerCount[layerID] = resultCount
+				}
+			}
+		}
 	}
+
+	// Convert the map to a slice of LayerInfo
+	layers := make([]LayerInfo, 0, len(layerMap))
+	i := 0
+	for layerID, layerCommand := range layerMap {
```
G-Rath, line 455: issue: I feel like this could be a bug, as Go purposely uses a random order when iterating over maps

I would expect currently repeatedly scanning the same image will give you a random layer order each time, unless I've missed a sort somewhere?

(that also means your index is kind of useless)

internal/output/html/filter_template.html
```diff
@@ -0,0 +1,85 @@
+<div id="summary-section">
+  <div id="filter-section">
+    {{ if .IsContainerScanning }}
+    <div class="filter-container">
+      <label for="layer-filter">Layer<br></label>
+      <div id="layer-filter" class="filter" onclick="toggleFilter('layer')">
+        <p value="all" id="layer-filter-selected" class="filter-selected">All layers ({{ .HTMLVulnCount.Called }})</p>
```
G-Rath, line 7: issue: `value` is not allowed on most HTML tags, but you seem to be using it quite a lot - are you maybe meaning to use `title` instead?

internal/output/html/package_table_template.html
```diff
@@ -1,53 +1,62 @@
...
+          {{ else }}
+          <div>{{ $element.HTMLPackageDetail.LayerCommand }}</div>
+          {{ end }}
+        </div>
+        <p><span class="package-detail-title">Layer ID:</span> {{ $element.HTMLPackageDetail.LayerID }}</p>
+        <p><span class="package-detail-title">In Base Image:</span> {{ $element.HTMLPackageDetail.InBaseImage }}</p>
         {{ end }}
-      </div>
+        {{ template "vuln_table_template.html" $element }}
+
```
G-Rath, line 59: issue: you've mistakenly removed the closing `div`:

```suggestion
        {{ template "vuln_table_template.html" $element }}
      </div>
```

internal/output/html/style.html
```diff
@@ -56,101 +61,437 @@
...
-  .vuln-details {
+  .package-details {
     text-align: left;
     padding: 0 50px;
     display: none;
+    padding-top: 20px;
+    padding-bottom: 30px;
+    border-bottom: 1px solid rgba(255, 255, 255, 0.12);
+    max-height: 800px;
+    overflow: overlay;
```
G-Rath, line 130: issue: [`overlay` is not an official property](https://caniuse.com/css-overflow-overlay) - it technically has a difference to `auto` (which MDN suggests it's an alias for), but ultimately I think we should try to avoid using it even if that requires more work:

```suggestion
    overflow: auto;
```

internal/output/html_test.go
G-Rath, line 1: question: why are we no longer testing this...?

suggestion: let's use `t.Skip` rather than commenting these out

G-Rath (commented):
awesome stuff!

I do wonder if it might be worth trying to split this out into at least two PRs as it feels like you're adding a lot of independent stuff, but will leave that for you to decide.

Also, you might already have spotted this but the "overall severity" element isn't responsive:

image

I know we're not expecting these reports to be viewed on mobile, but I had this occur on my 4k 150%-scaled monitor as I put the report side-by-side (so, half of the resolution) this pr for reviewing

internal/output/html.go
hogo6002: Thanks, updated!

internal/output/html_test.go
hogo6002: The generated snapshots are too large for reviewers to effectively review. This might also not be the best way to test the front end. For now, just want to make sure the function runs correctly to prevent any code breakage.

internal/output/html.go
hogo6002: After this PR is merged, I'll move the HTMLResult struct from html.go to table.go to create a package summary for table output during container scanning. I plan to refactor the code then and will try to make this function cleaner. 

(I also don't like pointers.)

hogo6002:
> I do wonder if it might be worth trying to split this out into at least two PRs as it feels like you're adding a lot of independent stuff, but will leave that for you to decide.

I wanted to split this into separate PRs, but it's a bit challenging. Any code changes would require corresponding changes in other files, and I can't show the outputted HTML without all the features working together. (I can imagine it might be a bit hard to review.)

> Also, you might already have spotted this but the "overall severity" element isn't responsive:

Yes, that's by design. Implementing dynamic "overall severity" without a frontend framework needs a bit of effort. I also prefer to avoid writing excessive JS code, as it's difficult to review and maintain. If we adopt a framework in the future, this feature will be much easier to implement. So, I think it's not worth to implement it for now.

> I know we're not expecting these reports to be viewed on mobile, but I had this occur on my 4k 150%-scaled monitor as I put the report side-by-side (so, half of the resolution) this pr for reviewing

Good catch! added a `min-width` setting. It should be good now.

another-rex (commented):
Quick note, can you target this to v2 rather than the base branch?

hogo6002:
> Quick note, can you target this to v2 rather than the base branch?

There are quite a few conflicts. I will merge this into the main and then create a separate PR to merge from main into v2.

internal/output/html.go
```diff
@@ -41,14 +44,15 @@ type HTMLSourceResult struct {
...
-	HTMLVulnCount    HTMLVulnCount
+	Name              string
+	Ecosystem         string
+	Source            string
+	CalledVulns       []HTMLVulnResult
+	UncalledVulns     []HTMLVulnResult
+	InstalledVersion  string
+	FixedVersion      string
+	HTMLVulnCount     HTMLVulnCount
+	HTMLPackageDetail HTMLPackageDetail
```
another-rex, line 55: Maybe rename this to HTMLPackageLayerDetail since it is all about layers?

internal/output/html.go
```diff
@@ -67,15 +71,23 @@ type HTMLVulnResultSummary struct {
 	SeverityScore    string
 }
 
+// HTMLPackageDetail represents detailed layer tracing information about a package.
+type HTMLPackageDetail struct {
+	LayerCommand        string
+	LayerCommandTooltip string
+	LayerID             string
+	InBaseImage         string
```
another-rex, line 79: The InBaseImage be a boolean, and the text should be decided in the template file

internal/output/html.go
```diff
@@ -91,8 +103,23 @@ type HTMLVulnCount struct {
 	UnFixed  int
 }
 
+type LayerInfo struct {
```
another-rex, line 106: rename to HTMLLayerInfo?

Also can you move this above HTMLVulnCount, since that's a field in this.

internal/output/html.go
```diff
@@ -91,8 +103,23 @@ type HTMLVulnCount struct {
...
+	Index        int
+	LayerCommand string
+	LayerID      string
+	Count        HTMLVulnCount
+}
+
+type VulnTypeCount struct {
```
another-rex, line 113: Rename to HTMLVulnTypeCount?

internal/output/html.go
```diff
@@ -103,6 +130,8 @@ var baseImages = []string{"Debian", "Alpine", "Ubuntu"}
 //go:embed html/*
 var templates embed.FS
 
+var isContainerScanning = false
```
another-rex, line 133: Prefer not to have a global variable here (especially since if this code is ran twice this will break, which will be hard to debug).

another-rex: We can have a separate function that does it's own loop and figure out whether there are image origin fields.

another-rex (commented):
Added some comments, not looking too closely at the HTML, just going to test the functionality for bugs.

internal/output/html/package_table_template.html
hogo6002: Updated

hogo6002:
Updated (https://hogo6002.github.io/scanner-report/):

UI:
- Renamed filter name from "Types" to "Filters". This is because "uncalled/unimportant" is not a specific type, and it allows for more flexibility in the future if we want to add more filters, such as showing fixable vulns only (I thought about moving "uncalled/unimportant" out as a separate one, but I really like the current layout. Also, "uncalled/unimportant" ones aren't really important, and users should ignore these vulns, so it's not worth giving them a separate, more obvious position).
- Updated the selected filter count when checking or unchecking checkboxes. This also helps show how many items have been filtered out.
Screenshot 2024-11-12 at 5 20 49 PM
- Ensured rounded corner consistency for elements on each row.
- Added prompt text to indicate uncalled/unimportant ones.

Go:
- Renamed some structs to maintain consistency.
- Removed unnecessary global variables.

internal/output/html/style.html
```diff
@@ -396,18 +400,27 @@
   }
 
   .severity-short p {
-    border-radius: 1px;
     text-align: center;
   }
 
+  .severity-short-first p {
+    border-top-left-radius: 2px;
```
another-rex, line 407: nit: Make border radius 4px for everything to match osv.dev

hogo6002: I set a 4px border-radius for most elements (all tags), except for the filter container, which has an 8px radius. I think it looks better this way (also osv.dev has different border-radius setting for different components).

another-rex (approved):

oliverchang (approved):

Issue #1378, opened 2024-11-05
Displaying unimportant vulnerabilities in HTML reports


hogo6002:
Currently, we remove all the unimportant vulnerabilities in [`filterPackageVulns()`](https://github.com/google/osv-scanner/blob/8509b99610d5c5d1ca431f94c47f6046a64c0cc1/pkg/osvscanner/osvscanner.go#L748C6-L748C24) (This is also only done for Debian. We need to do it for Ubuntu and others if they have their own way of deciding priority/urgency.). This means users don't see them in the results and don't know which ones were removed. 

A better way to implement this is to include the unimportant vulnerabilities in the HTML results but hide them by default. This is similar to how we handle vulnerabilities that are not called. Users can then choose to see them by checking a filter.

![Image](https://github.com/user-attachments/assets/b223e346-4bbb-4863-85c2-0bde4d1342e6)

PR #1385, opened 2024-11-07
feat(container-scanning): set unimportant vulns as uncalled

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1378

### DIFF
--- a/pkg/osvscanner/vulnerability_result.go
+++ b/pkg/osvscanner/vulnerability_result.go
@@ -70,22 +70,10 @@ func buildVulnerabilityResults(
 			}
 		}
 
-		for _, vuln := range pkg.Vulnerabilities {
-			if isUnimportant(pkg.Package.Ecosystem, vuln.Affected) {
-				for i, group := range pkg.Groups {
-					if slices.Contains(group.IDs, vuln.ID) {
-						if group.ExperimentalAnalysis == nil {
-							pkg.Groups[i].ExperimentalAnalysis = make(map[string]models.AnalysisInfo)
-						}
-						// Set unimportant vulns as uncalled
-						pkg.Groups[i].ExperimentalAnalysis[vuln.ID] = models.AnalysisInfo{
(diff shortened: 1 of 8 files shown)


G-Rath (approved):

G-Rath:
fwiw "solves" is not an issue-closing-trigger-word (a term I just made up 😄); ideally you should use either "resolves" or "fixes"

pkg/osvscanner/vulnerability_result.go
```diff
@@ -69,6 +69,25 @@ func buildVulnerabilityResults(
...
+		for _, vuln := range pkg.Vulnerabilities {
+			if isUnimportant(pkg.Package.Ecosystem, vuln.Affected) {
+				for i, group := range pkg.Groups {
+					if slices.Contains(group.IDs, vuln.ID) {
+						if group.ExperimentalAnalysis == nil {
+							pkg.Groups[i].ExperimentalAnalysis = make(map[string]models.AnalysisInfo)
+						}
+						// Set unimportant vulns as uncalled
+						pkg.Groups[i].ExperimentalAnalysis[vuln.ID] = models.AnalysisInfo{
+							Called: false,
```
oliverchang, line 82: nit: `Called` is misleading here, because it implies we did reachability analysis.

Is it possible for us to add a new field here to be clearer? I.e. just `Unimportant` ?

hogo6002: Added an `Unimportant` configuration option under `ExperimentalAnalysis`. One concern is that this will explicitly assign `Unimportant` to false for any other records that have call analysis results in the JSON output (see changes in `internal/output/__snapshots__/machinejson_test.snap`). I'm reusing the `IsCalled()` function to check if a vulnerability is important. I try to avoid adding a new function to prevent direct changes to `pkg/`.

oliverchang (approved):

PR #1386, opened 2024-11-07
fix: A quick fix to handle symbolic links

another-rex:
Probably should not merge this PR directly, but this is a fix to allow osv-scanner to scan Ubuntu:20.04 based images (e.g. osv.dev workers)


hogo6002:
I did some testing on this, and it unblocked the issue we encountered before, but the reported result is still not correct. It reported fewer vulnerabilities than expected and only included Ubuntu-related vulnerabilities, not any related to project types.

PR #1388, opened 2024-11-07
refactor: use `.gohtml` extension rather than suffixing templates with `_template`

G-Rath:
This means that IDEs like Goland automatically infer these files as Go templates, while still making it clear what files are templates vs "native" files


hogo6002:
Thanks! I'll close this PR for now since the original files have been modified (lots of conflicts). I'll make the changes in a follow-up PR where I'll also refactor the code a bit.

PR #1389, opened 2024-11-07
refactor: merge duplicate class selectors

G-Rath:
The styles from both of these still get applied, but we might as well just use a single selector block 🤷


hogo6002 (approved):
Thanks!

PR #1394, opened 2024-11-12
fix(guided remediation): handle extraneous/missing packages in package-lock.json more leniently

michaelkedar:
Changes two things for package-lock.json parsing:
1. Made it so packages installed in unknown locations (e.g. under a non-existent package) are ignored (following npm's behaviour), instead of triggering a panic.
2. Made missing dependencies in the lockfile *not* cause an error. npm would typically raise an error here, *unless* the missing dependency under a workspace's dependency graph. I think it's ok if we're more lenient than npm here.

### DIFF
--- a/internal/resolution/lockfile/npm.go
+++ b/internal/resolution/lockfile/npm.go
@@ -19,14 +19,17 @@ type NpmReadWriter struct{}
 
 func (NpmReadWriter) System() resolve.System { return resolve.NPM }
 
+type npmDependencyVersionSpec struct {
+	Version string
+	DepType dep.Type
+}
+
 type npmNodeModule struct {
-	NodeID       resolve.NodeID
-	Parent       *npmNodeModule
-	Children     map[string]*npmNodeModule // keyed on package name
-	Deps         map[string]string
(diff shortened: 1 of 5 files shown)


internal/resolution/lockfile/npm.go
```diff
@@ -19,14 +19,17 @@ type NpmReadWriter struct{}
...
+	DepType dep.Type
+}
+
 type npmNodeModule struct {
-	NodeID       resolve.NodeID
-	Parent       *npmNodeModule
-	Children     map[string]*npmNodeModule // keyed on package name
-	Deps         map[string]string
-	OptionalDeps map[string]string
-	DevDeps      map[string]string // dev dependencies are also included in Deps
```
oliverchang, line 28: can you explain a  bit more why we don't need `DevDeps` and `OptionalDeps` here anymore?

michaelkedar: `Deps` now also contains the `dep.Type`, which has whether it's optional or dev (or peer) in it.

oliverchang (approved):

cuixq (approved):

PR #1395, opened 2024-11-12
fix: add "slices" import back in

michaelkedar:
#1329 removed the slices import in `vulnerability_result.go`, because it was no longer using it.
At the same time, #1385 added a new usage of the module.

Add the import back to make osv-scanner build again.


G-Rath (approved):

hogo6002 (approved):
Thanks Michael!

PR #1396, opened 2024-11-13
chore: Update v2 to main

hogo6002:
Using a separate branch to resolve the merge conflicts, but other than that this is a straightforward update of the v2 branch up to the latest main branch.

PR #1397, opened 2024-11-13
feat(output): add a unified output result

hogo6002:
Refactored the [HTMLResult](https://github.com/google/osv-scanner/blob/9ea8aa57b45890a288a138b600900679ba6c98d1/internal/output/html.go#L19) into a unified structure that can be reused by other output formats. This new structure presents results hierarchically, grouping vulnerabilities by ecosystem, source, and package. Compare to the original vulnerabilityResult, it contains more information, such as vulnerability count (called/uncalled/fixable/unfixable/severity), package/vulnerability fix information, etc.

Follow-up PRs:
- Merge this into the v2 branch (via https://github.com/google/osv-scanner/pull/1396)
- Leverage this structure to generate a table package-view output.
- Update html.go and HTML templates to use the OutputResult.

### DIFF
--- a/internal/output/output_result.go
+++ b/internal/output/output_result.go
@@ -13,76 +13,78 @@ import (
 	"github.com/google/osv-scanner/pkg/models"
 )
 
-// OutputResult represents the vulnerability scanning results for output report.
-type OutputResult struct { //nolint:revive
-	Ecosystems []OutputEcosystemResult
+// Result represents the vulnerability scanning results for output report.
+type Result struct {
+	Ecosystems []EcosystemResult
 	// Container scanning related
 	IsContainerScanning bool
-	AllLayers           []OutputLayerInfo
-	VulnTypeCount       OutputVulnTypeCount
-	VulnCount           OutputVulnCount
+	AllLayers           []LayerInfo
+	VulnTypeCount       VulnTypeCount
+	VulnCount           VulnCount
 }
 
-// OutputEcosystemResult represents the vulnerability scanning results for an ecosystem.
-type OutputEcosystemResult struct { //nolint:revive
+// EcosystemResult represents the vulnerability scanning results for an ecosystem.
+type EcosystemResult struct {
 	Name    string
-	Sources []OutputSourceResult
+	Sources []SourceResult
 	IsOS    bool
 }
 
-// OutputSourceResult represents the vulnerability scanning results for a source file.
-type OutputSourceResult struct { //nolint:revive
+// SourceResult represents the vulnerability scanning results for a source file.
+type SourceResult struct {
 	Name             string
 	Ecosystem        string
-	PackageTypeCount OutputCallAnalysisCount
-	Packages         []OutputPackageResult
-	VulnCount        OutputVulnCount
+	PackageTypeCount CallAnalysisCount
+	Packages         []PackageResult
+	VulnCount        VulnCount
 }
 
-// OutputPackageResult represents the vulnerability scanning results for a package.
-type OutputPackageResult struct { //nolint:revive
+// PackageResult represents the vulnerability scanning results for a package.
+type PackageResult struct {
 	Name             string
 	InstalledVersion string
 	FixedVersion     string
-	CalledVulns      []OutputVulnResult
-	UncalledVulns    []OutputVulnResult
-	LayerDetail      OutputPackageLayerDetail
-	VulnCount        OutputVulnCount
+	CalledVulns      []VulnResult
+	UncalledVulns    []VulnResult
+	LayerDetail      PackageLayerDetail
+	VulnCount        VulnCount
 }
 
-// OutputVulnResult represents a single vulnerability.
-type OutputVulnResult struct { //nolint:revive
+// VulnResult represents a single vulnerability.
+type VulnResult struct {
 	ID             string
 	GroupIDs       []string
 	Aliases        []string
+	IsFixable      bool
 	FixedVersion   string
-	SeverityRating string
+	SeverityRating severity.Rating
 	SeverityScore  string
 }
 
-// OutputPackageLayerDetail represents detailed layer tracing information about a package.
-type OutputPackageLayerDetail struct { //nolint:revive
-	LayerCommand        string
-	LayerCommandTooltip string
-	LayerID             string
-	InBaseImage         bool
+// PackageLayerDetail represents detailed layer tracing information about a package.
+type PackageLayerDetail struct {
+	LayerCommand         string
+	LayerCommandDetailed string
+	LayerID              string
+	InBaseImage          bool
 }
 
-type OutputLayerInfo struct { //nolint:revive
+type LayerInfo struct {
 	Index        int
 	LayerCommand string
 	LayerID      string
-	Count        OutputVulnCount
+	Count        VulnCount
 }
 
-// OutputVulnCount represents the counts of vulnerabilities by severity and fixed/unfixed status
-type OutputVulnCount struct { //nolint:revive
-	SeverityCount     OutputSeverityCount
-	CallAnalysisCount OutputCallAnalysisCount
-	FixableCount      OutputFixableCount
+// VulnCount represents the counts of vulnerabilities by severity and fixed/unfixed status
+type VulnCount struct {
+	CallAnalysisCount CallAnalysisCount
+	// Only called vulnerabilities are included in the severity and fixable counts.
+	SeverityCount SeverityCount
+	FixableCount  FixableCount
 }
 
-type OutputSeverityCount struct { //nolint:revive
+type SeverityCount struct {
 	Critical int
 	High     int
 	Medium   int
@@ -90,22 +92,28 @@ type OutputSeverityCount struct { //nolint:revive
 	Unknown  int
 }
 
-type OutputVulnTypeCount struct { //nolint:revive
+type VulnTypeCount struct {
 	All      int
 	OS       int
 	Project  int
 	Uncalled int
 }
 
-type OutputCallAnalysisCount struct { //nolint:revive
+type CallAnalysisCount struct {
 	Called   int
 	Uncalled int
 }
 
-type OutputFixableCount struct { //nolint:revive
+type FixableCount struct {
 	Fixed   int
 	UnFixed int
 }
+
+const UnfixedDescription = "No fix available"
+const VersionUnsupported = "N/A"
+
+// baseImages is a list of OS images.
+var baseImages = []string{"Debian", "Alpine", "Ubuntu"}
 
 // PrintOutputResults prints the output to the outputWriter.
 // This function is for testing purposes only, to visualize the result format.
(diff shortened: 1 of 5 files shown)


internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+		IsContainerScanning: isContainerScanning,
+		AllLayers:           layers,
+	}
+}
+
+// processOutputSource processes a single source (lockfile or artifact) and returns an SourceResult.
+func processOutputSource(packageSource models.PackageSource) OutputSourceResult {
+	var sourceResult OutputSourceResult
+	packages := make([]OutputPackageResult, 0)
+	packageMap := make(map[string]bool)
```
another-rex, line 199: s/packageMap/packageSet

Also, since this is a set, we can use the `map[string]struct{}` type

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+	"strings"
+
+	"github.com/google/osv-scanner/internal/identifiers"
+	"github.com/google/osv-scanner/internal/semantic"
+	"github.com/google/osv-scanner/internal/utility/severity"
+	"github.com/google/osv-scanner/pkg/models"
+)
+
+// OutputResult represents the vulnerability scanning results for output report.
+type OutputResult struct { //nolint:revive
```
another-rex, line 17: Why is there nolint for all of these types, can we resolve the lint instead?

Guessing it's caused by stuttering? Do you know where outside this package requires these types, if it is not required, let's make all of these private.

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+	// Sort packageResults to ensure consistent output
+	slices.SortFunc(packages, func(a, b OutputPackageResult) int {
+		return cmp.Or(
+			cmp.Compare(a.Name, b.Name),
+			cmp.Compare(a.InstalledVersion, b.InstalledVersion),
+		)
+	})
+	sourceResult.Name = packageSource.Source.String()
+	sourceResult.Packages = packages
+	sourceResult.Ecosystem = ecosystemName
```
another-rex, line 235: I'm wondering if we even need to copy Ecosystem here, can we just get the ecosystem from the first package entry to be explicit about where this ecosystem is coming from.

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+
+	// Default to UnfixedDescription if no fix version is found.
+	if maxFixVersion == "" {
+		maxFixVersion = UnfixedDescription
+	}
+
+	return maxFixVersion
+}
+
+func updateVulnCount(original OutputVulnCount, newAdded OutputVulnCount) OutputVulnCount {
```
another-rex, line 415: A cleaner way to do this add is to make these member methods on each of the types, and just name it Add(). 

It looks like we are also just assigning the output back onto `original`, so we can even take it by pointer and update in place, though I can see that being less flexible compared to returning and assigning manually.

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+		severityCount.Low += 1
+	case "UNKNOWN":
+		severityCount.Unknown += 1
+	}
+
+	return severityCount
+}
+
+func isOSEcosystem(ecosystem string) bool {
+	for _, image := range baseImages {
```
another-rex, line 465: Where is the baseImages variable defined? Can we move it into this file if it is more appropriate.

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+	ecosystemPrefix := models.Ecosystem(strings.Split(ecosystem, ":")[0])
+	maxFixVersion := ""
+	var vp semantic.Version
+	for _, vuln := range allVulns {
+		if vuln.FixedVersion == VersionUnsupported {
+			return UnfixedDescription
+		}
+
+		// Skip vulnerabilities without a fixed version.
+		if vuln.FixedVersion == UnfixedDescription {
```
another-rex, line 389: All of these constants should be moved from html.go to this file in this PR.

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+	return minFixVersion
+}
+
+// getPackageFixedVersion determines the highest version that resolves the most known vulnerabilities for a package.
+func getPackageFixedVersion(ecosystem string, allVulns []OutputVulnResult) string {
+	ecosystemPrefix := models.Ecosystem(strings.Split(ecosystem, ":")[0])
+	maxFixVersion := ""
+	var vp semantic.Version
+	for _, vuln := range allVulns {
+		if vuln.FixedVersion == VersionUnsupported {
```
another-rex, line 384: Hmm I'm really not sure about doing string matches, can we define a enum as a field in OutputVulnResult which specify the state of the fixed version

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+		}
+	}
+
+	return false
+}
+
+func getAllLayerInfo(result []OutputEcosystemResult) []OutputLayerInfo {
+	layerMap := make(map[string]string)
+	layerCount := make(map[string]OutputVulnCount)
+	layerIndex := 0
```
another-rex, line 477: layerIndex not used?

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+
+	for _, ecosystem := range result {
+		for _, source := range ecosystem.Sources {
+			for _, packageInfo := range source.Packages {
+				layerID := packageInfo.LayerDetail.LayerID
+				layerCommand := packageInfo.LayerDetail.LayerCommand
+
+				// Check if this layer ID and command combination is already in the map
```
another-rex, line 485:
```suggestion
				// Check if this layer ID is already in the map
```

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+				if _, ok := layerMap[layerID]; !ok {
```
another-rex, line 486: Hmm looking at this some more, does the if statement need to be here at all? Feel like we can just remove it since both sides of the statement basically does the same thing

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+			maxFixVersion = vuln.FixedVersion
+			vp = semantic.MustParse(maxFixVersion, ecosystemPrefix)
+
+			continue
+		}
+
+		// Update if the current vulnerability's fixed version is higher
+		if vp.CompareStr(vuln.FixedVersion) < 0 {
+			maxFixVersion = vuln.FixedVersion
+			vp = semantic.MustParse(maxFixVersion, ecosystemPrefix)
```
another-rex, line 403: Can you add a comment here explaining that maxFixVersion will always be valid because it has been parsed before.

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+	VulnCount        OutputVulnCount
+}
+
+// OutputVulnResult represents a single vulnerability.
+type OutputVulnResult struct { //nolint:revive
+	ID             string
+	GroupIDs       []string
+	Aliases        []string
+	FixedVersion   string
+	SeverityRating string
```
another-rex, line 59: The rating should be a enum type with the specific ratings

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+		} else {
+			uncalledVulnMap[representID] = vuln
+			count.CallAnalysisCount.Uncalled += 1
+		}
+	}
+
+	return calledVulnMap, uncalledVulnMap, count
+}
+
+func updateVulnerabilityDetails(vulnPkg models.PackageVulns, vulnMap map[string]OutputVulnResult, count *OutputVulnCount) {
```
another-rex, line 316: Couple things here:
- Maybe rename to updateVulnMap to be clear which argument you are updating
- vulnMap should be the first argument, since that is the one you are modifying
- count maybe should be the second one
- Need a docstring here explaining that you are updating inplace.

another-rex: This function is also currently doing two separate things, getting the next fixed version, and updating the count.

Can we separate these out to two separate functions, and rename this function populateFixedVersion or something similar?

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+				}
+			}
+		}
+	}
+
+	return minFixVersion
+}
+
+// getPackageFixedVersion determines the highest version that resolves the most known vulnerabilities for a package.
+func getPackageFixedVersion(ecosystem string, allVulns []OutputVulnResult) string {
```
another-rex, line 379: maybe rename to calculatePackageFixedVersion, since some computing is needed beyond just getting a field.

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+		packageResult.LayerDetail = packageLayerDetail
+	}
+
+	return packageResult
+}
+
+// processOutputVulnerabilityGroups processes vulnerability groups within a package.
+// It populates the called and uncalled vulnerability maps and updates the vulnerability count
+// based on the severity and call analysis information of each group.
+func processOutputVulnerabilityGroups(vulnPkg models.PackageVulns) (map[string]OutputVulnResult, map[string]OutputVulnResult, OutputVulnCount) {
```
another-rex, line 280: In the docstring, please label each of the return values.

another-rex (commented):
Thanks! Added some comments.

internal/output/output_result.go
hogo6002: It's because of the naming.
`exported: type name will be used as output.OutputResult by other packages, and that stutters; consider calling this Result (revive)`
Any objection if I rename it from `OutputResult` to `Result`?

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+// This structured format facilitates generating various output formats (e.g., table, HTML, etc.).
+func BuildOutputResults(vulnResult *models.VulnerabilityResults) OutputResult {
+	var ecosystemMap = make(map[string][]OutputSourceResult)
+	var resultCount OutputVulnCount
+
+	for _, packageSource := range vulnResult.Results {
+		// Temporary workaround: it is a heuristic to ignore installed packages
+		// which are already covered by OS-specific vulnerabilities.
+		// This filtering should be handled by the container scanning process.
+		// TODO(gongh@): Revisit this after container scanning supports comprehensive functionality.
```
oliverchang, line 133: nit: let's be clearer here. "Revisit this once container scanning can distinguish these cases".

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+	Aliases        []string
+	FixedVersion   string
+	SeverityRating string
+	SeverityScore  string
+}
+
+// OutputPackageLayerDetail represents detailed layer tracing information about a package.
+type OutputPackageLayerDetail struct { //nolint:revive
+	LayerCommand        string
+	LayerCommandTooltip string
```
oliverchang, line 66: nit: do we need to call this "Tooltip"? It couples and reveals the implementation detail of the actual frontend in code, which we should try to avoid. 

Would it make sense to call this "LayerCommandDetailed" or something similar?

internal/output/output_result.go
```diff
@@ -0,0 +1,538 @@
...
+			continue
+		}
+		for _, affectedRange := range affected.Ranges {
+			for _, affectedEvent := range affectedRange.Events {
+				// Skip if it's not a fix version event or the installed version is greater than the fix version.
+				if affectedEvent.Fixed == "" || vp.CompareStr(affectedEvent.Fixed) > 0 {
+					continue
+				}
+
+				// Find the minium fix version
```
oliverchang, line 367: nit: spelling of minimum

internal/output/output_result.go
```diff
@@ -0,0 +1,546 @@
...
+type FixableCount struct {
+	Fixed   int
+	UnFixed int
+}
+
+const UnfixedDescription = "No fix available"
+const VersionUnsupported = "N/A"
+
+// baseImages is a list of OS images.
+var baseImages = []string{"Debian", "Alpine", "Ubuntu"}
```
oliverchang, line 116: nit: would `osEcosystems` be more accurate here?

internal/output/output_result.go
```diff
@@ -0,0 +1,546 @@
...
+			LayerID:      layerID,
+			Count:        layerCount[layerID],
+		})
+		i++
+	}
+
+	return layers
+}
+
+func getOutputVulnTypeCount(result []EcosystemResult) VulnTypeCount {
```
oliverchang, line 509: nit: just `getVulnTypeCount` to reflect the naming change.

oliverchang (approved):

internal/output/output_result.go
hogo6002: Changed name to "Result"

internal/output/output_result.go
hogo6002: Added a `IsFixable` bool field

internal/output/output_result.go
hogo6002: Yeah good catch

internal/output/output_result.go
hogo6002: Moved the vulnerability count updating logic to an individual function and renamed the current function to updateVuln(). This function is used to add any information from OSV records (e.g. description) to the vulnResult here. Although it's currently only populating the fix version, I still like to keep the name for more flexibility in the future.

hogo6002:
I'm going to merge this PR first to create the package view for the table output. Feel free to leave comments, and I'll address them in a separate PR.

PR #1399, opened 2024-11-14
chore: update snapshots

hogo6002:


G-Rath (approved):
maybe we should see if we can somehow remove or even just ignore `openssl`? 🤔 I know this happens with other stuff, but it feels like `openssl` is a very juicy target (which isn't a surprise)

michaelkedar (approved):

PR #1408, opened 2024-11-19
perf(maven): cache responses from Maven registries

cuixq:
Follow up on https://github.com/google/osv-scanner/issues/1045

Currently we cache the decoded Maven metadata and projects after fetching from Maven registries. We are not able to know if a request is Not Found or not. We can avoid making redundant not found requests if we cache the whole response from Maven registries.

This PR caches the response body as well as the statue code. This greatly improves the resolution performance - resolving [registry.xml](https://github.com/google/osv-scanner/blob/main/cmd/osv-scanner/fixtures/maven-transitive/registry.xml) improves from 20s to 5s.

A potential caveat is that the new cache may not work with the existing cache, however considering the cache expires after some time and this is still an experimental feature, I don't worry too much on this.


internal/resolution/datasource/maven_registry.go
```diff
@@ -30,8 +31,12 @@ type MavenRegistryAPIClient struct {
 	// Cache fields
 	mu             *sync.Mutex
 	cacheTimestamp *time.Time // If set, this means we loaded from a cache
-	projects       *RequestCache[string, maven.Project]
-	metadata       *RequestCache[string, maven.Metadata]
+	responses      *RequestCache[string, response]
+}
+
+type response struct {
```
cuixq, line 37: I think of naming this as `mavenResponse` but this structure is also quite general

michaelkedar (approved):
LGTM

cuixq:
after a offline discussion, we probably only want to cache only the not found status code, so I will mark this PR to draft.

PR #1418, opened 2024-11-27
ci: add a workflow to validate renovate config

cuixq:
This PR defines a new workflow to validate renovate config.

### DIFF
--- a/.github/workflows/renovate-validator.yml
+++ b/.github/workflows/renovate-validator.yml
@@ -4,11 +4,11 @@ on:
   push:
     branches: [main, v2]
     paths:
-      - 'renovate.json'
+      - "renovate.json"
   pull_request:
     branches: [main, v2]
     paths:
-      - 'renovate.json'
+      - "renovate.json"
 
 jobs:
   validate:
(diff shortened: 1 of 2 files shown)


cuixq:
I think what you mean is [this](https://docs.renovatebot.com/config-validation/#validation-of-renovate-config-change-prs)?

I didn't prefer this because we have to document this rule somewhere to ask people to follow it.

I didn't know rule set - this can be an option.

G-Rath:
Yeah, combined with [this](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets#push-rulesets) (which apparently apply across forks which is interesting).

I don't think the documentation requirement is a big deal - it should be coverable by a comment in the contrib doc (or maybe even in the config itself), and really how often are we actually expecting the config to change, and for that to not be done by one of us?

cuixq:
Since I have done the work to set up the workflow and it works, I may not spend more time investing the other option if there is some reason against this approach?

G-Rath:
Well the downside is it's pulling in more dependencies which scorecard will frown on (sorry forgot to mention that in my last comment).

At the least, you need to pin everything in the workflow to a sha, which includes the npx call

cuixq:
I have updated the file to pin the dependencies in the workflow, also updated the file to trigger the validation only on PR with changes on `renovate.json`.

One thing I don't like about the other approach is that we need to wait until next time renovate runs on the repository.

another-rex (approved):

PR #1422, opened 2024-12-04
fix: replace Debian package name with its source name

hogo6002:
The current v2 container scanning doesn't report any Debian-related vulnerabilities. The reason is that the extractor takes the package name and package version to match against OSV records. But OSV records store records with the source name. Debian packages may have a different source name than their package name (also source version). For example: 
image

The new change will convert the given package information to its corresponding source information for matching, if the source information is specified.

### DIFF
--- a/internal/image/scan.go
+++ b/internal/image/scan.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/google/osv-scalibr/extractor"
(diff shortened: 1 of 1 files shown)


another-rex (approved):
Thanks! Can you make an issue TODO to make debian container scanning tests to catch this in the future?

PR #1425, opened 2024-12-06
test(maven): remove the test of transitive scanning on native data source

cuixq:
This test makes a lot of unnecessary not found requests on `repository.apache.org`. This PR removes the test until we fix it.


another-rex (approved):

michaelkedar (commented):
I don't see this pom using the `repository.apache.org` repo.

I think it's also in this [pom](https://github.com/google/osv-scanner/blob/152731f19fc89757b267e6007c4496e449f236a9/internal/remediation/fixtures/zeppelin-server/parent/parent/pom.xml#L83-L86). I'm pretty sure the test that uses it uses a local universe, but we might want to remove/change the url from the pom (and make sure it still builds)

cuixq:
I think this may be one of the reasons slowing down the resolution on the mentioned POM. I like the idea to remove the URL, at least before we understand the full picture here.

Issue #1419, opened 2024-11-29
Unimportant vulnerabilities should be printed as unimportant, not uncalled


another-rex:
Followup to the TODO's in #1385.

Currently vulnerabilities marked as unimportant are printed out as Uncalled.  This should be updated to print Unimportant.

PR #1427, opened 2024-12-06
fix(output): correct unimportant vuln type

hogo6002:
resolves https://github.com/google/osv-scanner/issues/1419

- Print unimportant vulnerabilities as unimportant, not uncalled.
- Add an `IsGroupUnimportant()` function to identify if a vulnerability is unimportant.
- Refactor the output result structure from `CalledVulns` and `UncalledVulns` to `RegularVulns` and `HiddenVulns`.
- Add `VulnAnalysisType` to indicate if a vulnerability is regular, unimportant, or uncalled. This can also be expanded in the future when we have more reachability analysis. 

Project scanning (Changed from "uncalled vulnerabilities" to "unimportant vulnerabilities" for Debian package scanning):
Screenshot 2024-12-06 at 4 17 46 PM
Contianer scanning (Added a "filtered vulnerabilities" table with a filter reason):
Screenshot 2024-12-06 at 4 19 06 PM


internal/output/output_result.go
```diff
@@ -94,22 +96,30 @@ type SeverityCount struct {
...
+	All     int
+	OS      int
+	Project int
+	Hidden  int
 }
 
-type CallAnalysisCount struct {
-	Called   int
-	Uncalled int
+type AnalysisCount struct {
```
another-rex, line 105: Can you add a comment here and above on VulnTypeCount to explain why there is two different counts?

internal/output/table.go
```diff
@@ -134,6 +146,30 @@ func printContainerScanningResult(result Result, outputWriter io.Writer, termina
 			outputTable.Render()
 		}
 	}
+
+	if result.VulnTypeCount.Hidden != 0 {
+		// Add a newline
+		fmt.Fprintln(outputWriter)
+		fmt.Fprintln(outputWriter, "Filtered Vulnerabilities:")
+		outputTable := newTable(outputWriter, terminalWidth)
+		outputTable.AppendHeader(table.Row{"Package", "ecosystem", "Installed Version", "Filtered vuln count", "Filter Reason"})
```
another-rex, line 155:
```suggestion
		outputTable.AppendHeader(table.Row{"Package", "Ecosystem", "Installed Version", "Filtered Vuln Count", "Filter Reasons"})
```

another-rex (commented):
Good idea with labelling each count type!

internal/output/output_result.go
hogo6002: renamed `VulnTypeCount` to `VulnTypeSummary` and added comments. The `VulnTypeSummary` here is mostly used to implement the HTML vuln type filter.

another-rex (approved):

PR #1434, opened 2024-12-09
chore: Merge V2 back into main

another-rex:
Now that the last release is out, the main branch should be the v2 branch, and we can continue developing on main.

One caveat, scorecard github action has not updated yet to support the new ignore file format.


hogo6002 (approved):

cuixq (approved):
🎉

Issue #1280, opened 2024-09-26
Do not fetch snapshots from the Maven repository disabling that


cuixq:
Currently, by default, we fetch snapshots from all Maven repositories.

However, snapshot can be disabled for a repository ([ref](https://maven.apache.org/pom.html#repositories)).

We should take this into consideration when fetching projects from a Maven registry.

PR #1436, opened 2024-12-10
feat(maven): only fetch snapshots and releases when enabled

cuixq:
https://github.com/google/osv-scanner/issues/1280

In Maven pom.xml, download policies can be specified for repositories to decide whether to fetch releases/snapshots or not, which is currently not supported in MavenRegistryAPIClient. This causes us making many unnecessary requests. 

This PR defines a new data structure to hold information for a Maven registry including the download policies. For the default registry, we assume it is only allowed to download releases (it can be a TODO to support specifying this via CLI flags).


internal/resolution/manifest/maven.go
```diff
@@ -97,8 +97,15 @@ func (m MavenReadWriter) Read(df lockfile.DepFile) (Manifest, error) {
...
 	}
+
+	// TODO: there may be properties in repo.Releases.Enabled and repo.Snapshots.Enabled
 	for _, repo := range project.Repositories {
-		if err := m.MavenRegistryAPIClient.AddRegistry(string(repo.URL)); err != nil {
+		if err := m.MavenRegistryAPIClient.AddRegistry(datasource.MavenRegistry{
+			URL:              string(repo.URL),
+			ID:               string(repo.ID),
+			ReleasesEnabled:  repo.Releases.Enabled != "false",
+			SnapshotsEnabled: repo.Snapshots.Enabled != "false",
```
michaelkedar, line 107: What's the difference between this vs using `.Boolean()` as in `extractor.go`?

Also, maybe this should be `== "true"`? If it's something other than 'true' or 'false' I think it's safer if we ignore the repository.

internal/resolution/client/client.go
```diff
@@ -35,7 +35,7 @@ type DependencyClient interface {
 }
 
 type Registry struct {
-	URL string
+	EcosystemSpecific any
 }
```
michaelkedar, line 39: Might as well make this `type Registry any`/`interface{}` or remove the type altogether.

oliverchang (approved):

internal/resolution/datasource/maven_registry.go
```diff
@@ -64,20 +78,25 @@ func (m *MavenRegistryAPIClient) WithoutRegistries() *MavenRegistryAPIClient {
...
 
 // AddRegistry adds the given registry to the list of registries if it has not been added.
-func (m *MavenRegistryAPIClient) AddRegistry(registry string) error {
-	if slices.Contains(m.registries, registry) {
-		return nil
+func (m *MavenRegistryAPIClient) AddRegistry(registry MavenRegistry) error {
+	for _, reg := range m.registries {
+		if reg.URL == registry.URL {
+			return nil
+		}
```
michaelkedar, line 85: There's an edge case if you have two registries with the same URL but different enabled snapshots/releases.
It might be not too bad to double-up on URLs anyway since the duplicate requests should be cached.

internal/resolution/datasource/maven_registry.go
```diff
@@ -64,20 +78,25 @@ func (m *MavenRegistryAPIClient) WithoutRegistries() *MavenRegistryAPIClient {
 }
 
 // AddRegistry adds the given registry to the list of registries if it has not been added.
-func (m *MavenRegistryAPIClient) AddRegistry(registry string) error {
-	if slices.Contains(m.registries, registry) {
-		return nil
+func (m *MavenRegistryAPIClient) AddRegistry(registry MavenRegistry) error {
```
michaelkedar, line 81: Should this function be considering what happens when registries have the same `ID`?

michaelkedar (commented):
Nice

internal/resolution/manifest/maven.go
cuixq: opps! this is outdated - `Boolean()` is now introduced so I should update the code here.

internal/resolution/datasource/maven_registry.go
cuixq: I just updated the code to use the ID to decide if the registry should be added, which matches Maven's behavior.

internal/resolution/datasource/maven_registry.go
cuixq: With using ID to check if a registry should be added, this edge case should be handled now since registries with the same URL may still be both added if the IDs are different.

internal/utility/maven/maven.go
```diff
@@ -84,7 +84,12 @@ func MergeParents(ctx context.Context, mavenClient *datasource.MavenRegistryAPIC
 			return fmt.Errorf("failed to merge profiles: %w", err)
 		}
 		for _, repo := range proj.Repositories {
-			if err := mavenClient.AddRegistry(string(repo.URL)); err != nil {
+			if err := mavenClient.AddRegistry(datasource.MavenRegistry{
+				URL:              string(repo.URL),
+				ID:               string(repo.ID),
+				ReleasesEnabled:  repo.Releases.Enabled != "false",
+				SnapshotsEnabled: repo.Snapshots.Enabled != "false",
```
michaelkedar, line 91: This should also be `.Boolean()`?

michaelkedar (approved):

PR #1437, opened 2024-12-10
feat(guided remediation): non-interactive JSON output format

### LINKED ISSUE #1213: Better `osv-scanner fix --non-interactive` output
The current [guided remediation terminal output](https://google.github.io/osv-scanner/experimental/guided-remediation/#basic-usage) was mostly just thrown together as the easiest way to output the bare minimum information needed for the sample scripted usage.

We should put more thought into how and what information it outputs. Some initial thoughts:
- A better, machine-readable format, probably JSON
  - Option for other formats to match `osv-scanner scan` (i.e. a table format)?
- What information is useful / possible to output? e.g.
  - Something along the lines of what `osv-scanner scan` outputs
  - The possible patches & what vulns they fix
  - The impossible to fix vulns and what constraints are limiting it - #925
  - Anything else
- Multiple output verbosity levels.
- Better logging of the internals (with verbosity).
- Separate logs & output on stdout/stderr like the scan action does

Open to suggestions

michaelkedar:
Closes #1213
Adds `--format` flag to `osv-scanner fix --non-interactive` to choose between old `text` and new `json` format, which provides a more machine-readable description of the changes made by guided remediation. The `text` format should remain unchanged.

Also, renamed the `relock` strategy to `relax` (though maintained backwards-compatibility) to be a bit less ambiguous.

There are a few things that'll need to change with osv-scanner v2. In particular, the way the reporter interface is being used. I'll need to revisit this when those are refactored.


cmd/osv-scanner/fix/output.go
cuixq, line 1: can you add documentation to these newly defined structs?

cmd/osv-scanner/fix_test.go
```diff
@@ -60,7 +60,24 @@ func TestRun_Fix(t *testing.T) {
...
-		// TODO: add tests with the cli flags
+		{
+			name:     "fix non-interactive json in-place package-lock.json",
+			args:     []string{"", "fix", "--non-interactive", "--strategy=in-place", "--format=json"},
+			exit:     0,
+			lockfile: "./fix/fixtures/in-place-npm/package-lock.json",
+		},
+		{
+			name:     "fix non-interactive json relock package.json",
+			args:     []string{"", "fix", "--non-interactive", "--strategy=relock", "--format=json"},
```
cuixq, line 71: `--strategy=relax`?

cuixq (approved):
LGTM!

oliverchang (approved):

cmd/osv-scanner/fix_test.go
michaelkedar: good catch - fixed

cmd/osv-scanner/fix/output.go
michaelkedar: documentation added

oliverchang:
Is this ready to merge? 

We're probably still pending feedback from a few folks, but we can iterate on those right?

michaelkedar:
Yeah, this should be okay to merge

PR #1438, opened 2024-12-11
refactor: Model migration

another-rex:
# Changes in this PR:

## Temporary changes:
I have made several temporary changes to be reversed in follow up PRs to make this current PR more reviewable, the most notable of them are listed below, and these + the rest are all marked with TODO(v2) in the code:

- Disabled image scanning code
  - This includes commenting out all container scanning tests
  - This is because we are going to migrate to use the osvscalibr image scanning which has now been merged in
  - Not done in this PR as that would require updating osv-scalibr, which includes a number of breaking changes
- Disabled vendor scanning code
  - Vendor scanning code will be converted to the extractor interface, and use the new clients as part of V2
  - Not a core part of this refactor, so disabled for now.
- pomExtractor is initialized at the start and passed via a special argument to avoid cyclic dependencies occurring when piping through the TransitiveScannerActions. This will mostly likely need a more general solution long term
- Local db actually converts query back into lockfile.PackageDetails, which works fine as part of this refactor, but we will want to completely remove that conversion and use imodels.PackageInfo directly, so we can remove the lockfile package entirely, and follow the new client interface.

## Permanent changes:

These are the changes this refactor actually accomplishes, in no particular order

1. Unify all the intermediate types to use an internal models package (imodels) following the models refactor design doc. This made changes in:
    1. `scanresults.go` (Scan results need to be in a separate package, as it imports ScanParameters and configManager, which would cause a cyclic import if it is with the rest of imodels)
2. Use the new osvschema Ecosystem type, and made a new `ecosystem.Parsed` type, this made changes in:
3. Update lockfilescalibr to perform extraction more like osv-scalibr - Given a list of extractors, go through them and use FileRequired to determine the correct extractor to use for a given path.
    1. Translation between osv-scanner names and scalibr names are now done in the new `scanners` package.
    2. Inventory are sorted and deduped right after extraction from every extractor.
4. Create "extractors" for existing scanning methods that are not part of the extractor interface
    1. Created Git extractor, which is where the current git scanning capabilities (HEAD git hash, submodule scanning) has been migrated to.
5. Removed existing SBOM scanners with Scalibr extractors
6. Perform sorting
7. Moved filtering code and scanning code out of `osvscanner.go`
    1. Filtering has been moved to `filter.go`, and arguments updated to use the new types. Minimal changes has been made to logic.
    2. Scanning code has been moved to `scan.go`. This code has been significantly changed and should be reviewed.
8. Created the internal `scanners` package which contains the logic for each type of scanning.
9. Java transitive dependency extractor has it's name renamed to `java/pomxml`, same as the extractor in scalibr.

### File reference

This lists most of the files that have changed, and their corresponding feature that it is changed for
- `pkg/osv/osv.go`: Removed a completely redundant if statement as commit queries are made in another function (proven by having 0 test coverage in that if branch)
- `internal/config/config.go`: Feature 1.
- `internal/depsdev/license.go`: Feature 2.
- `internal/imodels/ecosystem/ecosystem.go`: Feature 2.
- `internal/imodels/imodels.go`: Feature 1.
- `internal/imodels/results/scanresults.go`: Feature 1.
- `internal/lockfilescalibr/errors.go`: Removed redundant error, and converted all of the removed error to ErrExtractorNotFound
- `internal/lockfilescalibr/invsort.go`: Feature 3b.
- `internal/lockfilescalibr/translation.go`: Feature 3a, translation.go is renamed to extract.go.
- `internal/lockfilescalibr/vcs/gitrepo/`: Feature 4a
- `pkg/osvscanner/filter.go` and `scan.go`: Feature 7
- `pkg/osvscanner/internal/scanners/lockfile.go`: Feature 8. Used specifically for single lockfile scanning, and contains the logic for the parseAs syntax of osv-scanner, and includes custom extractors that are not enabled for standard directory scanning.
  - Also contains the list of all (most) lockfile extractors
- `pkg/osvscanner/internal/scanners/sbom.go`: Feature 8. Used for osv-scanner --sbom flag
- `pkg/osvscanner/internal/scanners/walker.go`: Similar to how osv-scalibr does walking, builds a list of extractors depending on the args, then walk and pass every file to the list of extractors.
- `pkg/osvscanner/vulnerability_result.go`: Feature 1.
- `main_test.snap`:
  - There are significant changes to the main snapshot, though I tried to minimize any unnecessary changes as possible in this PR, with followup PRs to update/improve wording
  - SBOM extractors no longer state the name of the extractor (This will be reverted in the future to have  everything log the name of the extractor)
  - The new SBOM extractor also extracts some local/unscannable packages, so the package number is increased by 1, and then 1 package is filtered out later.
  - Ordering of some packages have been changed
  - Error message have been updated to include more details.
  - Some warning messages are no longer printed to logs, but instead printed directly to stderr. This needs to be examined further in future PRs.


internal/imodels/imodels.go
cuixq, line 1: can you add some documentation to the newly defined strcuts?

pkg/osvscanner/filter.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/google/osv-scanner/pkg/reporter"
+)
+
+// filterUnscannablePackages removes packages that don't have enough information to be scanned
+// e,g, local packages that specified by path
+func filterUnscannablePackages(r reporter.Reporter, scanResults *results.ScanResults) {
+	packageResults := make([]imodels.PackageScanResult, 0, len(scanResults.PackageScanResults))
+	for _, psr := range scanResults.PackageScanResults {
+		p := psr.PackageInfo
+		switch {
```
cuixq, line 19: can we just use `if` statement here?

another-rex: Hmm it looks quite ugly/unreadable with if statements here, need to look something like:
```
		if (p.Ecosystem.Ecosystem == "" || p.Name == "" || p.Version == "") &&
			(p.Commit == "") {
			continue
		}
```

internal/imodels/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if p.Suffix != "" {
+		str += ":" + p.Suffix
+	}
+
+	return str
+}
+
+// Parse parses a string into a constants.Ecosystem and an optional suffix specified with a ":"
+func Parse(str string) Parsed {
```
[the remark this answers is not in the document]
another-rex: Mmm... I renamed Parse to MustParse and added a function that does checking. I'm not sure if this is the right way to go though, we have to keep the list in osv-scanner updated with osv-schema changes, which could be difficult to do consistently, Also it's not clear what to do with the error, right now I just have it printing a log.

internal/imodels/results/scanresults.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/google/osv-scanner/internal/config"
+	"github.com/google/osv-scanner/internal/imodels"
+)
+
+type ScanResults struct {
+	PackageScanResults []imodels.PackageScanResult
+
+	// The user parameters for the scan
+	// ScanParameters
```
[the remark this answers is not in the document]
another-rex: Yes, added a TODO to this to enable once ScanResults have been moved out of osvscanner.go to models.go

internal/imodels/imodels.go
another-rex: Done, added comments to all the new structs in imodels package, hopefully I didn't miss any.

pkg/osvscanner/vulnerability_result.go
```diff
@@ -109,12 +98,32 @@ func buildVulnerabilityResults(
...
+			case imodels.SourceTypeProjectPackage:
+				sourceType = "lockfile"
+			case imodels.SourceTypeSBOM:
+				sourceType = "sbom"
+			case imodels.SourceTypeGit:
+				sourceType = "git"
+			case imodels.SourceTypeUnknown:
+				sourceType = "unknown"
+			default:
+				sourceType = "unknown"
```
oliverchang, line 120: does it make sense to just have a single `default` that also covers the `SourceTypeUnknown` case?

internal/lockfilescalibr/extract.go
```diff
@@ -0,0 +1,119 @@
...
+func ExtractWithExtractors(ctx context.Context, localPath string, extractors []filesystem.Extractor) ([]*extractor.Inventory, error) {
+	info, err := os.Stat(localPath)
+	if err != nil {
+		return nil, err
+	}
+
+	result := []*extractor.Inventory{}
+	extractorFound := false
+	for _, ext := range extractors {
+		if ext.FileRequired(localPath, info) {
```
oliverchang, line 62: nit: invert condition and continue to avoid additional indentation?

internal/imodels/imodels.go
```diff
@@ -0,0 +1,164 @@
...
+	Location   string // Contains Inventory.Locations[0]
+	SourceType SourceType
+
+	Commit     string
+	Repository string
+
+	// For package sources
+	DepGroups []string
+
+	// For OS packages
```
oliverchang, line 51: can you add a bit more details to the comments here to explain that for OS packages where the name is the source package, this is typically the binary name?

internal/lockfilescalibr/errors.go
```diff
@@ -5,5 +5,4 @@ import "errors"
 var ErrIncompatibleFileFormat = errors.New("file format is incompatible, but this is expected")
 var ErrNotImplemented = errors.New("not implemented")
 var ErrWrongExtractor = errors.New("this extractor did not create this inventory")
-var ErrExtractorNotFound = errors.New("could not determine extractor")
-var ErrNoExtractorsFound = errors.New("no extractors found to be suitable to this file")
+var ErrExtractorNotFound = errors.New("could not determine extractor suitable to this file")
```
oliverchang, line 8: nit: should we just rename `lockfilescalibr` to something else? "lockfile" has never been a great name. 

perhaps "scalibrextract" ?

internal/lockfilescalibr/vcs/gitrepo/extractor.go
```diff
@@ -0,0 +1,120 @@
...
+
+// FileRequired returns true for .package-lock.json files under node_modules
+func (e Extractor) FileRequired(path string, fi fs.FileInfo) bool {
+	return fi.IsDir() && filepath.Base(path) == ".git"
+}
+
+// Extract extracts packages from yarn.lock files passed through the scan input.
+func (e Extractor) Extract(_ context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	// Assume this is fully on a real filesystem
+	// TODO: Make this support virtual filesystems
```
oliverchang, line 77: Is there a lot of glue code that would have to be written to make this work?

oliverchang (approved):
very nice!! mostly looks good, just some minor nits and questions. 

Thanks for the very detailed description of changes in the PR description. This really helped with reviewing.

pkg/osvscanner/vulnerability_result.go
another-rex: It does, but it fails the linter since we no longer cover all the enum cases. Maybe it should panic at default since source type should always be valid here, but having it be unknown doesn't hurt.

internal/lockfilescalibr/errors.go
another-rex: Done

internal/lockfilescalibr/vcs/gitrepo/extractor.go
another-rex: yeah, git also uses a virtual filesystem, but with different signatures compared to our filesystem from what I can tell. Need to look into this more.

pkg/osvscanner/filter.go
cuixq: hmm I see - I was curious if this is for some future extensibility...

internal/imodels/ecosystem/ecosystem.go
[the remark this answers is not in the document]
cuixq: agree that it's difficult to keep osv-scanner updated with osv-schema. I don't think this blocks the refactoring, more like a thing good to have so either way sounds good to me.

cuixq (approved):

G-Rath (approved):
very exciting!

cmd/osv-scanner/main_test.go
```diff
@@ -750,115 +748,117 @@ func TestRun_Licenses(t *testing.T) {
...
-						t.Fatalf("%s does not exist - have you run scripts/build_test_images.sh?", arg)
-					}
-				}
-			}
-
-			testCli(t, tt)
-		})
-	}
-}
+// TODO(v2): Image scanning is not temporarily disabled
```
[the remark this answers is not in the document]
another-rex: Good idea

spencerschrock:
@another-rex after this commit (i noticed when bisecting another issue), git extraction seems to be bugged for me:

Before:
```
osv-scanner scan -r /tmp/dubbo
Scanning dir /tmp/dubbo
Scanning /tmp/dubbo/ at commit c6dbd5fce35fa38273f53cb79f44806acf0d29ea
```

After:
```
osv-scanner scan -r /tmp/dubbo
Scanning dir /tmp/dubbo
Error during extraction: (extracting as vcs/gitrepo) worktree not available in a bare repository
```

oliverchang:
> @another-rex after this commit (i noticed when bisecting another issue), git extraction seems to be bugged for me:
> 

I believe this is intentional and will be brought back by https://github.com/google/osv-scanner/pull/1451.

spencerschrock:
I'll note the repo in question isn't a bare repository either. Just a normal `git clone <uri>`

another-rex:
I think the error is coming from trying to extract submodules, does the git repository you cloned have submodules?

spencerschrock:
I don't believe so, the example repository is github.com/SpringCloud/spring-cloud-dubbo

Issue #1411, opened 2024-11-20
Do mutliple BatchQueries requests in parallel


michaelkedar:
BatchQueries have to be chunked into groups of 1000 packages per request.
We currently make each chunked BatchQuery sequentially:
https://github.com/google/osv-scanner/blob/8d59ca5d95eca2b5c3b5cb4cb0d519fd468b3129/pkg/osv/osv.go#L185-L191

We could get a decent performance improvement on large scans if we do these in parallel, though we might want to make sure the OSV API is okay with these kinds of spikes.

oliverchang:
@hogo6002 putting this in your queue in case you're ever blocked waiting on other things :)

PR #1440, opened 2024-12-11
feat(performance): parallelize batch query sending

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1411#event-15601233836

This change parallelizes BatchQuery sending with the maximum concurrent number set at 10.

Performance improvement: Using `rocket.chat.json` as an example (3 BatchQuery sending), it reduces the scanning time from 26 seconds to 17 seconds.

### DIFF
--- a/pkg/osv/osv.go
+++ b/pkg/osv/osv.go
@@ -27,9 +27,10 @@ const (
 	BaseVulnerabilityURL = "https://osv.dev/"
 	// maxQueriesPerRequest splits up querybatch into multiple requests if
 	// number of queries exceed this number
-	maxQueriesPerRequest  = 1000
-	maxConcurrentRequests = 1000
-	maxRetryAttempts      = 4
+	maxQueriesPerRequest       = 1000
+	maxConcurrentRequests      = 1000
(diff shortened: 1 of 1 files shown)


pkg/osv/osv.go
```diff
@@ -181,42 +182,65 @@ func MakeRequest(request BatchedQuery) (*BatchedResponse, error) {
...
-			return nil, err
-		}
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
 
-		totalOsvResp.Results = append(totalOsvResp.Results, osvResp.Results...)
+	var totalOsvResp BatchedResponse
+	for _, resutls := range totalOsvRespResults {
+		totalOsvResp.Results = append(totalOsvResp.Results, resutls...)
```
michaelkedar, line 243: nit: typo
```suggestion
	for _, results := range totalOsvRespResults {
		totalOsvResp.Results = append(totalOsvResp.Results, results...)
```

michaelkedar (approved):

cuixq (approved):

PR #1451, opened 2024-12-17
feat: Vendored extractor

another-rex:
This PR adds the vendored extractor. The contents under testdata is zlib vendored in with a couple of files removed. 

If there is a better way to do this without clogging up the number of files git istracking, please let me know. 

Importantly, this is not yet enabled in the actual extraction, as there are bugs in the walker which will be resolved in a separate PR.


internal/scalibrextract/filesystem/vendored/vendored.go
```diff
@@ -0,0 +1,179 @@
...
+	// Stat costs performance, so perform it after the name check
+	stat, err := fapi.Stat()
+	if err != nil {
+		return false
+	}
+
+	return stat.IsDir()
+}
+
+// Extract extracts packages from yarn.lock files passed through the scan input.
```
oliverchang, line 85: nit: update comment.

internal/scalibrextract/filesystem/vendored/vendored.go
```diff
@@ -0,0 +1,179 @@
...
+	return nil
+}
+
+// Ecosystem returns the OSV ecosystem ('npm') of the software extracted by this extractor.
+func (e Extractor) Ecosystem(_ *extractor.Inventory) string {
+	return ""
+}
+
+func queryDetermineVersions(repoDir string, fsys scalibrfs.FS, scanGitDir bool) (*osv.DetermineVersionResponse, error) {
+	fileExts := []string{
```
oliverchang, line 121: should we move this up as a global instead to be consistent with `vendoredLibNames` ?

internal/scalibrextract/filesystem/vendored/vendored.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+
+			if _, ok := vendoredLibNames[strings.ToLower(d.Name())]; ok {
+				// Ignore nested vendored libraries, as they can cause bad matches.
+				return filepath.SkipDir
+			}
+
+			return nil
+		}
+		for _, ext := range fileExts {
```
oliverchang, line 149: nit: should we use `slices.Contains` here instead?

internal/scalibrextract/filesystem/vendored/vendored.go
```diff
@@ -0,0 +1,179 @@
...
+
+// Version of the extractor.
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{}
+}
+
+// FileRequired returns true for .package-lock.json files under node_modules
```
oliverchang, line 66: nit: update comment.

oliverchang (approved):

internal/scalibrextract/filesystem/vendored/vendored.go
another-rex: Done, also updated some other comments in the other extractors as they were outdated as well.

internal/scalibrextract/filesystem/vendored/vendored.go
another-rex: Yep, done

cuixq (approved):

PR #1452, opened 2024-12-17
fix: walker now properly uses Root and passes non absolute paths

another-rex:
fs.FS is supposed to use only non absolute paths, and be rooted at the real path Root. This change does this.

As a consequence of this, some error messages will not be rooted at Root, so also find and replace without the initial leading /. I wonder how this will work on windows though. 

Follow up TODO is to update the pomxml extractor to use the virtual filesystem, rather than using the real filesystem with the Root shortcut.


oliverchang (approved):

internal/scalibrextract/extract.go
```diff
@@ -103,19 +115,37 @@ func extractWithExtractor(ctx context.Context, localPath string, info fs.FileInf
...
-		Root:   "/",
+		Root:   root,
 		Reader: reader,
 		Info:   fileInfo,
 	}
 
 	return &si, nil
 }
+
+func getRootDir(path string) string {
```
cuixq, line 141: can you add some comment to this function? I don't quite get the logic here (maybe due to lacking of context)

internal/scalibrextract/language/java/pomxmlnet/extractor.go
```diff
@@ -69,7 +69,7 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
 		}
 	}
 	// Merging parents data by parsing local parent pom.xml or fetching from upstream.
-	if err := mavenutil.MergeParents(ctx, e.MavenRegistryAPIClient, &project, project.Parent, 1, input.Path, true); err != nil {
+	if err := mavenutil.MergeParents(ctx, e.MavenRegistryAPIClient, &project, project.Parent, 1, filepath.Join(input.Root, input.Path), true); err != nil {
```
cuixq, line 72: shall we apply the change to other places calling `mavenutil.MergeParents`?

another-rex: I had a look at all the other instances of MergeParents, and none of them need this join, as they get the full path already (or are just empty).

It does look like we need to update the methods in osv-scanner fix to not depend on lockfile and move to the new interface.

internal/scalibrextract/extract.go
another-rex: Added some comments. This is just used to set the Root variable in scalibr. This is needed because go's fs.FS cannot use absolute paths, and always has to be relative to something. This gets the top level directory for it to be rooted at.

cuixq: I don't see new commits?

another-rex: Didn't push 🤦 , updated now

cuixq: thanks!

cuixq (approved):

PR #1456, opened 2024-12-18
fix: remove redundant calls from `PreFetch`

michaelkedar:
In every `DependencyClient` we have, `MatchingVersions()` is just a call to `Versions()` plus semver matching, which is computationally expensive. It also was being called on every edge of the pre-fetched graphs.

Removed it to reduce CPU usage and hopefully improve performance with Maven resolution / Guided Remediation.

I've also skipped fetching things with `MavenDependencyOrigin` set (e.g. dependencyManagement dependencies) since there's potentially hundreds of them.

### DIFF
--- a/internal/resolution/client/client.go
+++ b/internal/resolution/client/client.go
@@ -6,6 +6,7 @@ import (
 
 	pb "deps.dev/api/v3"
 	"deps.dev/util/resolve"
+	"deps.dev/util/resolve/dep"
 	"github.com/google/osv-scanner/internal/depsdev"
 	"github.com/google/osv-scanner/pkg/models"
 	"github.com/google/osv-scanner/pkg/osv"
@@ -62,6 +63,10 @@ func PreFetch(ctx context.Context, c DependencyClient, requirements []resolve.Re
 
 	// Use the deps.dev client to fetch complete dependency graphs of our direct imports
 	for _, im := range requirements {
+		// There are potentially a huge number of management/import dependencies.
+		if im.Type.HasAttr(dep.MavenDependencyOrigin) {
+			continue
+		}
 		// Get the preferred version of the import requirement
 		vks, err := c.MatchingVersions(ctx, im.VersionKey)
 		if err != nil || len(vks) == 0 {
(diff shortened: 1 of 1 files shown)


cuixq (approved):
Thanks Michael! Do you know how much this could improve the performance?

michaelkedar:
I'm not exactly sure - it would help CPU usage and the number of goroutines running, but I'm not sure how this translates to run time.

oliverchang (approved):

spencerschrock:
> Do you know how much this could improve the performance?
> but I'm not sure how this translates to run time.

For my specific use case, on my specific machine (sample repo I was using to debug), performance is:

v1.9.0: ~2 min
v1.9.1: ~12 min
#1436: ~8 min
This PR: ~8min

michaelkedar:
I added another change to this PR that should hopefully reduce the number of calls for Maven specifically.

cuixq (approved):

spencerschrock:
> I added another change to this PR that should hopefully reduce the number of calls for Maven specifically.

Much faster, even more than 1.9.0! Thanks!

spencerschrock:
I will note I'm getting more extraction errors on some repos (such as https://github.com/SpringCloud/spring-cloud-gray) though. I'm not a big Maven dev, so I'm not sure which is valid/right.

```
Scanning dir /tmp/spring-cloud/
Error during extraction: (extracting as vcs/gitrepo) worktree not available in a bare repository
Error during extraction: (extracting as java/pomxml) failed resolving {Maven:cn.springcloud.gray:spring-cloud-gray-client[Concrete:${revision}] {}}: failed parsing version constraint 'Maven:org.springframework:spring-jcl[Requirement:]': invalid `>` in `>=0.0.0`
Error during extraction: (extracting as java/pomxml) failed resolving {Maven:cn.springcloud.gray:spring-cloud-gray-client-netflix[Concrete:${revision}] {}}: failed parsing version constraint 'Maven:org.springframework:spring-jcl[Requirement:]': invalid `>` in `>=0.0.0`
Scanned /tmp/spring-cloud/spring-cloud-gray-code-component/pom.xml file and found 1 package
Scanned /tmp/spring-cloud/spring-cloud-gray-core/pom.xml file and found 1 package
...
```

michaelkedar:
Hmm those errors seems like it would have been introduced in v1.9.1 - might need to look into that further.

PR #1458, opened 2024-12-19
fix: Fix VCS extractor and suppresses vcs errors 

another-rex:
Currently the extractor is opening the .git directory, when the extractor should be opening the parent directory.

Also changed so on empty repositories, (or empty submodules) it will not return an error. 
This is following the osv-scalibr convention where if an extractor could fail to extract on valid files, and the error is sort of expected to happen some of the time, it should not return an error.

Ideally we might want to have this type of error return a special error type.

FYI  @spencerschrock This fixes the error discussed in #1438

### DIFF
--- a/internal/scalibrextract/vcs/gitrepo/extractor.go
+++ b/internal/scalibrextract/vcs/gitrepo/extractor.go
@@ -12,6 +12,8 @@ import (
 	"github.com/google/osv-scalibr/purl"
 )
 
+// Extractor extracts git repository hashes including submodule hashes.
+// This extractor will not return an error, and will just return no results if we fail to extract
 type Extractor struct{}
 
 var _ filesystem.Extractor = Extractor{}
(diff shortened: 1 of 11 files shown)


internal/scalibrextract/vcs/gitrepo/extractor.go
```diff
@@ -84,23 +86,26 @@ func (e Extractor) FileRequired(fapi filesystem.FileAPI) bool {
 func (e Extractor) Extract(_ context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
 	// Assume this is fully on a real filesystem
 	// TODO: Make this support virtual filesystems
```
oliverchang, line 88: nit: add a comment to explain why we call filepath.Dir?

oliverchang (approved):

internal/scalibrextract/vcs/gitrepo/extractor.go
another-rex: Done

PR #1464, opened 2024-12-23
feat(V2 Clients): OSV.dev client

another-rex:
Implementation of the OSV.dev client, which we'll want to move to osv.dev repository at some point.

Part of step 3 of the osv-scanner V2 refactor.

Thanks to VishalGawade1 for the makeRetryRequests testing code in #1390


internal/osvdev/config.go
```diff
@@ -0,0 +1,27 @@
...
+	MaxConcurrentBatchRequests int
+
+	MaxRetryAttempts           int
+	JitterMultiplier           float64
+	BackoffDurationExponential float64
+	BackoffDurationMultiplier  float64
+	UserAgent                  string
+}
+
+// Default make a default client config
```
cuixq, line 16: DefaultConfig

internal/osvdev/config.go
```diff
@@ -0,0 +1,27 @@
...
+func DefaultConfig() ClientConfig {
+	return ClientConfig{
+		MaxRetryAttempts:           4,
+		JitterMultiplier:           2,
+		BackoffDurationExponential: 2,
+		BackoffDurationMultiplier:  1,
+		UserAgent:                  "osv-scanner-v2-" + version.OSVVersion,
```
cuixq, line 23: shall we just use `osv-scanner-v` as prefix here?

cuixq (approved):
Nice!

internal/osvdev/config.go
another-rex: ah yep, meant to change this as this was just for testing, I think in osv.dev we'll update it to osv-go-client or something, and have osv-scanner override it. I'll change this back to our current one for now.

Issue #1447, opened 2024-12-16
Maven transitive scanning is slow due to the new registry support


spencerschrock:
Scorecard's public scanning infrastructure has been having some runtime issues since upgrading to v1.9.1. I'm hoping downgrade in our deployment to confirm the issue goes away, but it lines up with when we merged the dependabot PR.

`pprof` profiling shows a lot of time in `deps.dev/util/semver`, specifically around the Maven dependencies.

> resolve (* APIClient) MatchingVersions 31.57%
> resolve (* APIClient) mavenRequirements 31%
> semver (*Constraint) Match 24.33%

After chatting with the deps.dev team, it's know for `deps.dev/util/semver` to be a slow operation. I don't think its usage is new between v1.9.0 and v1.9.1, but perhaps it's used in a new way?

https://github.com/google/osv-scanner/compare/v1.9.0...v1.9.1

michaelkedar:
Hmm that does line up with a change I made to the Maven resolver https://github.com/google/deps.dev/commit/b3ba03369df10d7148cd6ae39e8147aa21b6454e

Though what benchmarks I did run on it found it should've been faster.

spencerschrock:
I profiled again before the v1.9.0 downgrade and after, and didn't see the maven code as a hotspot in either. So it's possible we were processing a lot of Java repositories at the time. 

However after deploying the new revision to GKE (where the only change was the downgrade), our CPU resources dropped significantly (from ~5.0 CPU to ~1.0 CPU)

cuixq:
@spencerschrock thank you for your updates! I suspect https://github.com/google/osv-scanner/pull/1286 might be the problematic change. I already have a draft [PR](https://github.com/google/osv-scanner/pull/1436) which should reduce the number of Maven repositories we fetch.

spencerschrock:
> Our p99 latencies for the scan also went from occasional spikes of 4000 seconds back to ~50 seconds. I'll see if I can isolate a problematic repo

One example is github.com/SpringCloud/spring-cloud-gray

which took 3-4 minutes to scan on v1.9.0 and on v1.9.1 I gave up after 15 minutes

spencerschrock:
[...]

PR #1466, opened 2024-12-23
fix(maven): do not track registries when fetching requirements

cuixq:
https://github.com/google/osv-scanner/issues/1447

When fetching requirements of a Maven dependency, we should not keep track of the registries defined in pom.xml of dependencies. Doing that will add a lot irrelevant registries to the client and sends unnecessary requests, and finally slows down the resolution.

This PR also disables `PreFetch` in Maven transitive scanning. `PreFetch` calls `MatchVersions` which sends requests to `maven-metadata.xml` for native Maven client. `maven-metadata.xml` is not necessarily needed for all dependencies. Disabling `PreFetch` will save us from making requests to these files. However, calling `PreFetch` should be fine for deps.dev client, and this can be a TODO in the future.

With this fix, resolving [pom.xml](https://github.com/google/osv-scanner/blob/main/internal/remediation/fixtures/zeppelin-server/pom.xml) is now shortened to 10-20 seconds.


internal/scalibrextract/language/java/pomxmlnet/extractor.go
```diff
@@ -132,7 +132,8 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
 	}
 	overrideClient.AddVersion(root, reqs)
 
-	client.PreFetch(ctx, overrideClient, reqs, filepath.Join(input.Root, input.Path))
+	// TODO: only run `PreFetch` for deps.dev client
+	// client.PreFetch(ctx, overrideClient, reqs, filepath.Join(input.Root, input.Path))
```
oliverchang, line 136: what's the impact on performance of this?

oliverchang (approved):

internal/scalibrextract/language/java/pomxmlnet/extractor.go
cuixq: this depends on which registry client we are using, for this [pom.xml](https://github.com/google/osv-scanner/blob/main/internal/remediation/fixtures/zeppelin-server/pom.xml):
 - with deps.dev client: the performance is worse, with runtime from 20s to 1m
 - with native registry: the runtime improved from more than 10m to 10s

for reference, it takes `mvn` about 25s to resolve without any local cache.

michaelkedar (approved):
LGTM - we can revisit whether prefetching some endpoints works for the native maven client in future.

PR #1467, opened 2024-12-24
chore: increase timeout for `golangci-lint`

cuixq:
Recently `golangci-lint` timeouts quite often, as suggested, increase the timeout to 5m.


another-rex (approved):

another-rex:
Not sure if this will fix it though, it doesn't seem to take that long when it is working correctly. (33s for this PR)

PR #1470, opened 2024-12-31
feat(V2 Clients): Client refactor

another-rex:
Second step of the client refactor after #1464 

Changes:
- vulnerability client interface as described in the design doc, this no longer performs paging by itself.
- "Matchers" as implementations to the vulnerability interface
  - osvmatcher: Matcher that takes in a osv.dev client and gets/hydrates the vulns. This also performs paging.
  - localmatcher: Replacement/migrated version of the local package.

Commented out Local client in resolve package to be completed in the followup.

Followup: Decide and implement the VulnClients in resolution.


internal/osvdev/models.go
```diff
@@ -28,12 +29,14 @@ type MinimalVulnerability struct {
 
 // Response represents a full response from OSV.
 type Response struct {
-	Vulns []models.Vulnerability `json:"vulns"`
+	Vulns         []models.Vulnerability `json:"vulns"`
+	NextPageToken string                 `json:"next_page_token"`
```
cuixq, line 33: do we decide to use the snake case? or camel case?

internal/clients/clientimpl/osvmatcher/osvmatcher.go
```diff
@@ -0,0 +1,153 @@
+package osvmatcher
```
cuixq, line 1: can you add some documentation about this new package?

internal/clients/clientimpl/localmatcher/localmatcher.go
```diff
@@ -0,0 +1,162 @@
+package localmatcher
```
cuixq, line 1: can you add some documentation for this package?

cuixq (commented):
Generally LGTM, just one question regarding the file structure: shall we put implementation the same folder as the interface? 
when we have more interfaces and implementation in the future, it will be tricky to track which interface an implementation implements.

internal/osvdev/models.go
another-rex: this matches our API's response, which is in snake_case (since it's in python).

another-rex:
> shall we put implementation the same folder as the interface?

I'm thinking that the interfaces will be moved to osv-scalibr at some point, so we shouldn't couple them in the same package. They are already under the same parent directory of clients,  I think that's probably as close as I can get them?

internal/clients/clientimpl/osvmatcher/osvmatcher.go
another-rex: Done

internal/clients/clientimpl/localmatcher/localmatcher.go
another-rex: Done.

another-rex: Added now to the struct rather than the package.

another-rex:
Paging impl has been extracted and moved to the matcher rather than the client.

internal/clients/clientimpl/osvmatcher/osvmatcher.go
```diff
@@ -0,0 +1,215 @@
...
+			batchQueryCtx, cancelFunc := context.WithDeadline(ctx, time.Now().Add(matcher.InitialQueryTimeout))
+			batchResp, err = queryForBatchWithPaging(batchQueryCtx, &matcher.Client, queries)
+			cancelFunc()
+		} else {
+			batchResp, err = queryForBatchWithPaging(ctx, &matcher.Client, queries)
+		}
+
+		if err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				deadlineExceeded = true
```
oliverchang, line 53: can you add a comment for why we special case DeadlineExceeded and let the logic continue to hydrate?

internal/clients/clientimpl/localmatcher/localmatcher.go
```diff
@@ -0,0 +1,164 @@
...
+		downloadDB: downloadDB,
+		r:          r,
+	}, nil
+}
+
+func (matcher *LocalMatcher) Match(ctx context.Context, invs []*extractor.Inventory) ([][]*models.Vulnerability, error) {
+	results := make([][]*models.Vulnerability, 0, len(invs))
+
+	// slice to track ecosystems that did not have an offline database available
+	var missingDbs []string
```
oliverchang, line 51: nit: missingDBs (capitalisation)

oliverchang (approved):

internal/clients/clientimpl/osvmatcher/osvmatcher.go
another-rex: Added a comment!

PR #1474, opened 2025-01-06
chore: use single quotes by default in Python scripts

G-Rath:
Apparently I've not been as consistent on this as I thought I was, so as punishment I've gone through all the scripts and made sure they're using single quotes by default - aside from still being used within `f` strings due to how that syntax works for string literals within interopolations, I have retained the usage of double quotes in four `print` calls which have single quotes in them to avoid doing an escape


hogo6002 (approved):
Thanks!!

cuixq (approved):

PR #1477, opened 2025-01-07
feat: support extracting package groups from v2+ `poetry.lock` files

G-Rath:
Poetry v2 lockfiles [now include the groups a package belongs to](https://github.com/python-poetry/poetry/pull/9427), meaning that we can support extracting them to allow filtering.

Similar to Maven and NPM, Poetry explicitly notes everything that isn't in an explicit group as being in the `main` group meaning for packages that are considered in that group we always use an empty slice instead since that's how we indicate the "production" group.

I generated the final fixture using the following `pyproject.toml`:

```
[project]
name = "mine"
version = "1.0.0"

requires-python = ">=3.9"

dependencies = [
  "proto-plus == 1.22.0; python_version >= '3.6' and python_version < '3.10'",
  "proto-plus == 1.23.0; python_version >= '3.10'",
  "six"
]

[project.optional-dependencies]
caching = ["redis>=1"]

[tool.poetry.group.dev.dependencies]
urllib3 = "*"
factory-boy = "*"

[tool.poetry.group.test.dependencies]
faker = "*"
```

Technically we don't need the `python_version` constraint on `proto-plus` for this feature, but I realised we don't have any existing fixtures for this situation so I figured I'd sneak it in as part of this given it seems like we already support the lockfile having a package multiple names with different versions

---

This is a sister PR to https://github.com/google/osv-scalibr/pull/376 since `osv-scanner` is still in the middle of transitioning to using `osv-scalibr` - the core logic is exactly the same, and either (or both) should be fine to land


G-Rath:
(fwiw I plan to port this to osv-scalibr tomorrow, so this might end up getting closed but the actual core implementation should still be reviewable)

cuixq (approved):

G-Rath:
Closing as this has been landed in `osv-scalibr`

PR #1480, opened 2025-01-07
feat: Experimental code to start java class reachability.

oliverchang:
This adds the beginnings of a Java reachability analyser that recursively parses class constant pools (https://docs.oracle.com/javase/specs/jvms/se22/html/jvms-4.html#jvms-4.7) for references to other classes. 

The intention is to see if this can be used to exclude transitive dependencies
from vulnerability scanning completely, if they can be proven to be
unreachable.

There's still a lot of TODOs remaining (recorded in the Go source as comments). This tool currently expects
all dependency class files to be available (passed via `--classpath`).

### DIFF
--- a/experimental/javareach/javaclass.go
+++ b/experimental/javareach/javaclass.go
@@ -179,10 +179,14 @@ func ParseClass(r io.Reader) (*ClassFile, error) {
 		switch kind {
 		case ConstantKindUtf8:
 			constant := &ConstantUtf8{}
-			// todo: limit len
 			err := binary.Read(r, binary.BigEndian, &constant.Length)
 			if err != nil {
 				return nil, err
+			}
+
+			const maxConstantLength = 32 * 1024
+			if constant.Length > maxConstantLength {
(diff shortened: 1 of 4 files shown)


experimental/javareach/javaclass.go
```diff
@@ -0,0 +1,440 @@
...
+		"sun/",
+		"org/ietf/",
+		"org/omg/",
+		"org/w3c/",
+		"org/xml/",
+	}
+)
+
+// ClassFile struct represents the overall structure of a Java class file.
+// This only contains the fields we care about for reachability analysis.
```
cuixq, line 40: I guess this corresponds to https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.1?

cuixq (approved):

experimental/javareach/javaclass.go
oliverchang: Yep!

PR #1481, opened 2025-01-08
feat: Update local matcher to the new requirements for resolve

another-rex:
Sets up logging to only return one error log per database that fails to load, and also adds a LoadEcosystem function to preload/preerror databases

imodels.go part can be safely ignored as that is removed in a followup PR.


michaelkedar (approved):
LGTM

PR #1482, opened 2025-01-09
refactor: imodels PackageInfo refactor to use functions

another-rex:
This PR refactors PackageInfo to be accessed via methods, which will perform necessary transformations on the underlying inventory as needed.

This also adds the ecosystemmock extractor which allows you set the return value of the Ecosystem() function.

TODO: Add tests for imodels. I will add tests in a followup PR once we are confident we are sticking with this approach and it works with the other refactors.

### DIFF
--- a/internal/imodels/imodels.go
+++ b/internal/imodels/imodels.go
@@ -90,15 +90,6 @@ func (pkg *PackageInfo) Version() string {
 	return pkg.Inventory.Version
 }
 
-func (pkg *PackageInfo) SetVersion() string {
-	// TODO(v2): SBOM special case, to be removed after PURL to ESI conversion within each extractor is complete
-	if pkg.purlCache != nil {
-		return pkg.purlCache.Version
-	}
-
-	return pkg.Inventory.Version
-}
-
 func (pkg *PackageInfo) Location() string {
 	if len(pkg.Inventory.Locations) > 0 {
 		return pkg.Inventory.Locations[0]
--- a/internal/imodels/imodels.go
+++ b/internal/imodels/imodels.go
@@ -32,123 +32,140 @@ var osExtractors = map[string]struct{}{
 	rpm.Extractor{}.Name():  {},
 }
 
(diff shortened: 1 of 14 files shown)


internal/imodels/imodels.go
```diff
@@ -32,123 +32,140 @@ var osExtractors = map[string]struct{}{
...
-	// For OS packages
-	// This is usually the BinaryName, while Name is the SourceName
-	OSPackageName string
+	if pkg.Ecosystem().Ecosystem == osvschema.EcosystemGo && pkg.Inventory.Name == "go" {
+		return "stdlib"
+	}
 
-	AdditionalLocations []string // Contains Inventory.Locations[1..]
+	if metadata, ok := pkg.Inventory.Metadata.(*dpkg.Metadata); ok {
+		// Debian uses source name on osv.dev
```
oliverchang, line 55: Do we need to add a TODO for any other distros?

internal/imodels/imodels.go
```diff
@@ -32,123 +32,140 @@ var osExtractors = map[string]struct{}{
...
+		return pkg.purlCache.Version
 	}
 
-	// Set DepGroups
-	if dg, ok := inventory.Metadata.(scalibrosv.DepGroups); ok {
-		pkgInfo.DepGroups = dg.DepGroups()
+	return pkg.Inventory.Version
+}
+
+func (pkg *PackageInfo) SetVersion() string {
```
oliverchang, line 93: This looks like it's unnecessary? It's not realy setting any values and hte contents are the same as `Verison()`.

oliverchang (approved):
nice!

internal/imodels/imodels.go
another-rex: Ah must have been a copy and paste error, removed it now.

internal/imodels/imodels.go
another-rex: I don't think so, at least not for any of the distros we currently support (redhat, alpine), SourceName is not in either of those metadatas.

PR #1485, opened 2025-01-10
feat(maven): improve PreFetch for Maven

cuixq:
We disabled `PreFetch` in https://github.com/google/osv-scanner/pull/1466 and this caused the resolution performance worse for deps.dev client.

To enable `PreFetch`, we need to avoid unnecessary requests of `maven-metadata.xml` files which is a result of calling `MatchingVersions` and `Versions`.

This PR improves the logic in `PreFetch` so that `MatchingVersions` is only called for non-soft Maven requirements and `Versions` is called for ecosystems other than Maven.

It used to take about 1m to resove [pom.xml](https://github.com/google/osv-scanner/blob/main/internal/remediation/fixtures/zeppelin-server/pom.xml) without `PreFetch` enabled and now it takes less than 20 seconds.


internal/resolution/client/client.go
```diff
@@ -65,19 +66,37 @@ func PreFetch(ctx context.Context, c DependencyClient, requirements []resolve.Re
...
-		for _, v := range vks {
-			if im.Version == v.Version {
-				vk = v
-				break
+			// We prefer the exact version for soft requirements.
+			for _, v := range vks {
+				if im.Version == v.Version {
+					vk = v
+					break
+				}
```
michaelkedar, line 99: We can get rid of this, since Maven is now being handled above.

michaelkedar (approved):
Nice

PR #1491, opened 2025-01-13
refactor: Licence matcher

another-rex:
Also partially removes the global user agent variable to be specificly passed in for every client.

Majority of the code in licencematcher.go is migrated  from the deleted license.go file.


internal/clients/clientimpl/licensematcher/licensematcher.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+const (
+	maxConcurrentRequests = 1000
+)
+
+// DepsDevLicenseMatcher implements the LicenseMatcher interface with a deps.dev client.
+// It sends out requests for every package version and does not perform caching.
+type DepsDevLicenseMatcher struct {
+	Client *datasource.DepsDevAPIClient
```
cuixq, line 23: I think clients in `internal/resolution/datasource` meant to be used for dependency resolution. considering we are going to use `DepsDevAPICleint` for license matching as well, shall we move `resolution/datasource` to `datasource`?

cuixq (approved):

oliverchang (approved):

internal/clients/clientimpl/licensematcher/licensematcher.go
another-rex: yeah probably a good idea, I'll create an issue for it to keep track, we can do it in a followup PR.

PR #1495, opened 2025-01-14
feat(guided remediation): support authentication for maven registries

michaelkedar:
Support for sending authorization headers with maven registry requests with the native client.
- Parsing username + password for each registry from `${maven.home}/conf/settings.xml` and `${user.home}/.m2/settings.xml`
- Created `HTTPAuthentication` struct to implement and execute auth header logic.
  - Implemented Basic, Bearer, and Digest authentication
- Refactored npm to use `HTTPAuthentication` struct


internal/resolution/datasource/auth.go
```diff
@@ -0,0 +1,285 @@
+package datasource
+
+import (
+	"bytes"
+	"context"
+	"crypto/md5"
```
oliverchang, line 6: This needs a nolint suppression.

internal/resolution/datasource/auth.go
```diff
@@ -0,0 +1,285 @@
...
+
+type AuthenticationMethod int
+
+const (
+	AuthBasic AuthenticationMethod = iota
+	AuthBearer
+	AuthDigest
+)
+
+type AuthenticationInfo struct {
```
oliverchang, line 25: Can you add some struct level code comments to explain this struct ? It seems to be used by both npm and Maven, so it's worth explaining how this all works for multiple ecosystems. Is this just generic http auth support?

internal/resolution/datasource/auth.go
```diff
@@ -0,0 +1,285 @@
...
+	Username    string // Basic & Digest, plain text.
+	Password    string // Basic & Digest, plain text.
+	BearerToken string
+	BasicAuth   string        // Base64-encoded username:password. Overrides Username & Password fields for Basic.
+	CnonceFunc  func() string `json:"-"` // Function used to generate cnonce string for Digest. OK to leave unassigned - mostly for use in tests.
+
+	lastUsed atomic.Value // The last-used authentication method - used when AlwaysAuth is false to automatically send Basic auth.
+}
+
+func (auth *AuthenticationInfo) GetRequest(ctx context.Context, httpClient *http.Client, url string) (*http.Response, error) {
```
oliverchang, line 42: nit: just `Get` to mirror the `net/http` naming conventions? `GetRequest` is also a little confusing because it's not clear if we're meant to be getting a request object, or making a get request.

oliverchang (commented):
nice!

internal/resolution/datasource/auth.go
michaelkedar: fixed this with other lint findings

internal/resolution/datasource/auth.go
michaelkedar: I've added a brief top-level comment - though I'm not really sure how much detail you wanted

internal/resolution/datasource/auth.go
michaelkedar: Renamed to `Get()`, though I find `authInfo.Get()` a bit confusingly named as well.
I thought of `authInfo.DoGet()`, but then if we were to add the `http.Do()` method I wouldn't want `authInfo.DoDo()`...

Maybe the struct should be renamed. Something like `authenticatedHTTP`?

internal/resolution/datasource/maven_registry.go
```diff
@@ -24,8 +24,9 @@ const MavenCentral = "https://repo.maven.apache.org/maven2"
 var errAPIFailed = errors.New("API query failed")
 
 type MavenRegistryAPIClient struct {
-	defaultRegistry MavenRegistry   // The default registry that we are making requests
-	registries      []MavenRegistry // Additional registries specified to fetch projects
+	defaultRegistry MavenRegistry                  // The default registry that we are making requests
+	registries      []MavenRegistry                // Additional registries specified to fetch projects
+	registryAuths   map[string]*AuthenticationInfo // Authentication for the registries, from settings.xml
```
cuixq, line 29: maybe state in the comment that this map is indexed by registry ID

internal/resolution/datasource/maven_settings.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"encoding/xml"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/google/osv-scanner/internal/cachedregexp"
+)
+
```
cuixq, line 11: add a link to the Maven documentation about settings.xml?

cuixq (approved):
LGTM!

oliverchang (approved):

michaelkedar:
FYI I've renamed this struct from `AuthenticationInfo` to `HTTPAuthentication` which I think is better.

oliverchang:
> FYI I've renamed this struct from `AuthenticationInfo` to `HTTPAuthentication` which I think is better.

Makes a lot of sense!

PR #1500, opened 2025-01-15
refactor: Unify where clients and interfaces are initialized, and how extractors are defined

another-rex:
After some discussions, this PR properly groups where clients, interfaces, and extractors so that they are all defined in sensible places that's easy to extend and modify in the future, and improve readability of the code.

The scanning process is now:
- Given CLI flags -> enable the clients + matchers we are able to. <sub>(We assume that initialising a client is not an expensive operation, nor does it make any network requests, so e.g. even if we don't scan a pomxml file, we can still initialise the Maven client)</sub>
- Given list of clients -> We build a list of extractors that can successfully function with the available clients
- Given list of extractors -> Perform the commanded scanning operations with the list of extractors to return an inventory list
- Given list of inventories -> Perform matching and enrichment with the list of available matchers.
- Build result into JSON format
- Return error if vuln found.

This PR also removes deprecated errors, renames errors to the proper format with Err at the front to remove the nolint flags.


pkg/osvscanner/osvscanner.go
```diff
@@ -56,32 +58,102 @@ type TransitiveScanningActions struct {
...
 }
 
-// NoPackagesFoundErr for when no packages are found during a scan.
-//
-//nolint:errname,stylecheck,revive // Would require version major bump to change
-var NoPackagesFoundErr = errors.New("no packages found in scan")
+type ExternalAccessors struct {
+	VulnMatcher            clientinterfaces.VulnerabilityMatcher
+	LicenseMatcher         clientinterfaces.LicenseMatcher
+	MavenRegistryAPIClient *datasource.MavenRegistryAPIClient
```
cuixq, line 64: maybe add a comment saying this is required to fetch pom.xml from upstream Maven repositories.

pkg/osvscanner/osvscanner.go
```diff
@@ -56,32 +58,102 @@ type TransitiveScanningActions struct {
 	MavenRegistry    string
 }
 
-// NoPackagesFoundErr for when no packages are found during a scan.
-//
-//nolint:errname,stylecheck,revive // Would require version major bump to change
-var NoPackagesFoundErr = errors.New("no packages found in scan")
+type ExternalAccessors struct {
+	VulnMatcher            clientinterfaces.VulnerabilityMatcher
+	LicenseMatcher         clientinterfaces.LicenseMatcher
```
cuixq, line 63: maybe add an empty line after LicenseMatcher to separate interfaces and implementations

pkg/osvscanner/osvscanner.go
```diff
@@ -56,32 +58,102 @@ type TransitiveScanningActions struct {
...
+	MavenRegistryAPIClient *datasource.MavenRegistryAPIClient
+	OSVDevClient           *osvdev.OSVClient
+	DependencyClients      map[osvschema.Ecosystem]client.DependencyClient
```
cuixq, line 66: also comment about the potential multiple DependencyClients in the future indexed by ecosystem

cuixq (approved):
🎉

oliverchang (approved):
nice!!

Issue #1492, opened 2025-01-13
Move `resolution/datasource` out to the top level


another-rex:
Since it is no longer just used for resolution.

PR #1501, opened 2025-01-15
refactor: move `datasource` out from `resolution`

cuixq:
https://github.com/google/osv-scanner/issues/1492

Since `datasource` clients are now used not only for resolution, this PR moves them out from `internal/resolution`.
Also to clearly differ from [DepsDevClient](https://github.com/google/osv-scanner/blob/a6d4ec81f975f2aded93d308a7999b2227e069f4/internal/resolution/client/depsdev_client.go#L14) and to better indicate it's a wrapper of `InsightsClient` with caches, this PR renames [DepsDevAPIClient](https://github.com/google/osv-scanner/blob/a6d4ec81f975f2aded93d308a7999b2227e069f4/internal/resolution/datasource/depsdev_api.go#L16) to `CachedInsightsClient`.


experimental/javareach/cmd/reachable/main.go
```diff
@@ -121,7 +121,7 @@ func enumerateReachabilityForJar(jarPath string) error {
 		}
 	}
 
-	for dep, _ := range reachableDeps {
+	for dep := range reachableDeps {
```
cuixq, line 124: This is a result of running `gofmt` - @oliverchang should this be `for _, dep := range reachableDeps`?

oliverchang: This change is OK -- the second "_" wasn't necessary.

another-rex (approved):

PR #1503, opened 2025-01-16
refactor: use new `VulnerabilityMatcher` in guided remediation

michaelkedar:
Following up on #1470 
- Made `ResolutionClient` use the `VulnerabilityMatcher` interface (and added helper function to convert deps.dev graphs into inventories)
- Deleted old `VulnerabilityClient`
- Created `CachedOSVMatcher` to re-implement performance improvements from the original `VulnerabilityClient` w.r.t. repeated queries.
- Re-enabled local database capability in `osv-scanner fix`


internal/remediation/in_place.go
```diff
@@ -249,7 +250,10 @@ func inPlaceVulnsNodes(cl client.VulnerabilityClient, graph *resolve.Graph) (inP
 
 	// Find all direct dependencies of vulnerable nodes.
 	for _, e := range graph.Edges {
-		if len(nodeVulns[e.From]) > 0 {
+		if e.From == 0 {
+			continue
+		}
+		if len(nodeVulns[e.From-1]) > 0 {
```
cuixq, line 256: is this `e.From-1` (as well as `nID+1` `nID+1`) related to converting graph to inventories? I am a bit confused about them.

internal/resolution/client/helper.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"deps.dev/util/resolve"
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/purl"
+)
+
+// GraphAsInventory is a helper function to convert a Graph into an Inventory for use with VulnerabilityMatcher.
+func GraphAsInventory(g *resolve.Graph) []*extractor.Inventory {
```
cuixq, line 11: `GraphToInventory`?

internal/remediation/in_place.go
michaelkedar: Previously, the returned `nodeVulns` was padded with an empty list at index 0 to correspond to the root of the graph, so `len(nodeVulns) == len(graph.Nodes)` and `nodeVulns[i]` is the vulns affecting `graph.Nodes[i]`.

An inventory shouldn't contain the root node, so it's now not included in `GraphAsInventory`. So now, `len(nodeVulns) == len(graph.Nodes)-1` and `nodeVulns[i]` is the vulns affecting `graph.Nodes[i+1]`. All these +-1's are accounting for the difference.

I might add the pad back in - the adjustments are confusing and error prone.

cuixq: I think it's a good idea to align the indices between `nodeVulns` and `graph.Nodes` (if no strong reason not to do this), which benefits readability and easier to maintain.

michaelkedar: I've added the padding element back in and undone the index changes.

cuixq (approved):

oliverchang (approved):

another-rex (approved):

PR #1505, opened 2025-01-17
feat: Use osv-scalibr to perform container scanning

another-rex:
This PR fully removes the image package in osv-scanner, and switches to use osv-scalibr to perform image scanning.

Also includes:
- Implementing base image identification, along with the baseimagematcher interface and implementation
- Some temporary placeholder values in output_result.go to allow it to compile. This is resolved in #1499 
- Adds image metadata types to show image metadata in the output.
- Minor refactors in osvscanner.go
  - determineReturnErr is an extracted function for the very end of DoScan
  - Move container scanning support out of DoScan, and into it's own DoContainerScan function
  - Move the helper methods that help pull docker images into it's own package
- Support checking if docker image exists locally.
- Support local paths with the docker flag.


internal/output/output_result.go
```diff
@@ -289,10 +289,10 @@ func processPackage(vulnPkg models.PackageVulns) PackageResult {
 
 	if vulnPkg.Package.ImageOrigin != nil {
 		packageLayerDetail := PackageLayerDetail{
-			LayerID:     vulnPkg.Package.ImageOrigin.LayerID,
-			InBaseImage: vulnPkg.Package.ImageOrigin.InBaseImage,
+			LayerID:     "",
```
another-rex, line 292: This is just placeholder code to keep things compiling, resolved in #1499

internal/clients/clientimpl/baseimagematcher/baseimagematcher.go
```diff
@@ -0,0 +1,160 @@
...
+			runningDigest = digest.FromBytes([]byte(runningDigest + " " + l.DiffID))
+		}
+
+		chainID := runningDigest
+		g.Go(func() error {
+			if ctx.Err() != nil {
+				return ctx.Err() // this value doesn't matter to errgroup.Wait(), it will be ctx.Err()
+			}
+
+			req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.deps.dev/v3alpha/querycontainerimages/"+chainID.String(), nil)
```
oliverchang, line 50: nit: should we put the URL in a const somewhere?

pkg/osvscanner/internal/scanners/artifacts.go
```diff
@@ -0,0 +1 @@
+package scanners
```
oliverchang, line 1: what is this file for?

internal/clients/clientimpl/baseimagematcher/baseimagematcher.go
```diff
@@ -0,0 +1,160 @@
...
+		//   The goal here is to replace empty layers that is currently categorized as the previous base image
+		//   with this base image if it actually belongs to this layer.
+		//
+		//   We do this by guessing the boundary of empty layers by checking for the following commands,
+		//   which are commonly the *last* layer.
+		//
+		//   Remember we are looping backwards in the outer loop,
+		//   so this backfill is actually filling down the layer stack, not up.
+		possibleFinalBaseImageCommands := []string{
+			"/bin/sh -c #(nop)  CMD",
```
oliverchang, line 135: How reliable is it that we're relying on there always being two? spaces before the `CMD` ?

pkg/osvscanner/internal/imagehelpers/imagehelpers.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Check if image exists locally, if not, pull from the cloud.
+	r.Infof("Checking if docker image (%q) exists locally...\n", dockerImageName)
+	cmd := exec.Command("docker", "images", "-q", dockerImageName)
+	output, err := cmd.Output()
+	if err != nil || string(output) == "" {
+		r.Infof("Image not found locally, pulling docker image (%q)...\n", dockerImageName)
+		err = runCommandLogError(r, "docker", "pull", "-q", dockerImageName)
+		if err != nil {
+			return "", fmt.Errorf("failed to pull container image: %w", err)
```
oliverchang, line 91: should we be deleting the temp dir in these error paths? the caller has no way to do it either since we're returning empty string in these cases.

oliverchang (approved):
nice!!

internal/clients/clientimpl/baseimagematcher/baseimagematcher.go
```diff
@@ -0,0 +1,160 @@
...
+				matcher.Reporter.Errorf("deps.dev API error: %s\n", err)
+				return nil
+			}
+			defer resp.Body.Close()
+
+			if resp.StatusCode == http.StatusNotFound {
+				return nil
+			}
+
+			if resp.StatusCode != http.StatusOK {
```
hogo6002, line 67: nit: this function is quite long. Could we separate all the requests into a separate function to improve readability?

internal/clients/clientimpl/baseimagematcher/baseimagematcher.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		chainID := runningDigest
+		g.Go(func() error {
+			if ctx.Err() != nil {
+				return ctx.Err() // this value doesn't matter to errgroup.Wait(), it will be ctx.Err()
+			}
+
+			req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.deps.dev/v3alpha/querycontainerimages/"+chainID.String(), nil)
+			if err != nil {
```
hogo6002, line 51: should we add some fault tolerance here (e.g. retries) in case of temporary service issues?

pkg/osvscanner/osvscanner.go
```diff
@@ -238,11 +255,118 @@ func DoScan(actions ScannerActions, r reporter.Reporter) (models.VulnerabilityRe
...
+	if accessors.LicenseMatcher != nil {
+		err = accessors.LicenseMatcher.MatchLicenses(context.Background(), scanResult.PackageScanResults)
+		if err != nil {
+			return models.VulnerabilityResults{}, err
+		}
+	}
+
+	results := buildVulnerabilityResults(r, actions, &scanResult)
+
+	filtered := filterResults(r, &results, &scanResult.ConfigManager, actions.ShowAllPackages)
```
hogo6002, line 354: Do we want to filter out vulnerabilities from `usr/` directories here?

hogo6002 (approved):

internal/clients/clientimpl/baseimagematcher/baseimagematcher.go
another-rex: I'm not 100% sure, as this is from the scalibr code. I believe it is relatively reliable, since it's added by build tools, and these two versions of CMD is supporting the current output of `docker build` and `docker buildx build` (aliased to `docker build` after a recent docker version) .

Perhaps this might change in a future release, but at that point we have no idea what the CMD could look like, so I think leaving it as this is fine for now.

oliverchang: SGTM

pkg/osvscanner/internal/imagehelpers/imagehelpers.go
another-rex: Ah good idea, done.

pkg/osvscanner/internal/scanners/artifacts.go
another-rex: Accidentally left in, deleted now.

internal/clients/clientimpl/baseimagematcher/baseimagematcher.go
another-rex: Added the retry code from osvdev client here as well, there's quite a bit of copy and pasting, but I think that's for the best for now, since there are slight differences, and we are planning to move the osvdev code out of osvscanner, so I'm not sure how we can share code.

another-rex: Also correctly set the user agent here.

pkg/osvscanner/osvscanner.go
another-rex: Added annotations to the under `usr/` directory.

internal/clients/clientimpl/baseimagematcher/baseimagematcher.go
another-rex: Done

hogo6002 (approved):
Thanks!

PR #1507, opened 2025-01-17
chore: remove snapshots from datasource tests

michaelkedar:
Redid tests that used snapshots to mostly use `cmp.Diff` instead, for easier migration into osv-scalibr


cuixq (approved):
Thanks!

PR #1508, opened 2025-01-17
fix: Correctly set ARG and revert docker hash back to original version

another-rex:
Turns out, ARG only applies to the current FROM section, so all of the ARGs in our test docker files were placebo and didn't do anything. The variable was just empty and just installed whatever version npm felt like haha, this meant recently the image building broke as npm started trying to install a newer version of itself that's not compatible with the version of node this image is pinned to. E.g. https://github.com/google/osv-scanner/actions/runs/12821471894/job/35752742949?pr=1505

Also revert the change where I updated the sha, now keeps using the existing older sha for more results.


oliverchang (approved):

G-Rath (approved):

G-Rath:
Well, whoops

PR #1518, opened 2025-01-21
feat: add more lockfile extractors

cuixq:
This PR adds more lockfile extractors supported in OSV-Scalibr:
 - Python `uv.lock`
 - .NET `deps.json`
 - Haskell `cabal.project.freeze`
 - Haskell `stack.yaml.lock`

Also adds a TODO to support Swift `Pacakge.resolved` - we need to figure out a workaround to map the extracted packages to SwiftURL in OSV.dev.


G-Rath:
fwiw if you wait a little longer we'll hopefully [be able to also add `bun.lock`](https://github.com/google/osv-scalibr/pull/379)

oliverchang:
Do we want to add any tests here? @another-rex 

Or can we assume that the extractors are already well tested in OSV-Scalibr and the integration here doesn't introduce any new points of failure?

another-rex:
We should copy some tests fixtures from scalibr into our main_test fixtures. We might not even need to add any new tests there, just update the snapshots after moving some new fixtures in.

cuixq:
@G-Rath yes I do plan to wait for `bun.lock` - sending out for some high-level review first.

another-rex (commented):
Can you add a comment around why Package.resolved is commented out?

cmd/osv-scanner/main_test.go
another-rex, line 1: Is it possible to update the lockfiles to pick a vulnerable version? I really want to test that the query is actually matching against packages on osv.dev's end and we haven't messed up the ecosystem for example.

cmd/osv-scanner/fixtures/locks-scalibr/stack.yaml.lock
```diff
@@ -0,0 +1,47 @@
...
+  original:
+    hackage: jose-jwt-0.10.0
+- completed:
+    hackage: postgresql-libpq-0.10.1.0@sha256:6b580c9d5068e78eecc13e655b2885c8e79cdacfca513c5d1e5a6b9dc61d9758,3166
+    pantry-tree:
+      sha256: ae81e7628a8f3d1ef33ace71fa0845c073c003ca7f1150cc9d9ba1e55fc84236
+      size: 1096
+  original:
+    hackage: postgresql-libpq-0.10.1.0
+- completed:
```
cuixq, line 35: the hashes of this package is not auto-generated so might not be accurate, we may need to update it.

cmd/osv-scanner/main_test.go
cuixq: I updated the lockfiles with one vulnerable Haskell package and one vulnerable .NET package.

another-rex (approved):
LGTM, thanks!

PR #1523, opened 2025-01-23
fix: Ignore linux source packages

another-rex:
Linux source packages are likely to be kernel packages, which generally do not apply in a container context,


hogo6002 (approved):
Thanks Rex!

oliverchang (approved):

PR #1531, opened 2025-01-24
feat: Add artifact extractors

another-rex:
- Adds the python wheel.egg and java jar archive extractors.

- renames artifacts source from lockfile to artifact, so output will actually be "artifact:<path to binary>".

- Reenable some of the image scanning tests accidentally removed from a previous PR. The unrelated snapshot changes will be from that.

This PR is not ready to be merged in yet, waiting on a osv-scalibr change to be merged in first. (https://github.com/google/osv-scalibr/pull/407)


internal/imodels/imodels.go
```diff
@@ -47,10 +60,28 @@ func (pkg *PackageInfo) Name() string {
...
+	// --- Make specific patches to names as necessary ---
+	// Patch Go package to stdlib
 	if pkg.Ecosystem().Ecosystem == osvschema.EcosystemGo && pkg.Inventory.Name == "go" {
 		return "stdlib"
 	}
 
+	// Patch python package names to be normalized
+	if pkg.Ecosystem().Ecosystem == osvschema.EcosystemPyPI {
+		// per https://peps.python.org/pep-0503/#normalized-names
+		return strings.ToLower(cachedregexp.MustCompile(`[-_.]+`).ReplaceAllLiteralString(pkg.Inventory.Name, "-"))
```
hogo6002, line 72: Should we add a TODO to remove this after the API query change is deployed to production?

pkg/osvscanner/osvscanner.go
```diff
@@ -360,10 +368,13 @@ func DoContainerScan(actions ScannerActions, r reporter.Reporter) (models.Vulner
...
 	}
 
-	// TODO: This is a heuristic, assume that packages under usr/ is an OS package
+	// TODO: This is a set of heuristics,
+	//    - Assume that packages under usr/ might be a OS package depending on ecosystem
+	//    - Assume python packages under dist-packages is a OS package
 	// Replace this with an actual implementation in OSV-Scalibr (potentially via full filesystem accountability).
 	for _, psr := range scanResult.PackageScanResults {
-		if strings.HasPrefix(psr.PackageInfo.Location(), "usr/") {
+		if (strings.HasPrefix(psr.PackageInfo.Location(), "usr/") && psr.PackageInfo.Ecosystem().Ecosystem == osvschema.EcosystemGo) ||
```
hogo6002, line 376: nit: This logic for annotation seems quite complex now. Can we add some examples in the comments to clarify which cases will be annotated?

hogo6002 (approved):
Thanks!

internal/imodels/imodels.go
another-rex: Yep good catch, I'll probably remove this entirely since I don't plan on merging this PR in until next week.

oliverchang (approved):

internal/imodels/imodels.go
another-rex: Hmm it looks like there's more work to be done here: 
1. Offline scanning also needs to this normalization
2. The code for identifying fix available/not available also needs this change.
Added a TODO.

pkg/osvscanner/osvscanner.go
another-rex: Added some comments.

PR #1535, opened 2025-01-28
docs: Image scanning docs and various refactors to docs

another-rex:
Update the docs to include information for image scanning + some restructuring of the docs. 

Specifically refactors usage into two sub pages and 1 parent page to match how the `scan` command works.

See https://another-rex.github.io/osv-scanner/usage/scan-source for a preview.

Need followup work to make this fully work with V2, including:
- [x] Mention usage of OSV-Scalibr
- [x] Add newly supported lockfiles
- More detail on the HTML output format

Also finally fixed the Note being really hard to read:
image


docs/scan-image.md
```diff
@@ -0,0 +1,138 @@
...
+<details open markdown="block">
+  <summary>
+    Table of contents
+  </summary>
+  {: .text-delta }
+- TOC
+{:toc}
+</details>
+
+OSV-Scanner analyzes container images by extracting package information and matching it against known vulnerability databases. This helps identify potential security risks in your containerized applications.
```
oliverchang, line 22: to be more explicit: "matching it against OSV.dev" ?

docs/scan-image.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Prerequisites
+
+- **Docker (Optional)**: If you want to scan images directly by name (e.g., my-image:latest) without exporting them first, the docker command-line tool must be installed and available in your system's PATH. If you choose to scan exported image archives, Docker is not required.
+
+All image scanning is done with the `scan image` subcommand:
+
+```bash
+osv-scanner scan image
```
oliverchang, line 31: nit: append `<image-name>` ?

docs/scan-image.md
```diff
@@ -0,0 +1,138 @@
...
+│ apt         │ Debian:12 │ 2.6.1                 │                   1 │ Unimportant    │
+│ binutils    │ Debian:12 │ 2.40-2                │                   8 │ Unimportant    │
+...
+│ util-linux  │ Debian:12 │ 2.38.1-5+deb12u2      │                   1 │ Unimportant    │
+╰─────────────┴───────────┴───────────────────────┴─────────────────────┴────────────────╯
+```
+
+</details>
+
+### Detailed Output:
```
oliverchang, line 123: nit: this should go first, before the table output, since I think this is a pretty nice game changer.

docs/scan-image.md
```diff
@@ -0,0 +1,138 @@
...
+
+For a more detailed view of vulnerabilities, including individual vulnerability details, use the HTML output format. You can enable it using:
```
oliverchang, line 125: nit: mention interactive layer exploration and other interactive features?

docs/supported_languages_and_lockfiles.md
```diff
@@ -18,9 +18,29 @@ nav_order: 2
 {:toc}
 </details>
 
-## Supported lockfiles
+### Core Concept
```
oliverchang, line 21: can we call out to osv-scalibr here? as the underlying library for these?

docs/scan-source.md
```diff
@@ -0,0 +1,134 @@
...
+permalink: /usage/scan-source
+nav_order: 2
+---
+
+# Source Scanning
+
+OSV-Scanner can be used to scan your project source and lockfiles to find vulnerabilities in your dependencies.
+
+```bash
+osv-scanner source scan <flags> [paths...]
```
oliverchang, line 14: this should be `scan source` right? there's a few references in this page that needs to be fixed.

docs/scan-image.md
another-rex: Because everything in the details is collapsed, it's actually just one line down. I'll make the screenshot no longer collapsed in <detail> to draw peoples eyes a bit more.

docs/scan-source.md
another-rex: Ah yes of course, fixed now.

docs/scan-image.md
```diff
@@ -0,0 +1,135 @@
...
+     podman save --format=docker-archive my-image:latest > my-image.tar
+
+     # Other image tools: Use the docker archive format to export the tar
+     ```
+
+### Usage Notes
+
+- **No other scan targets:** When using `scan image`, you cannot specify other scan targets (e.g., directories or lockfiles).
+
+- **Configuration Flags:** All the global configuration flags available for the `scan` command (as described in the [Usage documentation](https://www.google.com/url?sa=E&source=gmail&q=./usage.md)) can be used with the `scan image` subcommand. This includes flags for output format, verbosity, config files, and experimental features.
```
cuixq, line 68: is this URL correct?

docs/scan-image.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Detailed Output:
+
+For a more detailed view of vulnerabilities, including individual vulnerability details, use the HTML output format. You can enable it using:
+
+- `--format=html`: This will output the results to an HTML file.
+
+- `--serve`: This will generate an HTML report and host it locally on `localhost:8000`.
+
+See the [Output documentation](https://www.google.com/url?sa=E&source=gmail&q=./output.md) for more information on output formats.
```
cuixq, line 131: also this URL here

docs/scan-source.md
```diff
@@ -0,0 +1,134 @@
...
+
+# Source Scanning
+
+OSV-Scanner can be used to scan your project source and lockfiles to find vulnerabilities in your dependencies.
+
+```bash
+osv-scanner scan source <flags> [paths...]
+```
+
+As this is the most common usecase of osv-scanner, `source scan` is the default subcommand of osv-scanner, so the above is equivalent to:
```
cuixq, line 17: `scan source`?

docs/scan-image.md
cuixq: +1

docs/scan-source.md
```diff
@@ -0,0 +1,134 @@
...
+
+To enable call analysis in all languages, call OSV-Scanner with the `--call-analysis=all` flag. By default, call analysis in Go is enabled, but you can disable it using the `--no-call-analysis=go` flag.
+
+### Call analysis in Go
+
+OSV-Scanner uses the [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) library to analyze Go source code to identify called vulnerable functions.
+
+#### Additional Dependencies
+
+`go` compiler needs to be installed and available on `PATH`
```
cuixq, line 104: `.` at the end

docs/supported_languages_and_lockfiles.md
```diff
@@ -18,9 +18,29 @@ nav_order: 2
...
 
-A wide range of lockfiles are supported by utilizing this [lockfile package](https://github.com/google/osv-scanner/tree/main/pkg/lockfile).
+We found that when performing different forms of scanning, you are generally interested in different types of files. For example, when scanning your source project, you are much more interested in what your lockfiles and manifests contain, and less interested in what is installed on your development machine, or leftover compiled artifacts. However, if you are scanning a container, then what is installed is the vital piece of information, and lockfiles found on the system no longer matters if the artifacts they point to are not actually downloaded and installed.
+
+We make this distinction by splitting what we scan into two broad categories, **manifests** and **artifacts**.
```
cuixq, line 25: maybe move this sentence to the beginning of the paragraph? also I feel the first paragraph may not be necessary from a user perspective.

cuixq: also the order of `manifests` and `artifacts` should be the same as the following paragraphs.

cuixq (commented):
There are a bunch of URLs in weird formats - not sure if this is intentional?

docs/scan-image.md
another-rex: Fixed all the url issues I believe.

docs/supported_languages_and_lockfiles.md
another-rex: Updated this, PTAL

docs/scan-image.md
oliverchang: Ah, not sure I fully understood your reponse, but I was suggesting we promote the HTML output first (since it's the nicest) in the documentation, then the CLI output. WDYT?

another-rex: Hmm... I think it's better to have the default output first, and we can still have that while promoting the HTML output since it's so close and takes up much more space on the docs page. See https://another-rex.github.io/osv-scanner/usage/scan-image#output, you can easily see the HTML output as the primary thing we want people to use, while we still have the first line explaining the default behaviour.

oliverchang: SGTM. make sense now that I look at the final rendered output!

oliverchang (approved):

docs/supported_languages_and_lockfiles.md
another-rex: Made a call out to osv-scalibr here, though it is a bit janky. Perhaps we should have the osv-scanner and osv-scalibr integration better explained in another page.

cuixq (approved):
LGTM - but I still see two weird URLs.

another-rex:
You're right, I swear I found all of them, but nope turns out I didn't even fix the two you originally comment on 🤦

PR #1538, opened 2025-01-28
fix(guided remediation): reduce memory footprint by computing dependency subgraphs instead of chains

michaelkedar:
Guided remediation had been using `DependencyChains` to track paths to a vulnerable package (for computing things like depth and which direct dependencies to relax). It was computing *every* possible path in the graph to a dependency, which grows roughly exponentially with depth / connectivity. This was using an unreasonable amount of memory on some particularly large/complex projects.

I've changed the logic to instead compute one `DependencySubgraph` - the set of nodes and edges that would contain every path to a dependency. This should significantly reduce memory usage (and cpu usage from allocs) when running on larger projects.

This change has touched quite a few places in the code, and the logic is a bit complex. I've tried my best to check that everything still behaves as expected.


cuixq:
Any statistics about how much memory usage improves?

internal/resolution/dependency_subgraph.go
```diff
@@ -0,0 +1,232 @@
...
+				return false
+			}
+		}
+	}
+
+	return true
+}
+
+// ConstrainingSubgraph finds the subgraph of the subgraph with only edges that constrains the end dependency to a vulnerable version.
+//
```
cuixq, line 131: I think we can make the comment better explain what this function does (as what we did offline).

michaelkedar: I've expanded on the comment and added a brief example

michaelkedar:
> Any statistics about how much memory usage improves?

For a particularly gnarly `package-lock.json` (with almost 6000 packages), the old `ComputeChains` was allocating 130 GB (!!) vs new `ComputeSubgraphs` allocating 4.5 MB - around 1/30000th. The total allocations is 137 GB vs 5GB, mostly from writing the lockfile in-place.
(not sure about how much was in use at any given time, just total allocations)

cuixq (approved):
LGTM

internal/resolution/dependency_subgraph.go
```diff
@@ -0,0 +1,244 @@
...
+//	C resolves to C@1.9
+//
+// If the vuln affecting C is fixed in version 2.0, the constraining subgraph would only contain A,
+// since B would allow versions >=2.0 of C to be selected if not for A.
+//
+// This is a heuristic approach and may produce false positives.
+// If the constraining subgraph cannot be computed for some reason, returns the original DependencySubgraph.
+func (ds *DependencySubgraph) ConstrainingSubgraph(ctx context.Context, cl resolve.Client, vuln *models.Vulnerability) *DependencySubgraph {
+	// Just check if the direct requirement of the vulnerable package is constraining it.
+	// This still has some false positives.
```
oliverchang, line 147: What's the impact on our results of getting this wrong?

internal/resolution/dependency_subgraph.go
```diff
@@ -0,0 +1,244 @@
...
+		})
+	}
+
+	// groups == nil
+	// Check if any of the direct dependencies do not have the Dev attr.
+	for _, e := range ds.Nodes[0].Children {
+		if e.Type.HasAttr(dep.Dev) {
+			continue
+		}
+		if e.To == ds.Dependency {
```
oliverchang, line 116: can you explain what this case is in a comment also? why does this imply there is a non dev usage?

internal/resolution/dependency_subgraph_test.go
```diff
@@ -0,0 +1,335 @@
+package resolution_test
```
oliverchang, line 1: thanks for adding tests (this seems extremely tricky!)

oliverchang (approved):
awesome optimization!

internal/resolution/dependency_subgraph.go
michaelkedar: Added a comment and put this check after the workspace Children check to make more clear it's related.

internal/resolution/dependency_subgraph.go
michaelkedar: It would usually mean some direct dependencies that did not actually need to be relaxed end up being relaxed (or if they can't be relaxed further, the vuln deemed 'unfixable'). I've briefly noted this in the outer comment.

FWIW, this isn't new behaviour - this comment & logic is from the removed `DependencyChain` code. I also think this situation is unlikely to occur with npm.

PR #1541, opened 2025-01-28
docs: Update output and add HTML output as an option in the output.md doc.

another-rex:
Several small changes:
- Update the output doc to include HTML output
- Update Feedback link to point to the correct feedback discussion
- Clarify the error logging message to warn people to update osv-scanner if base image scanning fails.
- Actually fail base image scanning to avoid spamming people with the log.

### DIFF
--- a/docs/output.md
+++ b/docs/output.md
@@ -31,7 +31,7 @@ For every vulnerability found, OSV-Scanner will display the following informatio
 
 And if you are performing layer scanning, osv-scanner additionally returns:
 
-- Layers each package belongs to
+- Layer where a package was first introduced
 - Layer history and commands
 - Base images the image is based on
 - OS/Distro the container is running on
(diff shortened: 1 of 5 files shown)


docs/output.md
```diff
@@ -29,6 +29,13 @@ For every vulnerability found, OSV-Scanner will display the following informatio
 - Version: Package version
 - Source: Path to the sbom or lockfile where the package originated
 
+And if you are performing layer scanning, osv-scanner additionally returns:
+
+- Layers each package belongs to
```
oliverchang, line 34: "Layers where a package was first introduced" to be clearer?

oliverchang (approved):

docs/output.md
```diff
@@ -86,16 +93,49 @@ osv-scanner --format markdown your/project/dir
...
+- Full vulnerability advisory entries
+
+And additionally for container image scanning:
+
+- Layer filtering
+- Image layer information
+- Base image identification
+
+{: .note }
+This feature is in beta as part of OSV-Scanner v2, please share your feedback here: https://github.com/google/osv-scanner/discussions/1529
```
cuixq, line 118: maybe inline the link?

cuixq (approved):

docs/output.md
another-rex: Good idea, done.

PR #1546, opened 2025-01-29
fix: change module name to v2

michaelkedar:
Go requires major version 2 of a package to be under a `/v2` module.

Per https://go.dev/doc/modules/major-version, must change the go.mod file and all references to `github.com/google/osv-scanner` to `github.com/google/osv-scanner/v2` for our v2 release.


michaelkedar:
I've left [`github.com/google/osv-scanner/experimental/javareach`](https://github.com/google/osv-scanner/blob/main/experimental/javareach/go.mod) alone, since it's its own go module. Not sure if I should rename it to v2 anyway?

oliverchang:
> I've left [`github.com/google/osv-scanner/experimental/javareach`](https://github.com/google/osv-scanner/blob/main/experimental/javareach/go.mod) alone, since it's its own go module. Not sure if I should rename it to v2 anyway?

this is a temporary thing anyway. so totally fine to leave it alone.

oliverchang (approved):
thanks for fixing this!

another-rex (approved):

jess-lowe (approved):

PR #1547, opened 2025-01-30
fix(cmd): add experimental config support to image command

hogo6002:


another-rex (commented):
Can the code here be shared with `scan source`? I think most of this overlaps

cmd/osv-scanner/internal/helper/helper.go
```diff
@@ -55,6 +61,11 @@ var GlobalScanFlags = []cli.Flag{
 		Usage: "specify the level of information that should be provided during runtime; value can be: " + strings.Join(reporter.VerbosityLevels(), ", "),
 		Value: "info",
 	},
+	&cli.BoolFlag{
```
another-rex, line 64: I don't think we have git scanning in container scanning at the moment.

another-rex (commented):
LGTM, 1 small comment

cmd/osv-scanner/internal/helper/helper.go
hogo6002: Yes, it's not needed for container scanning, but the `experimental-offline` feature will [attempt to set this flag to false](https://github.com/google/osv-scanner/blob/8aa66748f4968e0679a914b54657be396dd10693/cmd/osv-scanner/internal/helper/helper.go#L66), which will cause errors if the flag doesn't exist. I considered skipping the false setting if the context lacks this flag, but this might alter the logic slightly. So I decided to leave this section unchanged and added the git flag globally.

another-rex: Ah ok, I think we need to add the check/skip, because this will show up on the --help, and if it doesn't do anything it'll be confusing for the user I think.

another-rex (approved):

PR #1548, opened 2025-01-30
docs: Update docs to point to latest

another-rex:
And fix an incorrect flag.

We can safely use @latest because v2 module only has prereleases at the moment, so latest will pick the latest prerelease.


hogo6002 (approved):

cuixq (approved):

jess-lowe (approved):
I think we should update the code snippets in the README to use code blocks instead but I can do that in a separate PR. Otherwise LGTM.

michaelkedar (approved):

PR #1552, opened 2025-01-31
docs: Polish up documentation CSS and font

another-rex:
- Update font to be more readable, Overpass is much harder to read for dense text, and bold text fades into the background. Use more standard text font of Source Sans Pro (3)
- Fix code blocks being very hard to read
- Fix Tables to not have gaps

New version of docs: https://another-rex.github.io/osv-scanner/supported-languages-and-lockfiles/
Old version of docs: https://google.github.io/osv-scanner/supported-languages-and-lockfiles/

### DIFF
--- a/docs/_sass/color_schemes/_variables.scss
+++ b/docs/_sass/color_schemes/_variables.scss
@@ -31,7 +31,9 @@ $red-200: #e94c4c !default;
 $red-300: #dd2e2e !default;
 
 $body-font-family:
-  "Overpass",
+  "Source Sans 3",
   system-ui,
   -apple-system,
   sans-serif;
+
+$mono-font-family: consolas, monospace;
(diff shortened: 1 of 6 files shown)


hogo6002 (approved):

cuixq:
the new heading looks not bold enough to me - can we improve this?

another-rex:
Hmm... I think it looks alright? If you look at just the docs https://just-the-docs.com/docs/ui-components/typography/ it's about the same, I can bump up the font weight a bit, but it's supposed to be thin.

jess-lowe:
I agree with Xueqin - it feels a little unbalanced. I think having the heading 1 font weight matching the heading 2 font weight feels better (so both at 500). That or bring back Overpass just for the headings -  I liked that font lol.

another-rex:
Reverted the heading fonts to overpass now :)

jess-lowe (approved):
Looks great!

cuixq (approved):

Issue #1528, opened 2025-01-24
Remaining Container Scanning Issues for v2 Official Release


hogo6002:
This issue tracker used to track container scanning issues specifically for the official v2 release.

Functionality:
- [x] https://github.com/google/osv.dev/issues/3082
- [x] #1571
- [x] #1572
- [x] bug: failed to handle tar entry with path /bin/envoy: file exceeds read limit
- [x] #1573

Testing:
- [x] https://github.com/google/osv-scanner/issues/1525

Performance:
- [x] #1574

Output:
- [x] #1575
- [x] #1576
- [x] Cleanup excessive scalibr logs
- [x] ~~--serve to automatically find a unused port around 8000 rather than immediately failing.~~ (don't need it, as we plan to allow users to specify the port)
- [x] #1577
- [x] Lift tooltip z-index so it does not get hidden by the container that it is in.
- [x] https://github.com/google/osv-scanner/issues/1555
- [x] https://github.com/google/osv-scanner/issues/1554
- [x] #1617

Any issues mentioned in discussion https://github.com/google/osv-scanner/discussions/1521

another-rex:
And issues listed here: https://github.com/google/osv-scanner/issues/1274#issuecomment-2620669917

PR #1561, opened 2025-02-03
fix: HTML report tinkering

another-rex:
Made a series of changes to resolve the issues identified in #1528

Hosted an example here: https://another-rex.github.io/TestPages/Vulnerability%20Scan%20Report.html

To make it easier to see the changes, when reviewing, use this link:
https://github.com/google/osv-scanner/pull/1561/files/bd8d5211e77612b1a0a68a4b00db1d40535fe400..8223593ef84cda34da84a57aecd12d3258ea1463
which select the diffs **Excluding** the first commit (use shift to select multiple commits). That moves the files around which breaks all of git's diffing. No change other than moving the files and reindenting is done in that first commit.

HTML:
- Move to actual .js and .css file rather than .html files.
- Alias and groupid tooltips now put each ID on a new line.
- Can now click on the entire filter box to change it, not just on the text part.

CSS: 
- Remove max-height in the inner tables, this was making it impossible to have tooltips that escape the container (at least I haven't figured out how to have both).
- Tooltip box sizing is now dynamic with max-width
- Tooltips now display upwards instead of downwards
-  Highlight source path better
- Minor refactor to how the search box is laid out
- Remove unused css lines.
- Make iframe bg color black instead of white to avoid flash banging people.

JS:

- Remove all style edits in javascript, state changes are made with classes now. (TIL `classList.toggle()` function)
- Basic linter pass (e.g. use const on variables, define all variables...etc)
- Run showAllVulns() function at page load.


hogo6002:
Thanks! can you host one sample HTML on Github

another-rex:
Hosted an example here: https://another-rex.github.io/TestPages/Vulnerability%20Scan%20Report.html

internal/output/html/style.css
```diff
@@ -495,26 +503,30 @@ header {
...
-  border-radius: 4px;
-  padding: 5px 5px;
-  white-space: nowrap;
-  width: fit-content;
+
+  white-space: normal;
   text-align: left;
+  line-height: 1.5;
+
+  max-width: 300px;
```
hogo6002, line 514: Setting max width is not good for all use cases:
Screenshot 2025-02-03 at 5 17 26 PM

hogo6002: Also, are there any reasons for placing the tooltip above the text? I feel it's more common to position it below.

internal/output/html/vuln_table_template.gohtml
```diff
@@ -15,7 +15,11 @@
       {{ else }}
       <div class="tooltip">
         <div class="clickable" onclick="openVulnInNewTab('{{ $element.ID }}')">{{ $element.ID }}</div>
-        <span class="tooltiptext">Group IDs: {{ join $element.GroupIDs ", " }}</span>
+        <span class="tooltiptext">Group IDs: <br>
```
hogo6002, line 18: This change also needs to be added to the "uncalled vulnerability" section below. Ideally, we should only have one template, but "called" and "uncalled" vulnerabilities are displayed slightly differently, so I've kept two sections for now. Perhaps we can just merge them.

internal/output/html/style.css
```diff
@@ -119,19 +124,22 @@ header {
...
 }
 
 .package-details {
   text-align: left;
-  padding: 0 50px;
-  display: none;
-  padding-top: 20px;
-  padding-bottom: 30px;
-  border-bottom: 1px solid rgba(255, 255, 255, 0.12);
-  max-height: 800px;
```
hogo6002: I still think we should keep the max height and make the package details a scrollable window. This would be a better user experience for packages with many vulnerabilities. I see the tooltip issue, but I feel it's not too much problem, the aliases/group ID tooltip window can still be viewed by scrolling down.

hogo6002 (commented):
Thanks for cleaning up the JS file with best practice!

hogo6002 (approved):
Thanks!

PR #1566, opened 2025-02-04
fix: offline scanning having different results than online

another-rex:
This fixes offline scanning having different results compared to using the API. Seems to have two causes:

- vulnerabilities.Include function tests whether a vulnerability is included in aliases as well, and offline scanning skips those vulns if it is included in aliases. However, this is not the behavior of online scanning, where we return all records that match and do not skip alised vulns.
- osv.dev query code has a special section to edit stdlib versions without a patch version to always assume it's at the max patch version. This code is not replicated in offline matcher code.
  - I initially placed it in the query code because I didn't want to show the patched version in the output. But I think it's probably a good thing that we show how we are interpreting a version rather than silently patching it, so moved it out to our PackageInfo struct.

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -1085,15 +1085,15 @@ Scanned <rootdir>/fixtures/call-analysis-go-project/go.mod file and found 4 pack
 +-------------------------------------+------+-----------+-----------------------------+---------+------------------------------------------+
 | https://osv.dev/GO-2023-1558        | 5.9  | Go        | github.com/ipfs/go-bitfield | 1.0.0   | fixtures/call-analysis-go-project/go.mod |
 | https://osv.dev/GHSA-2h6c-j3gf-xp9r |      |           |                             |         |                                          |
-| https://osv.dev/GO-2023-2375        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
-| https://osv.dev/GO-2023-2102        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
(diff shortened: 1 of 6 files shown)


hogo6002 (approved):
LGTM, just need to remove the unused functions.

cmd/osv-scanner/__snapshots__/main_test.snap
```diff
@@ -1085,15 +1085,15 @@ Scanned <rootdir>/fixtures/call-analysis-go-project/go.mod file and found 4 pack
...
-| https://osv.dev/GO-2023-2185        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
-| https://osv.dev/GO-2023-2382        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
-| https://osv.dev/GO-2024-2598        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
-| https://osv.dev/GO-2024-2599        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
-| https://osv.dev/GO-2024-2687        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
-| https://osv.dev/GO-2024-2887        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
-| https://osv.dev/GO-2025-3373        |      | Go        | stdlib                      | 1.19    | fixtures/call-analysis-go-project/go.mod |
+| https://osv.dev/GO-2023-2375        |      | Go        | stdlib                      | 1.19.99 | fixtures/call-analysis-go-project/go.mod |
```
hogo6002, line 1088: We probably don't want to change the version here right?

G-Rath (approved):

PR #1569, opened 2025-02-06
feat: Update scalibr and add support for bun.lock

another-rex:
Also reorder all the extractors so it's easier to read

We actually had a duplicate extractor defined in there.

This also fixes the issue with SBOM extraction discussed here: https://github.com/google/osv-scanner/discussions/1529#discussioncomment-12017733 

Though we still need to follow up with documentation updates.


cuixq (commented):
I am guessing the `bun.lock` fixture is not added to this PR...

cuixq (approved):

PR #1579, opened 2025-02-06
refactor: rewrite comments to be prefixed with the subject name

G-Rath:
This addresses most of the violations reported by Golang - the remaining ones are most todo comments.

Not all of these are perfect, but I think most of them are a slight improvement; there are a few that also made me think the "thing" in question might want revisiting, such as `IsAcceptenceTest` which should probably actually be named something like `IsAcceptenceTesting`.

### DIFF
--- a/cmd/osv-scanner/internal/helper/callanalysis_parser.go
+++ b/cmd/osv-scanner/internal/helper/callanalysis_parser.go
@@ -5,7 +5,7 @@ var stableCallAnalysisStates = map[string]bool{
 	"rust": false,
 }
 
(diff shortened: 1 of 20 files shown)


cuixq (approved):
Thank you!

another-rex (approved):

hogo6002 (approved):
Thanks!

another-rex:
Hmm it looks like the container image is not consistent every build, I'll fix it in another PR.

G-Rath:
@another-rex I think it's just that another vulnerability has come out - I've opened #1592 to update the snapshots

michaelkedar (approved):

PR #1582, opened 2025-02-07
feat(output): add container scanning related information to vertical output

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1575

This PR adds container scanning related information, such as base image and layer tracing, to the vertical output.
- Uses the unified output result structs to display extra information, such as a vulnerability count summary.
- Adds a base image summary at the beginning, similar to the HTML output.
- Adds the introduced layer for each package.
- Replaces "Uncalled" with the actual filter reason, such as "unimportant".
- The license violation code has been temporarily deleted and will be added in a separate PR. This is because the unified output result structs do not yet include license scanning results.

Sample:
Screenshot 2025-02-10 at 11 31 18 AM

### DIFF
--- a/internal/output/vertical.go
+++ b/internal/output/vertical.go
@@ -3,6 +3,7 @@ package output
 import (
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	"unicode"
 
@@ -38,9 +39,14 @@ func PrintVerticalResults(vulnResult *models.VulnerabilityResults, outputWriter 
 }
 
 func printSummary(result Result, out io.Writer) {
+	packageForm := Form(result.PackageTypeCount.Regular, "package", "packages")
+	vulnerabilityForm := Form(result.VulnTypeSummary.All, "vulnerability", "vulnerabilities")
+	fixedVulnForm := Form(result.VulnCount.FixableCount.Fixed, "vulnerability", "vulnerabilities")
+	ecosystemForm := Form(len(result.Ecosystems), "ecosystem", "ecosystems")
+
 	summary := fmt.Sprintf(
-		"Total %[1]d packages affected by %[2]d vulnerabilities (%[3]s, %[4]s, %[5]s, %[6]s, %[7]s) from %[8]s.\n"+
-			"%[9]d vulnerabilities have fixes available.\n",
+		"Total %[1]d %[10]s affected by %[2]d known %[11]s (%[3]s, %[4]s, %[5]s, %[6]s, %[7]s) from %[8]s.\n"+
+			"%[9]d %[12]s can be fixed.\n",
 		result.PackageTypeCount.Regular,
 		result.VulnTypeSummary.All,
 		text.FgRed.Sprintf("%d Critical", result.VulnCount.SeverityCount.Critical),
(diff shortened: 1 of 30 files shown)


internal/output/vertical.go
```diff
@@ -11,30 +11,91 @@ import (
...
+		if i < len(outputResult.Ecosystems)-1 {
 			fmt.Fprintln(outputWriter)
 		}
 	}
+
+	fmt.Fprintln(outputWriter)
+}
+
+func printSummary(result Result, out io.Writer) {
+	summary := fmt.Sprintf(
```
G-Rath, line 41: This needs to be adjusted to properly handle pluralization in the vent of 1 thing - hopefully you can do that with just `output.Form`, but note "have fixes" is tricky:
  - if there's only one fixable vulnerability and one fix, it should "1 vulnerability has a fix available"
  - if there's only one fixable vulnerability, with multiple fixes, it should be "1 vulnerability has fixes available"
  - if there's multiple fixable vulnerabilities, regardless of the number of fixes, it should be "x vulnerabilities have fixes available"

(alternatively, you could instead say "can be fixed" which I believe is valid for all three situations regardless of number of fixes or vulnerabilities)

internal/output/vertical.go
```diff
@@ -11,30 +11,91 @@ import (
...
+	for _, baseImage := range imageResult.AllBaseImages {
+		baseImageString := text.FgYellow.Sprintf("Base Image %d (%s)", baseImage.Index, getBaseImageName(baseImage))
+		if baseImage.Index == 0 {
+			baseImageString = text.FgYellow.Sprintf("Your Image")
+		}
+		fmt.Fprintf(out, "  %s:\n", baseImageString)
+		for _, layer := range baseImage.AllLayers {
+			layerCommand := formatLayerCommand(layer.LayerMetadata.Command)[0]
+			layerCommand = truncate(layerCommand, 100)
+			fmt.Fprintf(out, "    %s %s", text.FgCyan.Sprintf("Layer %d", layer.Index), text.Italic.Sprintf("%s", layerCommand))
```
G-Rath, line 69: nit: I feel like it would look slightly nicer if the numbers were padded to the largest length, so all the text aligned

i.e. in your screenshot:

```
Layer 8  /bin/sh ...
Layer 9  /bin/sh
Layer 10 /bin/sh ...
```

that makes my brain happy, and I don't think it'd be very expensive to calculate

internal/output/vertical.go
```diff
@@ -11,30 +11,91 @@ import (
...
-		Form(len(result.Packages), "package", "packages"),
+		text.FgMagenta.Sprintf("%s", result.Name),
+		text.FgYellow.Sprintf("%d", result.PackageTypeCount.Regular),
+		Form(result.PackageTypeCount.Regular, "package", "packages"),
 	)
 }
 
+func printVerticalPackageContainerInfo(pkg PackageResult, out io.Writer) {
+	baseImageName := getBaseImageName(pkg.LayerDetail.BaseImageInfo)
+	fmt.Fprintf(out, "    Layer introduced in: %s", text.FgCyan.Sprintf("# %d Layer", pkg.LayerDetail.LayerIndex))
```
G-Rath, line 92: this doesn't seem right to me - we're saying a layer is introduced in a layer?

Should this instead be just "Introduced in: # layer"?

(also I feel like the colon might be redundant if you've got the coloring + `#`, especially since we don't use the `#` when listing the layers above, so I think at least one of those characters should go)

G-Rath (changes requested):
looks awesome!

internal/output/vertical.go
hogo6002: Thanks for the suggestion. I didn't really handle any pluralization, and will add this to other output formats. Updated it using outptu.Form and decided to go with the simplest way that using "can be fixed"

internal/output/vertical.go
hogo6002: I didn't notice this. Thanks for pointing out. Added some padding based on the max number of digits

internal/output/vertical.go
hogo6002: Updated it

another-rex (approved):

PR #1584, opened 2025-02-07
feat: Make skip git the default behavior

another-rex:
Also removes the `--json` flag since that has been deprecated for a while now. The json tests are not deleted, as there are the same tests with the `--format json` flag.

Succeeds: #1311 
Closes: #1277


cuixq (approved):
Do we need to update docs about this behavior change? (not sure if skipping git is documented...)

another-rex:
Good point, I'll add the new flags to the docs.

cuixq (approved):

PR #1585, opened 2025-02-07
fix: Base image dedup

another-rex:
Deduplicate base images to always pick the latest layer when multiple layers returns with the same base image result.


hogo6002 (approved):
LGTM, but we really need to add some tests in the future

another-rex:
yeah, I'll actually take a look at making some JSON tests now.

PR #1586, opened 2025-02-07
refactor: use pomxmlnet extractor from scalibr

cuixq:
Considering that `pomxmlnet` extractor is available in scalibr, this PR switches to use scalibr for the extractor, and also removes it from the existing codebase.


pkg/osvscanner/internal/scanners/lockfile.go
```diff
@@ -21,9 +21,8 @@ var lockfileExtractorMapping = map[string]string{
 	"pnpm-lock.yaml":    "javascript/pnpmlock",
 	"yarn.lock":         "javascript/yarnlock",
 	"package-lock.json": "javascript/packagelockjson",
-	// This translation works for both the transitive scanning and non transitive scanning
-	// As both extractors have the same name
-	"pom.xml":                     "java/pomxml",
+	// Use the transitive extractor by default.
+	"pom.xml":                     "java/pomxmlnet",
```
another-rex, line 25: I think we need a special case for pom.xml to select the non transitive extractor here if offline scanning is enabled.  (Or even better, we just support offline scanning in pomxmlnet, a hacky way to do this would be to include the pomxml (not net) extractor in the pomxmlnet extractor, and if we don't pass in a dependency client, fallback to that)

another-rex (approved):

Issue #762, opened 2024-01-22
Support recommended security-severity property in SARIF file export


graemechristie:
The current rules in the exported Sarif file do not include the security-severity property. A per the docs below, this is recommended for security rules.  

https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#reportingdescriptor-object

Likewise, the precision property is also recommended and used in concert with the security severity to assess the impact of the recorded CVE's.

PR #1587, opened 2025-02-07
feat: Added support for severity in sarif report

mickem:
Adds severity to sarif reports.

This PR fixes https://github.com/google/osv-scanner/issues/762

PS. I am not a proficient GO programmer so any and all feedback and ideas for improving this is appreciated.

The new fields are `properties` and `severity` in the example below, I added severity as a string to the bundle as that is how I interpret the GitHub docs `A string representing a score that indicates the level of severity`.
If no severity is available the field is left out.

```json
          "rules": [
            {
              "id": "",
              "name": "",
              "shortDescription": {
                "text": ""
              },
              "fullDescription": {
                "text": "",
                "markdown": ""
              },
              "deprecatedIds": [
                "CVE-2022-24713",
                "RUSTSEC-2022-0013",
                "GHSA-m5pq-gvj9-9vr8"
              ],
              "help": {
                "text": "",
                "markdown": ""
              },
              "properties": {
                "security-severity": "7.5"
              }
            },

```

This should hopefully be according to GitHub specification here: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#reportingdescriptor-object


internal/output/sarif.go
```diff
@@ -2,10 +2,12 @@ package output
 
 import (
 	"fmt"
+	"github.com/google/osv-scanner/v2/internal/utility/severity"
```
another-rex, line 5: This needs to go below the stdlib imports.

another-rex (approved):
Thanks for the contribution! LGTM, just a minor nit.

internal/output/sarif.go
mickem: Should be fixed now.

mickem: I also checked pipelines and fixed some tests I had missed. There is however a series of "new vulnerabilities" found, should I update those snapshots in this PR or is there some process I should follow to get the build green?

mickem:
I also checked pipelines and fixed some tests I had missed. There is however a series of "new vulnerabilities" found, should I update those snapshots in this PR or is there some process I should follow to get the build green?

mickem:
The golangci-lint I think fails as a side effect of the other jobs failing and GitHub pulls the plug on the run.

The other jobs fail as there is a new CVE since I updated the PR, so if I know when we will run the builds next I can make sure the PR is green before that time. As it seems to check live data there are new CVE:s popping up on and off.

PR #1591, opened 2025-02-09
refactor: clean up HTML js script so its more consistent

G-Rath:
While looking through some other stuff, I noticed my IDE had some recommendations about the JavaScript, and I spotted a few other refactors I think are worth landing, so I did a bit of a cleanup.

At a high level the changes are about being consistent and improving readability - I've purposely tried to keep this light since ultimately the amount of work being done is very small and the whole file gets included in the binary so e.g. I've not refactored the order that we set some variables even though they could be made technically faster.

Overall, this reduces the size of the file by a couple of hundred bytes which is technically a win😅

### DIFF
--- a/internal/output/html/script.js
+++ b/internal/output/html/script.js
@@ -368,7 +368,7 @@ document.addEventListener("DOMContentLoaded", function () {
     "type-filter-option-container"
   );
 
-  typeFilterOptions.addEventListener("change", function () {
+  typeFilterOptions.addEventListener("change", function (event) {
     resetSearchText();
     const changedElement = event.target;
     const allTypesCheckbox = document.getElementById("all-type-checkbox");
--- a/internal/output/html/script.js
+++ b/internal/output/html/script.js
@@ -10,7 +10,7 @@ function quickFilterByLayer(DiffID, layerCommand) {
 }
 
 function showBaseImageLayer(imageID) {
-  const detailElementID = "base-image-details-" + imageID;
+  const detailElementID = `base-image-details-${imageID}`;
   const detailsElement = document.getElementById(detailElementID);
 
   const icon = document.querySelector(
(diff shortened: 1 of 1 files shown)


internal/output/html/script.js
```diff
@@ -252,16 +239,16 @@ function applyLayerFilter(selectedLayerID) {
   tableRows.forEach(row => {
     const rowLayerID = row.getAttribute("data-layer");
     if (selectedLayerID !== "all" && rowLayerID !== selectedLayerID) {
-      const packageDetails = document.getElementById(row.id + "-details");
+      const packageDetails = document.getElementById(`${row.id}-details`);
       const vulnElements = packageDetails.querySelectorAll(".vuln-tr");
       vulnElements.forEach(vuln => {
-        vuln.classList.toggle("hide-block", true);
+        vuln.classList.add("hide-block");
```
another-rex, line 245: Oh wow, TIL I've been over complicating this the whole time, I thought just using add will duplicate the class if it already exists, but apparently not: https://stackoverflow.com/questions/51214589/is-there-any-harm-in-adding-a-class-to-an-element-that-already-has-that-class-us

They should have named this classSet haha.

another-rex (approved):
LGTM, thanks! What IDE are you using?, I tried passing this file to eslint but it wasn't very helpful.

G-Rath:
> What IDE are you using?

I use IntelliJ ultimate, which has most of the JetBrains suite as plugins - so in this case it was WebStorm since its a JS file.

> I tried passing this file to eslint but it wasn't very helpful

Yeah ESLint will require some config/plugins to do anything, and that'd require making some choices on what you actually want to do - fwiw [Ackama has a solid config](https://github.com/ackama/eslint-config-ackama/blob/main/index.js) (maintained largely by yours truly, so also I'm biased 😅), which I'm happy to run over this if you want but tbh I'm not sure its worth it right now given its a single file and whatnot (likewise why I'd not pushed at setting up tools like ESLint or TypeScript).

But if you want I'm happy to go wholehog with all the toys on this if you like 😄

PR #1592, opened 2025-02-09
test: update image scanning snapshots

G-Rath:
I feel like ideally this shouldn't actually be in the Java test, but potentially eliminating it is out of my wheelhouse :)

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -3010,8 +3010,8 @@ failed to load image from tarball with path "./fixtures/oci-image/no-file-here.t
 Scanning local image tarball "../../internal/image/fixtures/test-java-full.tar"
 
 Container Scanning Result (Alpine Linux v3.21):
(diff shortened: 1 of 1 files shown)


another-rex (approved):
I think ideally I'll find a JRE image that I can pin this to. Installing the package in the lockfile causes changes to the image every build.

G-Rath:
Replying to:

              Hmm it looks like the container image is not consistent every build, I'll fix it in another PR.

_Originally posted by @another-rex in https://github.com/google/osv-scanner/issues/1579#issuecomment-2646620330_
            
~Not sure if I missed something in particular that made you think "inconsistent" rather than "new vulnerability just dropped", but fwiw I noticed the Java Dockerfile includes doing `apt update` and apk update` which could make things more "flakey" ...~

> I think ideally I'll find a JRE image that I can pin this to. Installing the package in the lockfile causes changes to the image every build.

Snap, that's where I was heading too (I think)

another-rex:
Just looking here: https://osv.dev/list?q=openssl&ecosystem=Alpine The latest one published by alpine is published Jan 20th. Maybe a new affected version could have been added more recently though.

PR #1595, opened 2025-02-09
ci: try building images in parallel

G-Rath:
Just curious to see if this has any impact to the build speed


G-Rath:
Unsurprisingly it does not 🤷

Issue #1567, opened 2025-02-04
Lower the time tests take to run


another-rex:
Our go tests are starting to take quite a long time now, we should look into performance improvements or gating tests behind flags so they are not always ran. (though caching should already be doing this)

Here are the tests that are currently taking more than 2 seconds, sorted in ascending order:

```
TestRun_MavenTransitive/scans_transitive_dependencies_by_specifying_pom.xml,2.04
TestRun_MavenTransitive/scans_transitive_dependencies_for_pom.xml_by_default,2.35
TestRun/config_file_can_be_broad,2.38
TestRun_Licenses/No_vulnerabilities_with_license_summary_in_markdown,2.42
TestRun_MavenTransitive/scans_dependencies_from_multiple_registries,2.52
TestRun/PURL_SBOM_case_sensitivity_(local),2.52
TestRun_MavenTransitive/resolve_transitive_dependencies_with_native_data_source,2.6
TestRun_Licenses/Some_packages_with_ignored_licenses,2.7
Test_runGovulncheck,2.9
TestRun/requirements.txt_can_have_all_kinds_of_names,2.97
TestRun_LocalDatabases/only_the_files_in_the_given_directories_are_checked_by_default_(no_recursion),3.39
TestRun_LocalDatabases/.gitignored_files,3.91
TestRun_LocalDatabases/all_supported_lockfiles_in_the_directory_should_be_checked#01,3.97
TestRun_LocalDatabases/nested_directories_are_checked_when_`--recursive`_is_passed,4.53
TestRun_Fix/fix_non-interactive_json_override_pom.xml,4.54
TestRun_Fix/fix_non-interactive_override_pom.xml,4.67
TestPrintCycloneDX14Results_WithVulnerabilities,4.87
TestPrintCycloneDX15Results_WithLicenseViolations,4.95
TestPrintOutputResults_WithVulnerabilities,4.96
TestFromCSVFile,5.05
TestRun_LocalDatabases/ignoring_.gitignore,5.08
TestRunCallAnalysis/Run_with_govulncheck,5.42
TestPrintTableResults_NoTerminalWidth_WithVulnerabilities,6.05
TestCVE202230633,6.12
TestRun/folder_of_supported_sbom_with_vulns,7.27
TestRun_Fix/fix_non-interactive_relax_package.json,8.87
TestRun_LocalDatabases/all_supported_lockfiles_in_the_directory_should_be_checked,9.46
[...]

PR #1598, opened 2025-02-09
fix(output): ensure that vulnerabilities are sorted by ID across groups in table output

G-Rath:
While prototyping some changes to help with #1567, I discovered that the table output format is not explicitly sorting vulnerabilities across their groups, meaning we're assumingly using the order that the API gives us, and means our current output does not actually follow a predictable order.

To address this, I've modified the table outputter to sort vulnerability groups by the first ID in each group as a group by definition will always have at least one ID and the first ID should be the one primary one since we already sort ids within each group as part of building the general results

### DIFF
--- a/internal/output/table.go
+++ b/internal/output/table.go
@@ -7,6 +7,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/google/osv-scanner/v2/internal/identifiers"
 	"github.com/google/osv-scanner/v2/internal/imodels/ecosystem"
 	depgroups "github.com/google/osv-scanner/v2/internal/utility/depgroup"
 	"github.com/google/osv-scanner/v2/internal/utility/results"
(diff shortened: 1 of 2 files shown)


internal/output/table.go
```diff
@@ -207,6 +209,11 @@ func tableBuilderInner(vulnResult *models.VulnerabilityResults, calledVulns bool
 				source.Path = sourcePath
 			}
 
+			// Ensure that groups are sorted consistently using the first ID in each group
+			slices.SortFunc(pkg.Groups, func(a, b models.GroupInfo) int {
```
G-Rath, line 213: note that this is actually modifying the value of `pkg.Groups` on the struct, which applies outside of this function - I actually thought Go didn't work like this, but apparently it does.

I have somewhat mixed feelings about doing it, as generally this is considered bad practice (especially as it's not obvious), but it saves us a few lines of code + a variable, and any bug that it causes I would expect to be because whatever is being impacted should itself be doing a sort if it cares about the order...

hogo6002: I think the best approach here is probably to replace everything with [Output/Result](https://github.com/google/osv-scanner/blob/5f2f7354b95c895ac7749696c80eec7fed62dab1/internal/output/output_result.go#L20), which is already sorted and contains all the data we need to display (even extras). I should probably update it for consistency (since container scanning result uses Output/Result).

another-rex: Should this sorting be done for the VulnerabilityResults rather than in the output?

another-rex (approved):

another-rex:
Hmm looks like the merge I did didn't work, need to run go formatter on it.

PR #1600, opened 2025-02-10
test: update snapshots again

G-Rath:
Here we go again...


another-rex (approved):
Thanks, I was looking at docker hub yesterday when I added this new java image, and saw that they spotted a new vulnerability that wasn't showing up in our results. 

Glad that it is now showing up! I think the delay is just because docker hub takes CVEs directly, and we wait for the OS advisory to triage them first.

cuixq (approved):

hogo6002 (approved):

another-rex:
Already out of date 😭

PR #1604, opened 2025-02-11
feat: Container json testing

another-rex:
Fork of #1602 

Adds a JSON matching method, and implements a path enumeration function to help mock values in the snapshot to avoid polluting the logs with very long JSON values that we are not necessarily testing.

Moved the normalisation step into the match snapshot function itself to ensure that it is always the **last** step taken.


another-rex:
@michaelkedar fyi this changes some of GR's snapshots to be technically incorrect (changes backslashes to forward slashes).

hogo6002 (approved):
Thanks!

PR #1605, opened 2025-02-11
fix(html): fix HTML filter

hogo6002:
converts NodeList to Array before using `.some()`


G-Rath (approved):
Ek whoops forgot about that 😅

another-rex (approved):

PR #1607, opened 2025-02-11
Update snapshot again

another-rex:
We really need to figure out a more stable way of testing, I think this is the fourth snapshot update in as many days.


cuixq (approved):
Considering we have the clients stabilized, I think we can start working on mocking clients for testing.

PR #1609, opened 2025-02-12
Output/use new struct

G-Rath:
This is an initial attempt to convert the table output to use the new `Output.Result` and co that is being used for container scanning.

Overall, I think I'm actually pretty close, but I'm going to put it down for now to focus on `slog` integration

### DIFF
--- a/internal/output/__snapshots__/table_test.snap
+++ b/internal/output/__snapshots__/table_test.snap
@@ -201,9 +201,9 @@
 ╭───────────────────────┬──────┬───────────┬─────────────┬─────────┬────────────────────────────╮
 │ OSV URL               │ CVSS │ ECOSYSTEM │ PACKAGE     │ VERSION │ SOURCE                     │
 ├───────────────────────┼──────┼───────────┼─────────────┼─────────┼────────────────────────────┤
+│ https://osv.dev/OSV-1 │      │ npm       │ mine1       │ 1.2.2   │ path/to/my/first/lockfile  │
 │ https://osv.dev/OSV-1 │      │ npm       │ mine1 (dev) │ 1.2.3   │ path/to/my/first/lockfile  │
 │ https://osv.dev/OSV-5 │      │ npm       │ mine1 (dev) │ 1.2.3   │ path/to/my/first/lockfile  │
-│ https://osv.dev/OSV-1 │      │ npm       │ mine1       │ 1.2.2   │ path/to/my/first/lockfile  │
 │ https://osv.dev/OSV-2 │      │ npm       │ mine2 (dev) │ 3.2.5   │ path/to/my/second/lockfile │
 │ https://osv.dev/OSV-3 │      │ npm       │ mine3       │ 0.4.1   │ path/to/my/second/lockfile │
 │ https://osv.dev/OSV-5 │      │ npm       │ mine3       │ 0.4.1   │ path/to/my/second/lockfile │
@@ -215,9 +215,9 @@
 ╭───────────────────────┬──────┬───────────┬─────────┬─────────┬────────────────────────────╮
 │ OSV URL               │ CVSS │ ECOSYSTEM │ PACKAGE │ VERSION │ SOURCE                     │
 ├───────────────────────┼──────┼───────────┼─────────┼─────────┼────────────────────────────┤
+│ https://osv.dev/OSV-1 │      │ npm       │ mine1   │ 1.2.2   │ path/to/my/first/lockfile  │
 │ https://osv.dev/OSV-1 │      │ npm       │ mine1   │ 1.2.3   │ path/to/my/first/lockfile  │
 │ https://osv.dev/OSV-5 │      │ npm       │ mine1   │ 1.2.3   │ path/to/my/first/lockfile  │
-│ https://osv.dev/OSV-1 │      │ npm       │ mine1   │ 1.2.2   │ path/to/my/first/lockfile  │
 │ https://osv.dev/OSV-2 │      │ npm       │ mine2   │ 3.2.5   │ path/to/my/second/lockfile │
 │ https://osv.dev/OSV-3 │      │ npm       │ mine3   │ 0.4.1   │ path/to/my/second/lockfile │
 │ https://osv.dev/OSV-5 │      │ npm       │ mine3   │ 0.4.1   │ path/to/my/second/lockfile │
@@ -244,12 +244,12 @@
 ╭───────────────────────┬──────┬───────────┬───────────────┬─────────┬────────────────────────────╮
 │ OSV URL               │ CVSS │ ECOSYSTEM │ PACKAGE       │ VERSION │ SOURCE                     │
 ├───────────────────────┼──────┼───────────┼───────────────┼─────────┼────────────────────────────┤
-│ https://osv.dev/OSV-1 │      │ Packagist │ author1/mine1 │ 1.2.3   │ path/to/my/first/lockfile  │
-│ https://osv.dev/OSV-5 │      │ Packagist │ author1/mine1 │ 1.2.3   │ path/to/my/first/lockfile  │
-│ https://osv.dev/OSV-1 │      │ npm       │ mine1         │ 1.2.2   │ path/to/my/first/lockfile  │
-│ https://osv.dev/OSV-2 │      │ NuGet     │ mine2         │ 3.2.5   │ path/to/my/second/lockfile │
 │ https://osv.dev/OSV-3 │      │ Packagist │ author3/mine3 │ 0.4.1   │ path/to/my/second/lockfile │
 │ https://osv.dev/OSV-5 │      │ Packagist │ author3/mine3 │ 0.4.1   │ path/to/my/second/lockfile │
+│ https://osv.dev/OSV-2 │      │ Packagist │ mine2 (dev)   │ 3.2.5   │ path/to/my/second/lockfile │
+│ https://osv.dev/OSV-1 │      │ npm       │ author1/mine1 │ 1.2.3   │ path/to/my/first/lockfile  │
+│ https://osv.dev/OSV-5 │      │ npm       │ author1/mine1 │ 1.2.3   │ path/to/my/first/lockfile  │
+│ https://osv.dev/OSV-1 │      │ npm       │ mine1         │ 1.2.2   │ path/to/my/first/lockfile  │
 ╰───────────────────────┴──────┴───────────┴───────────────┴─────────┴────────────────────────────╯
 
 ---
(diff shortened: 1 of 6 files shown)


internal/output/__snapshots__/table_test.snap
```diff
@@ -1024,12 +1024,12 @@
 ╭───────────────────────┬──────┬───────────┬───────────────┬─────────┬──────── ≈
 │ OSV URL               │ CVSS │ ECOSYSTEM │ PACKAGE       │ VERSION │ SOURCE  ≈
```
G-Rath, line 1025: I think this has revealed that our test cases are structured in a way that is not expected by the new data format - we've got at least one case where a `models.PackageSource` has packages from different ecosystems, which results in those packages getting associated with a single ecosystem due to `output.BuildResults` assuming all packages in a source will belong to the same ecosystem

@hogo6002 would you mind reviewing the "multiple sources with a mixed count of packages across ecosystems, and multiple vulnerabilities" case (among others) to confirm if its valid?

G-Rath:
(I think we might also have a sort bug or two, but first I want to confirm if the test cases are actually valid since that'll be a huge source of difference either way)

internal/output/__snapshots__/table_test.snap
hogo6002: @another-rex what do you think Rex

G-Rath: Personally I think we should be prepared to support it unless it's extremely hard, because it allows support for arbitrary "files" - though that handling could be done here or at the parser level.

First though, my example case: maybe one day we support an arbitrary CSV or JSON file, meaning I can say to the scanner "here is a single file that has multiple packages from different ecosystems".

On the one hand, that's a single file so technically it's from the same source and thus could end up here as a single source - on the other hand, we could decide to handle that at the extractor level i.e. we return a source for each ecosystem that all point to the same file.

While we've talked enough in the past about CSV support to know that that specifically is unlikely, but I still think its an easy example of how in future maybe some kind of file will come along that we do want to support.

Related, is this not actually a situation that's already possible with SBOM? can that not include packages across multiple ecosystems?

> I can also easily modify the outputResult code to handle this if we think this type of case is possible.

Again, I think if it's very easy then it's probably worth doing just so we don't have to every think about it again 🤷

internal/output/output_result.go
```diff
@@ -53,6 +53,8 @@ type PackageResult struct {
 	HiddenVulns      []VulnResult
 	LayerDetail      PackageContainerInfo
 	VulnCount        VulnCount
+	Path             string   `json:"-"`
```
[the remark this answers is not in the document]
G-Rath: yeah looks like it probably is - I was just trying to get things working enough that the snapshots weren't completely bogus before I started refining things.

fwiw, I do think the fields could do with some documentation to make it easier to find things - for example, whats the difference between `RegularVulns` and `HiddenVulns`? and why is `OSPackageNames` a thing - when should I use that instead of `Name`? etc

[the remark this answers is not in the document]
hogo6002: I'll add more comments to clarify. 
`RegularVulns` refers to the called vulnerabilities, and `HiddenVulns` are those we've filtered out for users. These hidden vulnerabilities might be unimportant (for OS packages), or they could be uncalled (source packages). The `OSPackageName` field exists because OS packages have both source name and package name (the binary file name). For example, the source package `krb5` on Debian might result in binary package names like `krb5-locales`and `libkrb5-3`. For project scanning, I think the package.name is enough

internal/output/__snapshots__/table_test.snap
hogo6002: I actually don't know about how we get ecosystems for packages in SBOM. I can just modify the OutputResult code to add support for this case, it's just a few lines code change.

internal/output/output_result.go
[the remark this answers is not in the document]
hogo6002: Added an [issue](https://github.com/google/osv-scanner/issues/1528?issue=google%7Cosv-scanner%7C1617) to track this.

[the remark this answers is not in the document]
hogo6002: I've resolved this issue and also added license scanning results to the output.Result structure. I think we can now fully replace the table output with this new structure.

G-Rath:
@hogo6002 I'm looking at picking this back up since it sounds like something we want overall?

I think there's two main points:
  1. I've switched to using `source.Name`, only caveat is that has the "type" included with a colon which'll be present even if there is no type; for now to maintain the existing output I'm trimming that colon prefix, but that might be something we want to change in future
  2. it looks like the old logic was sorting first by package source/path, whereas now we're sorting by ecosystem; personally I think it's better to be sorted by source first, but wanted to make sure we're in agreement on that before I look into changing it

(I'm not sure about the other sorting changes, but they look at least harmless enough to ignore them, though I imagine I'll probably find out the reason for them too if I dig into 2.)

internal/output/table.go
```diff
@@ -212,64 +209,63 @@ type tbInnerResponse struct {
...
+
+					name := pkg.Name
 
-				if pkg.Package.Ecosystem == "" && pkg.Package.Commit != "" {
-					pkgCommitStr := results.PkgToString(pkg.Package)
-					outputRow = append(outputRow, "GIT", pkgCommitStr, pkgCommitStr)
-					shouldMerge = true
-				} else {
-					name := pkg.Package.Name
 					// TODO(#1646): Migrate this earlier to the result struct directly
```
G-Rath, line 256: I'm not sure if I've actioned this as a happy accident, or if it's about moving the whole condition (and thus the need for my new `DepGroups` field) out earlier - if it is the latter though, is it meaning we append the dev marker elsewhere? that feels a bit weird as we'd be changing the "name" of the package and what if we need that for other logic...?

hogo6002:
> 1. I've switched to using `source.Name`, only caveat is that has the "type" included with a colon which'll be present even if there is no type; for now to maintain the existing output I'm trimming that colon prefix, but that might be something we want to change in future

sounds good to me. We can just trim the colon prefix first.

> 2. it looks like the old logic was sorting first by package source/path, whereas now we're sorting by ecosystem; personally I think it's better to be sorted by source first, but wanted to make sure we're in agreement on that before I look into changing it

Within each ecosystem, it also sorts by source name. It groups all sources from one ecosystem together, which benefits the display of container scanning results. For project scanning, I think we could also separate the table into multiple tables by ecosystem, similar to the [package table view](https://github.com/google/osv-scanner/blob/5462bdae82ee71e36bdc87e4c7fed6ce124f71b3/cmd/osv-scanner/snapshots/main_test.snap#L3176). Then, we can just remove the `ecosystem` column from the result table. If we don't want to make a big change, I think it's fine to show vulnerabilities from the same ecosystem together.

internal/output/table.go
hogo6002: I think we can just modify package.name earlier rather than adding DepGroups. This Result struct is only for displaying the result and is being used for other formats as well. I think it would be better that package.name matches the final output, and it also avoids extra handling in other formats

G-Rath:
> For project scanning, I think we could also separate the table into multiple tables by ecosystem, similar to the [package table view](https://github.com/google/osv-scanner/blob/5462bdae82ee71e36bdc87e4c7fed6ce124f71b3/cmd/osv-scanner/__snapshots__/main_test.snap#L3176)
> ...

I think you might have misunderstood - I'm asking if we want to be sorting first by source, or by ecosystem, as previously we were doing it by source but now we're doing it by ecosystem, and I personally think it would be better to sort by source first; if it's already been decided to not sort by source first intentionally, then so be it - but I'm not sure if that is the case, or if we've just defaulted to sorting by ecosystem (e.g. as a result of this new structure having been built for container scanning first, where the preference is by ecosystem).

The reason why I think "source first" would be better is that's one of the first things people think about as part of triaging and actioning - consider, you're scanning multiple codebases each with their own `package-lock.json`; you'd see say there's two entries for one of the lockfiles, and so you know that one has a low number of vulnerabilities, and for another lockfile there's a much larger list. Also, when starting to action those, you need to go to each `package-lock.json` file in their respective locations to run commands for actioning, and only then do I care about what those commands are (which are provided by the "ecosystem", in part).

> For project scanning, I think we could also separate the table into multiple tables by ecosystem, similar to the [package table view](https://github.com/google/osv-scanner/blob/5462bdae82ee71e36bdc87e4c7fed6ce124f71b3/cmd/osv-scanner/__snapshots__/main_test.snap#L3176)

I think this is a good idea in either case, but it doesn't answer if we do the output by ecosystem or by source.

---

This could be something that's actually worth a flag e.g. `--group-by=<source|ecosystem>`

internal/output/table.go
```diff
@@ -212,64 +209,63 @@ type tbInnerResponse struct {
...
-					if depgroups.IsDevGroup(ecosystem.MustParse(pkg.Package.Ecosystem).Ecosystem, pkg.DepGroups) {
+					if depgroups.IsDevGroup(ecosystem.MustParse(eco.Name).Ecosystem, pkg.DepGroups) {
 						name += " (dev)"
 					}
-					outputRow = append(outputRow, pkg.Package.Ecosystem, name, pkg.Package.Version)
+					outputRow = append(outputRow, name)
+					outputRow = append(outputRow, pkg.InstalledVersion)
+					outputRow = append(outputRow, strings.TrimPrefix(source.Name, ":"))
```
hogo6002, line 262: version here can be git commit, I think we can just use getInstalledVersionOrCommit(pkg)

hogo6002:
> I personally do still feel I'd prefer the order by source than ecosystem because it feels less surprising to me, and that a flag for that could be useful, but among other things I think I'm a little biased by our test suite using "one, two, three" type file names which make it easy to mentally sort so in the wild it might be less surprising 🤷

Yes, for cases where users scan multiple projects, sorting by source might make it easier to view the source with the highest number of vulnerabilities. The best approach, I think, is to follow the [HTML output](https://hogo6002.github.io/v2/ubuntu/) method: group all sources from one ecosystem together but separate each source into individual tables. That way, it's quite straightforward to review the number of vulnerabilities for each source. We've implemented this in both the vertical and HTML outputs. For this PR, I feel we can try to make less changes, which means keeping everything in one large table, as this is less surprising to users. If we receive complaints from users about this sorting, we can then separate them into multiple tables or add extra flags. For now, I think it's fine. (Personally, I wish users would use the vertical and HTML outputs more than the table output.)

G-Rath:
> The best approach, I think, is to follow the [HTML output](https://hogo6002.github.io/v2/ubuntu/) method: group all sources from one ecosystem together but separate each source into individual tables.

Personally I still think the other way around, especially for the HTML output - I tend to think about codebases/projects first, then ecosystems, but I think we can put a pin in this for now 🙃 

I might see if I can get the old implementation sorting by ecosystem first to reduce the diff here, but otherwise I think this is resolved for now.

> (Personally, I wish users would use the vertical and HTML outputs more than the table output.)

That's actually a really good reminder - weren't we going to make the vertical output the default?

G-Rath:
Yeah [we were](https://github.com/google/osv-scanner/pull/889#issuecomment-2207815086) - is that something we still want to do? cc @another-rex 

fwiw, I don't think that's actually technically a breaking change but we might as well do it now if we want before v2 goes out

internal/output/__snapshots__/table_test.snap
G-Rath: @hogo6002 fwiw, I've been working on some stuff for Drupal advisories and it sounds like one possible approach we might go with would having a dedicated `Drupal` ecosystem meaning extracting `composer.lock` files could give a mix of `Packagist` and `Drupal` packages, which would have been relevant here

hogo6002:
hey, is the PR ready for review?

G-Rath:
@hogo6002 yeah it should be - I think this ended up falling off my radar between other work and then you were off for a week or so and then I forgot about it again 😅 

I think it should be pretty much good to go, but ultimately I think it would be best if you just do a fresh review rather than us trying to remember what we were discussing 2+ months ago.

(I'm happy to do a new PR too if that'd be cleaner)

G-Rath:
yeah oh that's right, for some reason now the table output is using wrong widths for columns so I need to figure out what's happening there - still @hogo6002 feel free to look over the changes and let me know if there's anything else that you think should be actioned

hogo6002 (approved):
LGTM

hogo6002 (approved):

hogo6002:
I will merge it now. I think we need a followup PR to hide unimportant vulns when no `--all-vulns` flag. We can also add fixed version info into table output

G-Rath:
Yeah that is probably the way to go given how much rebasing this has needed anytime something else is merged in, and then we can work on building on this.

fwiw I'm starting to suspect our intermediate struct(s) is handling far too much stuff when instead it should be just holding structured data and instead the `output` package should be putting it together for output e.g. `source.Name` is being set to `source.Type` + `source.Path` only for us to later strip that out in `output` since we just want `source.Path` (for now) - I've not dug into this too deeply and wouldn't be surprised if at least some of this is the result of stuff out of our control, but worth keeping in mind for future work in this area I think

PR #1610, opened 2025-02-12
test: disable parallel testing

G-Rath:
(just seeing how slow this is)


G-Rath:
ok so under 8 minutes isn't super bad...

PR #1611, opened 2025-02-12
feat(output): add license scanning results into the new Result struct

hogo6002:
This PR
- Adds a `LicenseSummary` struct to store the count of each license.
- Adds `Licenses` and `LicenseViolations` fields
- Implements `calculateLicenseSummary`, mostly copying code from table.go/licenseSummaryTableBuilder().
- Adds `Licenses`  resutls to all formats

**Table output - container scanning:**
with `experimental-licenses-summary` flag:
image
with `experimental-licenses` flag:
image

Table output - project scanning (unchanged)

**Vertical output:**
with `experimental-licenses-summary` flag:
image
with `experimental-licenses` flag (adds it back, same as before):
image

**HTML output:**
with `experimental-licenses-summary` flag:
image
with `experimental-licenses` flag:
image
One issue here is that the HTML filter currently hides packages that have no vulnerabilities. So packages without vulnerabilities but with license violations will also be hidden.

Resolves https://github.com/google/osv-scanner/issues/1576


josieang:
Just checking that the plan is still:
- By default show summary and licenses by package in the `–experimental-licences` flag
   - `--experimental-licenses` turns on both summary and licenses by package.
   - Table package view: Show summary and all licenses by package.
   - Vertical view: Add a new “License information” section” and summary and all licenses by package.
   - Html output: Add license summary in the summary section. Per package as well.
-  remove summary flag.
- Change how the allowlist is accepted:
   - Change `–experimental-licences` it a bool flag instead of an allowlist
   - Add `–experimental-licenses-allowlist` flag
- Promote this feature to a non-experimental status.

josieang:
For the table and vertical summary output
- Where does the summary appear in relation to the other things outputted? I don't think it should go before vulnerabilities, but perhaps I'm missing context.

> One issue here is that the HTML filter currently hides packages that have no vulnerabilities. So packages without vulnerabilities but with license violations will also be hidden.

this is an interesting point. Perhaps we shouldn't support per-package licenses in the html packages view until we have a better idea of how to present this. In the summary, we can say "for more details on licenses, run with table/json/etc. output"

hogo6002:
@josieang yep, I still want to get license scanning out of experimental before we release v2. I talked to Rex, and I think we can combine those flags:
- Just keep the `--licenses` flag, but it can be used with or without extra args (as a generic flag type).
- As long as users just use `--licenses`, it'll show a summary of license scanning.
- If users use `--licenses [allowlist]`, we will treat the input as allowlist.

As we're making the summary the default for license scanning, I'll tweak the `VulnerabilityResults` struct to keep the JSON output clean.
- Summary does not imply all packages (filter out packages with no license/vuln issues)
- Store summary in the Result struct separately (pre-compute it)

what do you think?

cmd/osv-scanner/__snapshots__/main_test.snap
```diff
@@ -3004,8 +3005,9 @@ No issues found
 Scanning local image tarball "../../internal/image/fixtures/test-alpine.tar"
 
 Container Scanning Result (Alpine Linux v3.18):
-Total 2 packages affected by 41 vulnerabilities (2 Critical, 17 High, 14 Medium, 0 Low, 8 Unknown) from 1 ecosystems.
-41 vulnerabilities have fixes available.
+Total 2 packages affected by 41 known vulnerabilities (2 Critical, 17 High, 14 Medium, 0 Low, 8 Unknown) from 1 ecosystem.
+41 vulnerabilities can be fixed.
```
another-rex, line 3009: Let's keep the original wording here.

```
41 vulnerabilities have fixes available.
```

another-rex (approved):
Everything else LGTM!

cmd/osv-scanner/__snapshots__/main_test.snap
hogo6002: I think the current wording is fine, and it simplifies pluralization handling as discussed in https://github.com/google/osv-scanner/pull/1582#discussion_r1945861563.

hogo6002:
> this is an interesting point. Perhaps we shouldn't support per-package licenses in the html packages view until we have a better idea of how to present this. In the summary, we can say "for more details on licenses, run with table/json/etc. output"

By default, users will see license scanning results in the table output. The HTML output doesn't have many users and is mostly for container scanning. So, I don't think it's a big problem that some packages are hidden on HTML output. I've also created [an issue](https://github.com/google/osv-scanner/issues/1643) to track this. The best approach is probably just to add an extra filter to show all packages.

PR #1615, opened 2025-02-12
feat: filter out Ubuntu unimportant vulns

hogo6002:
resolves https://github.com/google/osv-scanner/issues/1571 and resolves https://github.com/google/osv-scanner/issues/1572


internal/grouper/grouper_models.go
```diff
@@ -26,6 +26,12 @@ func ConvertVulnerabilityToIDAliases(c []models.Vulnerability) []IDAliases {
 			idAliases.Aliases = append(idAliases.Aliases, v.Related...)
 		}
 
+		// For Ubuntu Security Advisory data,
+		// all related CVEs should be bundled together, as they are part of this USN.
+		if strings.Split(v.ID, "-")[0] == "USN" {
```
another-rex, line 31: hmm does this mean CVE in the related entries will show up as aliases in the output?

pkg/osvscanner/vulnerability_result.go
```diff
@@ -79,7 +79,8 @@ func buildVulnerabilityResults(
 
 		// For Debian-based ecosystems, mark unimportant vulnerabilities within the package.
 		// Debian ecosystems may be listed with a version number, such as "Debian:10".
-		if strings.HasPrefix(pkg.Package.Ecosystem, string(models.EcosystemDebian)) {
+		if strings.HasPrefix(pkg.Package.Ecosystem, string(models.EcosystemDebian)) ||
```
another-rex, line 82: Can we move the check into the isUnimportant function? This way the setUnimportant function is not tied to a specific ecosystem.

internal/grouper/grouper_models.go
hogo6002: nope, I think only `UBUNTU-CVE` will be shown in the group IDs (source CVEs shouldn't be present in the grouping results, as there's no Ubuntu data). And group IDs should only show one, which will be the USN. This is how we did it for DSA before.

pkg/osvscanner/vulnerability_result.go
hogo6002: I think it's better to keep the check here. This check determines whether the scanned package itself belongs to Ubuntu or Debian, while `isUnimportant` checks if a vulnerability has Ubuntu/Debian related information and is marked as unimportant. 

Once we have an ecosystem severity field (I saw some discussion on osv-schema about this), I can update code here to make it less hardcoded.

pkg/osvscanner/vulnerability_result.go
```diff
@@ -79,7 +79,8 @@ func buildVulnerabilityResults(
...
+			strings.HasPrefix(pkg.Package.Ecosystem, string(models.EcosystemUbuntu)) {
```
another-rex, line 83: Please use the osvschema Ecosystems here

hogo6002: Updated

another-rex (approved):

PR #1616, opened 2025-02-12
feat(output): display git scanning results

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1577
image


internal/output/output_result.go
```diff
@@ -401,10 +402,16 @@ func processPackage(vulnPkg models.PackageVulns) PackageResult {
...
 	packageFixedVersion := calculatePackageFixedVersion(vulnPkg.Package.Ecosystem, regularVulnList)
 
+	installVersion := vulnPkg.Package.Version
+	// If the ecosystem is not set, it's most likely from Git
+	if installVersion == "" && vulnPkg.Package.Ecosystem == "" {
+		installVersion = results.GetShortCommit(vulnPkg.Package.Commit)
+	}
+
 	packageResult := PackageResult{
 		Name:             vulnPkg.Package.Name,
```
another-rex, line 412: Can you also check if name is empty, but repo is not empty, set name to repo?

internal/output/output_result.go
```diff
@@ -401,10 +402,16 @@ func processPackage(vulnPkg models.PackageVulns) PackageResult {
 
 	packageFixedVersion := calculatePackageFixedVersion(vulnPkg.Package.Ecosystem, regularVulnList)
 
+	installVersion := vulnPkg.Package.Version
+	// If the ecosystem is not set, it's most likely from Git
+	if installVersion == "" && vulnPkg.Package.Ecosystem == "" {
+		installVersion = results.GetShortCommit(vulnPkg.Package.Commit)
```
another-rex, line 408: We should save the full commit here. This way in the HTML report we can have this as a hover tooltip.

another-rex: And if there is no package name/repo name, we should be showing the full commit.

internal/output/output_result.go
hogo6002: What do you mean by repo? we don't have repo information in `PackageVulns.PackageInfo` though.

internal/output/output_result.go
hogo6002: updated code to show short commit with the full commit in a hover tooltip. I think as long as the full commit is in the tooltip, it doesn't matter if there is a package name or not.

internal/output/output_result.go
another-rex: Ah ok we haven't passed that through, let's ignore this for now.

another-rex (approved):

PR #1619, opened 2025-02-13
test: update snapshots and upgrade golangci-lint

G-Rath:
At least this time it's a different snapshot 🤷 

Patch https://github.com/google/osv-scanner/pull/1623 so that workflows all pass

### DIFF
--- a/internal/sourceanalysis/__snapshots__/integration_test.snap
+++ b/internal/sourceanalysis/__snapshots__/integration_test.snap
@@ -227,7 +227,7 @@
             "filename": "\u003cAny value\u003e",
             "offset": -1,
             "line": -1,
-            "column": 18
+            "column": 16
           }
         },
         {
--- a/.github/workflows/lint-action/action.yml
+++ b/.github/workflows/lint-action/action.yml
@@ -22,5 +22,5 @@ runs:
       uses: golangci/golangci-lint-action@e60da84bfae8c7920a47be973d75e15710aa8bd7 # v6.3.0
       with:
         # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-        version: v1.62.2
+        version: v1.64.4
         args: --timeout=5m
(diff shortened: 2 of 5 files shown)


internal/sourceanalysis/__snapshots__/integration_test.snap
```diff
@@ -227,7 +227,7 @@
             "filename": "\u003cAny value\u003e",
             "offset": -1,
             "line": -1,
-            "column": 18
```
cuixq, line 230: I think this should not be updated - this is due to a different Go version.

G-Rath: whose version of Go specifically? CI seems to be happy with it...

cuixq: ah I think we are now at Go 1.23.6 due to https://github.com/google/osv-scanner/pull/1590 so this is expected.

cuixq (approved):

hogo6002 (approved):

G-Rath:
@cuixq I realised this morning I think the issue is because we've actually started using Go v1.24.0 in CI since we're using the "stable" version - that'll def be why `golangci-lint` started to fail, and I'm guessing it's why the Go call analysis fixture changed (and seems to be flakey).

I'm not quite sure what the right move here is between:
  - ignore it, while waiting for something to catchup (?)
  - update to Go v1.24 ourselves
  - switch some or all of our CI jobs to use `oldstable` instead of `stable` (this feels like it would be the right move, but I don't have a good handle on how Go versions interact to know if that'll result in other downsides e.g. maybe we'll miss out on linting rules, or something?)

cuixq:
I prefer to keep using `stable` and upgrading to Go 1.24 sounds good to me. hope https://github.com/google/osv-scanner/pull/1629 will solve the problem.

however, I do think we should have some presubmit set to check if any of the Go version gets updated in any PR - if so, the presubmit should fail and ask people to upgrade all Go versions (or do not update in that PR).

G-Rath:
I think the question is what is our Go policy version - previously it was the two latest versions, but technically the docs seemed to have been removed, so I'm not sure if that's because we're just giving up or a mistake.

If it is the last two versions, then we should remain on v1.23 and switch to using `oldstable`

another-rex:
We should be using the latest stable for building but ideally we have both stable and oldstable for tests (maybe we only do testing with stable at release time) to maintain compatibility.

PR #1620, opened 2025-02-13
refactor: replace experimental libraries with their std versions

G-Rath:
These became stable in Go v1.21, so we can stop using the `x/exp` versions

(this will be enforced by the `expotostd` linter which is being introduced in the next version of golanglint-ci)

### DIFF
--- a/cmd/osv-scanner/fix/noninteractive.go
+++ b/cmd/osv-scanner/fix/noninteractive.go
@@ -423,7 +423,7 @@ func makeResultVuln(vuln resolution.Vulnerability) vulnOutput {
 		vk := sg.Nodes[sg.Dependency].Version
 		affected[packageOutput{Name: vk.Name, Version: vk.Version}] = struct{}{}
 	}
-	v.Packages = slices.Collect(maps.Keys(affected))
+	v.Packages = slices.AppendSeq(make([]packageOutput, 0, len(affected)), maps.Keys(affected))
 	slices.SortFunc(v.Packages, func(a, b packageOutput) int {
 		if c := cmp.Compare(a.Name, b.Name); c != 0 {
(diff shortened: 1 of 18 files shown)


hogo6002 (approved):

cuixq (approved):

cuixq:
@G-Rath I think some tests break due to this change - can you help to fix?

another-rex (approved):

cmd/osv-scanner/fix/noninteractive.go
```diff
@@ -423,7 +423,7 @@ func makeResultVuln(vuln resolution.Vulnerability) vulnOutput {
 		vk := sg.Nodes[sg.Dependency].Version
 		affected[packageOutput{Name: vk.Name, Version: vk.Version}] = struct{}{}
 	}
-	v.Packages = slices.Collect(maps.Keys(affected))
+	v.Packages = slices.AppendSeq(make([]packageOutput, 0, len(affected)), maps.Keys(affected))
```
another-rex, line 426: Is this needed? It seems weird the .Collect doesn't reallocate slices. 

Even if it does not, I think it is better to stick with Collect here until we see that this causes a performance problem just for code clarity

G-Rath: Apparently [using `slices.AppendSeq` is the "simplest true equivalent"](https://github.com/golang/exp/blob/939b2ce775acaac3abfa605a85c57e43fc448693/maps/maps.go#L20), and there is a [performance difference](https://github.com/ldez/exptostd/issues/9).

It doesn't look like there's much difference between `slices.Collect` though, so I'm not fussed - I can do some benchmarking later too

PR #1623, opened 2025-02-13
ci: upgrade `golangci-lint` to v1.64

G-Rath:
For some reason our linting job has started failing without any actual error message to indicate the issue - I suspect it could be related to being on Go v1.23.6, but ultimately it seems upgrading to v1.64 of the linter resolves the issue.

As this introduces two new linters that want us to take actions, I've disabled them for now so that this can be landed as-is, but I've already got the patches for addressing their violations ready to go as dedicated PRs (of which #1620 is one)


cuixq (approved):

cuixq:
Updating the linter fails because of the outdated snapshots, and updating the snapshots fails because of the linter. :facepalm:

cuixq:
@G-Rath can you patch the snapshot updates in this PR so that they can get merged together?

PR #1626, opened 2025-02-14
test: update names to match their subjects

G-Rath:
This makes a lot of our tests align with the function under test, so that tooling can more accurately link tests with their subjects.

The main change is that the format for tests on private methods should be `Test_<method name>` whereas we've been doing `Test<public name version of method name>`

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -1,5 +1,5 @@
 
-[TestRun/#00 - 1]
+[Test_run/#00 - 1]
 NAME:
    osv-scanner scan - scans projects and container images for dependencies, and checks them against the OSV database.
 
@@ -19,11 +19,11 @@ OPTIONS:
 
 ---
 
-[TestRun/#00 - 2]
+[Test_run/#00 - 2]
 
 ---
 
-[TestRun/.gitignored_files - 1]
+[Test_run/.gitignored_files - 1]
 Scanning dir ./fixtures/locks-gitignore
 Scanned <rootdir>/fixtures/locks-gitignore/Gemfile.lock file and found 1 package
 Scanned <rootdir>/fixtures/locks-gitignore/subdir/yarn.lock file and found 1 package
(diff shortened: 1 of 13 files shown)


cuixq (approved):

cmd/osv-scanner/main_test.go
```diff
@@ -948,7 +948,7 @@ func TestRun_SubCommands(t *testing.T) {
 	}
 }
 
-func TestRun_InsertDefaultCommand(t *testing.T) {
+func Test_insertDefaultCommand(t *testing.T) {
```
hogo6002, line 951: Why not `Test_run_InsertDefaultCommand` here?

cmd/osv-scanner/__snapshots__/main_test.snap
```diff
@@ -1064,43 +1064,43 @@ Scanned <rootdir>/fixtures/locks-requirements/the_requirements_for_test.txt file
...
 
 ---
 
-[TestRun/version - 2]
+[Test_run/version - 2]
 
 ---
 
-[TestRunCallAnalysis/Run_with_govulncheck - 1]
+[Test_runCallAnalysis/Run_with_govulncheck - 1]
```
hogo6002, line 1103: I'm a bit confused by the naming here. I thought we were converting all test functions to the `Test_run_<OriginalFunctionName>` format. Why isn't this one named `Test_run_CallAnalysis`?

hogo6002: btw, why is `run` in lowercase?

cmd/osv-scanner/main_test.go
G-Rath: because this [is testing the function `insertDefaultCommand`, not `run`](https://github.com/google/osv-scanner/blob/main/cmd/osv-scanner/main_test.go#L1006) (this was actually the original test I noticed, which lead me to realizing about the others 😅)

cmd/osv-scanner/__snapshots__/main_test.snap
G-Rath: > Why isn't this one named Test_run_CallAnalysis?

I believe it should be, I've just mucked something up (things are a little hard to pin down right now between all the recent CI flake and the massive output from `osv-scalibr` 😅)

> btw, why is run in lowercase?

because `run` is a private method - also, just to be clear, if it was public then the name would be `TestRun`; the reason we add the underscore is that Go requires camel case names + prefixed with `Test`, and its possible to have a public and private thing with the "same" name i.e.

```
func Run() { } // -> TestRun
func run() { }  // -> Test_run
```

G-Rath:
I'm going to convert this to draft for now because apparently I've mucked up a couple of the snapshots, and it's a bit chaotic right now between the CI issues and all the output from osv-scalibr (which I think I've got an idea on how to possibly curb that short-term, which I'll do a PR for)

hogo6002 (approved):
Thanks!

another-rex (approved):

PR #1628, opened 2025-02-16
test: update snapshots and fixtures

G-Rath:
In addition to accounting for the new `musl` vulnerability (CVE-2025-26519), this also updates our Go call analysis tests to be stable on both Go 1.23 and 1.24, since the former is what we're using locally while the latter is what CI is using

### DIFF
--- a/cmd/osv-scanner/fixtures/locks-many/osv-scanner.toml
+++ b/cmd/osv-scanner/fixtures/locks-many/osv-scanner.toml
@@ -6,3 +6,7 @@ reason = "Test manifest file"
 [[IgnoredVulns]]
 id = "CVE-2022-48174"
 reason = "Test manifest file (alpine.cdx.xml)"
+
+[[IgnoredVulns]]
+id = "CVE-2025-26519"
+reason = "Test manifest file (alpine.cdx.xml)"
--- a/internal/sourceanalysis/__snapshots__/integration_test.snap
+++ b/internal/sourceanalysis/__snapshots__/integration_test.snap
@@ -227,7 +227,7 @@
             "filename": "\u003cAny value\u003e",
             "offset": -1,
             "line": -1,
-            "column": 16
+            "column": 18
           }
         },
         {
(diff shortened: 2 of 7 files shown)


internal/sourceanalysis/__snapshots__/integration_test.snap
```diff
@@ -227,7 +227,7 @@
             "filename": "\u003cAny value\u003e",
             "offset": -1,
             "line": -1,
-            "column": 16
+            "column": 18
```
G-Rath, line 230: I believe this is flipping back and forward because in `go.mod` (and thus locally) we're using Go v1.23, but CI is set to use `stable` which is now Go v1.24.

It looks like the Go versioning policy has been removed from the readme at some point, so not sure if we're intending to update to Go v1.24 now, or if we want to switch CI to use `oldstable`...

another-rex: Can we just mock this value so we don't store the column. Let's stick with using stable here.

G-Rath: Yeah probably - I think there's also a go vuln reported/not reported by this but we should be able to just ignore that too

cmd/osv-scanner/fixtures/locks-many/osv-scanner.toml
```diff
@@ -6,3 +6,7 @@ reason = "Test manifest file"
 [[IgnoredVulns]]
 id = "CVE-2022-48174"
 reason = "Test manifest file (alpine.cdx.xml)"
+
+[[IgnoredVulns]]
+id = "CVE-2025-26519"
```
G-Rath, line 11: I think ignoring this is the right call, at least for now, because we're using this fixture in a couple of different tests including some that look like they're expecting no vulnerabilities

(if I could go back in time, I think I'd push for us to include some form of readme in each of these fixtures to capture what the fixture is trying to provide)

cuixq (approved):

another-rex (approved):

PR #1629, opened 2025-02-16
build: upgrade to Go v1.24

G-Rath:
This puts us on the latest version of Go, which should hopefully resolve the recent issues we've seen with `golangci-lint` and the `sourceanalysis` snapshots which I believe are due to CI using the current `stable` version of Go which is now v1.24.

In the long-run I think we might want to actually switch to using `oldstable` in our CI, but for now we can just upgrade since we're in a beta period - we can downgrade again if needed once we've gotten the Go version policy sorted out

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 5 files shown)


cuixq (approved):

hogo6002 (approved):

G-Rath:
Closing in favor of #1628

PR #1630, opened 2025-02-17
test: don't skip acceptance tests when targeted with `-run`

G-Rath:
This is a neat little trick I thought of while cleaning up the snapshots - all its saving us from is having to stick `TEST_ACCEPTANCE=true` in front of our `go test` calls, but I think it's nice to have the test suite not require that when you're running a specific acceptance test.

Credit to `go-snaps` for proving it must be possible, since this is what it does to determine if a snapshot is obsolete or that the tests have just not been run.

I've gone with the slightly awkward name `isThisTestRunTarget` as I felt it was less ambiguous than `isThisTestRunning` and similar, since yes the test must be running if the function is being called... 😅

### DIFF
--- a/internal/testutility/utility.go
+++ b/internal/testutility/utility.go
@@ -1,6 +1,7 @@
 package testutility
 
 import (
+	"flag"
 	"os"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


internal/testutility/utility.go
```diff
@@ -34,16 +35,30 @@ func Skip(t *testing.T, args ...any) {
...
 }
 
+// isThisTestRunTarget tries to determine if the currently running test has been
+// targeted with the -run flag, by comparing the flags value to [testing.T.Name]
+//
+// Since this just does a direct comparison, it will not match for regex patterns
+func isThisTestRunTarget(t *testing.T) bool {
+	t.Helper()
+
+	runOnly := flag.Lookup("test.run").Value.String()
```
cuixq, line 45: the flag is `run` or `test.run`?

G-Rath: `test.run`, as that's actually what is used under the hood I believe because really things are being compiled into a binary, so this prefix helps avoid clashes with tests & binaries that include flags.

[From the docs](https://pkg.go.dev/cmd/go/internal/test):

> Each of these flags is also recognized with an optional 'test.' prefix,
> as in -test.v. When invoking the generated test binary (the result of
> 'go test -c') directly, however, the prefix is mandatory.

another-rex (approved):
Neat! Hopefully this means I can debug an acceptance test in vscode without having to comment out the "skip if not acceptance test" function now.

PR #1631, opened 2025-02-17
test: reduce noise in cmd cases using a custom `slog` handler

G-Rath:
Currently we got a lot of noise from `osv-scalibr` libraries which makes it hard to see what actually failed - since all of these are going through the `slog`, we can use our own custom handler to drop records with specific prefixes.

I believe there are some talks happening with the `osv-scalibr` team to see if some of these can be toned back, and also I think we will probably want to capture some of these at some point as part of valid tests, but short-term this should help make things more manageable.

I'm opening this as a draft as I don't know if we do actually want to land this, in part because I think this is going to almost immediately break as we've switching to `slog` ourselves which involves using a custom `slog` handler meaning this one will get overridden, but I still wanted to share the implementation either way as folks might like to cherry-pick it locally depending on what they're working on 🙂

### DIFF
--- a/cmd/osv-scanner/testmain_test.go
+++ b/cmd/osv-scanner/testmain_test.go
@@ -43,10 +43,6 @@ func newMuffledHandler(w io.Writer) *muffledHandler {
 	return &muffledHandler{TextHandler: *slog.NewTextHandler(w, nil)}
 }
 
-func Test_FailThis(t *testing.T) {
-	t.Fail()
-}
-
 func TestMain(m *testing.M) {
 	slog.SetDefault(slog.New(newMuffledHandler(log.Writer())))
 
--- a/cmd/osv-scanner/testmain_test.go
+++ b/cmd/osv-scanner/testmain_test.go
@@ -25,6 +25,9 @@ func (c *muffledHandler) Handle(ctx context.Context, record slog.Record) error {
 		"End status: ",
 		"Neither CPE nor PURL found for package",
 		"Invalid PURL",
+		"os-release[ID] not set, fallback to",
+		"VERSION_ID not set in os-release",
+		"osrelease.ParseOsRelease(): file does not exist",
 	} {
 		if strings.HasPrefix(record.Message, prefix) {
 			return nil
--- a/cmd/osv-scanner/testmain_test.go
+++ b/cmd/osv-scanner/testmain_test.go
@@ -1,14 +1,46 @@
 package main
 
 import (
+	"context"
+	"io"
+	"log"
+	"log/slog"
 	"os"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


cmd/osv-scanner/testmain_test.go
```diff
@@ -1,14 +1,40 @@
...
 	"github.com/google/osv-scanner/v2/internal/testutility"
 )
 
+// muffledHandler eats certain log messages to reduce noise in the test output
+type muffledHandler struct {
+	slog.TextHandler
+}
+
+func (c *muffledHandler) Handle(ctx context.Context, record slog.Record) error {
+	// todo: work with the osv-scalibr team to see if we can reduce these
```
cuixq, line 22: make a slice to have the prefixes and loop it for checking so we don't need to do `strings.HasPrefix` every time.

cuixq (approved):

cmd/osv-scanner/testmain_test.go
G-Rath: I don't really think it's worth doing that unless we've got 3-4+ items, because the number of chars are about the same until then 🤷

cmd/osv-scanner/testmain_test.go
```diff
@@ -1,14 +1,40 @@
...
+	if strings.HasPrefix(record.Message, "Neither CPE nor PURL found for package") ||
```
another-rex, line 23: Can you add the logging output that gets spammed when doing container scanning as well. 
e.g.
```
2025/02/19 15:37:54 Starting filesystem walk for root: 
2025/02/19 15:37:54 End status: 0 dirs visited, 1 inodes visited, 1 Extract calls, 154.071µs elapsed, 154.121µs wall time
```

another-rex:
Also for `slog` integration in general, let's remove snapshot testing slog output so we can keep the parallel tests.

cmd/osv-scanner/testmain_test.go
```diff
@@ -1,14 +1,55 @@
...
 	"github.com/google/osv-scanner/v2/internal/testutility"
 )
 
+// muffledHandler eats certain log messages to reduce noise in the test output
+type muffledHandler struct {
+	slog.TextHandler
+}
+
+func (c *muffledHandler) Handle(ctx context.Context, record slog.Record) error {
+	if record.Level < slog.LevelError {
```
G-Rath, line 22: I realised this is a good idea because as part of #1666 a log record with a level of error will trigger a non-zero exit code (effectively, that's the logic replacing the `HasErrored` path in the current `reporter` interface)

G-Rath:
look ma, no noise!

image

another-rex (approved):
Thanks, this is great!

PR #1634, opened 2025-02-17
fix: remove usage of `pkg/lockfile` from guided remediation

michaelkedar:
To unblock removing `pkg/lockfile` for the full v2 release.
Just copied the necessary parts into the internal directories. Guided remediation is being moved into OSV-Scalibr, so I'll do a proper refactor as part of that move.

There's still one usage left with [`VKToPackageDetails`](https://github.com/google/osv-scanner/blob/3bd1565fd31266b9899747840b8223ab6b7ed225/internal/resolution/util/depsdev.go#L15), which is needed in calls to [`vulns.IsAffected()`](https://github.com/google/osv-scanner/blob/3bd1565fd31266b9899747840b8223ab6b7ed225/internal/utility/vulns/vulnerability.go#L144).


cuixq (approved):
Thanks Michael!

another-rex (approved):

PR #1636, opened 2025-02-18
feat(updater): use upgrade config for Maven version updater

cuixq:
This PR aligns the flags of `update`, especially the `upgrade-config` with `fix` subcommand.

To be able to parse the upgrade config, this PR moves the parsing function to `upgrade` package.

The next step is to make [`comparisonFunctionWithWorkarounds`](https://github.com/google/osv-scanner/blob/e0206dedd17080021b768526f2a6fde635497c25/internal/remediation/override.go#L271) reusable and used in Maven updater as well.


internal/remediation/suggest/maven_test.go
```diff
@@ -382,22 +385,24 @@ func TestSuggestVersion(t *testing.T) {
...
+		requirement string
+		level       upgrade.Level
+		want        string
 	}{
-		{"1.0.0", false, "2.3.4"},
+		{"1.0.0", upgrade.Major, "2.3.4"},
 		// No major updates allowed
-		{"1.0.0", true, "1.2.3"},
+		{"1.0.0", upgrade.Minor, "1.2.3"},
+		// Only allow patch updates
```
cuixq, line 395: this test does not work now due to a  bug in `deps.dev/util/semver` - `1.0.1` is considered to diff from `1.0.0` by minor. I have made https://github.com/google/deps.dev/pull/191 to fix it.

cuixq: this is fixed.

cmd/osv-scanner/update/main.go
```diff
@@ -52,36 +63,58 @@ func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
...
+		switch system {
+		case resolve.Maven:
+			options.Client, err = client.NewMavenRegistryClient(ctx.String("maven-registry"))
+			if err != nil {
+				return nil, err
+			}
+		case resolve.NPM:
+		case resolve.UnknownSystem:
+		default:
+			return nil, fmt.Errorf("native data-source currently unsupported for %s ecosystem", system.String())
```
michaelkedar, line 116: Should these cases be `fallthrough`-ed?

internal/remediation/suggest/maven.go
```diff
@@ -107,7 +113,8 @@ func suggestMavenVersion(ctx context.Context, client resolve.Client, req resolve
 			// Skip versions smaller than the current requirement
 			continue
 		}
-		if _, diff := v.Difference(current); diff == semver.DiffMajor && noMajorUpdates {
+		_, diff := v.Difference(current)
+		if (level == upgrade.Minor && diff == semver.DiffMajor) || (level == upgrade.Patch && (diff == semver.DiffMajor || diff == semver.DiffMinor)) {
```
michaelkedar, line 117: I believe [`if !level.Allows(diff)`](https://github.com/google/osv-scanner/blob/ac5a74c38724001bc93e40e91ec31bfcf45cf0ae/internal/remediation/upgrade/level.go#L17) should work here

cuixq: this is useful!

cmd/osv-scanner/update/main.go
cuixq: done

michaelkedar (approved):

PR #1639, opened 2025-02-18
fix(output): handles cases where a single source contains packages from multiple ecosystems

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1617

### DIFF
--- a/internal/output/output_result.go
+++ b/internal/output/output_result.go
@@ -45,14 +45,17 @@ type SourceResult struct {
 
 // PackageResult represents the vulnerability scanning results for a package.
 type PackageResult struct {
-	Name             string
+	Name string
+	// OSPackageNames represents the actual installed binary names. This is primarily used for container scanning.
 	OSPackageNames   []string
 	InstalledVersion string
 	FixedVersion     string
-	RegularVulns     []VulnResult
-	HiddenVulns      []VulnResult
-	LayerDetail      PackageContainerInfo
-	VulnCount        VulnCount
+	// RegularVulns includes all vulnerabilities intended for display to users
+	RegularVulns []VulnResult
+	// HiddenVulns includes vulnerabilities that are filtered out for users, such as those deemed unimportant or uncalled.
+	HiddenVulns []VulnResult
+	LayerDetail PackageContainerInfo
+	VulnCount   VulnCount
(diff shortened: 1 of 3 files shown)


another-rex:
Does this change the output for the HTML report as well?

hogo6002:
> Does this change the output for the HTML report as well?

It doesn't need to change HTML code as the struct remains the same.

another-rex:
I meant does this affect the contents of the HTML output

internal/output/output_result.go
```diff
@@ -45,14 +45,17 @@ type SourceResult struct {
...
 	OSPackageNames   []string
 	InstalledVersion string
 	FixedVersion     string
-	RegularVulns     []VulnResult
-	HiddenVulns      []VulnResult
-	LayerDetail      PackageContainerInfo
-	VulnCount        VulnCount
+	// RegularVulns includes all vulnerabilities intended for display to users
+	RegularVulns []VulnResult
+	// HiddenVulns includes vulnerabilities that are filtered out for users, such as those deemed unimportant or uncalled.
```
G-Rath, line 55: I would suggest a phrasing more like this:

```suggestion
	// RegularVulns holds all the vulnerabilities that should be displayed to users
	RegularVulns []VulnResult
	// HiddenVulns holds all the vulnerabilities that should not be displayed to users, such as those deemed unimportant or uncalled.
```

There's two main points in this:
  1. I feel like "includes" is technically an open word, i.e. it could have _other_ vulns too; whereas "holds" feels more closed 
  2. My understanding of these is that they're effectively meant to be the opposite of each other, which is better implied if the description uses similar phrasing

(this can be pretty subjective though, so happy for both or either of these suggestions to be ignored if others disagree 🙂)

G-Rath (approved):

internal/output/output_result.go
hogo6002: [Updated](https://github.com/google/osv-scanner/pull/1639/commits/07ecabf874ea46ee3d0f881e17639ce037be9ba5#diff-8aab87430599a256552970809bea8790d62e02122ceb19293852fcce6214ebb4), thanks!

another-rex (approved):

PR #1647, opened 2025-02-19
feat(updater): use version comparison workarounds in Maven updater

cuixq:
This PR applies the Maven version comparison workarounds in the updater. To make the function reusable, moves the version comparison to `internal/utility/maven`.

Also considering that all version strings are parsed to semver first for comparison, this PR pre-parse the versions and store them in a map.

### DIFF
--- a/internal/remediation/override.go
+++ b/internal/remediation/override.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"slices"
-	"strings"
 
 	"deps.dev/util/resolve"
 	"deps.dev/util/resolve/dep"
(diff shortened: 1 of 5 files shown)


michaelkedar (approved):
Hmm I'm also going to need to copy this into osv-scalibr.
Maybe we should thing about adding this to the deps.dev repo?

cuixq:
I am happy with keeping it here - this sounds like our specific requirements when handling updates. 

does this PR bring any issue for you to move to scalibr?

michaelkedar:
> I am happy with keeping it here - this sounds like our specific requirements when handling updates.
> 
> does this PR bring any issue for you to move to scalibr?

This PR is fine - I just was wondering if we had an opportunity to prevent the duplicated code that'll end up in scalibr.

cuixq:
I see - I think it's a good idea to prevent the duplicated code (especially the Maven ones..) - let me open an issue to track this.

PR #1662, opened 2025-02-25
feat(javareach): Supported nested JARs.

oliverchang:
- Support loading of dependencies from nested JAR files (e.g. BOOT-INF for Spring Boot applications)
- Add flag to enable profiler for debugging performance issues.

### DIFF
--- a/experimental/javareach/jar.go
+++ b/experimental/javareach/jar.go
@@ -257,6 +257,7 @@ func GetMainClasses(manifest io.Reader) ([]string, error) {
 	// Extract the Main-Class specified in MANIFEST.MF:
 	// https://docs.oracle.com/javase/tutorial/deployment/jar/appman.html
 	const mainClass = "Main-Class:"
+	// Spring Boot specific metadata.
 	const startClass = "Start-Class:"
 	markers := []string{mainClass, startClass}
 
--- a/experimental/javareach/jar.go
+++ b/experimental/javareach/jar.go
@@ -253,22 +253,33 @@ func (f *DefaultPackageFinder) Classes(artifact string) ([]string, error) {
 }
 
 // GetMainClass extracts the main class name from the MANIFEST.MF file in a .jar.
-func GetMainClass(manifest io.Reader) (string, error) {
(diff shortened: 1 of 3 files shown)


experimental/javareach/jar.go
```diff
@@ -158,22 +253,33 @@ func (f *DefaultPackageFinder) Classes(artifact string) ([]string, error) {
 }
 
 // GetMainClass extracts the main class name from the MANIFEST.MF file in a .jar.
-func GetMainClass(manifest io.Reader) (string, error) {
+func GetMainClasses(manifest io.Reader) ([]string, error) {
 	// Extract the Main-Class specified in MANIFEST.MF:
 	// https://docs.oracle.com/javase/tutorial/deployment/jar/appman.html
 	const mainClass = "Main-Class:"
+	const startClass = "Start-Class:"
```
cuixq, line 260: is there documentation about this? the above docs does not say anything about `Start-Class`

experimental/javareach/reachable.go
```diff
@@ -63,22 +89,89 @@ func (r *ReachabilityEnumerator) EnumerateReachability(roots []*ClassFile) (*Rea
...
+		}
+		r.loadedJARs[jarPath] = zipr
 	}
-	classFile, err := os.Open(classFilepath)
+
+	zipr := r.loadedJARs[jarPath]
+	class, err := zipr.Open(className + ".class")
 	if err != nil {
+		if os.IsNotExist(err) {
+			// class not found in this .jar. not an error.
```
cuixq, line 116: is this the expected behaviour?

cuixq (approved):

experimental/javareach/jar.go
oliverchang: I can't find any public documentation, as it seems to be an implementation detail of spring boot. i'll add a comment.

experimental/javareach/reachable.go
oliverchang: yes, this is just needed for searching through classpaths for the relevant class. if it doesn't exist, we move on to searching the next one.

PR #1666, opened 2025-02-26
feat: replace custom `reporter` package with `slog`

G-Rath:
Our custom `reporter` package has served us well, but it's annoying to have to pass it through to every function especially when you're using the libraries, so this switches us over to using `slog` which uses the global default logger - the notable downside with this is as a global tests that depend on logging output cannot be run in parallel, but that shouldn't be a big deal for downstream consumers as they should be depending on the structured data returned by package functions (i.e. our packages only log information that could be useful for a human to debug stuff).

This downside does impact our CLI test suite since that currently _does_ depend on that logging output as part of feedback - for now I've just "fixed" this by removing `t.Parallel` from our CLI tests, which makes them take a lot longer but should make this landable straightaway, after which we'll be free to chop and change what we like since we'll have gotten rid of the public `reporter` package and log output is not covered by semver.

I've already got a few ideas following on from this on how we might be able to speed our CLI tests back up, starting with attempting to chop the tests up into actual small packages, since Go supports running package tests in parallel regardless of their dependency on globals.

---

I do still need to give this a manual review, but I believe the primary thing left before this could be landed (warts or not) is to ensure when we're outputting JSON all other output goes to `stderr`


cmd/osv-scanner/fix/output.go
```diff
@@ -62,63 +62,38 @@ type errorOutput struct {
...
-func (r *outputReporter) PrintResult(*models.VulnerabilityResults) error {
-	panic("not implemented")
+	return outputText(opts.Stdout, outputResult)
 }
 
-func outputText(w io.Writer, out fixOutput) error {
+func outputText(_ io.Writer, out fixOutput) error {
 	if len(out.Errors) > 0 {
-		fmt.Fprintf(w, "WARNING: encountered %d errors during dependency resolution:\n", len(out.Errors))
+		slog.Warn(fmt.Sprintf("WARNING: encountered %d errors during dependency resolution:\n", len(out.Errors)))
```
another-rex, line 75: I don't think we need any of the new lines now right? Though we might want to do that pass in a followup PR.

another-rex: Also, does this need Sprintf? I'm pretty sure we can just add args directly. This way slog can separate it out later as well if needed, rather than having an opaque string.

G-Rath: I'm not sure what you mean by "the new lines", but re needing `Sprintf`, yes I believe we should be able to do that but I'm keen to do it as a follow-up because this PR is already quite large and it shouldn't be a breaking change to action that whereas it is a breaking change to change our public function signatures.

fwiw, my current thinking is that this change overall further emphasises that we've really got two things: stand-alone libraries, and CLIs that happen to use those libraries and live in the same repo; the libraries should always be using plain structure-based `slog` for their logging, whereas the CLIs are where we handle nice output (and potentially even should almost never use `slog`?).

What I'm not immediately sure of is if our current implementation within some of the libraries work well with that, and how we want to handle that effectively e.g. the `internal/config` package outputs a "human" warning when there's duplicates - since that's actually a library, that should be a structured `slog` call which somewhere in the CLI we turn into the current "human" output.

Related, this is also why I've not bothered doing anything about prefixes - we've got a few different variations of how we structure logs for warnings and errors, such as with `cmd/osv-scanner/fix` here where they all seem to get prefixed with `WARNING:`, while elsewhere we use `warning`, and etc; now with `slog` that is reflected properly by the log level which will get included in the output in some cases already.

tl;dr my focus for this PR is to get the breaking change(s) done, so we can then deal with further improvements as quickly or slowly as we like without typing up other stuff

another-rex: That makes sense, by "the new lines" I mean the `\n` at the end of all of these sprintf's

I think those can be removed as part of this PR, as every log entry has them and we have to remove all of them anyway at some point.

G-Rath: > by "the new lines" I mean the \n at the end of all of these sprintf's

Right - are you meaning as in remove them from the strings, and handle them in the logging?

Reviewing the newline situation is one of the next things on my list after I've got it working, as it doesn't look like things are consistent and I'm not sure what the convention should be e.g. I'm pretty sure some of `osv-scalibr`s logs self-endline, but "Starting filesystem walk for root:" (or maybe its the "End status":) one doesn't which is why we've got a bunch of "<removed><removed><removed>..." lines.

So by extension I'm not sure if we can "just" stick a newline at the end of everything - but then again, it might actually be enough to just ensure there's always at least one newline at the end? Either way, this is why I'm ignoring it until after I've "gotten it working"😄

G-Rath:
... and there we go: completely moved off the public `reporter` package without any changes to the snapshots or cli behaviour 🎉

There are definitely ways that this can (and should) be improved which I've already got some ideas for that I'll start working on next, but I think I'll do _all_ of them as follow-ups since this should be the primarily (and hopefully only) breaking change that needs to be included in v2 specifically, and this PR is already very big.

Currently on my list:
  - review newline handling (though I think actually we should maybe leave newlines to the caller, otherwise we'll always have to ensure everything is single-lined...)
  - review logs in regard to having their level manually prefixed (we shouldn't do that I think in most cases)
  - review if it's appropriate to be using `slog` in `cmd` packages (it might be a bit nicer in some of those to just output?)
  - cleanup the `clilogger` "global" functions (I'd like to avoid the multiple doubled up functions)
  - cleanup `internal/reporter/verbosity.go` (we don't need the aliased type now)
  - consider renaming and merging `clilogger`, `output`, and `reporter` packages
    - I think they could be a single package now? though `output` is quite large, but `reporter` can definitely be merged with one of the other two
    - we could prefix all three with `cli` or `cmd` to group them, or maybe it'd be enough to just move them to `cmd/internal`?
  - (otherwise) see if we can avoid having to access `clilogger.HasErrored` in `output` package
  - look into switching the library packages into outputting actual structured logs, and then having the CLI side of things handle turning said structured records into "human" output lines
  - consider switching from "cli" to "cmd" (to match root level namespace), and likewise moving package into `cmd/internal` or similar
  - consider adding a `verbosity` flag to the other commands
  - consider if we want to introduce our own internal `logger` package, similar to `osv-scalibr` (it might be useful for managing our output vs external package output - by default, its global logger would be global `slog` one)

@another-rex let me know if you've got any thoughts on the above, or have anything to add to that list

G-Rath:
Also, this should be landed after https://github.com/google/osv-scanner/pull/1686 I think

cmd/osv-scanner/fix/output.go
another-rex: > Right - are you meaning as in remove them from the strings, and handle them in the logging?

No as in each log entry from slog already comes with a new line right? Then we don't also need to stick a newline at the end of the string, causing double new lines.

G-Rath: > No as in each log entry from slog already comes with a new line right?

hmm I didn't think it did, but easily double checked

another-rex (approved):
Looking great! It looks like the tests "only" took 4 minutes on ubuntu? That's surprisingly fast all things considered.

cmd/osv-scanner/fix/output.go
G-Rath: ok so it turns out the answer is a mix of both, sort of - `slog` itself doesn't do anything, the output is the responsibility of the logger.

In the case of `TextHandler`, it _escapes_ messages so that they're compatible with JSON, and outputs each log on its own line:

image

https://go.dev/play/p/xReNbtzltWX

So yes we should remove the newlines from our `slog` calls, and then add them back in our custom logger, but also no it does not result in a double newline 😄

(this also means `osv-scalibr` should not include newlines in its logs, which I believe it is already doing in most cases)

another-rex: Thanks!

spencerschrock:
Browsing patch notes as I make the migration to v2 and saw this PR.

> the notable downside with this is as a global tests that depend on logging output cannot be run in parallel

FYI @G-Rath, [a proposal ](https://www.github.com/golang/go/issues/59928)(marked as likely accept) could help with this in a future Go version.

G-Rath:
@spencerschrock unfortunately I don't think that proposal will help - if I understand correctly, it's about the output from tests being printed which can get messy especially when tests are run in parallel, whereas our issue is that we're using a global and `t.Parallel` works by using goroutines

PR #1668, opened 2025-02-26
feat: move transitive scanning flags from experimental

cuixq:
This PR moves a few transitive scanning flags from experimental and aligns with `fix`:
 - `experimental-no-resolve` -> `no-resolve`
 - `experimental-resolution-data-source` -> `data-source`
 - `experimental-maven-registry` -> `maven-registry`

### DIFF
--- a/cmd/osv-scanner/fix/main.go
+++ b/cmd/osv-scanner/fix/main.go
@@ -58,8 +58,8 @@ type osvFixOptions struct {
 func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 	return &cli.Command{
 		Name:        "fix",
-		Usage:       "scans a manifest and/or lockfile for vulnerabilities and suggests changes for remediating them",
-		Description: "scans a manifest and/or lockfile for vulnerabilities and suggests changes for remediating them",
+		Usage:       "[EXPERIMENTAL] scans a manifest and/or lockfile for vulnerabilities and suggests changes for remediating them",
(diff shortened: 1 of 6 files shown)


cmd/osv-scanner/fix/main.go
```diff
@@ -58,8 +58,8 @@ type osvFixOptions struct {
 func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 	return &cli.Command{
 		Name:        "fix",
-		Usage:       "[EXPERIMENTAL] scans a manifest and/or lockfile for vulnerabilities and suggests changes for remediating them",
-		Description: "[EXPERIMENTAL] scans a manifest and/or lockfile for vulnerabilities and suggests changes for remediating them",
+		Usage:       "scans a manifest and/or lockfile for vulnerabilities and suggests changes for remediating them",
```
hogo6002, line 61: Do we want to move guided remediation out of experimental?

cuixq: yes I would like to - added @michaelkedar to review.

cuixq: actually I am making another CL to make non-interactive mode the default and will move this over there.

cmd/osv-scanner/scan/source/main.go
```diff
@@ -55,7 +55,7 @@ var projectScanFlags = []cli.Flag{
 
 var projectScanExperimentalFlags = []cli.Flag{
```
hogo6002, line 56: it seems we don't need this variable anymore. We can just move these two into the `projectScanFlags`.

hogo6002 (approved):

PR #1670, opened 2025-02-26
feat: make non-interactive the default mode of guided remediation

cuixq:
This PR makes non-interactive the default mode for guided remediation:
 - adds `--interactive` flag for the interactive mode;
 - deprecates `--non-interactive` flag;
 - updates tests and docs accordingly.

Also moves guided remediation from experimental.


cmd/osv-scanner/fix/main.go
```diff
@@ -89,12 +89,22 @@ func Command(stdout, stderr io.Writer, r *reporter.Reporter) *cli.Command {
 				Name:  "maven-registry",
 				Usage: "URL of the default Maven registry to fetch metadata",
 			},
-
 			&cli.BoolFlag{
 				Name:  "non-interactive",
-				Usage: "run in the non-interactive mode",
+				Usage: "[DEPRECATED] run in the non-interactive mode",
 				Value: !term.IsTerminal(int(os.Stdin.Fd())), // Default to non-interactive if not being run in a terminal
```
michaelkedar, line 95: This flag doesn't actually do anything now, does it?
Maybe make it `Hidden`, and set the `Value` to true.

docs/guided-remediation.md
michaelkedar, line 1: Maybe we should add `--interactive` to the Remediation Flags section?

michaelkedar (approved):

Issue #1652, opened 2025-02-21
Add Ubuntu JSON tests


hogo6002:
Same as https://github.com/google/osv-scanner/pull/1604

PR #1678, opened 2025-02-27
feat: add test for Ubuntu images

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1652


internal/image/fixtures/test-ubuntu.Dockerfile
```diff
@@ -0,0 +1 @@
+FROM ubuntu:22.04@sha256:ed1544e454989078f5dec1bfdabd8c5cc9c48e0705d07b678ab6ae3fb61952d2
```
another-rex, line 1: Can we pick a slim image to reduce the number of packages?

hogo6002: I don't see ubuntu has official slim images, but this one is not big, only 28mb similar size to Debian slim.

another-rex (approved):

PR #1691, opened 2025-03-04
fix(javareach): parse `/META-INF/services/` correctly

hogo6002:
Parse /META-INF/services/ correctly by handling cases like files defining `class = org.jline.terminal.impl.exec.ExecTerminalProvider` and `sun.io.StandardCodec # Standard codecs for the platform`

### DIFF
--- a/experimental/javareach/cmd/reachable/main.go
+++ b/experimental/javareach/cmd/reachable/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/fs"
 	"log/slog"
 	"maps"
 	"os"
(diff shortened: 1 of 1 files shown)


oliverchang:
> Skips directorates within `/META-INF/services/ `.
> 
> `/META-INF/services/` may contain classes within subdirectories (e.g. `/META-INF/services/org/jline/terminal/provider/exec`). These classes will also fail to be located by findClass() later, so just skips them here.

What is inside `services/org/jline/terminal/provider/exec`? Is it a text file? Or .class files?

hogo6002:
> > Skips directorates within `/META-INF/services/ `.
> > `/META-INF/services/` may contain classes within subdirectories (e.g. `/META-INF/services/org/jline/terminal/provider/exec`). These classes will also fail to be located by findClass() later, so just skips them here.
> 
> What is inside `services/org/jline/terminal/provider/exec`? Is it a text file? Or .class files?

It's a text file, but inside defines `class = org.jline.terminal.impl.exec.ExecTerminalProvider`

oliverchang:
> > > Skips directorates within `/META-INF/services/ `.
> > > `/META-INF/services/` may contain classes within subdirectories (e.g. `/META-INF/services/org/jline/terminal/provider/exec`). These classes will also fail to be located by findClass() later, so just skips them here.
> > 
> > 
> > What is inside `services/org/jline/terminal/provider/exec`? Is it a text file? Or .class files?
> 
> It's a text file, but inside defines `class = org.jline.terminal.impl.exec.ExecTerminalProvider`

In that case, the correct behaviour might be to parse this file instead. 

i.e. we do a https://pkg.go.dev/path/filepath#Walk on META-INF/services instead, and handle both the case of a top level file and nested files like these (which seems to have a different format).

However, while I can find documentation on what's already handled (https://docs.oracle.com/javase/8/docs/technotes/guides/sound/programmer_guide/chapter13.html), I can't find anything that documents cases like `services/org/jline/terminal/provider/exec`. which JAR has this? I wonder if it's some custom, non-standard extension.

oliverchang:
Side note: we probably also don't handle this case properly inside META-INF/services/foo.class:

```
sun.io.StandardCodec    # Standard codecs for the platform
```

We only handle the case when a line starts with "#", not when it is inline.

hogo6002:
> However, while I can find documentation on what's already handled (https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#Service_Provider), I can't find anything that documents cases like `services/org/jline/terminal/provider/exec`. which JAR has this? I wonder if it's some custom, non-standard extension.

I saw three JARs (flix.jar, scio-repl.jar and wallet-cli.jar) have it (`/META-INF/services/org/jline/terminal/provider/`). 
I tried to walk through the directory, but it will still be failed after due to `failed to find class class = org/jline/terminal/impl/exec/ExecTerminalProvider: class not found"`. The class just doesn't not exist under the root.

hogo6002:
Update code to handle special cases, such as `class = org.jline.terminal.impl.exec.ExecTerminalProvider` and `sun.io.StandardCodec    # Standard codecs for the platform`

oliverchang (approved):

PR #1692, opened 2025-03-04
fix: Update test action to not fail on error

another-rex:
Codecov has failed randomly quite a bit over the past year, it seems to have popped back up again. Going to turn off fail ci until it improves.


G-Rath:
It's seemed pretty stable to me...? I think I had the first failure in a long time the other day, which isn't bad across all my recent pushes...

fwiw #1689 might be because it's dependabot - could you confirm that the codecov token has been made available to dependabot first?

another-rex:
Ah I see, though I thought I also saw the same error in multiple PRs, but I might just be mistaken. Updated dependabot secrets to have the code cov tokens to see if that works.

PR #1693, opened 2025-03-04
fix(javareach): handle cases where META-INF/services classes can't be found

hogo6002:
Classes defined in META-INF/services might not exists in the JAR.

### DIFF
--- a/experimental/javareach/reachable.go
+++ b/experimental/javareach/reachable.go
@@ -59,12 +59,21 @@ func NewReachabilityEnumerator(
 	}
 }
 
-func (r *ReachabilityEnumerator) EnumerateReachabilityFromClasses(mainClasses []string) (*ReachabilityResult, error) {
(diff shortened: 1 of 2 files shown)


experimental/javareach/reachable.go
```diff
@@ -59,12 +59,21 @@ func NewReachabilityEnumerator(
 	}
 }
 
-func (r *ReachabilityEnumerator) EnumerateReachabilityFromClasses(mainClasses []string) (*ReachabilityResult, error) {
+func (r *ReachabilityEnumerator) EnumerateReachabilityFromClasses(mainClasses []string, serviceClasses []string) (*ReachabilityResult, error) {
```
oliverchang, line 62: actually I'd recommend renaming `serviceClasses` to `optionalRootClasses` or something similar to decouple the logic here a bit.

PR #1708, opened 2025-03-13
fix(javareach): maven validation and correct manifest class name parsing

hogo6002:
Fixes a bunch of Java readability issues:
- remove `com/` from stdlib prefixes
- exit program with errors if running against a non maven JAR
- handle wrapped class names in the manifest file
- replace some non-fatel error messages with debug messages to reduce noise

### DIFF
--- a/experimental/javareach/cmd/reachable/main.go
+++ b/experimental/javareach/cmd/reachable/main.go
@@ -20,6 +20,10 @@ import (
 	"github.com/google/osv-scanner/experimental/javareach"
 )
 
+const MavenDepDirPath = "META-INF/maven"
+const ManifestFilePath = "META-INF/MANIFEST.MF"
+const ServiceDirPath = "META-INF/services"
+
 // Usage:
 //
 //	go run ./cmd/reachable path/to/file.jar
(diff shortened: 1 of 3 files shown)


experimental/javareach/cmd/reachable/main.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/google/osv-scanner/experimental/javareach"
 )
 
+const MavenDepDirPath = "META-INF/maven"
```
oliverchang, line 23: nit while we're here: can we make this agnostic to platform? and use filepath.Join here?

experimental/javareach/cmd/reachable/main.go
```diff
@@ -99,6 +103,14 @@ func enumerateReachabilityForJar(jarPath string) error {
 		return err
 	}
 
+	// Reachability analysis is limited to Maven-built JARs for now.
+	// Check for the existence of the Maven metadata directory.
+	_, err = os.Stat(filepath.Join(tmpDir, MavenDepDirPath))
+	if err != nil {
+		slog.Error("reachability analysis is only supported for JARs built with Maven.")
+		return err
```
oliverchang, line 111: nit: we should return a more specific error (e.g. a new error type we define) rather than the stat error.

oliverchang (approved):

PR #1712, opened 2025-03-14
feat: make `vertical` output the default

G-Rath:
I can't find anywhere that we actually said explicitly "yes let's make this the default", but I'm pretty sure that was the plan and we didn't say _not_ to do it.

There was some discussion on #889 about this - I'm also pretty sure we were comfortable saying this isn't actually a breaking change, but since we've working on v2 it seems like a good time to change this all the same.


hogo6002:
We added a package view table output for container scanning (showing affected packages instead of vulnerabilities). This is because we don't want users to see hundreds of vulnerabilities from a single run. I feel switching to vertical output doesn't really address this issue. My thought is to keep the table output as a brief view of the results and prompt users to use the vertical/HTML output for more detailed results. What do you think @another-rex

PR #1720, opened 2025-03-18
chore: explicitly allow linting errors from `package-comments`

G-Rath:
By default `golangci-lint` ignores output from `package-comments` unless you explicitly allow it, even if you're already explicitly enabling the underlying rule, as they consider it too noisy for most codebases

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -121,6 +121,8 @@ linters-settings:
     block-size: 2
 
 issues:
+  include:
+    - EXC0015 # revive package-comments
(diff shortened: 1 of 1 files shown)


hogo6002 (approved):

cuixq:
I think the linter is doing its work - we need to add comments to these packages.

G-Rath:
@cuixq yup that's the next step - I'm going to be working on that in the background, but its currently a low priority

G-Rath:
not stale, just have not gotten around to it yet 😅

PR #1725, opened 2025-03-20
refactor: Maintain consistent warning messages with slog.Warn

Abdorithm:
Replace `fmt.Fprintf(stderr, ...)` with `slog.Warn(fmt.Sprintf(...))` in the `warnIfCommandAmbiguous` function and modify tests accordingly. This change:
- Addresses the existing TODO
- Makes warning messages consistent with the rest of the codebase
- Updates snapshots, the warning now appears in stdout instead of stderr

### DIFF
--- a/cmd/osv-scanner/__snapshots__/main_test.snap
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap
@@ -16,11 +16,11 @@
 ---
 
 [Test_insertDefaultCommand - 5]
+Warning: `scan` exists as both a subcommand of OSV-Scanner and as a file on the filesystem. `scan` is assumed to be a subcommand here. If you intended for `scan` to be an argument to `default`, you must specify `default scan` in your command line.
 
 ---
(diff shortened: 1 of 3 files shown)


another-rex:
I'm not 100% sure we want slog for CLI output, but I suppose using slog right now is better than what we are currently doing. @G-Rath thoughts?

G-Rath:
the first part of your comment is exactly what I was just typing out - it's why the todo has a "maybe..." in it, though I'm not sure about the "better than what we're currently doing".

My take is that this is just an attempt to action the todo, rather than fixing an external issue i.e. is anyone actually complaining that this message is going to stderr instead of stdout? (because no one should be complaining that the output isn't going through `slog` since this lives in the CLI code which is private, so anyone calling it programmatically from Go is breaking our api contracts anyway)

another-rex:
Closing this for now as it seems like we don't want to have this output with slog.

Issue #1564, opened 2025-02-03
Generated CycloneDX file is invalid


anthonyharrison:
Using osv-scanner to scan a SBOM and produce a CycloneDX 1.5 report produces an invalid file.

File has been validated using the [CycloneDX online validator](https://cyclonedx.github.io/cyclonedx-web-tool/validate)

Issue appears to be with repeated references to the same component in the vulnerability 

Sample of invalid entry

```json
      "description": "libwebp: OOB write in BuildHuffmanTable",
      "detail": "Heap buffer overflow in libwebp allow a remote attacker to perform an out of bounds memory write via a crafted HTML page. ",
      "advisories": [
        {
          "url": "https://nvd.nist.gov/vuln/detail/CVE-2023-4863"
        }
      ],
      "published": "2023-09-12T15:30:20Z",
      "updated": "2024-08-07T19:29:17Z",
      "credits": {
        "organizations": []
      },
      "affects": [
        {
          "ref": "pkg:cargo/libwebp-sys2"
        },
        {
          "ref": "pkg:cargo/libwebp-sys"
        },
        {
          "ref": "pkg:npm/electron"
        },
        {
          "ref": "pkg:npm/electron"
        },
        {
          "ref": "pkg:npm/electron"
        },
        {
          "ref": "pkg:npm/electron"
        },
        {
          "ref": "pkg:npm/electron"
        },
        {
          "ref": "pkg:nuget/SkiaSharp"
        },
        {
          "ref": "pkg:golang/github.com/chai2010/webp"
        },
        {
          "ref": "pkg:pypi/pillow"
        },
        {
          "ref": "pkg:cargo/webp"
        },
        {
          "ref": "pkg:nuget/magick.net-q16-anycpu"
        },
        {
          "ref": "pkg:nuget/magick.net-q16-hdri-anycpu"
        },
        {
          "ref": "pkg:nuget/magick.net-q16-x64"
        },
        {
          "ref": "pkg:nuget/magick.net-q8-anycpu"
        },
        {
          "ref": "pkg:nuget/magick.net-q8-openmp-x64"
        },
        {
          "ref": "pkg:nuget/magick.net-q8-x64"
        }
      ]
```

another-rex:
Seems like a fix would be to de-dup the output results in the cycloneDX reporter.

PR #1726, opened 2025-03-20
fix: De-dup the output results in the cycloneDX reporter (#1564)

kuscar:
As previously discovered by @another-rex and @anthonyharrison, a file is deemed invalid if multiple identical references are present. This has been fixed by ensuring only one instance of each reference is kept.


another-rex:
Thank you! Can you run `./script/run_tests.sh` with the environment `UPDATE_SNAPS=true`? This will update the snapshots with this new change.

another-rex (approved):

another-rex:
Yep that's correct, the error was from an unrelated snapshot change, updating the branch fixed it.

PR #1737, opened 2025-03-24
refactor: move cli building into a dedicated internal package to enable relocating tests

G-Rath:
As we've switched to using a global logger, we can no longer run tests that involve checking that output in parallel which greatly impacts our cmd tests since that's all about what we output - however, `go test` always runs tests across packages in parallel, meaning we can improve our test suite performance by refactoring so that our cmd tests are spread across multiple (sub)packages, which this does.

The key change to enabling this is moving the build-and-run of `cli.App` from the `main` package in `cmd/osv-scanner` into a dedicated internal package, since that results in an internally public function we can import elsewhere from other packages including tests under the guise of testing the specific command each subpackage represents.

Overall for me locally this takes `TEST_ACCEPTANCE=true go ./cmd/osv-scanner/... -count 1` from ~6 minutes to ~3 minutes, and also should hopefully mean caching happens slightly more frequently (though it's still unlikely to be great).

While I still have some reservations about this, overall I feel like its a good move due to a couple of small bonus positives such as now having the tests in their own `_test` packages and being more explicitly tied with the public `Command` of each package.

I think my remaining reservations are around how unlike other unitish tests this blurs the line on what is actually being tested since each command package is really still testing the full `cli.App` with all commands mounted, meaning behaviour unrelated to the command/package can be inadvertently tested - that's not the end of the world, but ideally it would be nice if we could untangle things some more.

We already see that a bit with the remaining tests in `cmd/osv-scanner/main_test.go` which are arguably testing `cmd/osv-scanner/internal/cmd#Run` and so could live there, but they're calling `testcmd.Run` - we could alternatively call `cmd/osv-scanner#run` but then that'd ideally mean replicating some of the logic in `testcmd`...

A change we could do that might help with this is switch to passing `cmd.Run` the array of commands to mount, meaning each package could be more explicitly based on testing their particular thing and that would also mean improve cache reuse a little (since right now a change to any command means the tests for all command packages are technically changed), but then we'd arguably be testing less and in theory increase the chance of missing bugs related to the interactions between different commands (though that might be a thing at all 🤷).

Ultimately, this does improve the test suite performance and its all internal stuff so it's probably fine (and maybe best) to aim at landing it as-is to start with, and then we can follow-up with changes later as our feelings come out in the wash.

Finally, there is some further refactors with the test suite that I'm going to look into as a follow-up such as possibly moving the fixtures into the subpackages and deduplicating the methods in `testcmd` as I think we can get away with just the `testcmd.Run` method with a few tweaks.


G-Rath:
I think this is fine to review even though I do need to clean up the version and commit variables - they should be very straightforward though as they're just two values 🤷

Feel free to give opinions on package naming and if we want to try landing this as multiple PRs - we could split up the "introduce `internal/cmd`", "move the tests into subpackges", and "rename subpackage files and tests from `main` to `command`"

G-Rath:
I think this should be good to go - @another-rex and I caught up offline and decided on a few follow-up changes after this:
  - prefixing cli based packages in `internal` with `cmd` (i.e `cmdoutput`, `cmdreporter, etc)
  - move the `scan source` specific tests into that package, and update the suite to explicitly specify the subcommand
  - move the `commit` and `date` variables back into `cmd/osv-scanner/main.go`
  - refactor `internal/cmd.Run` to take an array of commands, and have the command tests pass in just their specific command

@another-rex would you mind looking into the CI failures for me? I can't reproduce them locally no matter what I try, and I feel like I've ran into this before with these specific image tests

internal/testutility/normalize.go
```diff
@@ -61,6 +61,19 @@ func normalizeRootDirectory(t *testing.T, str string) string {
 
 	cwd = normalizeFilePaths(t, cwd)
 
+	// todo: currently this is a workaround for the osv-scanner/cmd/scan tests to
+	//  avoid having to move the fixtures and make other changes right now but
+	//  we should be able to remove this in future, e.g. by moving the fixtures
+	if strings.Contains(cwd, "osv-scanner/cmd") {
```
another-rex, line 67: nit: add an example for what this does to make this easier to read.

another-rex (approved):
LGTM, thanks!

another-rex:
Ah I know the issue with the tests, you need to rebuild the image, but also remove the cached image from your docker cache so it will run `pip install` again.

It is a bit annoying that the dependencies installed might change, but I'm not sure what can be done about it unless we pin each dep by hash instead of version in the requirements.txt (which we probably should do)

G-Rath:
ah nuts, apparently Windows has been failing for a completely different (and explosive...) reason - for some reason the tests are timing out there 😕

PR #1740, opened 2025-03-24
ci: ensure that `go.sum` is tidy

G-Rath:
This should ensure this never slips through again 😄


G-Rath:
This will fail until #1741 is landed

another-rex (commented):
Good idea!

another-rex (approved):

PR #1747, opened 2025-03-25
fix: osv-reporter should be print a warning instead of an error for these messages

another-rex:
Printing an error results in a 127 return code, when we want to return 0 when these warnings happen. This fixes a unintented behaviour change made as part of the slog refactor.

Fixes https://github.com/google/osv-scanner-action/issues/65


hogo6002 (approved):

G-Rath (approved):
whoops 😅

PR #1748, opened 2025-03-25
chore: add `--force` and `--no-cache` flags to script for building test images

G-Rath:
This should make it easier to ensure that you've got the same images as in CI - it'd be good if someone on macOS could test these for me, as opt-parsing has a bit patchy between OSs

Usages:

```
scripts/build_test_images.sh --force
scripts/build_test_images.sh --no-cache
scripts/build_test_images.sh --force --no-cache
```

### DIFF
--- a/scripts/build_test_images.sh
+++ b/scripts/build_test_images.sh
@@ -2,12 +2,16 @@
 
 set -e
 
+# this is inverted because docker build takes "--no-cache",
+# so "false" here means that the layer cache will be used
+no_layer_cache=false
+
 function build_docker_image_fixture {
   image_name="$1"
   output_tar="internal/image/fixtures/$image_name.tar"
 
   if [ ! -f "$output_tar" ]; then
-    docker build internal/image/fixtures/ -f "internal/image/fixtures/$image_name.Dockerfile" -t "osv-scanner/$image_name:latest"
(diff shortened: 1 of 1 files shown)


scripts/build_test_images.sh
```diff
@@ -16,8 +20,31 @@ function build_docker_image_fixture {
...
+      ;;
+  esac
+done
+
 for dockerfile in internal/image/fixtures/*.Dockerfile; do
   image_name=$(basename "$dockerfile" .Dockerfile)
 
+  if [ "$force" = true ]; then
+    echo "Removing existing tar file for $image_name..."
+    rm -f "internal/image/fixtures/$image_name.tar"
```
another-rex, line 46: Is -f needed here?

another-rex (commented):
Hmm does this script check whether the caller is in the right directory? Want to avoid deleting the wrong files.

scripts/build_test_images.sh
G-Rath: na it shouldn't be

G-Rath:
> Hmm does this script check whether the caller is in the right directory

it doesn't but I feel like the paths are specific enough that it shouldn't be easy to have the mistakenly remove something

another-rex (approved):

another-rex:
@hogo6002 or @cuixq  Can you try running the script on Mac to see if it works correctly? Thanks!

PR #1757, opened 2025-03-27
ci: cache test images for reuse across runs

G-Rath:
While it only tends to take 2-3 minutes, they completely block our three test jobs and in theory it should be very straightforward to cache our test images as (again, in theory) they should be very stable.

Right now I'm just redownloading the tars and skipping the build entirely, though we could also possibly try re-importing them so that we always do a build but they're available as cached layers, which _might_ be a better middleground

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -91,22 +91,29 @@ jobs:
       contents: read # to fetch code (actions/checkout)
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


G-Rath:
@another-rex what do you think about this? It's actually technically not working on this pull request as even though its reporting it's pushed up the cache, it never finds it when doing the restore, but I think that's because this is coming from a fork - I'm curious to see what happens if we just land it

PR #1760, opened 2025-03-27
test: do everything through `testcmd.RunAndMatchSnapshots` including JSON replacements

G-Rath:
The initial motivation for doing this is that we're wanting to explore passing in the specific commands under test for better caching and separation, which is easier to do if we've got less public methods, and I decided to pull everything else through as it felt to be particularly specific to the cmd tests.


cmd/osv-scanner/internal/testcmd/run.go
```diff
@@ -20,3 +23,36 @@ func Run(t *testing.T, tc Case) (string, string) {
 
 	return stdout.String(), stderr.String()
 }
+
+func Run(t *testing.T, tc Case) {
```
G-Rath, line 27: should we expand this name to try make it more assert-y, e.g. `testcmd.RunAndMatchSnapshots`?

cmd/osv-scanner/internal/testcmd/run.go
```diff
@@ -20,3 +23,36 @@ func Run(t *testing.T, tc Case) (string, string) {
...
+	t.Helper()
+
+	stdout, stderr := run(t, tc)
+
+	if tc.isOutputtingJSON() {
```
G-Rath, line 32: do we want to factor in if there are any JSON replacement rules? I think doing so technically would be more performant, as `normalizeJSON` always calls `json.Indent` even if there are no rules, though that's probably not even measurable....

another-rex: I think that's fine to always call json.Indent, it'll make the diffs easier to read as well (maybe? I don't quite remember how the diffs work for JSON).

cmd/osv-scanner/internal/testcmd/run.go
another-rex: Yep, I like more descriptive names.

another-rex (approved):
Good idea!

Issue #1758, opened 2025-03-27
Improve OS status file scanning output


another-rex:
We should use the new summarised view. 

Rename PACKAGE column to SOURCE PACKAGE
Add column of BINARY PACKAGE

Related #1293

mnaser:
Thanks, happy to do follow up and testing here for this.

PR #1761, opened 2025-03-27
feat: improve OS status scanning output

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1758

table output: 
image

vertical output:
image

HTML output:
image

### DIFF
--- a/internal/output/output_result.go
+++ b/internal/output/output_result.go
@@ -43,6 +43,7 @@ type EcosystemResult struct {
 // SourceResult represents the vulnerability scanning results for a source file.
 type SourceResult struct {
 	Name                   string
+	Type                   models.SourceType
 	Ecosystem              string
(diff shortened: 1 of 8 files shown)


internal/output/output_result.go
```diff
@@ -831,3 +832,19 @@ func getInstalledVersionOrCommit(pkg PackageResult) string {
 
 	return result
 }
+
+func isOSResult(sourceName string) bool {
+	return strings.Split(sourceName, ":")[0] == OSSourcePrefix
```
another-rex, line 837: I'm not a fan of doing a string split here. Can we pass through the enum and match on that directly?

hogo6002: updated

another-rex (approved):

Issue #1745, opened 2025-03-25
Unable to parse cyclonedx 1.6 sbom when filename is not `bom.json`


ketan:
Unsure if this is related to #1744. But parsing this file also gives a secondary error when the filename is not `bom.json`. Arguably the file contains enough hints on what parser can be used. In particular the `bomFormat` and `specVersion` should be sufficient information?

```shell

$ ~/.gradle/osv-scanner --version
osv-scanner version: 2.0.0
commit: 0e986b49c4e7ee5aa545531c4a8908455f8a9e82
built at: 2025-03-17T01:50:11Z


$ ~/.gradle/osv-scanner scan source --sbom ./foo.json
Failed to parse SBOM "/tmp/foo.json" with error: could not determine extractor suitable to this file
could not determine extractor suitable to this file
```

```json
{
  "bomFormat" : "CycloneDX",
  "specVersion" : "1.6",
  "version" : 1,
  "metadata" : {
    "timestamp" : "2025-03-24T18:02:51Z",
    "tools" : {
      "components" : [
        {
          "type" : "application",
          "author" : "CycloneDX",
          "name" : "cyclonedx-gradle-plugin",
          "version" : "2.2.0"
        }
      ],
      "services" : [ ]
    },
    "component" : {
      "type" : "library",
      "bom-ref" : "pkg:maven/com.example.group/fooBar@1.2.3?project_path=%3A",
      "group" : "com.example.group",
      "name" : "fooBar",
      "version" : "1.2.3",
      "purl" : "pkg:maven/com.example.group/fooBar@1.2.3?project_path=%3A",
      "modified" : false,
      "externalReferences" : [
        {
          "type" : "vcs",
          "url" : "https://github.com/znsio/specmatic-gradle-plugin.git"
        }
      ]
    },
    "licenses" : [ ]
  },
  "dependencies" : [
    {
      "ref" : "pkg:maven/com.example.group/fooBar@1.2.3?project_path=%3A",
      "dependsOn" : [ ]
    }
  ]
}
```

ketan:
Closing - this is documented behavior - https://google.github.io/osv-scanner/usage/scan-source#specify-sbom, but the error message could have been more useful.

PR #1762, opened 2025-03-28
fix: include a hint when sbom extractor cannot be found about the file name being important

### LINKED ISSUE #1746: Improve error message when SBOM name is not in the correct format
Related #1745

G-Rath:
I'm open to suggestions about the actual message - I'm a little on the fence about including a list of the file types or external links as they could change (and CycloneDX also has `bom.json` and `bom.xml` meaning it's technically not as easy as saying "it must end with these file types"), but maybe that's worth doing...

I think ideally it would be good if `osv-scalibr` could indicate the reason why the extractor can't be found, but I don't think that's compatible with the current API and also somewhat niche.

Resolves #1746
Relates to #1745

### DIFF
--- a/cmd/osv-scanner/scan/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/__snapshots__/command_test.snap
@@ -911,7 +911,7 @@ No issues found
 ---
 
 [TestCommand/one_file_that_does_not_match_the_supported_sbom_file_names - 1]
-Failed to parse SBOM "<rootdir>/osv-scanner/fixtures/locks-many/composer.lock" with error: could not determine extractor suitable to this file
+Failed to parse SBOM "../osv-scanner/fixtures/locks-many/composer.lock" with error: could not determine extractor suitable to this file
 If you believe this is a valid SBOM, make sure the filename follows format per your SBOMs specification.
 
 ---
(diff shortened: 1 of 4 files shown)


another-rex (commented):
I think the current message is fine, though maybe we can link to the documentation about supported sbom types: https://google.github.io/osv-scanner/usage/scan-source#specify-sbom  

I'm not sure whether the message should be here though, I do have a TODO on line 79 to make this function less SBOM focused,and move the logging to the level above.

G-Rath:
yeah I've got mixed feelings about including a link like that because what if it changes in someway? though having said that, I was about to suggest we could say something like "check the documentation for more" but that's probably _worse_ because things could still change, but you'd not even have a link you could wayback-machine and similar to try and find out where to go in future...

I think the better thing could be to have an explainer of the file types, similar to the docs but that way it's bound to the version so if there are new file formats allowed in future, a particular version of the scanner would always accurately say what ones _it_ supports (and if there were new ones that we forgot to add to the list, e.g. after updating `osv-scalibr`, then that's just a hidden bonus 🤷).

I've checked what we have for the flag help text, and we just say:

> scan sbom file on this path, the sbom file name must follow the relevant spec

another-rex:
That makes sense, let's keep the current message then!

another-rex:
Huh, seems like the paths are slightly off

another-rex (approved):

PR #1763, opened 2025-03-28
feat: Made the source type enum public

another-rex:
Make source type enum public


internal/scalibrextract/ffa/ffa.go
```diff
@@ -0,0 +1 @@
+package ffa
```
hogo6002, line 1: do you want to check in this file?

hogo6002 (approved):

PR #1776, opened 2025-04-01
feat: support outputting results in `spdx` format

### LINKED ISSUE #1767: Support Outputting in both CycloneDX and SPDX SBOM formats
## Overview 

Currently we only support the CycloneDX SBOM format with our own implementation in OSV-Scanner.
This work entry is to specifically support adding the SPDX SBOM generation without changing our current CycloneDX SBOM output.

Longer term we would want to unify the two CycloneDX implementations in both SCALIBR and OSV-Scanner, but that's beyond the scope of this issue.

## Details: 

Scalibr’s SBOM outputs take in a `[]Inventory` and write the output directly. An implementation could be to attach the `[]Inventory` that we receive from the scans to the JSON output of osv-scanner, but omit them from the JSON to avoid bloating the actual JSON output. 

Our internal formatters can still pass `[]Inventory` to SCALIBR’s SPDX outputer.

> [!NOTE]
> The `Inventory[]` type will change very soon (hopefully), once https://github.com/google/osv-scalibr/pull/523 is merged. All the fields will still be present, just moved around slightly. Might want to wait till this is merged and we are migrated before implementing this.

## Tasks:

- [ ] Migrate SCALIBR SPDX outputter into a OSV-Scanner formatter
- [ ] Modify output struct and attach `[]Inventory` to it

G-Rath:
This adds support for outputting scanner results in SPDX v2.3 format, using the `osv-scalibr` library to convert the actual results in an `spdx.Document`.

Resolves #1767


internal/output/helpers_test.go
G-Rath, line 1: ideally I'd like to have a smarter helper to make this less repetitive but at this point I don't think it's worth the time as it'll be tedious to refactor and it's unlikely these will change in the near future

internal/output/spdx.go
```diff
@@ -0,0 +1,28 @@
...
+func PrintSPDXResults(vulnResult *models.VulnerabilityResults, outputWriter io.Writer) error {
+	scanResult := &scalibr.ScanResult{}
+
+	for _, source := range vulnResult.Results {
+		for _, pkg := range source.Packages {
+			scanResult.Inventories = append(scanResult.Inventories, pkg.Package.Inventory)
+		}
+	}
+
+	doc := converter.ToSPDX23(scanResult, converter.SPDXConfig{})
```
another-rex, line 22: Let's add a follow up TODO here for configuration, though that's probably some time away.
```suggestion
  // TODO(#1783): Allow user configuration
	doc := converter.ToSPDX23(scanResult, converter.SPDXConfig{})
```

internal/output/spdx_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/google/osv-scanner/v2/internal/cachedregexp"
+	"github.com/google/osv-scanner/v2/internal/output"
+	"github.com/google/osv-scanner/v2/internal/testutility"
+	"github.com/jedib0t/go-pretty/v6/text"
+)
+
+func normalizeSPDXOutput(t *testing.T, str string) string {
```
another-rex, line 13: I wonder if it's better to do the JSON matching, and then use the json normalization code.

I don't know if that's actually any better than this though, so happy to merge it as is.

another-rex (approved):

internal/output/spdx.go
G-Rath: whoops yeah good call, as I'd completely forgotten about the config myself 😅

internal/output/spdx_test.go
G-Rath: I don't think it makes much of a difference at this point, as I've already implemented it this way and it never felt like I was pushing things uphill - we can revisit if/when the output becomes more complicated

PR #1781, opened 2025-04-03
chore: upgrade `golangci-lint` to v2

G-Rath:
The new major primarily was focused on overhauling the config structure along with introducing the concept of formatters as being different from linters, though it also seems to have super charged `staticcheck` as that has a number of new findings mainly around being able to remove embedded fields.

For now I have disabled the failing checks and will look to review + tackle them in follow up PRs

Also see [the migration guide for details on the changes](https://golangci-lint.run/product/migration-guide/)


G-Rath:
Actually I just realised the "new" `staticcheck` violations are because v2 does not have "false positives" overrides, so most if not all of these new violations are from [these](https://golangci.github.io/legacy-v1-doc/usage/false-positives/)

G-Rath:
The linter check is currently failing due to https://github.com/golangci/golangci-lint/pull/5692

hogo6002 (approved):

PR #1787, opened 2025-04-07
feat(xml): keep the extra space in self-closing tags

cuixq:
Previously we don't preserve any space when marshaling a self-closing tag, however this results a few diffs if there is a space in the tag.

This PR adds `Space` to the `StartElement` to indicate whether any space is spotted, so we are able to write an extra space to minimize the diff.

Also to minimize the diff from escaping quotes in a text, this PR skips escaping the single quote for CharData.


michaelkedar (approved):
Could you also duplicate this to https://github.com/michaelkedar/xml

PR #1797, opened 2025-04-11
chore: Update scalibr to use the new inventory format.

another-rex:
This PR attempts to make the minimal number of changes required to support the new SCALIBR format. Follow up PRs will better integrate the new format into OSV-Scanner.

As part of updating scalibr, the following changes are made:
- Update to go 1.24
  - This brings in some new lints, primary one is the introduction of `t.Context()` to replace `context.Background()` for tests.
- Support extracting the actual binary name from gobinaries.
- Fix broken link in documentation


cuixq:
Can you also mention the changes of using `t.Context()` in tests in the PR description?

cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
```diff
@@ -1314,6 +1314,18 @@ Scanning local image tarball "../../../../internal/image/fixtures/test-python-fu
             "GO-2025-3563"
           ],
           "groups": 8
+        },
```
cuixq, line 1317: is this change due to using the new inventory format? or because of any fixes?

cuixq (approved):

spencerschrock:
> * Update to go 1.24

Is it still osv-scanner's goal to support the two supported Go versions (1.24 and 1.23 currently)? Does osv-scalibr (I presume the root cause of this bump) share these aspirations?

another-rex:
Since v2.0 we only support the latest Go version, as osv-scalibr also only tracks the latest version.

Issue #1794, opened 2025-04-10
License Summary do not respect package license override


jaksky:
When using configuration option for package license override like e.g.

[[PackageOverrides]]
ecosystem="PyPI"
name= "isodate"
license.override = ["BSD-3-Clause"]

the license summary does not respect those overrides and counts it against non-standard when tool called with "--licenses="BSD-3-Clause" though licence_violation for given package in resulted json output.

cuixq:
Hi @jaksky could you share more details about how you run the license scanning with overrides? 

I did some local testing and it works for me:
```
overriding license for package PyPI/isodate/x.x.x with BSD-3-Clauses
```

jaksky:
Hi @cuixq,

I am using version 2.0.0. The specific spdx version override doesn't matter - the same applies for different licences as well. 

Config mentioned above is placed to osv.toml and than command run following:
```
osv-scanner -r --format=json --licenses="UNKNOWN,BSD-3-Clause" --config=./osv.toml  .
```

In the resulting json in license summary you see still count for UNKNOWN even though you override the config in osv.toml file.

I see the same log line you mentioned but the summary license stats in json do not respect that configuration and still see UNNKNON counts.

cuixq:
is it possible to share the manifest that is being scanned? or the part of it containing `isodate`? 

it will also be helpful if you could share the part of result json that you think is not respecting the configuration.

jaksky:
@cuixq thank you! Glad that issue is replicated

cuixq:
@jaksky thanks for reporting the issue and it should be fixed with the next release!

PR #1806, opened 2025-04-15
fix: override licenses in scan results

cuixq:
https://github.com/google/osv-scanner/issues/1794

Previously the license overrides are not reflected in the scan results so the license summary is based on the licenses before overriding. This PR fixes the issue by updating the package scan result with the license overrides.

### DIFF
--- a/cmd/osv-scanner/scan/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/__snapshots__/command_test.snap
@@ -389,7 +389,9 @@ overriding license for package Packagist/league/flysystem/1.0.8 with 0BSD
 +---------+-------------------------+
 | LICENSE | NO. OF PACKAGE VERSIONS |
 +---------+-------------------------+
-| UNKNOWN |                      20 |
+| MIT     |                      14 |
+| 0BSD    |                       2 |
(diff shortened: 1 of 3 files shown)


another-rex (approved):
I didn't realize we still used scanResults after buildVulnerabilityResults.

Feel like longer term we should fix that, but this fix LGTM.

cmd/osv-scanner/scan/__snapshots__/command_test.snap
```diff
@@ -1651,7 +1655,8 @@ overriding license for package Packagist/league/flysystem/1.0.8 with 0BSD
 +---------+-------------------------+
 | LICENSE | NO. OF PACKAGE VERSIONS |
 +---------+-------------------------+
-| UNKNOWN |                      17 |
```
another-rex, line 1654: What happened here? How did 17 package versions turn to 4...

another-rex: Ah we are correctly ignoring packages now

cuixq:
Yes I also agree that we should be building the license summary based on the vulnerability results

PR #1813, opened 2025-04-22
chore: Update snapshots

another-rex:
I am a bit confused as to why the positions of existing entries keep changing positions. Need to look into that a bit more.


cuixq (approved):

G-Rath (approved):
looks like it's been moved from important to not important, which seems correct based on my brief check of the code + the advisory (https://api.osv.dev/v1/vulns/CVE-2024-2236 has `"urgency": "unimportant"` in all three `affected` entries).

My first thought would be is it possible that there's some "last advisory" wins type condition, with a non-consistent order meaning that the advisory being marked as unimportant is not always the case - to that end, here's a copy of the advisory as of today, so we can compare if/when this changes: [CVE-2024-2236.json](https://github.com/user-attachments/files/19841864/CVE-2024-2236.json)

PR #1825, opened 2025-04-24
feat(xml): store the original text in CharData

cuixq:
We can not fully rely on `escape` flag to decide whether to escape texts in CharData or not since there can be both escaped and unescaped texts. 

When decoding a token, escape sequences are replaced by the characters that they represent. To be able to write the original text, we add the slice of bytes to CharData to represent the texts before any replacement. And when encoding a token, we use the original bytes to minimize the diff.


cuixq:
Still need to test a few edge cases

michaelkedar:
We've made so many changes to the standard `encoding/xml` by now, it feels like we're trying to force it into a use case that it really was not designed for.

Maybe we should start looking into using another xml library?
I just discovered [github.com/antchfx/xmlquery](https://pkg.go.dev/github.com/antchfx/xmlquery) which seems promising

oliverchang:
> We've made so many changes to the standard `encoding/xml` by now, it feels like we're trying to force it into a use case that it really was not designed for.
> 
> Maybe we should start looking into using another xml library? I just discovered [github.com/antchfx/xmlquery](https://pkg.go.dev/github.com/antchfx/xmlquery) which seems promising

That's reasonable as a long term, but it's probably still worthwhile seeing how far we can go with forking the original `encoding/xml` in the shorter term. We should record all the changes we have to make and the edge cases we encounter, so we have a clear list of requirements once we look for a better long term.

cuixq:
I think it makes sense to look for some different approach for long term. However I am not sure if these package can meet our requirements to minimize the diff (assuming they depend on `encoding/xml`).

Considering the time requirement of the XML feature, I would prefer adding the patches first then look for the long term solution at the same time.

`github.com/antchfx/xmlquery` seems only for querying but not for writing?

michaelkedar:
> `github.com/antchfx/xmlquery` seems only for querying but not for writing?

[It does have an example of constructing and outputting xml](https://pkg.go.dev/github.com/antchfx/xmlquery#readme-create-xml-document-without-call-xml-marshal), and the Data of the nodes are editable.

But I agree, for the short term if we can get this fork to do what we need it should be fine.

cuixq:
the [outputConfiguration](https://github.com/antchfx/xmlquery/blob/master/node.go#L56) looks interesting

michaelkedar (approved):

PR #1831, opened 2025-04-26
test: only include specific command under test in `cmd` package tests

G-Rath:
Currently because all commands are included in all `cmd` tests any code change will invalid the test cache for all `cmd` packages regardless of if the change is actually impactful to them - we can avoid this by having the setup for each `cmd` package explicitly set the commands to include in the test, as the cost of a minor change to the app builder function


another-rex (approved):
Good idea!

PR #1835, opened 2025-04-26
test(scan): add license cases for when invalid and offline

G-Rath:
Among other things, this covers the fact that we currently check if we're offline _after_ we validate the licenses since we could flip that to simplify our implementation but currently doing so would not reveal that it changes our behaviour

### DIFF
--- a/cmd/osv-scanner/scan/source/command_test.go
+++ b/cmd/osv-scanner/scan/source/command_test.go
@@ -600,9 +600,29 @@ func TestCommand_Licenses(t *testing.T) {
 			Exit: 1,
 		},
 		{
-			Name: "Licenses with invalid expression",
+			Name: "Licenses with invalid licenses in flag",
+			Args: []string{"", "source", "--licenses=MIT,something-something", "./fixtures/locks-licenses/package-lock.json"},
+			Exit: 127,
+		},
+		{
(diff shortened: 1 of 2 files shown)


cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -584,10 +584,30 @@ func TestCommand_Licenses(t *testing.T) {
...
+			Exit: 1,
+		},
+		{
+			Name: "When offline licenses cannot be checked",
+			Args: []string{"", "scan", "source", "--offline", "--licenses=MIT", "--config=./fixtures/osv-scanner-empty-config.toml", "./fixtures/locks-many/package-lock.json"},
+			Exit: 1,
+		},
+		{
+			Name: "When offline licenses are still validated",
+			Args: []string{"", "scan", "source", "--licenses=MIT,something-something", "./fixtures/locks-many/package-lock.json"},
```
hogo6002, line 608: do you want to add a `offline` flag here

G-Rath: yup, apparently I forgot to finish testing the actual cases work before opening this 🤦

hogo6002 (approved):

PR #1842, opened 2025-04-28
fix(gh-action): call analysis doesn't work for go 1.24

shahar-h:
When using `osv-scanner` with a go project that contains the new go 1.24 [tool](https://tip.golang.org/doc/go1.24#tools) directive, call analysis fails with:
```log
Failed to run code analysis (govulncheck) on '/github/workspace/go.mod' because govulncheck: loading packages: err: exit status 1: stderr: go: errors parsing go.mod:
go.mod:523: unknown block type: tool
```

Since osv-scanner image is based on alpine image and latest native alpine go package is 1.23.8 apk, I installed go official release instead with latest version(1.24.2).


another-rex:
Rather than adding Go ourselves, let's just switch to using the go base image: `golang:1-alpine`. This way renovatebot will also automatically update to the latest go version and alpine version.

shahar-h:
> Rather than adding Go ourselves, let's just switch to using the go base image: `golang:1-alpine`. This way renovatebot will also automatically update to the latest go version and alpine version.

Should I use `golang:1.24.2-alpine3.21@sha256:7772cb5322baa875edd74705556d08f0eeca7b9c4b5367754ce3f2f00041ccee` to be aligned with [Dockerfile](https://github.com/google/osv-scanner/blob/main/Dockerfile#L15) and [action.dockerfile](https://github.com/google/osv-scanner/blob/main/action.dockerfile#L15)?

another-rex:
Yes please! It looks like renovatebot can also update base images to different tags, not just within the same tag, so automated updates should still work.

another-rex (approved):

PR #1844, opened 2025-04-28
feat: upgrade `urfave/cli` to v3

### LINKED ISSUE #1773: Update/Migrate to cli/v3
It looks like the cli package has released v3, https://github.com/google/osv-scanner/pull/1732 , it's unclear what exactly the new features are as they make it really hard to find on their repo. But one of the features is that arguments can now be inputted in any order, which is really nice feature for us as we want to be able to do: `osv-scanner scan image <image-name> --arg1 --arg2` which is impossible (or really hard) in v2

**Migration guide:** https://cli.urfave.org/migrate-v2-to-v3/

G-Rath:
I believe the main change (for us) is that a native `context.Context` is now used, with all the "cli" stuff living in `cli.Command`

Resolves #1773


hogo6002 (approved):

G-Rath:
hmm looks like the tests are flakey, so will need to dig into that

G-Rath:
ok so this has been interesting - it looks like the parallel logger is just "missing" so output ends up to the wrong logger, though I don't know why - maybe there's some memory address reuse going on that isn't expected..?

However when attempting to craft a small "debug" command to try and dig into that further, I managed to cause a nil pointer panic in `urfave/cli` around its flag handling when running the small crafted command at high counts (e.g. `-count 10000`), which I think I've narrowed down to being related to [this code](https://github.com/urfave/cli/blob/main/command_setup.go#L203-L211) (at the least, if I set `HideHelp` on my command, the panics seem to stop no matter how high I set the count) which makes some sense because it looks like their manipulating a global.

There's also another global (`disableSliceFlagSeparator`) being set that `-race` reports though as that is just a bool it doesn't seem like it'll be a big issue, but overall combined with what I've seen in the issues it looks like we shouldn't be treating `urfave/cli` as thread-safe - I don't actually think that is a huge deal for us, and they would hopefully be still open to trying and avoiding globals, but I wanted to write this up anyway so it was _somewhere_ 😅

another-rex (approved):

PR #1845, opened 2025-04-28
test(osv-scanner/fix): actually import command from package

G-Rath:
This managed to sneak in through #1832


another-rex (commented):
Huh, I'm a bit confused as to how this still passed the test.

another-rex (approved):

PR #1846, opened 2025-04-28
feat: allow controlling which extractors are enabled

### LINKED ISSUE #1768: Support selectively enabling/disabling individual plugins

G-Rath:
This allows more flexibility by enabling users to control which extractors are enabled (or disabled), similar to call scan analysis.

Resolves #1768


G-Rath:
todos:
- [] add documentation
- [] rename cli flags (it should be something like `--disable-extractors`, or maybe even "extractor" since you pass them in one at a time?)
- [] switch to passing the extractors themselves in `osvscanner.ScannerActions` rather than just their names
  - while the issue originally called for the names as strings, the end goal is to be able to pass in custom extractors/plugins, and so I'm pretty sure it costs us very little to start doing that now which might save us some deprecation work
- [] sort out what to do with sboms, as we currently have a unique flag for them which in the long-run we probably ideally want to get merged into `-L` and co

another-rex:
> rename cli flags (it should be something like --disable-extractors, or maybe even "extractor" since you pass them in one at a time?)
I'm pretty sure `cli` allows you to pass it in comma separated?

G-Rath:
Yeah I'm going to refresh my memory on that, because I'm pretty sure you're right but then we were doing custom splitting for our generic license flag and so maybe we actually have a bug with that flag? so I'm going to double check what we're actually doing for all our list flags to make sure its consistent 😅

pkg/osvscanner/scan.go
```diff
@@ -7,19 +7,66 @@ import (
...
+			Disabled:   accessors.OSVDevClient == nil,
+		})
+	}
+}
+
+func getExtractors(defaultExtractorNames [][]string, accessors ExternalAccessors, actions ScannerActions) []filesystem.Extractor {
+	names := actions.ExtractorNames
+
+	if len(names) == 0 {
+		for _, preset := range defaultExtractorNames {
```
another-rex: nit: slices.Concat can be used here. I think that means we can get rid of the [][]string and just have a single [] and have the caller call slices.Concat themselves.

pkg/osvscanner/scan.go
```diff
@@ -7,19 +7,66 @@ import (
...
+				DependencyClient:       accessors.DependencyClients[osvschema.EcosystemMaven],
+				MavenRegistryAPIClient: accessors.MavenRegistryAPIClient,
+			})
+		}
+
+		// todo: the "disabled" aspect should probably be worked into the extractor being present in the first place
+		//  since "IncludeRootGit" is always true
+		gitrepo.Configure(tor, gitrepo.Config{
+			IncludeRootGit: actions.IncludeGitRoot,
+			Disabled:       !actions.IncludeGitRoot,
```
another-rex, line 36: This makes me realize that this might just be a bug at the moment, I don't think I intended skip root git to also skip subdirectories as well, but at the same time I guess we currently have no flags to disable skipping submodule scanning before this PR is landed.

Now that we do have a way to disable it, can you change this to also scan submodules? (We definitely should mention this behavior change in our changelog.

another-rex (commented):
This approach LGTM!

I do quite like the configure idea, I'll have a look and see if this will work for configuring plugins in scalibr as well.

As for passing extractors as types directly into the DoScan function, this does mean we'll have to make all the extractors public right? I'm not sure we want to do that in this repo, and probably will have to migrate it into OSV-Scalibr.

cmd/osv-scanner/internal/helper/extractors_parser.go
```diff
@@ -0,0 +1,53 @@
...
+	"lockfile":  scalibrextract.ExtractorsLockfiles,
+	"directory": scalibrextract.ExtractorsDirectories,
+	"artifact":  scalibrextract.ExtractorsArtifacts,
+}
+
+func ResolveEnabledExtractors(enabledExtractors []string, disabledExtractors []string) []string {
+	toDisable := make(map[string]bool)
+
+	for _, disabled := range disabledExtractors {
+		if names, ok := presets[disabled]; ok {
```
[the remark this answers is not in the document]
G-Rath: you're right that we don't need the condition for the disabled section, but we do need it for the enabled section otherwise the preset name itself will end up being added as an extractor

pkg/osvscanner/scan.go
G-Rath: Yeah I was going to talk to you about that + the wider "we should probably make our extractors public in some way"

if it's ok with you, I'd like to leave our extractors alone in this PR and change them as a follow-up so that they've got their own changelog entries

docs/experimental/managing-extractors.md
G-Rath, line 1: this is a very initial start of a doc, which has ended up being mainly just a capturing of what I think are key things for us to dive into as I realized while this PR does do what we want, there's at least a couple of key changes we should make around it like cleaning up our private extractors and having all extractors run (aka merge the `lockfiles`, `directories`, and `sboms` presets) which would make this documentation much simpler.

on an aside, I've added this in a manner that leverages Jekyell more (notably using subdirectories to imply the subpath) which we might like to actually do - if so I can do a restructure PR but want to check we actually like moving away from a flat structure

pkg/osvscanner/scan.go
another-rex: yep that sounds good

docs/experimental/managing-extractors.md
another-rex: Let's mark this page as hidden for now to get the logic merged in (I think function wise it's still mostly the same right? (if they just don't enable/disable extractors), and once we clean up how our different flags work we can re-enable it.

cmd/osv-scanner/internal/helper/extractors_parser.go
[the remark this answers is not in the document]
another-rex: Ah makes sense.

another-rex (approved):
LGTM, let's get this merged in as it is, and we can work on a few of the followup things:
- go generate for generating a list of available plugins
- Cleanup SBOM flags
- Add docs
- skip git behavior

G-Rath:
@another-rex I've removed both the doc and generator files for now - they're captured by GitHub in this PRs history so they'll be easy to bring back

another-rex:
Seems like this just needs a snapshot update? I'm not particularly sure why a vuln would get removed, I'm guessing it's because Debian added "unimportant" to the vuln.

PR #1853, opened 2025-04-30
ci: add job to ensure that Docker image should work

G-Rath:
Currently we're actually just testing the Linux AMD64 `osv-scanner` image, but it's still better than nothing

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -130,3 +130,41 @@ jobs:
         uses: ./.github/workflows/test-action
         with:
           codecov_token: ${{ secrets.CODECOV_TOKEN }}
+  docker:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+    runs-on: ubuntu-latest
+    env:
+      # Required for buildx on docker 19.x
+      DOCKER_CLI_EXPERIMENTAL: "enabled"
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


G-Rath:
This will fail until #1855 is landed

.github/workflows/checks.yml
```diff
@@ -130,3 +130,41 @@ jobs:
         uses: ./.github/workflows/test-action
         with:
           codecov_token: ${{ secrets.CODECOV_TOKEN }}
+  docker:
```
another-rex, line 133: I think this check should be part of the pre-release rather than every PR right (or maybe only when the .dockerfiles change)?

G-Rath: yeah I had similar thoughts - I think doing it as a prerelease and when the dockerfile changes would be the ideal for saving compute time but I figured folks might prefer to just run it all the time because it's still a faster job than our main test suite, and I wanted it up to show that https://github.com/google/osv-scanner/pull/1855 did indeed fix things.

I can adjust when we run this either way, so feel free to merge or wait at your leisure

another-rex: Yeah, I guess you're right, since everything is running in parallel it doesn't matter too much time wise, just extra compute. I'll merge it as is and we can come back to look at it later if we find issues with it.

another-rex (approved):

PR #1857, opened 2025-05-01
fix: update pkg assignment in vulnerability_result

hogo6002:
https://github.com/google/osv-scanner/pull/1776 added inventory to `Package`, but this didn't actually work because the package was subsequently overwritten when assigned `pkg.Package` to a new struct. 

Updates the package assignment to prevent this issue in the future.

### DIFF
--- a/pkg/osvscanner/vulnerability_result.go
+++ b/pkg/osvscanner/vulnerability_result.go
@@ -50,12 +50,10 @@ func buildVulnerabilityResults(
 		}
 
 		if p.Version() != "" && !p.Ecosystem().IsEmpty() {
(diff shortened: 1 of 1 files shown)


G-Rath:
Is it possible to write a test to cover this? I didn't notice any issues when I originally authored #1776

another-rex (approved):
Thanks! +1 Gareth's comment, it would be nice to add a test to cover this

hogo6002:
> Thanks! +1 Gareth's comment, it would be nice to add a test to cover this

We actually have tests on it, just updated the snaps.

PR #1859, opened 2025-05-02
fix(customgitignore) parsing to respect repository boundaries

### LINKED ISSUE #209: Gitignore parsing does not respect repository boundaries
The current Gitignore parsing does not take into account the boundaries of git repositories, meaning:
- `.gitignore` files that are outside of git repositories are parsed and applied.
- `.gitignore` files from the parent repository are applied to files in git subrepositories.

x-stp:
Fixes #209

## Problem
The current gitignore parsing doesn't respect repository boundaries:
1. Gitignore files outside of git repositories are parsed and applied
2. Gitignore files from parent repositories are applied to files in git subrepositories

## Changes
- Modified `readIgnoreFilesFromParents` function to detect repository boundaries
- Added `isSubRepository` function to check for different git repositories
- Added comprehensive tests for subrepositories and non-repo directories
- Ensures gitignore files only apply to their own repositories

## Testing
Added two new test cases:
- `TestParsingGitignoreFilesWithSubrepository` - Verifies parent repo gitignores aren't applied to subrepositories
- `TestGitignoreFilesOutsideRepoAreNotParsed` - Verifies gitignore files outside repos are ignored

All existing and new tests **should** pass. 

Thanks in advance for the review!

### DIFF
--- a/internal/customgitignore/dir_test.go
+++ b/internal/customgitignore/dir_test.go
@@ -627,6 +627,153 @@ func TestParsingGitRepoWithoutGitignoreFiles(t *testing.T) {
 	}
 }
 
+func TestParsingGitignoreFilesWithSubrepository(t *testing.T) {
+	t.Parallel()
+
+	// Create a main git repo
+	mainRepo := setupGitRepo(t)
+
+	// Create a subdirectory that will be a subrepository
+	subRepoPath := filepath.Join(mainRepo, "subrepo")
+	if err := os.MkdirAll(subRepoPath, 0755); err != nil {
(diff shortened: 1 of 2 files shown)


another-rex (approved):
Thanks for the contribution, but it's not very clear to me what the change accomplishes. 

I commented out the isSubRepository() function and all the newly added tests still passes just fine?

x-stp:
> Thanks for the contribution, but it's not very clear to me what the change accomplishes.
> 
> I commented out the isSubRepository() function and all the newly added tests still passes just fine?

Thank you for pointing this out; think that my tests need a more comprehensive case covering nested repo's.

PR #1866, opened 2025-05-13
test: Ignore goroutine logs

another-rex:
Moved the muffler logic into the newly made testlogger handler.

Currently, our tests will crash when a log is emitted from a goroutine, as we cannot find the test caller in the call stack of the goroutine. 

This happens right now from the status printer of osv-scalibr, which is in its own goroutine, and will be called if a filesystem walk lasts longer than 2 seconds, which only happens in windows occasionally, causing sproatic test errors for our windows tests. E.g. https://github.com/google/osv-scanner/actions/runs/14958771377/job/42017981090?pr=1863

---

This fixes that issue in 2 ways:
- Move the muffling logic up and muffle the status output. This actually allows us to remove the mufflerWriter entirely.
- Create a void logger that will write logs into the void, and use that logger when we find that we are calling the log from a goroutine. This is "fine" in that we **never** want to be snapshot testing logs made in goroutines anyway, since they don't have a deterministic order, so may constantly change each run.

To be extra safe that we don't accidently add logging in goroutines and assume that we are snapshot testing them, I also check whether the void log would have been muffled anyway, and if it was not muffled, panic with a more useful message. This will let us make sure we are consciously muffling those logs. 

> Technically the voidLogger is actually useless here I believe, since it will never get called.


G-Rath (approved):
awesome! fwiw, this should be prefixed with `test` rather than `fix` since its not changing any runtime code

michaelkedar (approved):

PR #1868, opened 2025-05-19
chore: enable `redundant-test-main-exit` rule for `revive` linter

G-Rath:


G-Rath:
ugh I was hoping https://github.com/golangci/golangci-lint/pull/5692 would land out of band, but apparently not so this is blocked by upgrading to v2

hogo6002 (approved):

PR #1870, opened 2025-05-20
ci: create workflow for updating snapshots automatically

G-Rath:
This workflow will run at 12:47 UTC each day to check if our snapshots have changed, and if so it will ensure a pull request exists with the updated result to make it easy to ensure CI is always passing.

In order to have CI run for this pull request, the workflow manually triggers a workflow run  on its own pull request using the GitHub API


another-rex (approved):
Pretty neat! Let's try it (Maybe those github actions might be blocked by the org, let's test it out and see how it goes)

PR #1872, opened 2025-05-21
test(output): add cases of packages with commits

G-Rath:
I realized when revisiting #1609 that we don't have this covered


another-rex (approved):
Thanks!

PR #1877, opened 2025-05-22
feat: transitive extractor for Python requirements

cuixq:


another-rex:
#1846 Going to wait for this one to get merged in first, then we need to do some work to support having both extractor types similar to what that PR does for the pomxml extractor.

PR #1881, opened 2025-05-24
ci: make "update snapshots" workflow actually work

G-Rath:
This job looks to have run correctly but it technically failed due to the `bot/update-snapshots` branch not existing as there were no changes to the snapshots to commit - I didn't really think about this case as my prior art was triggered by an event that meant it always had changes so the branch would always end up existing 😅


.github/workflows/update-snapshots.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           TEST_ACCEPTANCE: true
           UPDATE_SNAPSHOTS: true
```
another-rex, line 32:
```suggestion
          UPDATE_SNAPS: true
```

another-rex (changes requested):
Hmm... it looks like the env variable is incorrect, so it's just failing the test.

https://github.com/google/osv-scanner/actions/runs/15238121054

Also, can you add a `name:` to the top to make it easier to find in the Actions tab, thanks!

G-Rath:
> Hmm... it looks like the env variable is incorrect, so it's just failing the test.

Yup that's a different issue🤦I'll push some updates up shortly, as I've learned a few things to sort out

.github/workflows/snapshots.yml
```diff
@@ -25,12 +27,12 @@ jobs:
         with:
           go-version: stable
           check-latest: true
-      - run: scripts/build_test_images.sh
```
G-Rath, line 28: this is run by `scripts/run_tests` when `TEST_ACCEPTANCE` is `true`, so we can omit it here

.github/workflows/snapshots.yml
G-Rath, line 1: I renamed this on a bit of a whim, but I think the shorter name is a bit nicer + fits nicely with "`snapshots` workflow has a job named `update`", and I figured this would be the best time to do that 😅

G-Rath:
in saying that, I'm considering also adding `if: github.repository == 'google/osv-scanner'` to the job so that it doesn't run on forks by default 🤔

another-rex:
Hmm... catch 22, can't merged this until snapshot is updated, but I want to use this to test updating snapshots haha. I'll just force merge this.

G-Rath:
hmm I have a feeling that `go-snaps` looks for the `CI` variable in the env first when deciding if it should write snapshots 😞

G-Rath:
yup: https://github.com/gkampitakis/go-snaps/blob/1bd0d839f3997f8b41e6c24eb0f306aee266fe8a/snaps/utils.go#L120-L130

PR #1884, opened 2025-05-26
ci: only fail `docker` job if we get a non-zero exit code that does not mean "no vulns found"

G-Rath:
While we generally shouldn't have any vulnerabilities, it does sometimes happen due to merge delays or false positives, so we shouldn't fail this job as a result of that as it's just using our root `go.mod` to verify the image is working.

Note that by default GitHub Actions runs inline commands with `bash -e` meaning we need to extract the exit code on the same line rather than getting it afterwards


another-rex (approved):
Thanks!

hogo6002 (approved):

PR #1898, opened 2025-05-28
ci: allow snapshots to be updated

G-Rath:
This should hopefully be the last blocker 🎉 

Also see https://github.com/gkampitakis/go-snaps/issues/125


another-rex (approved):

cuixq (approved):

another-rex:
Hmm... not quite it seems: https://github.com/google/osv-scanner/actions/runs/15313152556

another-rex:
Do I need to give the workflow permission somewhere?

another-rex:
Ah the permissions key is empty, we need to add 
```
permissions:
  contents: write
  pull-requests: write
```

PR #1905, opened 2025-05-29
test: update snapshots

another-rex:


cmd/osv-scanner/__snapshots__/main_test.snap
another-rex, line 1: Hmm this doesn't look correct

PR #1908, opened 2025-05-29
feat: add javareach as an enrichment plugin

hogo6002:
This is an initial PR to add javareach as an enrichment plugin. The follow-up PR will resolve things listed in https://github.com/google/osv-scanner/issues/1886, after which this feature will be merged into osv-scalibr.

### DIFF
--- a/internal/scalibrenricher/javareach/jar.go
+++ b/internal/scalibrenricher/javareach/jar.go
@@ -17,7 +17,6 @@ package javareach
 import (
 	"archive/zip"
 	"bufio"
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


internal/scalibrenricher/javareach/javareach.go
```diff
@@ -0,0 +1,472 @@
...
+	manifest, err := jarRoot.Open(ManifestFilePath)
+	if err != nil {
+		return err
+	}
+
+	mainClasses, err := GetMainClasses(manifest)
+	if err != nil {
+		return err
+	}
+	log.Debug("Found", "main classes", mainClasses)
```
another-rex, line 202: Are these logs still relevant?

internal/scalibrenricher/javareach/javareach.go
```diff
@@ -0,0 +1,472 @@
...
+	}
+
+	// Slow path: make sure parent exists and then call Mkdir for path.
+
+	// Extract the parent folder from path by first removing any trailing
+	// path separator and then scanning backward until finding a path
+	// separator or reaching the beginning of the string.
+	i := len(path) - 1
+
+	for i >= 0 && path[i] == PathSeparator {
```
another-rex: This doesn't look right, what if it looks like this? `//\\//`, it will get trimmed down to `//` but then keep going?

internal/scalibrenricher/javareach/javareach.go
```diff
@@ -0,0 +1,472 @@
...
+// Version returns the version of the enricher.
+func (Enricher) Version() int {
+	return 0
+}
+
+// Requirements returns the requirements of the enricher.
+func (Enricher) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{
+		Network:  plugin.NetworkOnline,
+		DirectFS: false,
```
another-rex: DirectFS is false by default

internal/scalibrenricher/javareach/javareach.go
```diff
@@ -0,0 +1,472 @@
...
+	}
+
+	nestedJARs, err := unzipJAR(jarPath, input, jarRoot)
+	if err != nil {
+		return err
+	}
+
+	// Reachability analysis is limited to Maven-built JARs for now.
+	// Check for the existence of the Maven metadata directory.
+	_, err = os.Stat(path.Join(jarRoot.Name(), MavenDepDirPath))
```
another-rex: I don't think os.Stat is the correct function here, everything should be under root.

internal/scalibrenricher/javareach/javareach.go
```diff
@@ -0,0 +1,472 @@
...
+	// In windows, the zipPath still contains back slash
+	path := filepath.Join(zipPath, file)
+	if !strings.HasPrefix(path, filepath.Clean(zipPath)+string(os.PathSeparator)) {
+		return "", fmt.Errorf("directory traversal: %s", path)
+	}
+
+	return path, nil
+}
+
+func mkdirAll(root *os.Root, path string, perm os.FileMode) error {
```
another-rex: I would put mkdirAll to a separate file, and link where the code is from. I think we can simplify this by quite a bit though.

another-rex: We don't need to handle all the edge cases, since they won't show up in a jar zip anyway, and if they do it's a suspicious zip that we shouldn't extract.

internal/scalibrenricher/javareach/jar.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+// mapRootClasses maps class files to the root application where we can determine that association.
+func mapRootClasses(jarDir string, classMap map[string][]string, artifactMap map[string][]string) error {
+	// Spring Boot.
+	// TODO(#787): Handle non-Spring Boot applications. We could add heuristic for
+	// detecting root application classes when the class structure is flat based
+	// on the class hierarchy.
+	bootInfClasses := path.Join(jarDir, BootInfClasses)
+	if _, err := os.Stat(bootInfClasses); err != nil {
```
another-rex: This should not be using os.Stat

internal/scalibrenricher/javareach/jar.go
```diff
@@ -0,0 +1,372 @@
...
+		if os.IsNotExist(err) {
```
another-rex: Same here, should not be using os

internal/scalibrenricher/javareach/javareach.go
hogo6002: yes, i would like to keep those debug messages

internal/scalibrenricher/javareach/javareach.go
hogo6002: it copied from os.mkdirAll. Also os.Root is adding mkdirAll() function, I will replace it after it's available

internal/scalibrenricher/javareach/javareach.go
hogo6002: updated to root.Stat

internal/scalibrenricher/javareach/jar.go
```diff
@@ -237,7 +236,7 @@ func extractClassMappings(ctx context.Context, inv *extractor.Package, classMap
 
 // NewDefaultPackageFinder creates a new DefaultPackageFinder based on a set of
 // inventory.
-func NewDefaultPackageFinder(ctx context.Context, inv []*extractor.Package, jarDir string, client *http.Client) (*DefaultPackageFinder, error) {
+func NewDefaultPackageFinder(ctx context.Context, inv []*extractor.Package, jarDir *os.Root, client *http.Client) (*DefaultPackageFinder, error) {
```
another-rex: let's keep all the the osRoot the same name, probably jarRoot

internal/scalibrenricher/javareach/jar.go
another-rex: https://pkg.go.dev/os#IsNotExist

> This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

internal/scalibrenricher/javareach/reachable.go
```diff
@@ -0,0 +1,416 @@
...
+		if err != nil {
+			return nil, err
+		}
+		r.loadedJARs[jarPath] = zipr
+	}
+
+	zipr := r.loadedJARs[jarPath]
+	class, err := zipr.Open(className + ".class")
+	if err != nil {
+		if os.IsNotExist(err) {
```
another-rex: Same here

internal/scalibrenricher/javareach/reachable.go
```diff
@@ -0,0 +1,416 @@
...
+		}
+
+		// Look inside the class directory.
+		classFilepath := path.Join(classPath, className)
+
+		if !strings.HasSuffix(classFilepath, ".class") {
+			classFilepath += ".class"
+		}
+
+		if _, err := jarRoot.Stat(classFilepath); os.IsNotExist(err) {
```
another-rex: And here

internal/scalibrenricher/javareach/jar.go
```diff
@@ -0,0 +1,369 @@
...
+		if err != nil {
+			return nil, err
+		}
+		jarPaths = jarPaths[1:]
+	}
+
+	return zipr, nil
+}
+
+func checkNestedJARContains(inv *extractor.Package) ([]string, bool) {
```
another-rex: Is this function being tested? As in is one of the testdata jars a nested jar?

another-rex: Also, this seems like a good standalone function to write a unit test for.

hogo6002: I will create an issue on osv-scalibr side to add more unit tests for javareach. But for this initial PR, I would like to keep this for now.

hogo6002: i decided to remove nested jar check for now. Now i just focus on mapping class files with maven central. I will add the issue number to this todo when merging code into osv-scalibr

another-rex (approved):

PR #1919, opened 2025-06-03
fix: include actual error if Maven registry response body cannot be read

G-Rath:
Currently we're creating a local variable `err` which shadows the outer variable, so while we do still correctly go off the happy path in this case, the error we return does not include the underlying error message since we use the outer `err` variable which at this point will hold `nil`.

This was found by the `nilnesserr` linter which is a new linter included in v2 that "combines `nilness` and `nilerr`", and that I've enabled as part of this

### DIFF
--- a/internal/datasource/maven_registry.go
+++ b/internal/datasource/maven_registry.go
@@ -224,11 +224,11 @@ func (m *MavenRegistryAPIClient) get(ctx context.Context, auth *HTTPAuthenticati
 		}
 
 		b, err := io.ReadAll(resp.Body)
-		if err == nil {
-			return response{StatusCode: resp.StatusCode, Body: b}, nil
+		if err != nil {
+			return response{}, fmt.Errorf("failed to read body: %w", err)
 		}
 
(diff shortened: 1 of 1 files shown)


internal/datasource/maven_registry.go
```diff
@@ -223,7 +223,8 @@ func (m *MavenRegistryAPIClient) get(ctx context.Context, auth *HTTPAuthenticati
 			return response{}, fmt.Errorf("%w: Maven registry query status: %d", errAPIFailed, resp.StatusCode)
 		}
 
-		if b, err := io.ReadAll(resp.Body); err == nil {
+		b, err := io.ReadAll(resp.Body)
```
another-rex, line 226: I think this whole thing should just be inverted right, the last line should be the valid response, and all the branches are the errors.

another-rex (approved):
Good catch! I think we should just swap the branches though to be more consistent.

internal/datasource/maven_registry.go
G-Rath: done

PR #1920, opened 2025-06-03
test(cmd): give `cmd/osv-scanner/scan` its own dedicated copy of the fixtures

### LINKED ISSUE #1896: git fixtures in our tests causes a race condition for tests
We create a git repository in our top level tests, and remove it when the top level tests are complete. However, the sub command tests also use the same git directory if it is present, causing a race condition to occur.

https://github.com/google/osv-scanner/blob/ffe33914b385206086077226aa0724a929021a67/cmd/osv-scanner/testmain_test.go#L21

I'm not 100% sure what's the best way to fix this.

G-Rath:
Turns out we still had some cmd fixture sharing going on which I bet is why we've still got a race condition related to the `.git` setup.

Resolves #1896


cuixq (approved):
Which folder do the fixtures come from? Can we get rid of the original copies?

G-Rath:
They were copied from `cmd/osv-scanner/fixtures` which is where the fixtures originally lived - we decided to duplicate them to avoid issues around locking and snapshot normalizations, since these fixtures are just for confirming the subcommands work (we could probably remove a few files to make them smaller, it's just not been a priority 🤷)

PR #1922, opened 2025-06-03
fix: prevent panic when batchResp is nil in MatchVulnerabilities

laojianzi:
### Problem
The `MatchVulnerabilities` method in `osvmatcher.go` could panic when accessing `batchResp.Results` if `batchResp` is nil.

### Solution
- Added nil check for `batchResp` before accessing its `Results` field

### Testing
- Added test case `"Timeout returns deadline exceeded error"` to verify the timeout behavior
- Test ensures the method returns `context.DeadlineExceeded` error without panicking

### Changes
- `internal/clients/clientimpl/osvmatcher/osvmatcher.go`: Added nil check and documentation
- `internal/clients/clientimpl/osvmatcher/osvmatcher_test.go`: New test file with timeout test case

### DIFF
--- a/internal/clients/clientimpl/osvmatcher/osvmatcher.go
+++ b/internal/clients/clientimpl/osvmatcher/osvmatcher.go
@@ -58,7 +58,7 @@ func (matcher *OSVMatcher) MatchVulnerabilities(ctx context.Context, pkgs []*ext
 			}
 		}
 
-		// No results found - this could be due to a timeout or no vulnerabilities being found
+		// No results found - this could be due to a timeout
 		if batchResp == nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


internal/clients/clientimpl/osvmatcher/osvmatcher.go
```diff
@@ -56,6 +57,11 @@ func (matcher *OSVMatcher) MatchVulnerabilities(ctx context.Context, pkgs []*ext
 				return nil, err
 			}
 		}
+
+		// No results found - this could be due to a timeout or no vulnerabilities being found
```
another-rex: Only from timeout, no vulnerabilities found will result in `batchResp.Result = []`
```suggestion
		// No results found - this could be due to a timeout
```

Issue #1769, opened 2025-03-30
Migrate OSV-Scanner to use SCALIBR’s main Scan() function


another-rex:
## Overview

OSV-Scanner will use OSV-Scalibr’s Scan function directly rather than using our own filesystem walker.

This is currently blocked by several features missing in OSV-Scalibr:'

- [x] Git based .gitignore files for filtering what files to walk through for recursive scans
- [x] Support non recursive scans.

## Tasks

- [ ] Setup `Config` for calling OSV-Scalibr
- [ ] Convert directory scan to use `Scan()`
- [ ] Convert single file scan to use `Scan()`

another-rex:
Looking into this more, I'm not sure we want to convert single file scanning to use Scan(), as it doesn't make too much sense, since Scan is also directly calling the underlying Extractor, and it just makes it difficult to do the extractor name translation. 

The biggest benefit would be to use just convert directory scanning to use Scan, so we can share the filesystem walking logic.

PR #1935, opened 2025-06-10
chore: Update osv-scalibr

another-rex:
Update osv-scalibr to the latest version to unblock various osv-scanner feature developments, including:
- OSV-Scanner/OSV-Scalibr convergence #1769
- Guided Remediation to use scalibr migrated code.

OSV-Scalibr changed the package format to use a combination of PURLType and Plugin Name string rather than a pointer to the extractor itself. This means to get ecosystem we need to set PURLType and Metadata in our mocks.


cuixq:
Can you update the description with the reason that we update osv-scalibr now (instead of waiting for the next release)? Also what is being changed with this update.

another-rex:
Done, and updating to the latest version of scalibr with your fix seemed to have solved this specific issue.

cuixq (approved):

PR #1936, opened 2025-06-10
feat: Use OSV-Scalibr scan function directly

### LINKED ISSUE #1769: Migrate OSV-Scanner to use SCALIBR’s main Scan() function

another-rex:
Closes #1769 

Behaviour changes / notes:
- OSV-Scalibr's Scan() function does DFS rather than BFS, so the order of files being read in the snapshot changes slightly.
- Their Scan() function also does not deduplicate the output of the extractor, so the SBOM with duplicates will return double the results. I think it's fine to have that behavior change for now unless someone complains about it.
- Unfortunately, we also double log the path to the file in the error, because both the extractor and the walker both prepend the path the error originated from. This is a TODO in osv-scalibr to remove input path references in the extractor errors. (https://github.com/google/osv-scalibr/issues/821)


another-rex:
Annoyingly it seems like osv-scalibr's directory walk is not deterministic on different systems (as it uses the order the filesystem gives when iterating through a directory).

I'm not sure how to fix this for our tests, do we just avoid snapshotting any of the walking logs?

G-Rath:
> I'm not sure how to fix this for our tests, do we just avoid snapshotting any of the walking logs?

I think those logs are an important part of our snapshots since they're output we definitely want to be showing to the user (compared to some of the other logs we're currently filtering) and they aid in debugging since they tell us clearly what files are being scanned and the number of packages extracted.

Is there anyway to make the walking deterministic, even if its controlled via a flag or env variable so that we can opt-in?

pkg/osvscanner/stats.go
```diff
@@ -15,6 +15,10 @@ type FileOpenedPrinter struct {
 var _ stats.Collector = &FileOpenedPrinter{}
 
 func (c FileOpenedPrinter) AfterExtractorRun(_ string, extractorstats *stats.AfterExtractorStats) {
```
G-Rath, line 18: a possible workaround could be to try buffering the logging, and try to dump it immediately if we error - we could also do the same logging here under a different level like `DEBUG` so that you could see the actual unbuffered order...

e.g. something like:

```
// NOT ACTUAL GO CODE
scannedLogs := []string{}

err := scanner.Scan({
	StatsAfterExtractorRun(stats) {
		slog.Debug(fmt.Sprintf(
			"Scanned %s file and found %d %s",
			extractorstats.Path,
			pkgsFound,
			output.Form(pkgsFound, "package", "packages"),
		))

		scannedLogs = append(scannedLogs, fmt.Sprintf(
			"Scanned %s file and found %d %s",
			extractorstats.Path,
			pkgsFound,
			output.Form(pkgsFound, "package", "packages"),
		))
	}
})

for line := range slices.Sort(scannedLogs) {
	slog.Info(line)
}

if err != nil {
	slog.Error(err)
}
```

pkg/osvscanner/scan.go
```diff
@@ -108,13 +115,74 @@ func scan(accessors ExternalAccessors, actions ScannerActions) ([]imodels.Packag
...
+		rootMap[root] = append(rootMap[root], absPath)
+	}
+
+	for root, paths := range rootMap {
+		capabilities := plugin.Capabilities{
+			DirectFS:      true,
+			RunningSystem: true,
+		}
+
+		if actions.CompareOffline {
```
G-Rath, line 152: nit: the `else` branches here could be the initial values in the struct

pkg/osvscanner/stats.go
another-rex: I ended up with something similar to the buffering approach, just on the logs instead so we don't have to touch the code functions that much.

another-rex:
@G-Rath 

> I think those logs are an important part of our snapshots since they're output we definitely want to be showing to the user (compared to some of the other logs we're currently filtering) and they aid in debugging since they tell us clearly what files are being scanned and the number of packages extracted.

Yeah I agree.

> Is there anyway to make the walking deterministic, even if its controlled via a flag or env variable so that we can opt-in?

I am very hesitant to change the actual code logic for tests, because then we are not really testing for the actual code paths someone running osv-scanner will run into.

---

I ended up with printing a few markers at the start and end of the directory walk (When testing is true), then sorting the lines in-between the markers to check for presence, but not the specific order. 

PTAL

G-Rath:
> > Is there anyway to make the walking deterministic, even if its controlled via a flag or env variable so that we can opt-in?
>
> I am very hesitant to change the actual code logic for tests, because then we are not really testing for the actual code paths someone running osv-scanner will run into.

fwiw I agree - I was meaning we do that for the actual scanner logic not just for tests

another-rex:
The windows test is currently blocked on a bug in osv-scalibr's recursive flag on windows, fix here: https://github.com/google/osv-scalibr/pull/828

internal/scalibrextract/language/python/requirementsenhancable/requirementsenhanceable.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/plugin"
+)
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "python/requirementsenhanceable"
+)
+
+// Extractor extracts Maven packages from pom.xml files.
```
oliverchang: nit: wrong comment?

oliverchang (approved):
nice! looks good generally to me

internal/scalibrextract/language/python/requirementsenhancable/requirementsenhanceable.go
another-rex: fixed!

PR #1943, opened 2025-06-12
feat: add flag to suppress "no package sources found" error

G-Rath:
Currently we always consider it an error if the scanner does not find any supported files to scan as we assume in most cases it's expected the scanner will be check something and this avoids thinking that there are no vulnerabilities.

While this is a reasonable assumption for the default behaviour, it's annoying for enterprise types that want to use the GitHub Actions action as a required "always on" action for every repository regardless of what they actually hold - to workaround this currently the action eats the 128 exit code, but that's not ideal as it introduces the risk we're trying to avoid by having the error in the first place.

To give more flexibility, I've introduced a new flag to the scanner so that users can explicitly signal they don't want "no lockfiles being found" to be treated as an error, and in turn letting us deprecate our existing exit code overriding workaround in the GHA action in favor of having this flag passed in


another-rex (approved):
LGTM, thanks!

G-Rath:
@another-rex this is failing CI - I was still trying to sort the test suite out

PR #1945, opened 2025-06-13
test: remove unneeded `strings.ReplaceAll` call

G-Rath:
Someone should make a linter for this..


another-rex:
Umm is that duplicate? It's replacing a different text, though the tests didn't fail, so maybe the initial version is incorrect.

G-Rath:
> Umm is that duplicate? It's replacing a different text, though the tests didn't fail

Nope, I'm apparently just starting to lose my eyesight 🤦

G-Rath:
fwiw here's a run with this entire normalizer removed: https://github.com/google/osv-scanner/actions/runs/15626775880/job/44022528538?pr=1945

what's weird is that the replace I'm removing looks like it should match, but the tests were passing before - I'll dig into that next week

G-Rath:
ok it looks like either way this is now needed by the test I introduced in #1943 😅

PR #1947, opened 2025-06-13
ci: add job to see how fast Windows tests run without a cache

G-Rath:
A large amount of CI time for the Windows "test suite" is actually spent extracting the module cache to the point that I suspect it might be faster to skip caching - to test this, I'm adding a new job that runs our tests on Windows without the cache so that after a few months we'll be able to compare the average runtime of the jobs in the metrics tab:

image

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -131,6 +131,33 @@ jobs:
         uses: ./.github/workflows/test-action
         with:
           codecov_token: ${{ secrets.CODECOV_TOKEN }}
+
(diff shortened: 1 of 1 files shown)


G-Rath:
Just for future us, this is the link for reviewing the results: https://github.com/google/osv-scanner/actions/metrics/performance?tab=jobs&filters=job_name%3A%22Run+unit+tests+%28windows-latest%2C+no+cache%29%22%2C%22Run+unit+tests+%28windows-latest%29%22

Issue #1790, opened 2025-04-08
Filtering output does not filter out the source package from table output


another-rex:
Output filtering does not result in the source tables being filtered out:

```
Container Scanning Result (Distroless):
Total 1 package affected by 1 known vulnerability (0 Critical, 0 High, 0 Medium, 0 Low, 1 Unknown) from 1 ecosystem.
0 vulnerabilities can be fixed.


Debian:12
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Source:os:var/lib/dpkg/status.d/gcc-12-base                                                                                  │
├────────────────┬───────────────────┬───────────────┬────────────┬─────────────────────────┬──────────────────┬───────────────┤
│ SOURCE PACKAGE │ INSTALLED VERSION │ FIX AVAILABLE │ VULN COUNT │ BINARY PACKAGES (COUNT) │ INTRODUCED LAYER │ IN BASE IMAGE │
├────────────────┼───────────────────┼───────────────┼────────────┼─────────────────────────┼──────────────────┼───────────────┤
╰────────────────┴───────────────────┴───────────────┴────────────┴─────────────────────────┴──────────────────┴───────────────╯
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Source:os:var/lib/dpkg/status.d/libc6                                                                                        │
├────────────────┬───────────────────┬───────────────┬────────────┬─────────────────────────┬──────────────────┬───────────────┤
│ SOURCE PACKAGE │ INSTALLED VERSION │ FIX AVAILABLE │ VULN COUNT │ BINARY PACKAGES (COUNT) │ INTRODUCED LAYER │ IN BASE IMAGE │
├────────────────┼───────────────────┼───────────────┼────────────┼─────────────────────────┼──────────────────┼───────────────┤
╰────────────────┴───────────────────┴───────────────┴────────────┴─────────────────────────┴──────────────────┴───────────────╯
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Source:os:var/lib/dpkg/status.d/libgcc-s1                                                                                    │
├────────────────┬───────────────────┬───────────────┬────────────┬─────────────────────────┬──────────────────┬───────────────┤
[...]

PR #1949, opened 2025-06-15
fix: filter regular PackageTypeCount

Vialathor:
Resolves https://github.com/google/osv-scanner/issues/1790

Added an if statement when looping over an ecosystem's source to check if there are any regular vuln counts and break the loop if there is.


internal/output/table.go
```diff
@@ -118,9 +118,11 @@ func printSummaryResult(result Result, outputWriter io.Writer, terminalWidth int
 	}
 
 	for _, ecosystem := range result.Ecosystems {
-		fmt.Fprintln(outputWriter, ecosystem.Name)
-
 		for _, source := range ecosystem.Sources {
+			if source.PackageTypeCount.Regular == 0 {
+				break
```
hogo6002: one ecosystem contains multiple sources, so we need to do `continue` here instead of `break` the entire for loop.

internal/output/table.go
```diff
@@ -118,9 +118,11 @@ func printSummaryResult(result Result, outputWriter io.Writer, terminalWidth int
...
+			}
+			fmt.Fprintln(outputWriter, ecosystem.Name)
```
hogo6002: let's keep this statement outside the for loop because an ecosystem can have multiple sources. If you want to avoid printing the ecosystem name when all sources lack regular package types, consider creating a helper function. This function could check for if regular packages exist in any source and then use to determine whether to print the ecosystem name.

hogo6002 (commented):
Thanks for contributing! I have let some comments

hogo6002:
Make sure to run `TEST_ACCEPTANCE=true ./scripts/run_tests.sh` to trigger all tests and see if code changes break any existent function. 
If code change does require to update test snapshots, add `UPDATE_SNAPS=true`. 
more details: https://github.com/google/osv-scanner/blob/main/CONTRIBUTING.md#running-tests

internal/output/output_result.go
```diff
@@ -850,3 +850,15 @@ func containsOSResult(result Result) bool {
 
 	return false
 }
+
+func getEcosysRegVulnCount(ecosystem EcosystemResult) int {
```
hogo6002: the function name doesn't seem correct. The thing you are counting here is the number of sources containing regular vulns, not exact the vuln count. We can just return a bool here to check if a ecosystem contains any regular vulns (count > 0). Maybe rename to something like: `ecosystemHasRegVuln`

hogo6002 (approved):

Issue #1960, opened 2025-06-16
Remove dedicated SBOM flag in favor of `-L` and friends


G-Rath:
The [`--sbom` flag](https://google.github.io/osv-scanner/usage/scan-source#specify-sbom) was introduced while we were still figuring out how to handle scanning in different contexts and stuff like resolving patterns to files (like with `requirements.txt`) rather than just 1:1 mappings.

Nowadays we shouldn't need a dedicated flag (in fact right now you can use `-L` instead of `-sbom` and get the same results), and it makes it hard to move forward with #1846 due to its slightly special-but-unneeded behaviour.

We should start by deprecating the `--sbom` flag in favor of `-L`, and go from there

G-Rath:
<img width="300" height="169" alt="Image" src="https://github.com/user-attachments/assets/304d59d6-1bc2-4e9f-a889-b0471f333857" />

PR #1961, opened 2025-06-16
feat: deprecate `--sbom` in favor of just using `-L`

G-Rath:
I think the only difference between the two flags is that `--sbom` will give you an extra message if you provide a file which we fail to make to an extractor we give you an extra message about making sure to check the filename, which I actually think is arguably better with `-L` because imo you can't reasonably assume the scanner will be able to magically figure out e.g. `my.awesome.file.json` is meant to be an SBOM unlike with the current flag?

This is the first of part #1960

### DIFF
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -40,7 +40,6 @@ import (
 
 type ScannerActions struct {
 	LockfilePaths      []string
-	SBOMPaths          []string
 	DirectoryPaths     []string
 	GitCommits         []string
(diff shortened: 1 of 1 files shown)


cuixq:
@G-Rath this PR fails the linter - could you help to fix it?
```
Error: cmd/osv-scanner/scan/source/command.go:128:2: SA1019: scannerAction.SBOMPaths is deprecated: in favor of LockfilePaths (staticcheck)
  	scannerAction.SBOMPaths = cmd.StringSlice("sbom")
```

PR #1963, opened 2025-06-16
feat: make various experimental fields as stable

G-Rath:
These fields were marked as stable as part of the v2 release but we forgot to move the actual struct fields out of the experimental struct, which this does.

Technically this isn't breaking because we declare anything in the experimental struct as not being covered by semver 😅

### DIFF
--- a/cmd/osv-scanner/internal/helper/helper.go
+++ b/cmd/osv-scanner/internal/helper/helper.go
@@ -256,14 +256,21 @@ func GetScanLicensesAllowlist(cmd *cli.Command) ([]string, error) {
 	return allowlist, nil
 }
 
-func GetExperimentalScannerActions(cmd *cli.Command, scanLicensesAllowlist []string) osvscanner.ExperimentalScannerActions {
-	return osvscanner.ExperimentalScannerActions{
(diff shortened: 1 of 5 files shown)


cmd/osv-scanner/internal/helper/helper.go
```diff
@@ -256,14 +256,21 @@ func GetScanLicensesAllowlist(cmd *cli.Command) ([]string, error) {
 	return allowlist, nil
 }
 
-func GetExperimentalScannerActions(cmd *cli.Command, scanLicensesAllowlist []string) osvscanner.ExperimentalScannerActions {
-	return osvscanner.ExperimentalScannerActions{
-		LocalDBPath:           cmd.String("local-db-path"),
-		DownloadDatabases:     cmd.Bool("download-offline-databases"),
-		CompareOffline:        cmd.Bool("offline-vulnerabilities"),
+func GetCommonScannerActions(cmd *cli.Command, scanLicensesAllowlist []string) osvscanner.ScannerActions {
```
G-Rath, line 259: note I think we could also handle getting the `scanLicensesAllowlist` here, but I've left that for another PR because it would involve moving some other code around in order to support returning an error

PR #1965, opened 2025-06-17
ci: create snapshots pull request even if tests fail

G-Rath:
Any test failure that isn't fixed by (re)generating the snapshots should be present in the CI for the resulting pull request and if there are snapshot updates needed having those already generated will save some work - for example, the workflow is currently failing because one of our tests needs its exit code updated as previously it didn't have any vulnerabilities


G-Rath:
@another-rex you happy if we land this and rerun the workflow to see if it works as I expect, before we actually get CI passing again?

Issue #1968, opened 2025-06-18
Ignoring all unimportant vulnerabilities


jayvdb:
I am recently seeing the following table with 13 entries

```
+-----------------------------------------+------+-----------+---------+-----------+--------------+
| OSV URL                                 | CVSS | ECOSYSTEM | PACKAGE | VERSION   | SOURCE       |
+-----------------------------------------+------+-----------+---------+-----------+--------------+
| Unimportant vulnerabilities             |      |           |         |           |              |
+-----------------------------------------+------+-----------+---------+-----------+--------------+
| https://osv.dev/UBUNTU-CVE-2018-1000886 | 5.5  | Ubuntu    | nasm    | 2.16.03-1 | sbom.cdx.xml |
..
```

I dont see any docs on how I can configure the scanner to ignore them all, except by adding an entry for each of them.

jayvdb:
This didnt work
```toml
[[PackageOverrides]]
ecosystem = "ubuntu"
group = "unimportant"
ignore = true
```

oliverchang:
@hogo6002 can you take a look?

hogo6002:
Currently we don't have a way to ignore all unimportant ones from table output. We do plan to add a new flag to remove unimportant ones from the results. It would probably takes a bit of time as we need to wait for next release cycles. But meanwhile, feel free to use our HTML output (run with `--serve` flag) which filters out unimportant ones by default.

jayvdb:
The table or HTML isnt important for me.

It is the exit code that matters.  osv-scanner fails unless I explicitly list these self-declared unimportant vulns.

another-rex:
~~In that case, it should already have your desired behavior?~~

~~Right now if there are only uncalled vulns the exit code would be 0.~~

another-rex:
Ah I see we have a separate code path for unimportant, that should be also returning 0, I'll add a fix for that.

PR #1970, opened 2025-06-18
feat: hide unimportant vulns by default

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1968

Hides unimportant/uncalled vulns by default, and adds a `--all-vulns` flag to show all.
Returns 0 if only unimportant vulns.
Add an ignorePacakge filter to DoContainerScan()

### DIFF
--- a/cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
@@ -157,6 +157,7 @@ You can also view the full vulnerability list in your terminal with: `osv-scanne
 [TestCommand_OCIImage/Empty_Ubuntu_20.04_image_tar_with_no_vulns_shown - 1]
 Scanning local image tarball "../../../../internal/image/fixtures/test-ubuntu-20-04.tar"
 Package Ubuntu:20.04/coreutils/8.30-3ubuntu2 has been filtered out because: Just want to test only unimportant vulns
+Package Ubuntu:20.04/dpkg/1.19.7ubuntu3.2 has been filtered out because: Just want to test only unimportant vulns
 Package Ubuntu:20.04/gcc-10/10.5.0-1ubuntu1~20.04 has been filtered out because: Just want to test only unimportant vulns
 Package Ubuntu:20.04/gnupg2/2.2.19-3ubuntu2.4 has been filtered out because: Just want to test only unimportant vulns
 Package Ubuntu:20.04/glibc/2.31-0ubuntu9.17 has been filtered out because: Just want to test only unimportant vulns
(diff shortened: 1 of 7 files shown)


G-Rath:
Is there a reason why you've gone with a flag vs adding support for this as a config option? as the latter sounds like it would address the reason behind #1968. Not saying it's wrong, just keen to hear a bit more about the offline discussions that have happened on this before I review.

fwiw I'm also a little hesitant on having this be the default, but that might just be because I work in languages that don't have call analysis and the idea is it'll be far more mainstream in future? (I'm use to dealing with security frameworks that say "all vulns _must_ be patched" which without CA means truly patching anything that gets called a vuln by an advisory...)

hogo6002:
> Is there a reason why you've gone with a flag vs adding support for this as a config option?

I think we should add both.
The idea here is to first add a flag to provide a quick solution to this issue, which allows users to enable or disable all unimportant/uncalled vulnerabilities. Then, for each specific ecosystem or package, we can support a config option that allows users to override.
I will make a second PR to adjust the exit code. The exit code should match the final output. If no vulnerabilities are printed, we should exit with zero.

cmd/osv-reporter/main.go
```diff
@@ -137,12 +137,12 @@ func run(args []string, stdout, stderr io.Writer) int {
 				diffVulns = ci.DiffVulnerabilityResults(oldVulns, newVulns)
 			}
 
-			if errPrint := reporter.PrintResult(&diffVulns, "table", stdout, termWidth); errPrint != nil {
+			if errPrint := reporter.PrintResult(&diffVulns, "table", stdout, termWidth, false); errPrint != nil {
```
another-rex: I think you need to add the flag to the reporter as well.

We'll also need to pass it through the github action, though that can be a separate PR.

another-rex:
@G-Rath Just want to understand what you meant by:

> fwiw I'm also a little hesitant on having this be the default, but that might just be because I work in languages that don't have call analysis and the idea is it'll be far more mainstream in future?

I think this PR will have no effect on those languages right? As there won't be CA to mark vulns as unimportant, so there won't be any change.

G-Rath:
Right _now_ there isn't CA but the goal is to have that for as many languages as possible ay? it might be (a lot) harder for some languages than others to the point that it could take decades but the goal is still to have it is my understanding.

I think my feelings on this are due to uncertainty in the confidence of the CA - similar to how there are tools for detecting "secrets" in code that are literally just searching for the word "secret" in variable names; what if someone starts shipping a CA for a new language that is similar but maybe a bit smarter than that? with this being the default that means suddenly a bunch of vulns are no longer being signaled to me because some persons `grep` doesn't find any references to the function name?

But I think that's actually already accounted for by us humans doing the building - yes what I've said is _technically_ possible, but in practice we'd decide that that CA isn't reliable enough yet to "trust" with being used in this context as a default and so wouldn't adopt it; and likewise I expect as we got closer to having CA in a new language we'd consider how it gets rolled out including "whats the confidence" and probably use an experimental flag for a bit, and so on.

so tl;dr happy to discuss my feelings more but I don't think it needs to be a blocker to making this a default 🙂

cmd/osv-reporter/main.go
hogo6002: the new flag has already been added https://github.com/google/osv-scanner/pull/1970/commits/ae7544ec293f12d766bf4c5b9f4560f18d246e44

another-rex: Oh, oops this is an old comment I forgot to submit.

hogo6002:
> I think we should at least print a line "Hiding X number of vulnerabilities deemed unimportant, use --all-vulnsto show them." if there are unimportant vulns so at least they are not silently dropped. Other wise LGTM

Added this message into table container view and vertical output. I will add it to the regular table output after we replace it with internal structure (https://github.com/google/osv-scanner/pull/1609).

PR #1982, opened 2025-06-19
fix: Parse ubuntu suffix

Vialathor:
Removed ":Pro" ":LTS" from appearing in string comparison method for getNextFixVersion(...) for Ubuntu environments.


hogo6002:
can you update branch and then run the test case locally to update the snapshots?

PR #1985, opened 2025-06-20
chore: Update go-sarif to v3 from v2

another-rex:
Updated sarif library to v3 from v2. Still outputting SARIF 2.1 as that is what github supports.

I also changed how sarif_test snapshots are generated in the first commit to make it easier to see the diff, so when reviewing, please see the differences between the first commit and the last one for the snapshots, don't look at the full thing.

~~Currently it seems like there's an issue with ruleIndex being populated with -1 rather that the actual rule index, not sure why yet, I'll post an issue on the libraries repo.~~

Solved! Updated to v3.2.0 which resolved the above issue.


internal/output/sarif_test.go
```diff
@@ -80,7 +81,14 @@ func TestPrintSARIFReport(t *testing.T) {
 			if err != nil {
 				t.Errorf("Error writing SARIF output: %s", err)
 			}
-			tt.want.MatchText(t, bufOut.String())
+
+			jsonStructure := map[string]any{}
```
cuixq: can we factor this out?

cuixq (approved):

internal/output/sarif_test.go
another-rex: Good idea! Done.

cuixq (approved):

PR #2003, opened 2025-06-27
feat: Experimental output format for reporter

another-rex:
Playing around with output formats for the reporter, perhaps we introduce it to osv-scanner --output flag in the future to allow for multiple outputs.


cuixq (approved):
We should update the documentation as well.

another-rex:
This is not really public, it's only used in the github action as an implementation detail. But I added a page for osv-reporter under the experimental section now.

PR #2011, opened 2025-07-02
chore: Update scalibr and remove javareach

another-rex:
Javareach has already been migrated into osv-scalibr, we'll just need a follow up PR to add the functionality into osv-scanner.

Currently we are still using the deprecated Annotation field. That will be migrated after govulncheck and rust call analysis is moved to osv-scalibr.

Current windows test failure is fixed here: https://github.com/google/osv-scanner/pull/2011


pkg/osvscanner/vulnerability_result.go
```diff
@@ -134,7 +134,7 @@ func buildVulnerabilityResults(
 
 			groupedBySource[source].pvs = append(groupedBySource[source].pvs, pkg)
 			// Overwrite annotations as it should be the same for the same package.
-			groupedBySource[source].annotations = p.Annotations
+			groupedBySource[source].annotations = p.AnnotationsDeprecated
```
hogo6002, line 137: are we still gonna use annotation here?

another-rex: I think this is for govulncheck and rust call analysis. Once those are moved to osv-scalibr, that's when we can get rid of annotations here.

PR #2020, opened 2025-07-06
refactor: cleanup internal cmd helper package

G-Rath:
The current single file is getting a bit big in no small part due to the huge slice of common flags in the middle, so I've split things across three files:
  * `flags.go` which holds helpers related to CLI flags, namely our internal `allowedLicensesFlag` flag and "common scan flags builder" (which I've renamed to no longer qualify as a "getter" in addition to just being overall more accurate)
  * `getters.go` which holds functions that "get" stuff from the `*cli.Command`
  * `misc.go` which holds everything else


cuixq (approved):

another-rex:
Hmm merging the natural language PR caused conflicts with this one I think

PR #2021, opened 2025-07-07
feat: support running detectors

G-Rath:


G-Rath:
@another-rex turns out I just completely forgot about configuring the snapshot normalizers for JSON 🤦

I think this is good to go, and probably best to land as-is so it can be used, and get things improved in follow up, specifically:
  - supporting it in more output formats
  - reusing the detector building, in particular trying to leverage `list.DetectorsFromName()`
    - (I've started looking into this but frankly `InitMap` has been giving me a headache and I'm thinking it might be easier to tackle once we've refactored extractors in a similar manner per our convo the other day)

G-Rath:
hmm so the failures on Windows are from detectors such as `weakcreds/filesystem` in the container tests, which is interesting as that detector should only run on a running Linux system which we're not in - I think this is because we are not actually doing any filtering and scalibr only applies its filtering in some situations? But I think it would be good to get a review and check-in from you either way so I'll wait to hear from you before digging deeper

cmd/osv-scanner/internal/helper/detectors_parser.go
```diff
@@ -0,0 +1,52 @@
...
+
+			detectors[detectorOrPreset] = enabled
+		}
+	}
+
+	asSlice := make([]detector.Detector, 0, len(detectors))
+
+	// todo: rethink this life choice...
+	for name, enabled := range detectors {
+		if name != "" && enabled {
```
another-rex: We can probably invert some of these if statements

cmd/osv-scanner/scan/image/command_test.go
```diff
@@ -283,6 +319,36 @@ func TestCommand_OCIImage_JSONFormat(t *testing.T) {
...
+				testcmd.OnlyIDVulnsRule,
+				testcmd.OnlyFirstBaseImage,
+				testcmd.AnyDiffID,
+			},
+		},
+		{
+			Name: "scanning_insecure_alpine_image_with_detector_preset",
+			Args: []string{
+				"", "image", "--format=json",
+				"--experimental-detectors", "weakcreds",
```
another-rex, line 341: I'm surprised this is running at all, since in ValidatePluginRequirements in scalibr it should fail if we are running some of these extractors on containers.

another-rex: Oh, ScanContainer is not checking for capabilities... that's not intended I believe.

another-rex (commented):
To make the detector presets more useful, I think we need to do the capabilities filtering ourselves on OSV-Scanner's side, and auto-disable detectors that have requirements that doesn't match.

I'm a bit confused on the windows specific errors, since for container scanning we should be scanning a linux system, even on windows, so I'm not sure why there are those errors appearing.

cmd/osv-scanner/scan/image/command_test.go
another-rex: *and we are also not passing in any capabilities

another-rex (approved):

another-rex (approved):

PR #2024, opened 2025-07-07
test: rename `scan image` JSON format cases

G-Rath:
Only one of these tests actually uses `--all-packages` but they all output in JSON format so I think this is a better name and that also means its reasonable for us to add more cases such as in #2021


G-Rath:
@another-rex it happened in CI! (cli v3 nil pointer panic)

image

PR #2029, opened 2025-07-07
fix(output): don't include package type in name for vertical format

G-Rath:
It doesn't seem like it's that useful and we don't include it in other formats - i fact right now we're explicitly removing it as part of the table format

### DIFF
--- a/internal/output/output_result.go
+++ b/internal/output/output_result.go
@@ -372,7 +372,7 @@ func processSource(packageSource models.PackageSource) map[string]SourceResult {
 	// If no packages with issues are found, mark the ecosystem as empty.
 	if len(packageSource.Packages) == 0 {
 		sourceResults[""] = SourceResult{
-			Name:      packageSource.Source.String(),
+			Name:      packageSource.Source.Path,
 			Type:      packageSource.Source.Type,
 			Ecosystem: "",
 			Packages:  []PackageResult{},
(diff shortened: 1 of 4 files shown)


internal/output/output_result.go
```diff
@@ -372,7 +372,7 @@ func processSource(packageSource models.PackageSource) map[string]SourceResult {
 	// If no packages with issues are found, mark the ecosystem as empty.
 	if len(packageSource.Packages) == 0 {
 		sourceResults[""] = SourceResult{
-			Name:      packageSource.Source.String(),
```
hogo6002, line 375: HTML output uses this to print out the type. If you want to remove the type from Name, you might need to also modify https://github.com/google/osv-scanner/blob/dbac12ea035c7fcf8badb958e915ca38246bb8d9/internal/output/html/package_view_template.gohtml#L7 to add the type back.

I don't think it's bad to have type printing out though. This is helpful for container scanning.

hogo6002: I think we don't include the type in the table output mainly because of the width of the table. But we don't have this concern for vertical or HTML outputs, so I think we can try to provide more details there.

G-Rath: Two issues with that:
  1. this function is currently assuming all formats want to display that when that's not the case, so the table format then needs to undo this with a chain of ugly `TrimPrefix` calls which feels more brittle than having the other formats handle this themselves (as they already have access to both the `Name` and the `Type`)
  2. our output fixtures are not entirely representative of the real world as I believe we always provide the absolute path which can be a lot longer, so even for the vertical output there's a good chance this can be get pushed over 80 chars so including the type has a not-zero cost (see #2034 for more, though it still doesn't entirely show this off because we replace most of the path with `<rootdir>` as part of our snapshot normalizations)

G-Rath:
@hogo6002 I'm going to close this for now as it only changes the vertical output in a bikesheddable way which I don't think is that critical right now - I do still stand by my comments that it's a smell that we're having to undo this in the table formatter specifically, but addressing that would involve a bigger change to our intermediate output struct and again I don't think that's critical right now.

We should keep this in mind though for future changes to the formats and struct for opportunities where we could do some of this cleanup at the same time

PR #2034, opened 2025-07-07
test(output): make paths absolute

G-Rath:
We expect the source information given to `output` to always be an absolute path, but our internal fixtures are currently using relative paths, so this changes that.

Note that this is about making the existing snapshots accurate to our current output even if that isn't something we like - if anyone feels something should be changed as a result of this, please file an issue


G-Rath:
@another-rex so I think I've got this sorted for the `output` package, but it looks like container scanning does pass through relative paths:

image

This seems to be by design e.g. for `lib/apk/db/installed`, that is [the exact path being checked for by the `os/apk` extractor](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/os/apk/apk.go#L91-L93), which makes sense as the `Path` property is meant to be [relative to the `Root`](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/filesystem.go#L73-L74) - so if we're saying `output` should always be given absolute paths, we might need to have the container path sort that out...

That shouldn't be a blocker for this though so I can create an issue and we can go from there - there's also the question on if we want to include any runtime or test-only checks to help catch this in future

(since I'm about to remove the check, here's a link to the run: https://github.com/google/osv-scanner/actions/runs/16154653469/job/45594387172?pr=2034)

another-rex:
It should be a simple fix, as all container scanning should be relative to root, we just need to add a loop in ScanContainer to add `/` to all Locations that pop out of ScanContainer.

another-rex:
Or try setting StoreAbsolutePath to be true in the config for ScanContainer() call.

another-rex:
Need to merge and rerun the snapshots, because of #2064

PR #2036, opened 2025-07-08
feat(grouping): Include upstream field in grouping

### LINKED ISSUE #2027: Use `upstream` field to group vulns
OSV.dev has added an `upstream` field to demonstrate the relationship between Linux distribution vulnerabilities and source CVEs. 

Previously, OSV.dev used `related` for this. The current implementation for grouping IDs is that records with `DSA` and `USN` prefixes, we add [their `related` group into the `aliases` group ID, ](https://github.com/google/osv-scanner/blob/2e440822df1bdb65b3466a960d5060f53b1fd185/internal/grouper/grouper_models.go#L31) allowing them to be merged into one group. We then give [USN and DSA higher priority](https://github.com/google/osv-scanner/blob/dc11dde17fb536c570c5d084f152d65a52ea9dbb/internal/identifiers/identifiers.go#L9) so those IDs are used as the primary IDs, and all other upstream CVEs are hidden.

Image

To support the `upstream` field, we first need to [add `upstream IDs` into `alias IDs` for all ecosystems](https://github.com/google/osv-scanner/blob/2e440822df1bdb65b3466a960d5060f53b1fd185/internal/grouper/grouper_models.go#L34), also make sure no duplicates occur.

Next, we need to compute the priority of each ID prefix to select a primary ID from the list. This can be done by continuing with the [hardcoding method](https://github.com/google/osv-scanner/blob/dc11dde17fb536c570c5d084f152d65a52ea9dbb/internal/identifiers/identifiers.go#L9), where we maintain a list of Linux distro advisory prefixes and assign them a higher priority score. Alternatively, a more programmatic approach would be to re-do the sort logic for the ID list and place the ones with the most upstream (`USN` and `DSA`) as their primary ID. Both methods work, but the first is much easier to implement.

jess-lowe:
Closes #2027 and reduces reliance on hard coding prefix ordering. 

Instead, we are utilizing the logic that the furthest downstream vuln will always have more upstreams than its upstream to order and display the furthest downstream report.

### DIFF
--- a/internal/grouper/grouper_models.go
+++ b/internal/grouper/grouper_models.go
@@ -1,8 +1,10 @@
 package grouper
 
 import (
+	"slices"
 	"strings"
 
+	"github.com/google/osv-scanner/v2/internal/identifiers"
 	"github.com/ossf/osv-schema/bindings/go/osvschema"
 )
 
@@ -13,6 +15,9 @@ type IDAliases struct {
 
 func ConvertVulnerabilityToIDAliases(c []osvschema.Vulnerability) []IDAliases {
 	output := []IDAliases{}
+
+	slices.SortFunc(c, identifiers.MostUpstreamsOrder)
+
 	for _, v := range c {
 		idAliases := IDAliases{
 			ID:      v.ID,
@@ -23,12 +28,14 @@ func ConvertVulnerabilityToIDAliases(c []osvschema.Vulnerability) []IDAliases {
 		// all related CVEs should be bundled together, as they are part of this DSA.
 		// TODO(gongh@): Revisit and provide a universal way to handle all Linux distro advisories.
 		if strings.Split(v.ID, "-")[0] == "DSA" {
+			idAliases.Aliases = append(idAliases.Aliases, v.Upstream...)
 			idAliases.Aliases = append(idAliases.Aliases, v.Related...)
 		}
 
 		// For Ubuntu Security Advisory data,
 		// all related CVEs should be bundled together, as they are part of this USN.
 		if strings.Split(v.ID, "-")[0] == "USN" {
+			idAliases.Aliases = append(idAliases.Aliases, v.Upstream...)
 			idAliases.Aliases = append(idAliases.Aliases, v.Related...)
 		}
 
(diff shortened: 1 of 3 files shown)


internal/grouper/grouper_models.go
```diff
@@ -23,12 +28,14 @@ func ConvertVulnerabilityToIDAliases(c []osvschema.Vulnerability) []IDAliases {
 		// all related CVEs should be bundled together, as they are part of this DSA.
 		// TODO(gongh@): Revisit and provide a universal way to handle all Linux distro advisories.
 		if strings.Split(v.ID, "-")[0] == "DSA" {
+			idAliases.Aliases = append(idAliases.Aliases, v.Upstream...)
```
hogo6002: We should be able to apply this to all vulnerabilities right? Not only for DSA and USN. This also means we should stop appending related ones to it. But given the fact that Ubuntu has updated everything from "related" into "upstream," maybe we should write a "todo" here to remind ourselves.

hogo6002 (approved):
LGTM

internal/output/output_result.go
```diff
@@ -491,9 +491,6 @@ func processVulnGroups(vulnPkg models.PackageVulns) (map[string]VulnResult, map[
 	hiddenVulnMap := make(map[string]VulnResult)
 
 	for _, group := range vulnPkg.Groups {
-		slices.SortFunc(group.IDs, identifiers.IDSortFunc)
-		slices.SortFunc(group.Aliases, identifiers.IDSortFunc)
```
hogo6002, line 495: One thing we need to confirm with the HTML output is that we want to show the CVE ID as the first alias. I know a CVE isn't really an alias of a USN or DSA, but we still want to display its CVE ID because many scanners default to printing CVEs as their vulnerability ID. We want to avoid confusing people when comparing our results with others. We can rename "Aliases" to something more reasonable, but just ensure the HTML output displays the CVE as the first item under the current "Aliases" column. (this is also why we want to print out aliases there)
image

jess-lowe:
will also close #2039

PR #2042, opened 2025-07-09
chore: add a basic Makefile

G-Rath:
This should make it a bit easier to get going with the codebase since you can easily run the most popular commands required for finalizing changes, and hopefully will also make it easier to discover our other scripts for more specific and advanced tasks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ test:
 
 clean:
 	rm -f osv-scanner
+	rm -r internal/image/fixtures/test-*.tar
 
 update-snapshots:
 	UPDATE_SNAPS=true scripts/run_tests.sh


jess-lowe (approved):
Yay! LGTM

cuixq (approved):

another-rex (commented):
Can you also add some commands for:
- Building images
- Updating snapshots
- Rebuild images

G-Rath:
@another-rex when you say images, are you meaning the test images or the release images?

If you mean the test images, I'm on the fence about having make commands for those as the images are already built by the test script if required (i.e. when running acceptance tests), and the build script itself has a few flags (`--force` and `--no-cache`) so I feel like having a make command could start us down that path of having a command for every possible option (with fun long names) when it would be better to just have people use the scripts.

I think it could make sense though for the `clean` command to include removing the images, which would make it easier for us to say to people "run `make clean test`"

another-rex:
Yeah I was referring to test images, but agree that we probably shouldn't add that many commands. Happy with adding the clean command for cleaning test images as well.

Issue #1986, opened 2025-06-20
govulncheck Enricher migration


another-rex:
We need to migrate govulncheck source code scanning to the enricher interface. Currently govulncheck is called during buildResults function to find all the go.mod source files that we scanned, and essentially running `govulncheck ./... -db [vulns-we-found-in-gvc-format]`

For the MVP of it being an enricher, it should:

- Loop through the packages for go.mod
- Find the path in the actual file system (joining scan root and path together), currently govulncheck doesn't support virtual filesystem.
- Run govulncheck with `path/...`, but without the `-db`, as we currently are not executing enrichers in order to get the vulns to pass in, so without db it'll query directly against the go vulndb for all the vulns.
- **can't do this part yet until enrichers are properly hooked up** Filter vuln findings based on the govulncheck result.

PR #2045, opened 2025-07-09
feat: add govulncheck enricher

hogo6002:
Adds `VulnerableCodeNotInExecutePath` signals to uncalled Go vulns.

closes https://github.com/google/osv-scanner/issues/1986


another-rex (commented):
Can you implement it end to end and replace the existing govulncheck? I think you can just pass it through, and it's a big benefit of implementing it in osv-scanner first.

hogo6002:
> Can you implement it end to end and replace the existing govulncheck? I think you can just pass it through, and it's a big benefit of implementing it in osv-scanner first.

I don't think we can replace the existing govulncheck for now. If i recall correctly, we're currently matching vulnerabilities after calling osv-scalibr, right? We will need to pass a list of vulns into this enricher first to add signals.
Also the existing govulncheck handles aliases, but the new enricher doesn't do it.

this PR also only focuses on adding an govulncheck enricher. If we're able to replace govulncheck, I'll do it in a separate PR.

internal/scalibrenricher/govulncheck/source/govulncheck_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/google/osv-scalibr/extractor/filesystem/language/golang/gomod"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/inventory/vex"
+	"github.com/google/osv-scalibr/purl"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+)
+
+const testdata = "./testdata"
+const reachableVuln = "GO-2023-1558"
```
another-rex: nit: rename to reachableVulnID  same with the unreachable one. It was a bit confusing to me looking at the switch statement later

internal/scalibrenricher/govulncheck/source/govulncheck.go
```diff
@@ -0,0 +1,194 @@
...
+			})
+		}
+	}
+}
+
+func (e *Enricher) runGovulncheck(ctx context.Context, absModDir string, goVersion string) (map[string][]*Finding, error) {
+	cmd := scan.Command(ctx, "-C", absModDir, "-json", "./...")
+	var b bytes.Buffer
+	cmd.Stdout = &b
+	cmd.Env = append(os.Environ(), "GOVERSION=go"+goVersion)
```
another-rex, line 148: Hmm... is this necessary? if so, you need to get the go version "per location", as different mod files might have different goVersions.

hogo6002: Updated the Enrich() function to get go versions from different mod files and pass into this function.

another-rex (approved):
LGTM

Btw osv-scalibr has now added the Findings/Vulns to Package link field.

PR #2051, opened 2025-07-10
fix: properly flag if running on Linux or Mac OSs

G-Rath:
The `OSUnix` value is only meant to be used by plugins to indicate they're fine working with either Linux or macOS, but [the OS still needs to be one of those values](https://github.com/google/osv-scalibr/blob/ed2676c6ae1b/plugin/plugin.go#L119-L125) when scanning.

This is not an issue currently because we're just using extractors, but there are a few detectors that require a Unix-like OS

### DIFF
--- a/pkg/osvscanner/scan.go
+++ b/pkg/osvscanner/scan.go
@@ -148,21 +148,19 @@ func scan(accessors ExternalAccessors, actions ScannerActions) ([]imodels.Packag
 	}
 
 	testlogger.BeginDirScanMarker()
+	osCapability := determineOS()
+
 	// For each root, run scalibr's scan() once.
 	for root, paths := range rootMap {
 		capabilities := plugin.Capabilities{
 			DirectFS:      true,
 			RunningSystem: true,
 			Network:       plugin.NetworkOnline,
-			OS:            plugin.OSUnix,
+			OS:            osCapability,
 		}
 
 		if actions.CompareOffline {
 			capabilities.Network = plugin.NetworkOffline
-		}
-
-		if runtime.GOOS == "windows" {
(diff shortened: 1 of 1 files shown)


pkg/osvscanner/scan.go
```diff
@@ -240,3 +238,16 @@ func getRootDir(path string) string {
...
+func determineOS() plugin.OS {
+	switch runtime.GOOS {
+	case "windows":
+		return plugin.OSWindows
+	case "darwin":
+		return plugin.OSMac
+	case "linux":
+		return plugin.OSLinux
+	default:
+		return plugin.OSAny
```
G-Rath, line 253: I'm pretty sure its safe to return this rather than error, though (assuming we don't actually want to error) it could be nice if scalibr had a `plugin.OSUnknown` either as a new value or an alias for this one, since that's what we want to represent here

G-Rath:
@another-rex I found this while trying to come up with tests for the detectors - https://github.com/google/osv-scanner/pull/2021/commits/afa06df7472a11e9f00fa8802eac01ba474a1cc7 shows that this is broken.

I've opened this so I don't forget, but I think it's probably best if we first get the snapshot failures on #2021 sorted before we dig into everything else including this

another-rex (approved):

pkg/osvscanner/scan.go
another-rex: Hmm, can this log a warning that OS "xyz" is not officially supported.

another-rex:
Talked with other Scalibr folks, I think we're happy to add a OSUnknown in scalibr that just only matches OSAny.

PR #2053, opened 2025-07-10
test(cmd): add some more cases

G-Rath:


cmd/osv-scanner/update/__snapshots__/command_test.snap
```diff
@@ -1,4 +1,157 @@
...
+
+---
+
+[TestCommand/errors_with_invalid_data_source - 2]
+unsupported data-source "github" - must be one of: deps.dev, native
+
+---
+
+[TestCommand/errors_with_invalid_data_source - 3]
+<?xml version="1.0" encoding="UTF-8"?>
```
another-rex, line 12: Huh, shouldn't this be empty if the input flags are invalid? @cuixq

cuixq: I think the snapshot above is what expected:
```
unsupported data-source "github" - must be one of: deps.dev, native
```

another-rex: Oh, the -3 is just testing the file, in this case nothing has changed. I think there might be  a clearer way to do this, but that's fine for now.

G-Rath:
turns out I had a few leftover snapshots and our CI doesn't actually flag that - I've opened https://github.com/gkampitakis/go-snaps/issues/130 to see if we can make that a thing

(it's not the biggest deal for us though as our "update snapshots" workflow should catch this as a side effect)

PR #2058, opened 2025-07-11
test(cmd): introduce helper for copying file-based flags

G-Rath:
This is branched off #2053


another-rex (approved):
~~I think this might be a bit too clever?~~

Oh that was the old comment

PR #2063, opened 2025-07-14
ci: adjust permissions of reusable workflow

G-Rath:
Nested workflows can only be "the same or more restrictive" so by setting all permissions to none at the workflow level, we remove the jobs access to whatever permissions the initial caller workflow allowed

Closes https://github.com/google/osv.dev/pull/3663


jess-lowe (approved):

cuixq (approved):

another-rex:
I think the files that need to be updated are in osv-scanner-action repository, not this one, though it's probably still good to update it.

another-rex:
Huh wait I'm confused, osv-scanner-action always had these permissions at the top level, so what was the issue?

PR #2065, opened 2025-07-14
fix: Go releaser nightly

another-rex:
Turns out the nightly config I made didn't actually use the nightly config 🤦

### DIFF
--- a/.github/workflows/goreleaser-nightly.yml
+++ b/.github/workflows/goreleaser-nightly.yml
@@ -4,6 +4,7 @@ on:
   schedule:
     # we want the nightly builds only on work days
     - cron: "0 0 * * 2-6"
+  workflow_dispatch:
 
 # Restrict jobs in this workflow to have no permissions by default; permissions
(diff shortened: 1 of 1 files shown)


jess-lowe (approved):

.github/workflows/goreleaser-nightly.yml
```diff
@@ -4,6 +4,7 @@ on:
   schedule:
     # we want the nightly builds only on work days
     - cron: "0 0 * * 2-6"
+  workflow_dispatch:
```
cuixq, line 8: what is this for? and it's empty?

another-rex: It just allows us to manually run the workflow, rather than fully reliant on the schedule. You can see it for our snapshot workflow:

https://github.com/google/osv-scanner/blob/main/.github/workflows/snapshots.yml

another-rex: I'll add a comment for it

PR #2068, opened 2025-07-14
fix(output): change links to go to vuln page instead of list page

### LINKED ISSUE #2066: Point to vuln record not list view in scanner output
Image

As we now have redirect logic in OSV.dev that redirects to the search page if the record isn't found, it makes more sense to direct directly to the vuln record, which should have links to everything that relates to that vuln.

jess-lowe:
Closes #2066


jess-lowe:
> looks good, though I just realized we're doing `https://osv.dev/<id>` for other outputs like table which gets redirected to /vuln - is there a reason why we weren't doing that here before and also now?
> 
> (we do want to keep using it in places like the table output because its shorter, which isn't as important for these outputs but it would still be a little less data and consistent)

Not sure why, but updated to use the shorter vers regardless.

another-rex (approved):

another-rex:
We were not doing that before here because before  the CVE didn't exist in OSV, and we didn't have the redirect to the search page, and just 404s. Now we have the redirect we can now point directly to the short link.

PR #2072, opened 2025-07-14
add text field into shortdescription and fulldescription, update snap…

### LINKED ISSUE #2071: Invalid SARIF, shortDescription requires property "text", fullDescription requires property "text"
Hi there,

After a bit of investigation, I think there's a regression in the latest 2.1.0 version (at least, assuming that GitHub's sarif submission function is correct!)

It's been failing my scheduled GitHub actions since July 12, 05:13 - which was my next run after the latest release dropped on July 11. I automatically pick up the latest version, so I started seeing the problem over the weekend and picked it up today.

- [GitHub actions runs evidence])(https://github.com/brabster/terraform-bootstrap-gcp/actions?query=event%3Aschedule+)
- [A specific run](https://github.com/brabster/terraform-bootstrap-gcp/actions/runs/16257878961)

Prior runs uploaded fine, now I now see this error:

```
Unable to upload "osv_scan_results.sarif" as it is not valid SARIF:
- instance.runs[0].tool.driver.rules[0].shortDescription requires property "text"
- instance.runs[0].tool.driver.rules[0].fullDescription requires property "text"
- instance.runs[0].tool.driver.rules[1].shortDescription requires property "text"
...
```

brabster:
…shots

Fixes #2071 

I think the code and snapshots were updated as part of #1985, but the required "text" property in the *description fields was lost and not detected.

I think this change restores the previous behaviour with respect to these fields - I can't yet be sure there are no other breaking changes

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -265,7 +265,8 @@ Loaded Alpine local db from <tempdir>/osv-scanner/Alpine/all.zip
                 "GHSA-whgm-jr23-g3j9"
               ],
               "fullDescription": {
-                "markdown": "This affects all versions of package ansi-html. If an attacker provides a malicious string, it will get stuck processing the input for an extremely long time."
+                "markdown": "This affects all versions of package ansi-html. If an attacker provides a malicious string, it will get stuck processing the input for an extremely long time.",
+                "text": "This affects all versions of package ansi-html. If an attacker provides a malicious string, it will get stuck processing the input for an extremely long time."
               },
               "help": {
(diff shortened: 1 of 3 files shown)


brabster:
Hm, just checked an attempt to upload to GitHub, seems there might be more issues than those two.

https://github.com/brabster/terraform-bootstrap-gcp/actions/runs/16271328991/job/45939640440

another-rex (approved):

another-rex:
Huh interesting, I'm not sure how you got the field "instance", as it doesn't show up in any of the tests snapshots we have in osv-scanner.

another-rex:
Seems to be working: https://github.com/google/osv.dev/actions/runs/16281807862/job/45972784285 :)

another-rex:
I'll push an update to osv-scanner-action to make sure people using the github action gets the fix immediately, and we'll followup with a patch release next week. There has been some other changes merged in the mean time that we want to have additional bake time.

PR #2074, opened 2025-07-14
test(ci): add additional cases

G-Rath:
This takes us to 100% coverage for this package 🎉


internal/ci/vulnerability_result_diff_test.go
```diff
@@ -67,6 +67,14 @@ func TestDiffVulnerabilityResults(t *testing.T) {
 			// opposite of above test case, result should be empty
 			wantPath: "fixtures/vulns/test-vuln-diff-b-c.json",
 		},
+		{
+			args: args{
+				oldRes: testutility.LoadJSONFixture[models.VulnerabilityResults](t, "fixtures/vulns/test-vuln-results-a.json"),
+				newRes: testutility.LoadJSONFixture[models.VulnerabilityResults](t, "fixtures/vulns/test-vuln-results-d.json"),
+			},
+			// `newRes` has a package that is not in `oldRes` and has no vulns, so result should just be empty
+			wantPath: "fixtures/vulns/test-vuln-diff-b-c.json",
```
G-Rath: @another-rex its not a big deal, but it feels like we could use snapshots to handle the "want" path rather than needing to have dedicated files?

another-rex: Yeah good idea! let's update that to be snapshots.

G-Rath: haha I already did this in https://github.com/google/osv-scanner/pull/717, I just apparently didn't remove the old fixtures

another-rex (approved):

another-rex:
For a followup PR, can the snapshot matching be replaced with MatchText with JSONReplacements? There's a lot of text crowding out the actual diffs making it hard to see, and since it's just the standard osv-scanner JSON output, we should be able to reuse the existing replacements we got.

G-Rath:
@another-rex I think it would be better to just make the input fixtures smaller since JSONReplacements is currently internal to the CMD test package as we've not had a use for it elsewhere, and I think having smaller fixtures will make things even easier to manage

Issue #1964, opened 2025-06-16
Update Ubuntu unimportant filter


hogo6002:
We currently rely on the ubuntu_priority tag from the ecosystem_specific field to determine if an Ubuntu CVE is negligible. Since the new OSV schema (https://github.com/ossf/osv-schema/pull/337) is moving this tag into the severity field, we'll need to update our osv-scanner logic to ensure our 'unimportant' vulnerability filtering still works as intended.

PR #2076, opened 2025-07-15
feat: support ubutnu severity type

hogo6002:
Resolves https://github.com/google/osv-scanner/issues/1964

Note: some Ubuntu records have moved their priority tags from ecosystem-specific to severity fields, but osv.dev has failed to provide a correct severity type. For now, we check both empty severity tags and Ubuntu severity tags to see if a vulnerability has been marked as ineligible


cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
```diff
@@ -2243,9 +2247,9 @@ Scanning local image tarball "../../../../internal/image/fixtures/test-node_modu
             }
           },
           "vulnerabilities": [
+            "UBUNTU-CVE-2022-3219",
```
hogo6002, line 2250: this looks like is caused by upstream ordering? Do we put the one with the most amount of upstream at the top or the one with the least upstream at the top? UBUNTU-CVE-2022-3219 recently added an upstream field. But i think we wanted to put USN at the top, right? @jess-lowe

internal/image/fixtures/ubuntu20-04-unimportant-config.toml
```diff
@@ -67,3 +67,8 @@ ignore = true
 name = "gnutls28"
```
another-rex, line 67: Can you try removing this? I think we don't need this once you add the severity check back in

hogo6002: well, it will fail if I remove it. It has 4 vulns from that package, so I think the new one was actually important vulns.

Issue #2077, opened 2025-07-15
[BUG] Error dependencies in setup.cfg


komgritz:
We have facing issue on OSV docker image version v2.1.0 when scan python source code so There is error message below

`(extracting as python/requirementsenhanceable) failed resolving: sdist: dependencies in setup.cfg, not in PKG-INFO`

But we works fine on previous version v2.0.3

martijnmumbles:
I've been looking into the exact same issue. I want to further specify it's exclusive to django projects for me. My vanilla python projects seem unaffected. This could be because of related dependencies, but perhaps that's something to go by.

Attached is the most sparse requirements.txt that started failing for me with version 2.1.0

[requirements.txt](https://github.com/user-attachments/files/21228191/requirements.txt)

cuixq:
thank you both for the feedback.

This is due to us enabling the [transitive extractor](https://github.com/google/osv-scanner/blob/main/internal/scalibrextract/language/python/requirementsenhancable/requirementsenhanceable.go#L8) for requirements.txt and we rely on deps.dev's `pypi` package to [extract metadata](https://github.com/google/deps.dev/blob/main/util/pypi/sdist.go#L65) which does not support dependencies in setup.cfg unfortunately.

One potential improvement I can think of is to fallback to the offline extractor if the dependency resolution fails so at least we have some results returned.

cuixq:
This should be fixed in v2.2.0!

PR #2079, opened 2025-07-16
feat: fall back to offline extractor if the transitive one fails

cuixq:
https://github.com/google/osv-scanner/issues/2077

Currently, if the extractor is enhanced but the extraction failed, no result will be returned. A better user experience is to fall back to the offline extractor so at least the direct dependencies are returned.

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -1220,65 +1220,68 @@ Scanned <rootdir>/fixtures/locks-requirements/requirements.prod.txt file and fou
 Scanned <rootdir>/fixtures/locks-requirements/requirements.txt file and found 13 packages
 Scanned <rootdir>/fixtures/locks-requirements/the_requirements_for_test.txt file and found 1 package
 Scanned <rootdir>/fixtures/locks-requirements/unresolvable-requirements.txt file and found 3 packages
-+-------------------------------------+------+-----------+------------+---------+-----------------------------------------------------------+
-| OSV URL                             | CVSS | ECOSYSTEM | PACKAGE    | VERSION | SOURCE                                                    |
-+-------------------------------------+------+-----------+------------+---------+-----------------------------------------------------------+
-| https://osv.dev/PYSEC-2023-62       | 8.7  | PyPI      | flask      | 1.0     | fixtures/locks-requirements/my-requirements.txt           |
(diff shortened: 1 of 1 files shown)


cuixq:
@another-rex @G-Rath can you take a look on this PR see if this looks reasonable to you? thanks.

cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -880,6 +880,11 @@ func TestCommand_Transitive(t *testing.T) {
 			Args: []string{"", "source", "--config=./fixtures/osv-scanner-empty-config.toml", "./fixtures/locks-requirements/requirements.txt"},
 			Exit: 1,
 		},
+		{
+			Name: "uses native data source for requirements.txt",
```
G-Rath:
```suggestion
			Name: "uses_native_data_source_for_requirements_txt",
```

internal/scalibrextract/language/java/pomxmlenhanceable/pomxmlenhanceable.go
```diff
@@ -17,11 +17,15 @@ const (
 
 // Extractor extracts Maven packages from pom.xml files.
 type Extractor struct {
-	actual filesystem.Extractor
+	base     filesystem.Extractor
```
G-Rath: nit: I'd probably call these `online` and `offline` since that's the heart of them

internal/scalibrextract/language/python/requirementsenhancable/requirementsenhanceable.go
```diff
@@ -17,12 +17,14 @@ const (
 
 // Extractor extracts python packages from requirements.txt files.
 type Extractor struct {
-	actual filesystem.Extractor
+	base     filesystem.Extractor
```
G-Rath: nit: like with the pom extractor, I'd probably call this `online` and `offline` too

G-Rath (approved):
this looks fine, though I think it speaks to how useful it'd be to have enrichers supported 😅

another-rex (approved):

moritzschmitz-oviva:
@cuixq is it possible this introduced a bug?

Please see my question here: https://github.com/google/osv-scanner/discussions/2163.

In the discussion attached there is the `quarkus-vertx` dependency explicitly mentioned. It cannot be found on deps.dev.

As a result in V2.1 the run fails. In v2.2 it doesn't fail no more, but it also doesn't list the other vulnerabilities any more.

cuixq:
@moritzschmitz-oviva I replied to that question before seeing the comment here so I copied the information there. I will post my replies in that thread!

PR #2084, opened 2025-07-17
fix: store absolute paths when scanning containers

G-Rath:
Turns out container scanning is not using absolute paths in its results, which isn't breaking anything but is an assumption of our `output` package


another-rex:
Interesting, it looks like the store absolute path option breaks layer tracing

another-rex:
Opened this to fix it: https://github.com/google/osv-scalibr/pull/916

hogo6002 (approved):

PR #2086, opened 2025-07-18
chore(lints+comments): Add package comments to everything

another-rex:
Spent a bit of time to just add comments to all the offending packages.

Closes: #1720


internal/datasource/cache.go
```diff
@@ -1,3 +1,4 @@
+// Package datasource provides data sources for Guided Remediation
```
cuixq:
```suggestion
// Package datasource provides data sources for dependency resolution used by transitive scanning and guided remediation.
```

internal/remediation/suggest/suggest.go
```diff
@@ -1,3 +1,4 @@
+// Package suggest provides functionality for suggesting remediation for vulnerabilities.
```
cuixq:
```suggestion
// Package suggest provides functionality for suggesting version updates.
```

michaelkedar (approved):

PR #2095, opened 2025-07-21
fix: pass through plugin capabilities

another-rex:
Also added a test for No vulnerabilities but with license violations. (about #2093, it looks like this doesn't actually trigger the removed codepath either suprisingly)

### DIFF
--- a/cmd/osv-scanner/scan/source/command_test.go
+++ b/cmd/osv-scanner/scan/source/command_test.go
@@ -761,6 +761,11 @@ func TestCommand_Licenses(t *testing.T) {
 			Exit: 1,
 		},
 		{
+			Name: "No vulnerabilities but license violations with allowlist",
+			Args: []string{"", "source", "--licenses=Apache-2.0", "--config=./fixtures/osv-scanner-empty-config.toml", "./fixtures/locks-many/yarn.lock"},
+			Exit: 1,
+		},
+		{
 			Name: "Vulnerabilities and all license violations allowlisted",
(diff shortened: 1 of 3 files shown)


pkg/osvscanner/osvscanner.go
```diff
@@ -340,16 +340,18 @@ func DoContainerScan(actions ScannerActions) (models.VulnerabilityResults, error
...
 	}
 
-	plugins = plugin.FilterByCapabilities(plugins, &plugin.Capabilities{
+	capabilities := &plugin.Capabilities{
 		DirectFS:      true,
 		RunningSystem: false,
 		OS:            plugin.OSLinux,
-	})
+	}
+	plugins = plugin.FilterByCapabilities(plugins, capabilities)
```
cuixq, line 348: does this make any functional change?

cuixq (approved):

cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -760,6 +760,11 @@ func TestCommand_Licenses(t *testing.T) {
 			Args: []string{"", "source", "--licenses=MIT", "--config=./fixtures/osv-scanner-empty-config.toml", "./fixtures/locks-many/package-lock.json"},
 			Exit: 1,
 		},
+		{
+			Name: "No vulnerabilities but license violations with allowlist",
```
G-Rath, line 764: I think this is still good to have but fyi we've already got this covered with "some packages with license violations in json"  about 4 cases down, which is why #2093 doesn't get covered by this 🤷

pkg/osvscanner/osvscanner.go
another-rex: Because we are filtering already, it doesn't really change anything functional.

PR #2096, opened 2025-07-21
feat: Allow log handler overriding

### LINKED ISSUE #2081: Allow configuration of the library's logging
The OpenSSF Scorecard project uses osv-scanner as a library, specifically `osvscanner.DoScan`. I experimented briefly with the migration to v2 recently and started getting a flood of logs to our stderr which are originating from osv-scanner.

```
2025/07/16 11:03:49 INFO Scanning dir <...>
2025/07/16 11:04:15 ERROR Error during extraction: (extracting as java/pomxmlnet) failed resolving
2025/07/16 11:04:39 INFO Scanned <...>/package-lock.json file and found 1568 packages
```

Is there anyway to introduce a control to disable `cmdlogger`'s output? It seems most of the codebase uses `cmdlogger.Info` (and friends), which write to the default (global) slog handler.
https://github.com/google/osv-scanner/blob/6515dc09a74355444b56c7dbf85b025bf9161b94/internal/cmdlogger/fmt.go#L8-L22

I can set the slog default to something which would ignore it, like [slog.DiscardHandler](https://pkg.go.dev/log/slog#DiscardHandler), but this would break many of our own logging.

another-rex:
Resolves #2081

### DIFF
--- a/cmd/osv-scanner/scan/image/command_test.go
+++ b/cmd/osv-scanner/scan/image/command_test.go
@@ -370,6 +370,7 @@ func TestCommand_OCIImage_JSONFormat(t *testing.T) {
 
 func TestCommand_HtmlFile(t *testing.T) {
 	t.Parallel()
+	testutility.SkipIfNotAcceptanceTesting(t, "Needs container image")
 
 	testDir := testutility.CreateTestDir(t)
 
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -548,7 +548,7 @@ func overrideGoVersion(scanResults *results.ScanResults) {
 	}
 }
 
-// SetCmdLogger sets the global slog handler for the cmdlogger.
-func SetCmdLogger(handler slog.Handler) {
+// SetLogger sets the global slog handler for the cmdlogger.
+func SetLogger(handler slog.Handler) {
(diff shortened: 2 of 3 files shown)


cmd/osv-scanner/scan/image/command_test.go
```diff
@@ -370,6 +370,7 @@ func TestCommand_OCIImage_JSONFormat(t *testing.T) {
 
 func TestCommand_HtmlFile(t *testing.T) {
 	t.Parallel()
+	testutility.SkipIfNotAcceptanceTesting(t, "Needs container image")
```
G-Rath, line 373: oops yeah good point 😅

pkg/osvscanner/osvscanner.go
```diff
@@ -546,3 +547,8 @@ func overrideGoVersion(scanResults *results.ScanResults) {
 		}
 	}
 }
+
+// SetCmdLogger sets the global slog handler for the cmdlogger.
+func SetCmdLogger(handler slog.Handler) {
```
G-Rath: I feel like `SetLogger` would be more idiomatic, as the "cmd" part is an implementation detail right? and our goal is _avoid_ having more than one logger that would necessitate more of these functions...

another-rex: Good idea, updated this.

PR #2098, opened 2025-07-22
ci: have `osv-scalibr` always kept up to date

G-Rath:
This should have Renovate do dedicated PRs for updating `osv-scalibr`, allowing us to ensure we can upgrade to the latest version


G-Rath:
@another-rex I think this is mostly what we want, because of [`updateNotScheduled`](https://docs.renovatebot.com/configuration-options/#updatenotscheduled) which defaults to `true` and effectively means "schedule" controls when branches are _created_ - so once a PR has been created, by default Renovate will keep it up to date ASAP regardless of the schedule

another-rex (approved):

PR #2100, opened 2025-07-22
feat: move resolving extractors from names into `osvscanner`

G-Rath:
This puts the API of our CLI and public package back in sync and make it easier to build on since we should not have to do as much double-handling.

As this is expected to be one of a couple of changes to this aspect of the scanner, I've introduced the `scalibrplugin` package as part of this which I expect to eventually house stuff like our presets and builder related code

### DIFF
--- a/cmd/osv-scanner/internal/helper/getters.go
+++ b/cmd/osv-scanner/internal/helper/getters.go
@@ -48,10 +48,8 @@ func GetCommonScannerActions(cmd *cli.Command, scanLicensesAllowlist []string) o
 
 func GetExperimentalScannerActions(cmd *cli.Command) osvscanner.ExperimentalScannerActions {
 	return osvscanner.ExperimentalScannerActions{
-		Extractors: ResolveEnabledExtractors(
-			cmd.StringSlice("experimental-extractors"),
-			cmd.StringSlice("experimental-disable-extractors"),
-		),
+		ExtractorsEnabled:  cmd.StringSlice("experimental-extractors"),
+		ExtractorsDisabled: cmd.StringSlice("experimental-disable-extractors"),
 		Detectors: ResolveEnabledDetectors(
 			cmd.StringSlice("experimental-detectors"),
 			cmd.StringSlice("experimental-disable-detectors"),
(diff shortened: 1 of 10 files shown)


cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
```diff
@@ -1467,6 +1467,7 @@ Total 0 packages affected by 0 known vulnerabilities (0 Critical, 0 High, 0 Medi
 
 [TestCommand_ExplicitExtractors/scanning_directory_with_an_extractor_that_does_not_exist - 2]
 unknown extractor "custom/extractor"
+unknown extractor "custom/extractor"
```
G-Rath, line 1470: This is being printed twice now as it's printed by `builders.BuildExtractors` which gets called when we resolve the enabled/disabled extractors which we do for the lockfiles section and then again for the directories section, whereas previously we'd handle that once in the command.

While not ideal I don't think it's worth blocking this since it should only happen for people trying to enable specific extractors and its an error anyway so they'll have to change their input no matter how many times we print this message.

This will eventually get addressed through some of the follow-up changes, probably even before the next release anyway 🤷

G-Rath:
hmm so I've just realised we take a `map[string]bool` for `CallAnalysisStates` which is very similar to what we're providing here - I think we should be consistent in our package interface, though since these are experimental flags and I want to get into the other changes we have planned, I might just stick with what I have here and revisit changing to a map later down the track

cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
G-Rath: Actually, this should be taken care of by #2102 which can be landed before or after this one

cmd/osv-scanner/internal/helper/getters.go
```diff
@@ -48,13 +48,10 @@ func GetCommonScannerActions(cmd *cli.Command, scanLicensesAllowlist []string) o
...
 	return osvscanner.ExperimentalScannerActions{
-		Extractors: ResolveEnabledExtractors(
-			cmd.StringSlice("experimental-extractors"),
-			cmd.StringSlice("experimental-disable-extractors"),
-		),
-		Detectors: ResolveEnabledDetectors(
-			cmd.StringSlice("experimental-detectors"),
-			cmd.StringSlice("experimental-disable-detectors"),
-		),
+		ExtractorsEnabled:  cmd.StringSlice("experimental-extractors"),
```
another-rex, line 51: I think eventually we want to think about merging these into a single enabled/disabled plugins list to match the scalibr behavior, but happy to leave it as it is for now until we get enrichers working.

G-Rath: yeah that's what I was planning to do next and I've actually started on that locally - I think I'm about halfway through but then I realized having the extractors and detectors using slightly different structures (string vs `InitMap`) made it harder, so then I looked into https://github.com/google/osv-scanner/pull/2103 which looks mostly working

PR #2103, opened 2025-07-23
feat: use extractor presets provided by scalibr

G-Rath:


cmd/osv-scanner/internal/helper/extractors_parser.go
```diff
@@ -1,17 +1,59 @@
...
+		},
+	),
+	"directory": {gitrepo.Name: {gitrepo.New}, vendored.Name: {vendored.New}},
+	"artifact":  concat(list.Artifact, list.InitMap{nodemodules.Name: {nodemodules.New}}),
+}
+
+func without(initMap list.InitMap, omit []string) list.InitMap {
+	result := list.InitMap{}
+
+	for name := range initMap {
```
another-rex: nit: Just maps.Clone and then delete might be better (clearer and less allocations)

G-Rath:
Note this is branched off https://github.com/google/osv-scanner/pull/2108 which I opened first as I wanted to check everything was passing - I've marked that as ready for review since theres no reason _not_ to land it, and that would make the diff for this smaller, but we can also just land this 🤷

internal/scalibrplugin/extractors_parser.go
```diff
@@ -2,17 +2,72 @@
...
-	"directory": scalibrextract.ExtractorsDirectories,
-	"artifact":  scalibrextract.ExtractorsArtifacts,
+var extractorPresets = map[string]list.InitMap{
+	"sbom": list.SBOM,
+	"lockfile": concat(
+		without(list.SourceCode, []string{
+			pomxml.Name, pomxmlnet.Name,
+			requirements.Name, requirementsnet.Name,
+			// this is actually a detector
+			secrets.Name,
```
G-Rath: @another-rex @erikvarga if you agree this should be a detector, I can have a crack at making that move (and that would also mean we'd have a simple well-tested detector for our e2e tests as a nice bonus)

another-rex: Is this a detector? Having a quick look at it, it seems to be correctly implemented as an extractor right?

G-Rath: ugh yeah ok looking over it a bit more I can see how you could classify it as an extractor - I guess then the issue is more that the scanner expects all extractors to give inventories aka packages, whereas this extractor is giving secrets? (and yup I didn't dig deep enough: `Inventory` literally has a `Secrets` field 🤦)

Ultimately, unless anyone really wants this particular extractor enabled by default right now, I think it's still best for us to exclude it for now and have a separate discussion at some point on how to handle it

(my thinking for it being a detector was that it was flagging secrets which shouldn't be in files, but its not its just extracting secrets)

cmd/osv-scanner/internal/helper/extractors_parser.go
G-Rath: oh yeah I forgot about `delete` 😅

internal/scalibrplugin/extractors_parser.go
G-Rath: for reference, this is why I don't think we should have it enabled by default right now

image

another-rex: Yeah agreed, let's remove it from the default

another-rex (approved):
LGTM, thanks!

PR #2118, opened 2025-07-28
chore: adjust pseudo constraint for `osv-scalibr`

G-Rath:
It seems if you use 0.0.0 for a pseudo version, then renovate will want to upgrade to the latest tagged version rather than the latest digest, and that there's generally no explicit way to force renovate to only use digests.

I expect this means we'll need to adjust this again once the next version of osv-scalibr is tagged, but for now we can just set the version back to v0.3.2 🤷

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/go-git/go-billy/v5 v5.6.2
 	github.com/go-git/go-git/v5 v5.16.2
 	github.com/google/go-cmp v0.7.0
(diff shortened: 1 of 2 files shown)


G-Rath:
oh apparently this is a "feature" of `go get -u`? https://github.com/renovatebot/renovate/discussions/24841

> renovate will run go get -u but per go get behavior it will not not look for digest release if the current version is tagged.
>
> To get the latest digest, go mod must target a git branch from which to get the latest digest.

another-rex (approved):

PR #2122, opened 2025-07-29
feat: support Drupal packages

G-Rath:


G-Rath:
This revealed #2126 😄

G-Rath:
This is no longer needed as we're not going to use a dedicated ecosystem for Drupal - I have opened #2323 expanding our test data to cover Drupal packages

PR #2123, opened 2025-07-29
test: fail if there are dirty snapshots or if there is an issue cleaning snapshots

G-Rath:
[`go-snaps` have added support](https://github.com/gkampitakis/go-snaps/issues/130) for signaling if there are dirty snapshots or if there was an issue with cleaning snapshots, so now our snapshot clean utility function will fail the tests if that is the case.

It feels a little meh to be doing an `os.Exit` in a function like this but the alternative would be to do this in every `TestMain` that uses snapshots, and we only have three packages whose `TestMain` does something after cleaning snapshots so its probably fine (for now at least)

### DIFF
--- a/cmd/osv-scanner/scan/source/testmain_test.go
+++ b/cmd/osv-scanner/scan/source/testmain_test.go
@@ -28,7 +28,7 @@ func TestMain(m *testing.M) {
 	testcmd.CommandsUnderTest = []cmd.CommandBuilder{source.Command}
 	m.Run()
 
-	testutility.CleanSnapshots(m)
-
 	os.RemoveAll("./fixtures/.git")
+
+	testutility.CleanSnapshots(m)
 }
--- a/cmd/osv-scanner/scan/testmain_test.go
+++ b/cmd/osv-scanner/scan/testmain_test.go
@@ -28,7 +28,7 @@ func TestMain(m *testing.M) {
 	testcmd.CommandsUnderTest = []cmd.CommandBuilder{scan.Command}
 	m.Run()
 
-	testutility.CleanSnapshots(m)
-
 	os.RemoveAll("./fixtures/.git")
+
+	testutility.CleanSnapshots(m)
 }
--- a/cmd/osv-scanner/testmain_test.go
+++ b/cmd/osv-scanner/testmain_test.go
@@ -34,7 +34,7 @@ func TestMain(m *testing.M) {
 	}
 	m.Run()
 
-	testutility.CleanSnapshots(m)
-
 	os.RemoveAll("./fixtures/.git")
+
+	testutility.CleanSnapshots(m)
 }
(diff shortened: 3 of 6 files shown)


G-Rath:
<details>
<summary>Here is a patch that has a dirty snapshot if folks want to play, and a change to the snapshot cleaner script to enforce the clean happens at the end</summary>

```
Subject: [PATCH] test: clean snapshots last
---
Index: cmd/osv-scanner/__snapshots__/main_test.snap
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/cmd/osv-scanner/__snapshots__/main_test.snap b/cmd/osv-scanner/__snapshots__/main_test.snap
--- a/cmd/osv-scanner/__snapshots__/main_test.snap	(revision 021f00f5b458d3c0a97bbe9f22c80ed536511a7c)
+++ b/cmd/osv-scanner/__snapshots__/main_test.snap	(date 1753828273050)
@@ -33,6 +33,17 @@
 
 ---
 
+[Test_run/version2 - 1]
+osv-scanner version: 2.1.0
+commit: n/a
+built at: n/a
+
+---
+
+[Test_run/version2 - 2]
+
+---
+
 [Test_run_SubCommands/scan_with_a_flag - 1]
 Scanning dir ./fixtures/locks-one-with-nested
 Scanned <rootdir>/fixtures/locks-one-with-nested/nested/composer.lock file and found 1 package
Index: scripts/report_uncleaned_snapshots.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/scripts/report_uncleaned_snapshots.py b/scripts/report_uncleaned_snapshots.py
--- a/scripts/report_uncleaned_snapshots.py	(revision 021f00f5b458d3c0a97bbe9f22c80ed536511a7c)
+++ b/scripts/report_uncleaned_snapshots.py	(date 1753828658103)
@@ -12,9 +12,13 @@
 def does_clean_snapshots(pkg_dir):
   try:
     with open(f'{pkg_dir}/testmain_test.go', 'r') as file:
+      found_cleaning_call = False
+
       for _, line in enumerate(file):
+        if found_cleaning_call:
+          return line == '}\n'
         if '	testutility.CleanSnapshots(m)' in line:
-          return True
+          found_cleaning_call = True
     return False
   except FileNotFoundError:
     return False
```
</details>

Alternatively we could `defer` the cleanup call 🤷

hogo6002 (approved):

another-rex:
We really should get that segfault in the CLI fixed.

G-Rath:
yeah I was going to ask you about it the other week but then forgot - I'll dig back into it next week

PR #2131, opened 2025-08-01
Add the main file for package reachability level

p1gc0rn:


experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	importRegex       = regexp.MustCompile(`^\s*import\s+([a-zA-Z0-9_.]+)(?:\s+as\s+([a-zA-Z0-9_]+))?`)
+	fromImportRegex   = regexp.MustCompile(`^\s*from\s+([a-zA-Z0-9_.]+)\s+import\s+(.+)`)
+	importItemRegex   = regexp.MustCompile(`([a-zA-Z0-9_.*]+)(?:\s+as\s+([a-zA-Z0-9_]+))?`)
+	memberImportRegex = regexp.MustCompile(`import (\w+)\.(\w+)`)
+)
+
+func init() {
+	flag.StringVar(&directory, "directory", "directory", "directory to scan")
+	flag.StringVar(&pythonFile, "python_file", "example.py", "python file to scan")
+	flag.StringVar(&requirementsFile, "requirements_file", "requirements.txt", "requirements.txt to read")
```
cuixq: do we support getting dependencies from `requirements.txt`?

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+
+	return nil
+}
+
+func main() {
+	flag.Parse()
+
+	// 1. Looking for files with main entry point
+	pythonFiles, _ := findMainEntryPoint(directory)
```
cuixq: let's print the error

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+
+	err = filepath.WalkDir(absDir, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || !strings.HasSuffix(d.Name(), ".py") {
+			return nil
+		}
+
+		err = func() error {
```
cuixq: maybe move this function to a helper function to improve readability.

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+}
+
+func main() {
+	flag.Parse()
+
+	// 1. Looking for files with main entry point
+	pythonFiles, _ := findMainEntryPoint(directory)
+	for _, file := range pythonFiles {
+		// 2. Collect libraries from poertry.lock file
+		poetryLibraryInfos, err := parsePoetryLock(filepath.Join(filepath.Dir(file), "poetry.lock"))
```
cuixq: let's check what is the manifest/lockfile in the directory e.g. `requirements.txt` or `peotry.lock`, etc. For manifests other than `peotry.lock`, log something like "dependency extraction not supported for xxx".

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+	defer r.Close()
+
+	input := &filesystem.ScanInput{
+		FS:     fsys,
+		Path:   fpath,
+		Reader: r,
+	}
+	extractor := poetrylock.New()
+	inventory, err := extractor.Extract(context.Background(), input)
```
cuixq: I would recommend having the context in main() and pass to this function.

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	libraryInfos := []*LibraryInfo{}
+	for _, i := range inventory.Packages {
+		libraryInfos = append(libraryInfos, &LibraryInfo{Name: i.Name, Version: i.Version})
+	}
+
+	return libraryInfos, nil
+}
+
+// libraryFinder scans the Python file for import statements and returns a list of LibraryInfo.
+func libraryFinder(file *os.File, poetryLibraryInfos []*LibraryInfo) ([]*LibraryInfo, error) {
```
cuixq: use `io.Reader` since this is what `bufio.NewScanner` needs and we are only reading the file

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+
+		// Skip empty lines and comments
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		// Find the existing imported libraries, if they don't exist, create a new one.
```
cuixq: can you add more comments to the following block for readability?
also it confuses me a bit that sometimes `importedLibraries` is not checked found or not - probably comments will help explain this.

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+			var pypiResponse PyPIResponse
+			err = json.Unmarshal(body, &pypiResponse)
+			if err != nil {
+				errs <- fmt.Errorf("failed to unmarshal json for %s: %w", library.Name, err)
+				return
+			}
+
+			requiresDist := pypiResponse.Info.RequiresDist
+			for _, dep := range requiresDist {
+				re := regexp.MustCompile(`^[^<>=~!]+`)
```
cuixq: use `deps.dev/util/pypi` to parse the requirement

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	for _, library := range libraryInfos {
+		if library.Version == "" {
+			continue // Skip modules without a version
+		}
+		wg.Add(1)
+
+		// Use goroutine for each library to fetch dependencies concurrently.
+		go func(library *LibraryInfo) {
+			defer wg.Done()
+			url := fmt.Sprintf("https://pypi.org/pypi/%s/%s/json", library.Name, library.Version)
```
cuixq: I would recommend to move this to a separate helper function.

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+		}
+		defer pythonFile.Close()
+
+		// 3. Find libraries imported in the main file that are defined in poetry.lock
+		importedLibraries, err := libraryFinder(pythonFile, poetryLibraryInfos)
+		if err != nil {
+			log.Printf("Error finding libraries in file %s: %v\n", file, err)
+		}
+
+		// 4. Update dependencies of the imported libraries
```
cuixq: `Get dependencies of the imported libraries`

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+
+	fileLibraryInfos := make([]*LibraryInfo, 0, len(importedLibraries))
+	for _, lib := range importedLibraries {
+		fileLibraryInfos = append(fileLibraryInfos, lib)
+	}
+	return fileLibraryInfos, nil
+}
+
+// getPackageDependencies fetches the dependencies of each library from PyPI and updates the libraryInfo.
```
cuixq: the comment is a bit confusing, maybe change it to something "gets the names of the direct dependencies of each library"

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+			if lib.Version == "" {
+				continue
+			}
+			err = retrievePackageSource(lib)
+			if err != nil {
+				log.Printf("Get source of lib error: %v\n", err)
+			}
+		}
+
+		// 6. Traverse directory of the source code and ook for Python files where they define the imported items
```
cuixq: look

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+
+	s := strings.ToLower(fmt.Sprintf(`%s\-%s\.tar\.gz`, libraryInfo.Name, libraryInfo.Version))
+	scanner := bufio.NewScanner(resp.Body)
+	for scanner.Scan() {
+		line := strings.ToLower(scanner.Text())
+		re := regexp.MustCompile(s)
+		matches := re.MatchString(line)
+		if matches {
+			re = regexp.MustCompile(`<a href="([^"]+)"`)
```
cuixq: there is also JSON API available https://docs.pypi.org/api/index-api/, see https://github.com/google/osv-scalibr/blob/main/clients/datasource/pypi_registry.go for reference.

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+			for _, err := range errs {
+				log.Printf("Error getting package dependencies: %v\n", err)
+			}
+		}
+		// 5. Download the source code of the libraries
+		for _, lib := range importedLibraries {
+			if lib.Version == "" {
+				continue
+			}
+			err = retrievePackageSource(lib)
```
cuixq: since we are sending requests to simple API anyway, we can try to get the dependencies from the files (instead of sending requests to JSON API) - also see https://github.com/google/osv-scalibr/blob/main/clients/datasource/pypi_registry.go for reference.

experimental/main_test.go
```diff
@@ -0,0 +1,231 @@
...
+			expectError:   false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// 1. Setup: Create a temporary directory for the test
+			tmpDir := t.TempDir()
+
+			// Create all the files and subdirectories needed for the test case
```
cuixq: add a `testdata` folder with these test files so no need to create them in testing.

experimental/main_test.go
```diff
@@ -0,0 +1,231 @@
...
+// A simple sort utility for comparing slices of LibraryInfo
+func sortLibraries(libs []*LibraryInfo) {
+	sort.Slice(libs, func(i, j int) bool {
+		return libs[i].Name < libs[j].Name
+	})
+}
+
+func TestParsePoetryLibrary(t *testing.T) {
+	// A sample valid poetry.lock content for testing.
+	validPoetryLockContent := `
```
cuixq: also move this to `testdata`

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+type Response struct {
+	ImportedLibrary string   `json:"imported_library"`
+	UsedModule      []string `json:"used_module"`
+}
+
+var (
+	directory         string
+	pythonFile        string
+	requirementsFile  string
+	mainEntryRegex    = regexp.MustCompile(`^\s*if\s+__name__\s*==\s*['"]__main__['"]\s*:`)
```
cuixq: regex is usually expensive so we probably can add a TODO to find alternatives for regex.

experimental/main.go
```diff
@@ -0,0 +1,615 @@
...
+	importRegex       = regexp.MustCompile(`^\s*import\s+([a-zA-Z0-9_.]+)(?:\s+as\s+([a-zA-Z0-9_]+))?`)
+	fromImportRegex   = regexp.MustCompile(`^\s*from\s+([a-zA-Z0-9_.]+)\s+import\s+(.+)`)
+	importItemRegex   = regexp.MustCompile(`([a-zA-Z0-9_.*]+)(?:\s+as\s+([a-zA-Z0-9_]+))?`)
+	memberImportRegex = regexp.MustCompile(`import (\w+)\.(\w+)`)
+)
+
+func init() {
```
cuixq: instead of `flag.StringVar`, we can also do something like
```
var species = flag.String("species", "gopher", "the species we are studying")
```
which is from https://pkg.go.dev/flag#example-package.

so we can define the flags when we define variable and no need for `init()`

cuixq:
can you add some description for this PR?

experimental/main.go
p1gc0rn: Removed the flag for requirements.txt

experimental/main.go
p1gc0rn: Updated with comment and fixing the check on `importedLibraries`. The map `importedLibraries` ensure the imported libraries info are unique with key is `libraryName`

experimental/main.go
p1gc0rn: Updated with printing the error

experimental/main.go
p1gc0rn: Created a var context in main() and parsed to the function.

experimental/main.go
p1gc0rn: Updated accordingly

experimental/main.go
p1gc0rn: fixed

experimental/main.go
p1gc0rn: Updated comment with suggestion

experimental/main.go
p1gc0rn: Add a TODO for these regexs

experimental/main.go
p1gc0rn: Cleaned up init() and used flag.String instead

experimental/main.go
p1gc0rn: Changed to `io.Reader`.

experimental/main.go
p1gc0rn: moved it to another function

experimental/main.go
p1gc0rn: resolved with `deps.dev/utl/pypi`

experimental/main.go
p1gc0rn: utilised JSON API

experimental/main.go
p1gc0rn: created switch case for that.

experimental/main.go
p1gc0rn: changed with pypi_registry

experimental/main.go
p1gc0rn: Add a helper function.

experimental/main_test.go
p1gc0rn: added folder.

experimental/main_test.go
p1gc0rn: added to testdata

experimental/pythonreach/example/poetry.lock
```diff
@@ -0,0 +1,415 @@
+# This file is automatically @generated by Poetry 2.1.3 and should not be changed by hand.
```
cuixq, line 1: I don't see this example project is being referenced anywhere in test?

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+// LibraryInfo represents a Python library and its dependencies
+type LibraryInfo struct {
+	Name         string        // Library name as it appears in imports
+	Alias        string        // Alias used when importing the entire library
+	Version      string        // Version from poetry.lock
+	Modules      []*ModuleInfo // Specific modules or functions imported from this library
+	Dependencies []string      // Direct dependencies declared in library's metadata
+}
+
+// Constants for terminal output formatting
```
cuixq: this comment should be removed?

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+			// Find the imported libraries in the files where the imported items are defined.
+			err = findImportedLibrary(lib)
+			if err != nil {
+				log.Printf("Error finding imported items: %v\n", err)
+			}
+		}
+
+		// 6. Comparison between the collected imported libraries and the PYPI dependencies of the libraries
+		// to find the reachability of the PYPI dependencies.
+		for _, library := range importedLibraries {
```
cuixq, line 774: let's make a struct for the output so that we can test the reachability result properly as well.

experimental/pythonreach/main_test.go
```diff
@@ -0,0 +1,138 @@
...
+			name:           "Malformed poetry.lock - Parser error",
+			fpathInTestDir: "./testdata/tmultifileswithentrypoint/poetry.lock",
+			expectedResult: nil,
+			expectError:    true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.Background()
```
cuixq: you may use `t.Context()` for context in testing

experimental/pythonreach/main_test.go
```diff
@@ -0,0 +1,138 @@
...
+				if err == nil {
+					t.Errorf("Expected an error, but got none")
+				}
+			} else {
+				if err != nil {
+					t.Errorf("Did not expect an error, but got: %v", err)
+				}
+			}
+
+			// To compare slices, we need a canonical order.
```
cuixq, line 63: I assume the main entry point will be a single string so we probably don't need to sort it.

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+
+	for _, file := range files {
+		if file.IsDir() {
+			continue
+		}
+		fileName := file.Name()
+		if slices.Contains(supportedManifests, fileName) {
+			manifestFiles = append(manifestFiles, fileName)
+		} else if slices.Contains(unsupportedManifests, fileName) {
+			return nil, fmt.Errorf("unsupported manifest file found: %s", fileName)
```
cuixq: I think we probably should return an error if any unsupported manifest is found. For example, there may be the case that both requirements.txt and poetry.lock exists in the folder, and error will be returned if requirements.txt is found before poetry.lock. 

I would suggest to just check the supported manifests and we don't need to worry about manifests that we don't support.

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return mainFiles, nil
+}
+
+// findManifest searches a directory for a supported manifest file.
+func findManifestFiles(dir string) ([]string, error) {
```
cuixq, line 201: could you explain a bit why we return slice of string here? do we plan to support more manifests in the future?

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+		} else if slices.Contains(unsupportedManifests, fileName) {
+			return nil, fmt.Errorf("unsupported manifest file found: %s", fileName)
+		}
+	}
+
+	return manifestFiles, nil
+}
+
+// parsePoetryLock reads the poetry lock file and updates  libraryInfo with versions.
+func parsePoetryLock(ctx context.Context, fpath string) ([]*LibraryInfo, error) {
```
cuixq, line 233: we may pass dir to parsePoetryLock instead of file path (see comment at line 581)

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+		log.Printf("Error finding manifest files: %v\n", err)
+		return
+	}
+
+	poetryLibraryInfos := []*LibraryInfo{}
+	for _, manifestFile := range manifestFiles {
+		switch manifestFile {
+		case "poetry.lock":
+			// Parse the poetry.lock file to get library information.
+			poetryLibraryInfos, err = parsePoetryLock(ctx, filepath.Join(*directory, manifestFile))
```
cuixq: considering you are hard coding "poetry.lock" in `parsePoetryLock`  we can probably just pass the directory to the function

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+
+	fileLibraryInfos := make([]*LibraryInfo, 0, len(importedLibraries))
+	for _, lib := range importedLibraries {
+		fileLibraryInfos = append(fileLibraryInfos, lib)
+	}
+	return fileLibraryInfos, nil
+}
+
+// libraryFinder scans the Python file for import statements and returns a list of LibraryInfo,
```
cuixq:
```suggestion
// findLibrariesPoetryLock scans the Python file for import statements and returns a list of LibraryInfo,
```

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+	// Get the HTTP response
+	resp, err := http.Get(downloadLink)
+	if err != nil {
+		os.Remove(tempFile.Name())
+		return "", fmt.Errorf("failed to get URL: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		os.Remove(tempFile.Name())
```
cuixq: how about we create the temp file only with http status OK?

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+
+// retrieveSourceAndCollectDependencies fetches the source code of a library from PyPI, extracts the compressed source file and
+// collect dependencies of the imported library.
+func retrieveSourceAndCollectDependencies(ctx context.Context, libraryInfo *LibraryInfo) error {
+	reg := datasource.NewPyPIRegistryAPIClient("")
+	response, _ := reg.GetIndex(ctx, libraryInfo.Name)
+	downloadURL := ""
+	fileName := strings.ToLower(fmt.Sprintf(`%s-%s.tar.gz`, libraryInfo.Name, libraryInfo.Version))
+	for _, file := range response.Files {
+		//fmt.Printf("Found file %s with URL %s\n", file.Name, file.URL)
```
cuixq: delete the unused comment.

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+
+	return nil
+}
+
+// retrieveSourceAndCollectDependencies fetches the source code of a library from PyPI, extracts the compressed source file and
+// collect dependencies of the imported library.
+func retrieveSourceAndCollectDependencies(ctx context.Context, libraryInfo *LibraryInfo) error {
+	reg := datasource.NewPyPIRegistryAPIClient("")
+	response, _ := reg.GetIndex(ctx, libraryInfo.Name)
```
cuixq: it's a good practice to check the error

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+	downloadURL := ""
+	fileName := strings.ToLower(fmt.Sprintf(`%s-%s.tar.gz`, libraryInfo.Name, libraryInfo.Version))
+	for _, file := range response.Files {
```
cuixq: you may use `pypi.SdistVersion` to check what is the version of the file name, see https://github.com/google/osv-scalibr/blob/629500d6fb6a1392afe40e0c9327c48a787358d8/clients/resolution/pypi_registry_client.go#L95

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+		//fmt.Printf("Found file %s with URL %s\n", file.Name, file.URL)
+
+		//fmt.Printf("Looking for file %s\n", fileName)
+		if file.Name == fileName {
+			downloadURL = file.URL
```
cuixq: break the loop? since the file is found

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+		}
+	}
+
+	downloadFileSource, err := downloadPackageSource(downloadURL)
```
cuixq: you may find https://github.com/google/osv-scalibr/blob/629500d6fb6a1392afe40e0c9327c48a787358d8/clients/datasource/pypi_registry.go#L91 useful

experimental/pythonreach/testdata/multifileswithentrypoint/script.sh
```diff
@@ -0,0 +1 @@
+#!/bin/bash
```
cuixq: is there any particular reason that we want to keep this shell script?

experimental/pythonreach/testdata/multifileswithentrypoint/main.py
```diff
@@ -0,0 +1,3 @@
+print('hello')
```
cuixq, line 1: let's add newline to end of these test files if that won't bring any trouble

cuixq:
we probably need `go.mod` as well

experimental/pythonreach/main.go
cuixq: https://github.com/google/osv-scalibr/blob/main/enricher/reachability/java/java.go can be a reference about what the results should look like.

experimental/pythonreach/main.go
p1gc0rn: thank you, I changed to only check the supported manifest files

experimental/pythonreach/main.go
p1gc0rn: thank you for the link!

experimental/pythonreach/main.go
p1gc0rn: thanks to your suggestion, this function can be removed

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,685 @@
...
+
+	for {
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break // End of archive
+		}
+		if err != nil {
+			return fmt.Errorf("failed to read tar header: %w", err)
+		}
+		targetPath := filepath.Join(".", header.Name)
```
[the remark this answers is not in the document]
p1gc0rn: the input of the function `extractCompressedPackageSource` is a temporary directory

experimental/pythonreach/main.go
```diff
@@ -0,0 +1,702 @@
...
+		file, closer, err := safeOpenFile(path)
+		if err != nil {
+			return err
+		}
+		defer closer()
+
+		return scanFile(file, func(line string) error {
+			for _, module := range libraryInfo.Modules {
+				searchTerm := fmt.Sprintf("def %s(", module.Name)
+				if strings.Contains(line, searchTerm) {
```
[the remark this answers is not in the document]
p1gc0rn: add TODO

experimental/pythonreach/main.go
p1gc0rn: Yes, we can support other manifest files like requirements.txt, pyproject.toml, pipfile.lock

experimental/pythonreach/example/poetry.lock
p1gc0rn: this is for running with the main.go in CLI. `go run main.go --directory=example`

experimental/pythonreach/testdata/multifileswithentrypoint/script.sh
p1gc0rn: Nah, I will remove it

experimental/pythonreach/main.go
p1gc0rn: Thank you, I have updated the code base to use the function.

experimental/pythonreach/main_test.go
p1gc0rn: Done

cuixq (approved):

PR #2146, opened 2025-08-06
feat: merge `extractor` and `detector` flags into `plugins`

G-Rath:
This removes the external distinction between extractors and detectors in favor of just plugins, which matches how they're represented in scalibr


another-rex (approved):
LGTM as well, though need to resolve the conflicts before the diffs are correct.

another-rex (approved):

Issue #2151, opened 2025-08-07
OSV Scanner scanning Podfile.lock and throws 400 Bad Request Invalid ecosystem since 2.2.0


Ianrisch:
The previous version (2.1.0) did not detect Podfile.lock, but the new one does and throws an error.
This issue is reproducible in Flutter when scanning recursively.

Error:
```
Scanned /src/ios/Podfile.lock file and found 50 packages
Scanned /src/pubspec.lock file and found 256 packages
End status: 534 dirs visited, 2574 inodes visited, 2 Extract calls, 121.639539ms elapsed, 121.639839ms wall time
error when retrieving vulns: client error: status="400 Bad Request" body={"code":3,"message":"Invalid ecosystem."}
client error: status="400 Bad Request" body={"code":3,"message":"Invalid ecosystem."}
```

another-rex:
Thanks for the report, this should be fixed in the next patch release, which we'll try to release soon.

For now you should be able to workaround this issue with the new disable-plugins flag, e.g.:

```
osv-scanner scan --experimental-disable-plugins swift/podfilelock,swift/packageresolved /src/
```

another-rex:
OSV-Scanner 2.2.1 fixes this issue :)

PR #2156, opened 2025-08-07
fix: invalid ecosystem error

another-rex:
Fixes: #2151

It looks like osv.dev will return invalid ecosystem for all queries in a batchquery if 1 query has an invalid ecosystem. This means we have to add a check in osv-scanner to prevent any invalid ecosystems from being sent. 

This does mean if we release osv-scanner with extractors to support future ecosystems that are currently invalid, people will have to update to get support for those ecosystems when supported in osv.dev in the future, instead of being able to use existing version of osv-scanner. I don't think this is a common case, so this is acceptable.

Changes in this PR:
- Added Podfile.lock to test cases, should return 127 if scanned alone
- Excude Podfile.lock and Package.resolved because we don't support the Swift Cocoapods ecosystem  (SwiftURL is for the Swift package manager, which is different from Cocapods?)
- Added a IsValid() check for ecosystems, which contains all the current ecosystems within osvschema. Missing ecosystems would be caught by the exhaustive linter in the future.
- Filter on IsValid ecosystems so we don't send queries with invalid ecosystems.
- Fix what seemed to be a bug with unknown Maven packages not being skipped.

### DIFF
--- a/internal/imodels/ecosystem/ecosystem.go
+++ b/internal/imodels/ecosystem/ecosystem.go
@@ -91,6 +91,53 @@ func (p Parsed) String() string {
 	return str
 }
 
+func (p Parsed) IsValid() bool {
+	// Missing ecosystems here would be caught by the "exhaustive" linter
+	switch p.Ecosystem {
+	case osvschema.EcosystemAlmaLinux,
+		osvschema.EcosystemAlpaquita,
+		osvschema.EcosystemAlpine,
+		osvschema.EcosystemAndroid,
+		osvschema.EcosystemBellSoftHardenedContainers,
+		osvschema.EcosystemBioconductor,
+		osvschema.EcosystemBitnami,
+		osvschema.EcosystemChainguard,
+		osvschema.EcosystemConanCenter,
+		osvschema.EcosystemCRAN,
+		osvschema.EcosystemCratesIO,
+		osvschema.EcosystemDebian,
+		osvschema.EcosystemGHC,
+		osvschema.EcosystemGitHubActions,
+		osvschema.EcosystemGo,
+		osvschema.EcosystemHackage,
+		osvschema.EcosystemHex,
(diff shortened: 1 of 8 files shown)


internal/imodels/ecosystem/ecosystem.go
```diff
@@ -91,6 +91,53 @@ func (p Parsed) String() string {
 	return str
 }
 
+func (p Parsed) IsValid() bool {
+	// Missing ecosystems here would be caught by the "exhaustive" linter
+	switch p.Ecosystem {
+	case osvschema.EcosystemAlmaLinux,
```
hogo6002, line 97: This looks like we need to update every time when new ecosystem gets added. 
Not for this PR but I am wondering if we could add the check function to the schema repo or make this function to auto scale.

hogo6002 (approved):

pkg/osvscanner/filter.go
```diff
@@ -12,22 +12,29 @@ import (
...
-		case p.Ecosystem().Ecosystem == osvschema.EcosystemMaven && p.Name() == "unknown":
 		default:
 			continue
 		}
 
+		switch {
+		// If **any** of the following cases are true, skip this package
+		case p.Ecosystem().Ecosystem == osvschema.EcosystemMaven && p.Name() == "unknown",
+			!p.Ecosystem().IsValid() && !p.Ecosystem().IsEmpty():
+			continue
```
michaelkedar, line 35: Wasn't before `p.Ecosystem().Ecosystem == osvschema.EcosystemMaven && p.Name() == "unknown"` would **not** continue (since there's no implicit `fallthrough` in go), and now after it is continuing?

Am I interpreting this right?

another-rex: yes, this is a behavior change, I think the previous behavior is a bug.

michaelkedar (approved):

Issue #2157, opened 2025-08-08
`--all-packages` flag should also show unscannable filtered packages.


another-rex:
Currently `--all-packages` flag only adds/includes packages that don't have any vulnerabilities. It should also add packages that have been filtered out because it can't be scanned for whatever reason.

In `osvscanner.go` after the initial scanning phase we do multiple filtering steps (most of the logic will be in `filter.go`). Specifically before querying packages are filtered out according to the config rules and according to whether they are scannable or not. 

These filtered packages should be set aside somewhere and then re-added back into the final result if the `--all-packages` flag is passed in.

SavvyHex:
can I please be assigned to this issue?

another-rex:
Sure, assigned to you.

SavvyHex:
Hi, I've fixed the issue and `--all-packages` now shows all the unscannable filtered packages. Would you like me to make it so that it shows all of the packages that would normally be filtered out? I'm talking about the following functions :
- `filterNonContainerRelevantPackages`
- `filterIgnoredPackages`
- `filterResults`
- `filterPackageVulns`

another-rex:
Thanks, I think the rest is still good to keep filtered, as they are mostly user defined. I can see `filterNonContainerRelevantPackages` as a potential one to also allow, but let's leave it for now.

SavvyHex:
Okay, understood. I've submitted a PR for this issue. Please tell me if I need to change anything.

PR #2170, opened 2025-08-11
fix: Include unscannable packages with --all-packages flag (#2157)

SavvyHex:
The `--all-packages` flag previously excluded packages that couldn’t be scanned (unscannable packages). This fix updates the `filterUnscannablePackages` function in `pkg/osvscanner/filter.go` to respect the `showAllPackages` flag, ensuring these packages are included in the output and providing a more complete package list.

A simple conditional check was added to skip filtering unscannable packages when `--all-packages` is set.

The tests fail initially because a new log message,
"Skipping filtering of unscannable packages as --all-packages is set."
was added. This causes snapshot mismatches. Updating the snapshots will resolve the test failures.

### DIFF
--- a/pkg/osvscanner/filter.go
+++ b/pkg/osvscanner/filter.go
@@ -15,7 +15,11 @@ import (
 // filterUnscannablePackages removes packages that don't have enough information to be scanned or
 // are not a supported ecosystem
 // e,g, local packages that specified by path
-func filterUnscannablePackages(scanResults *results.ScanResults) {
+func filterUnscannablePackages(scanResults *results.ScanResults, showAllPackages bool) {
+	if showAllPackages {
+		cmdlogger.Infof("Skipping filtering of unscannable packages as --all-packages is set.")
+		return
+	}
 	packageResults := make([]imodels.PackageScanResult, 0, len(scanResults.PackageScanResults))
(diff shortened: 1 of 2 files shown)


another-rex:
Can you add "Fixes #2157" to the PR description? Thanks!

pkg/osvscanner/filter.go
```diff
@@ -15,7 +15,11 @@ import (
 // filterUnscannablePackages removes packages that don't have enough information to be scanned or
 // are not a supported ecosystem
 // e,g, local packages that specified by path
-func filterUnscannablePackages(scanResults *results.ScanResults) {
+func filterUnscannablePackages(scanResults *results.ScanResults, showAllPackages bool) {
+	if showAllPackages {
```
another-rex, line 20: I think this needs to be a bit more complicated than this. We still don't want to send the unscannable packages for vulnerability querying, but we do want to keep the packages for the output. 

You would have to copy the list of packages if ShowAllPackages is true, and merge it together after vulnerability matching.

SavvyHex: Okay, that makes more sense. I'll add that logic right now.

SavvyHex:
I've now made it so that the function copies all packages from the original scan results into a new slice `allPackages`, to preserve the original for output.

PR #2172, opened 2025-08-11
Add tests for spdx

another-rex:
Branched off of #2171 

Add examples to the docs for SPDX and CycloneDX output

### DIFF
--- a/docs/output.md
+++ b/docs/output.md
@@ -239,7 +239,8 @@ osv-scanner scan --format json -L path/to/lockfile > /path/to/file.json
 osv-scanner scan --format sarif your/project/dir
 ```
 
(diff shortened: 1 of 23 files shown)


docs/output.md
```diff
@@ -385,6 +386,220 @@ Outputs the result in the [SARIF](https://sarifweb.azurewebsites.net/) v2.1.0 fo
...
 
+### SPDX
+
+```bash
+osv-scanner scan --format spdx-2-3 --all-packages your/project/dir
+```
+
+Outputs the result in the [SPDX](https://spdx.dev/) v2.3 format. This matches OSV-Scalibr's SPDX output format.
+
+{: .note }
```
cuixq, line 397: I think this does not look right in the rich diff - can you double check that?

another-rex: Ran this locally and it looks fine, I think the github markdown renderer is different from the jekyll one github pages uses, so it doesn't support the same note blocks syntax.

another-rex: image

cuixq (approved):

Issue #2174, opened 2025-08-11
Support osv-scanner-custom.json as a scannable file type.


another-rex:
Support `osv-scanner-custom.json` as a scannable file type. This should be as simple as adding that file name to the osv-scanner-custom extractor's FileRequired. 

Related #1938

PR #2177, opened 2025-08-12
fix: Support osv-scanner-custom.json as a scannable file type (#2174)

kuscar:
Added support for recognizing osv-scanner-custom.json with the -r option. I wasn’t sure if `"osv-scanner-custom.json": {osvscannerjson.Name},`
also needs to be added to lockfile.go, since it seems to work without it. Please review and advise on any improvements. I will update the snapshots after.
Do I also need to update the webpage that lists supported files?

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -439,18 +439,16 @@ overriding license for package Alpine/scanelf/1.3.5-r1 with MIT
 overriding license for package Packagist/sentry/sdk/2.0.4 with 0BSD
 overriding license for package Alpine/ssl_client/1.36.1-r27 with MIT
 overriding license for package Alpine/zlib/1.2.13-r0 with MIT
-Total 4 packages affected by 4 known vulnerabilities (1 Critical, 0 High, 0 Medium, 0 Low, 3 Unknown) from 5 ecosystems.
-3 vulnerabilities can be fixed.
+Total 2 packages affected by 2 known vulnerabilities (1 Critical, 0 High, 0 Medium, 0 Low, 1 Unknown) from 5 ecosystems.
+1 vulnerability can be fixed.
 
 
-+-------------------------------------+------+-----------+------------------+----------+---------------+-------------------------------------------------+
-| OSV URL                             | CVSS | ECOSYSTEM | PACKAGE          | VERSION  | FIXED VERSION | SOURCE                                          |
-+-------------------------------------+------+-----------+------------------+----------+---------------+-------------------------------------------------+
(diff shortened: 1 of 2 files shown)


internal/scalibrplugin/resolve.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/google/osv-scanner/v2/internal/scalibrextract/language/javascript/nodemodules"
 	"github.com/google/osv-scanner/v2/internal/scalibrextract/language/python/requirementsenhancable"
 	"github.com/google/osv-scanner/v2/internal/scalibrextract/vcs/gitrepo"
+	"github.com/google/osv-scanner/v2/internal/scalibrextract/language/osv/osvscannerjson"
```
another-rex: The imports doesn't seem formatted/ordered correctly here.

kuscar:
I also updated the orders, but there are still some issues with the lint script. Since this custom file is not part of the Scalibr package, I wasn’t able to resolve them. Could you please take a look?

Also, please check if the documentation is sufficient.

internal/scalibrplugin/resolve.go
```diff
@@ -36,6 +37,8 @@ func resolveFromName(name string) (plugin.Plugin, error) {
 		return vendored.New(), nil
 	case gitrepo.Name:
 		return gitrepo.New(), nil
+	case "osv/osvscannerjson":
```
G-Rath:
```suggestion
	case osvscannerjson.Name:
```

docs/supported_languages_and_lockfiles.md
```diff
@@ -135,3 +135,11 @@ Then pass this to `osv-scanner` with this:
 ```
 osv-scanner --lockfile osv-scanner:/path/to/osv-scanner.json
 ```
+
+For automatic scan detection, you can create an osv-scanner-custom.json file using the same procedure described above for osv-scanner.json.
```
G-Rath:
```suggestion
For automatic scan detection, you can create an `osv-scanner-custom.json` file using the same procedure described above for `osv-scanner.json`.
```

G-Rath:
@kuscar it looks like the remaining linting issues are with `gofmt` - running `gofmt -w internal/scalibrplugin/*` should resolve them.

You'll also need to run `npx prettier --write .` to format the docs

kuscar:
Thank you, that was insanely fast. Please let me know if there’s anything else I can improve.

another-rex:
Known limitation with this current fix, it will fail for the `--format=spdx` output, but that will require much more work to fix, so leaving it as is for now. @michaelkedar Can you take a look?

michaelkedar:
> Known limitation with this current fix, it will fail for the `--format=spdx` output, but that will require much more work to fix, so leaving it as is for now. @michaelkedar Can you take a look?

We might want to document this limitation somewhere in the docs

PR #2179, opened 2025-08-13
fix(test): replace stat with GetFileAtributesEx

hogo6002:


internal/testutility/normalize.go
```diff
@@ -110,6 +110,7 @@ func normalizeErrors(t *testing.T, str string) string {
 	str = strings.ReplaceAll(str, "The system cannot find the file specified.", "no such file or directory")
 	str = strings.ReplaceAll(str, "CreateFile", "lstat")
 	str = strings.ReplaceAll(str, "\nstat ./fixtures/", "\nlstat ./fixtures/")
+	str = strings.ReplaceAll(str, "stat <rootdir>", "GetFileAttributesEx <rootdir>")
```
another-rex: We don't need the `<rootdir>`. Also this is the wrong way around. See line 111 for similar example.

internal/testutility/normalize.go
```diff
@@ -110,7 +110,7 @@ func normalizeErrors(t *testing.T, str string) string {
 	str = strings.ReplaceAll(str, "The system cannot find the file specified.", "no such file or directory")
 	str = strings.ReplaceAll(str, "CreateFile", "lstat")
 	str = strings.ReplaceAll(str, "\nstat ./fixtures/", "\nlstat ./fixtures/")
-	str = strings.ReplaceAll(str, "stat <rootdir>", "GetFileAttributesEx <rootdir>")
+	str = strings.ReplaceAll(str, "GetFileAttributesEx", "lstat")
```
another-rex:
```suggestion
	str = strings.ReplaceAll(str, "GetFileAttributesEx", "stat")
```

another-rex: I'm so confused at how the windows tests are passing right now...

another-rex (changes requested):

another-rex (approved):

internal/testutility/normalize.go
another-rex: Oh I was looking at an outdated diff, that's why haha.

PR #2180, opened 2025-08-14
ci: Update golangci version to support 1.25

another-rex:


G-Rath (changes requested):
you need to bump the version in `scripts/run_lints.sh` and there will be some violations about `os/exec` not using context

hogo6002 (approved):

PR #2200, opened 2025-08-25
test: use `go-vcr` to mock osv.dev bulk query calls

G-Rath:
This is an attempt to make our CI easier to work with day-to-day by introducing `go-vcr` as a proxy to record and replay certain HTTP requests, with our "update snapshots" workflow being updated to also update these records.

The idea is this will let us continue effectively testing against real external APIs but let us decide when those external APIs change by way of updating the HTTP records (terms cassettes, in keeping with the VCR metaphor).

For now I've just focused on the osv.dev client, setting up `go-vcr` to record interactions with any endpoint aside from `/v1/vulns/<id>` since those requests have huge responses making the cassettes easily balloon beyond 5mb per file - while these do somewhat frequently change, I'm hoping that their day-to-day changes won't be as noisy and especially considering its the response of the bulk query endpoint that leads to the specific vuln lookups so since they're being recorded things should be a lot more predictable.

I've structured things so that a recording client is created per top-level test so that a single cassette is used for any subtests, and a `X-Test-Name` header is included in each request via a second `http.Client` with a custom `RoundTripper` that gets created in each subtest wrapping the recording client, which allows us to sort the interactions afterwards giving us a stable diff when regenerating.

I have also included a custom hook to remove unneeded headers to both further reduce the noise in diffs and the size of the cassettes overall; the only header I'm on the fence about is the `Date` header since that relates to caching and its technically possible for code to check that for determining if data should be re-fetched, but I think it's best to omit it for now given its present in every single request (ideally I'd like to have it be updated only when something else changes, but that doesn't seem possible without a lot more work).

Finally, one notable downside for us to keep an eye on is with keeping the cassettes up to date: the nature of this functionality means its inherently hard for such a library to determine if a particular record is obsolete or needs to be generated as (among other things) that would most likely require making the request in the first place e.g. if we change what headers are excluded but don't explicitly regenerate everything, I don't think CI will complain - thankfully, the worse case is that this should get caught by our "update snapshots" workflow


G-Rath:
hmm so (unsurprisingly, in hindsight) this really moves the problem rather than solve it outright as now our CI can fail due to other external conditions, and it doesn't present itself super obviously.

We have these failing tests:

- `TestCommand_OCIImage_JSONFormat/Scanning_python_image_with_some_packages`
- `TestCommand_OCIImage_JSONFormat/Scanning_python_image_with_some_packages`
- `TestCommand_LockfileWithExplicitParseAs/"apk-installed"_is_supported`
- `TestCommand_LockfileWithExplicitParseAs/"dpkg-status"_is_supported`
- `TestCommand/requirements.txt_can_have_all_kinds_of_names`

The first two are really the same failure because its the same underlying image just with a different output format, and the issue is that a new version of `certifi` has been published

image

This is addressed by running `scripts/build_test_images.sh --force --no-cache`, and something we already knew about.

The next two I'm guessing are a result of the underlying OS extractors pulling information from the local environment which for me is Ubuntu 22.04 but CI is e.g. Ubuntu 24.04

image

And then the last one I'm guessing is the result of `typing-extensions` having a new version, and that that is a transient dependency of one of our fixtures since its not mentioned anywhere in our codebase

image

CONTRIBUTING.md
```diff
@@ -94,6 +94,17 @@ You can regenerate snapshots by setting `UPDATE_SNAPS=true` when running tests:
...
 
+`cmd` tests use [`go-vcr`](https://github.com/dnaeon/go-vcr) to provide a custom `http.Client` for osv.dev requests to the `querybulk` endpoint which uses
+snapshots of requests called cassettes to reduce noise from changes to advisories while still providing a high degree
+of confidence.
+
+You can control the recording behaviour by setting the `TEST_VCR_MODE` environment variable to one of the [supported modes](https://github.com/dnaeon/go-vcr/blob/v4/pkg/recorder/recorder.go#L51),
+specified either by its name without the `Mode` suffix or by its int value.
+
+The default mode locally is `ModeReplayWithNewEpisodes`, meaning existing interactions will be replayed while any new ones will
+be recorded and added to the existing cassette; when running in CI, the default mode is `ReplayOnly` meaning an error will be
```
cuixq, line 105:
```suggestion
be recorded and added to the existing cassette; when running in CI, the default mode is `ModeReplayOnly` meaning an error will be
```

CONTRIBUTING.md
```diff
@@ -94,6 +94,17 @@ You can regenerate snapshots by setting `UPDATE_SNAPS=true` when running tests:
 UPDATE_SNAPS=true ./scripts/run_tests.sh
 ```
 
+`cmd` tests use [`go-vcr`](https://github.com/dnaeon/go-vcr) to provide a custom `http.Client` for osv.dev requests to the `querybulk` endpoint which uses
+snapshots of requests called cassettes to reduce noise from changes to advisories while still providing a high degree
+of confidence.
+
+You can control the recording behaviour by setting the `TEST_VCR_MODE` environment variable to one of the [supported modes](https://github.com/dnaeon/go-vcr/blob/v4/pkg/recorder/recorder.go#L51),
```
cuixq, line 101: I think it worth mentioning that we can set the flag with text or number, and probably list the available text options.

cuixq (approved):

CONTRIBUTING.md
G-Rath: I purposely didn't list the options in our docs because I didn't want us to be maintaining our own list

CONTRIBUTING.md
G-Rath: This is the value we're passing in, rather than the enum

CONTRIBUTING.md
G-Rath: having said that we're actually mapping without the `Mode` - I might instead link to our own code and then from there link to the recorder states in the library

G-Rath: actually, I've already got that documented, the diff just made it non-obvious 😅 

> You can control the recording behaviour by setting the `TEST_VCR_MODE` environment variable to one of the [supported modes](https://github.com/dnaeon/go-vcr/blob/v4/pkg/recorder/recorder.go#L51),
specified either by its name without the `Mode` suffix or by its int value.

CONTRIBUTING.md
G-Rath: ah but its inconsistent with the mode mentioned earlier in the sentence - will update that

CONTRIBUTING.md
cuixq: > I purposely didn't list the options in our docs because I didn't want us to be maintaining our own list

we are maintaining the list of the modes in our code anyway - it was not clear to me the flag should be `ModeReplayOnly` or `ReplayOnly` but the next paragraph now is clear about the mode texts so I think it's fine.

another-rex (approved):
It works!

@G-Rath PTAL, and I think we might need to move update script to a bot or similar so we can update the snapshots without having to run windows.

.github/workflows/checks.yml
```diff
@@ -185,3 +185,27 @@ jobs:
               exit $exit_code
             fi
           done
+  update:
```
G-Rath: this should be removed

.github/workflows/snapshots.yml
```diff
@@ -31,15 +31,16 @@ jobs:
       - run: ./scripts/run_tests.sh || true
         env:
           TEST_ACCEPTANCE: true
+          TEST_VCR_MODE: recordonly
```
G-Rath, line 34: @another-rex this is updating the snapshot bot to also do the cassettes

cmd/osv-scanner/internal/testcmd/vcr.go
```diff
@@ -0,0 +1,224 @@
...
+		var buffer bytes.Buffer
+		if _, err := buffer.ReadFrom(r.Body); err != nil {
+			return false
+		}
+
+		r.Body = io.NopCloser(bytes.NewBuffer(buffer.Bytes()))
+
+		body2, err := normalizeBody(buffer.Bytes())
+		if err != nil {
+			// This file will only run in tests, and this should be very
```
G-Rath: be very what...?

another-rex:
One thing I want to do before merging is to do the formatting similar to how it's done on osv.dev. Right now it's really hard to see the diff when updating cassettes.

.github/workflows/snapshots.yml
another-rex: Hmm... but this will not work for nonlinux. Probably not a big issue for now.

G-Rath: yes but like I said in a comment I apparently typed out but then didn't actually post (😅):

currently the only OS specific tests we have are for the detectors, which don't make queries recorded by `go-vcr` so the only difference in their cassettes should be the test name header, and I don't expect we'll have that many OS-specific tests (at least for now).

So the main time those cassettes should need updating is when a new version is released, as the user agent will have changed - outside of that, any changes that do happen should be easily addressed by just copying and pasting the updated cassette and then manually updating the test names in the header 🤷

another-rex: Oh I removed the user agent, so we don't even need to worry about that.

PR #2203, opened 2025-08-26
feat: update `osv-scalibr`

G-Rath:


cuixq (approved):

cuixq:
I think the osv-scalibr version in snapshots needs to be updated to reflect this change.

G-Rath:
@another-rex @cuixq actually any chance we could get https://github.com/google/osv-scalibr/pull/1196 landed so it can be included in this

another-rex:
@G-Rath Can you update the snapshots? I think a scalibr change (probably semantic) fixed the ability to find the fixed version here.

G-Rath:
ooh yeah you're right, it's because we now can handle Hackage versions so we can determine the next fixed version 🎉

cuixq (approved):

Issue #2194, opened 2025-08-18
Release v2.2.1 action.dockerfile broken due go version mismatch


tuminoid:
v2.2.1 is including a go.mod that has requirement of v1.24.6 in `go.mod`, but the `action.dockerifle` is bumped to same Go version 6 commits AFTER cutting v2.2.1.

<img width="1419" height="351" alt="Image" src="https://github.com/user-attachments/assets/0f681ee2-3743-4771-bf11-b0f43402a63d" />

We use the osv-scanner action, and when running the action pinned to v2.2.1, it fails to build `action.dockerfile` as go wants 1.24.6 but the dockerfile only has 1.24.5.

Please cut v2.2.2.

/kind bug

another-rex:
Ah, hmm, we'll likely remove the go.mod patch version constraint to avoid this in the future. 

For this current issue, are you able to use our prebuilt osv-scanner-action image? (ghcr.io/google/osv-scanner-action:v2.2.1) It should be basically the same as the `action.dockerfile`, except quite a bit faster since it doesn't have to build from scratch every time.

lentzi90:
We [pin to the commit sha](https://docs.github.com/en/actions/reference/security/secure-use), and I think that is why it is built from scratch each time. I couldn't find it in the github docs, but I assume this is why the pinning is for the commit sha, rather than the image sha. Using the pre-built action should work fine (but it comes with the risk that the tag is overwritten with new content).

another-rex:
You can pin the sha256 of the docker image that we publish as well to avoid the problem of the tag being changed. I believe renovatebot and depenabot can automatically update it to the next sha whenever we do a release.

tuminoid:
> Ah, hmm, we'll likely remove the go.mod patch version constraint to avoid this in the future.

The controversial `go.mod` version bumping for no reason :) 

> For this current issue, are you able to use our prebuilt osv-scanner-action image? (ghcr.io/google/osv-scanner-action:v2.2.1) It should be basically the same as the `action.dockerfile`, except quite a bit faster since it doesn't have to build from scratch every time.

So, we're using the report action like this, and here in the PR it is being bumped from v2.2.0 to v2.2.1. This is the expected way of using Github actions, right? We specify no image pins, we just use your action and dependabot manages the SHA.

```yaml
    - name: "Run OSV Scanner Reporter"
      uses: google/osv-scanner/actions/reporter@04a8728383d8f286cf2b0f110e68482e8773df6e # v2.2.1
[...]

PR #2204, opened 2025-08-26
fix: Add warning to use the correct action

another-rex:
People are getting confused which action to use. See #2194


cuixq (approved):

G-Rath:
@another-rex should we just have these workflows error like the others? That'll ensure people migrate

another-rex:
I mean it still mostly works, so I'm not sure we need to intentionally break it, it'll break on its own eventually and then folks will be forced to migrate.

PR #2210, opened 2025-08-28
feat: make `--plugins` additive by default, and introduce `--no-default-plugins`

### LINKED ISSUE #1958: Create documentation for (experimental) custom extractors
We should have some documentation for #1846 explaining:
  * how to use the CLI flags
  * how to use the fields programmatically 
  * how to have the scanner run just your own custom extractor
  * how to match the existing scanner behaviour + your own extractor

It's probably best though to wait until we've landed some of the other changes as they'll simplify what we need to document by removing edge cases.

Also see https://github.com/google/osv-scanner/pull/1846/commits/8edae54f46cb62a34bb4859d2691b0097f5c0c8f#diff-f6e4ddeec588fb949027adbba185137af1f443ceb182b703b70e14e9c0715f7c for some rough notes I put together at the time

### LINKED ISSUE #2178: Make it clearer that the `--experimental-plugins` flag disables existing plugins
Couple of things that need to be done here:
- Update the usage text of the flag to mention that we are  https://github.com/google/osv-scanner/blob/e28c9bc59852fb4599e79077e46c73d110066853/cmd/osv-scanner/internal/helper/flags.go#L172
- Update documentation to list out the default set of plugin presets we use, so people can reenabled it if they want to using and add additional plugins
- When using the flag, log out the default preset that was disabled, so it's easy for users to know which preset to re-add, and they are more aware of this behavior.

G-Rath:
Resolves #1958
Resolves #2178

### DIFF
--- a/cmd/osv-scanner/internal/helper/flags.go
+++ b/cmd/osv-scanner/internal/helper/flags.go
@@ -184,5 +184,9 @@ func BuildCommonScanFlags(defaultExtractors []string) []cli.Flag {
 			Name:  "experimental-disable-plugins",
 			Usage: "list of specific plugins and presets of plugins to not use",
 		},
+		&cli.BoolFlag{
+			Name:  "experimental-no-default-plugins",
+			Usage: "disable default plugins, instead using only those enabled by --experimental-plugins",
+		},
 	}
 }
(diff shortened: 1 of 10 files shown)


another-rex (approved):
LGTM, thanks!

PR #2216, opened 2025-09-02
feat: warn when a config has ignored vulns that were not found during scanning

### LINKED ISSUE #2206: Warn/error when an IgnoredVulns.id was not encountered
If the config file contains a `IgnoredVulns.id` that has been fixed, it would be useful to emit a warning by default , with the option of being able to convert these warnings to errors.

G-Rath:
Resolves #2206


G-Rath:
@another-rex I'm wondering if we should introduce a new hidden `--config-file-name` flag that we can use in our tests to have the scanner look for e.g. `osv-scanner-test.toml` instead of the regular filename, so that way we can have tests with implicit config files but still explicitly ignore all the testdata vulns for scorecard.

what do you think?

G-Rath:
This is waiting on #2254 which should make things a lot easier

G-Rath:
hmm I think it might still be worth restructuring our tests as we have a lot of usages of single files within the `locks-many` test fixtures, so introducing a `locks-many-with-vulns` version would reduce our snapshot sizes 🤔

another-rex:
Is this ready for review?

G-Rath:
yup, I was just waiting for the tests to pass in between meetings 😄

another-rex (approved):

PR #2234, opened 2025-09-14
test: add RockyLinux image cases

### LINKED ISSUE #254: Add RPM/Red Hat ecosystem support

G-Rath:
Resolves #254

### DIFF
--- a/cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
@@ -636,6 +636,108 @@ You can also view the full vulnerability list in your terminal with: `osv-scanne
 
 ---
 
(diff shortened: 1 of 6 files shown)


G-Rath:
blocked by https://github.com/google/osv-scalibr/pull/2450

PR #2240, opened 2025-09-21
fix: update `osv-schema` bindings

G-Rath:
This should hopefully make #2231 happy

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/jedib0t/go-pretty/v6 v6.6.8
 	github.com/muesli/reflow v0.3.0
 	github.com/opencontainers/go-digest v1.0.0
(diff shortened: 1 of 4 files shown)


cuixq:
I think we probably should fix the breaking change in `osv-schema` bindings - though we want to rely on automation to generate constants but we also want to avoid breaking changes for existing constants.

G-Rath:
right ok, I'll have a look at that

another-rex:
osv-schema is updated now, just triggered a rerun on renovate bot

G-Rath:
@another-rex unless I'm missing something, we need https://github.com/ossf/osv-schema/pull/428 landed before that'll work..?

G-Rath:
ah right I see this has already been done - I forgot I don't get notifications for that repo 😅

PR #2243, opened 2025-09-22
chore: update `golangci-lint` to v2.5.0 and ensure godoc comments are consistent

G-Rath:
This adds:
  - https://golangci-lint.run/docs/linters/configuration/#godoclint (Checks Golang's documentation practice)
  - https://golangci-lint.run/docs/linters/configuration/#iotamixing (Checks if iotas are being used in const blocks with other non-iota declarations)
  - https://golangci-lint.run/docs/linters/configuration/#unqueryvet (Detects SELECT * in SQL queries and SQL builders, preventing performance issues and encouraging explicit column selection)


internal/resolution/client/maven_registry_client.go
```diff
@@ -54,6 +54,8 @@ func (c *MavenRegistryClient) Version(ctx context.Context, vk resolve.VersionKey
 	return resolve.Version{VersionKey: vk, AttrSet: attr}, nil
 }
 
+// Versions does something...
```
G-Rath: happy to take suggestions on what a good comment here would be 😅

cuixq:
```
Versions returns all available versions of the given package from the maven-metadata.xml file
```

cuixq (approved):

internal/resolution/client/maven_registry_client.go
G-Rath:
```suggestion
// Versions returns all available versions of the given package from the maven-metadata.xml file
```

G-Rath: thanks!

PR #2241, opened 2025-09-22
perf(local): only load advisories that are about the packages being s…

G-Rath:
This updates the local matcher to have it skip advisories that don't have at least one affected entry with a package name matching one of the packages being scanned in the current run, which can greatly reduce the peak memory usage for databases like Ubuntu (going from something like 10gb down to 1gb).

Since we cache databases based on their ecosystem only, this does mean subsequent calls to `LocalMatcher#MatchVulnerabilities` will not give any results for packages that were not present in the first call - while this shouldn't be a problem currently since we handle creating the `VulnerabilityMatcher` as part of scanning, I've added a basic guard that returns an error if the function is called with any "partial" database cached to catch this (be it on purpose or because of a bug)

This should not impact guided remediation since it explicitly loads the database before doing any work meaning this change won't help it but should also not hurt it

Resolves #2217 (again)

PR #2244, opened 2025-09-22
test(fix): add cases for `--offline-vulnerabilities`

G-Rath:
I realized we were missing these while working on #2241


another-rex (approved):

G-Rath:
@michaelkedar I don't suppose you've got the `package.json` for `cmd/osv-scanner/fix/testdata/in-place-npm`? I'd like to add a case for covering preferring the directory of the manifest over the lockfile, and also interested in seeing if it might be possible to make the lockfile a bit smaller (though I'm less committed to that 😅)

michaelkedar:
> @michaelkedar I don't suppose you've got the `package.json` for `cmd/osv-scanner/fix/testdata/in-place-npm`? I'd like to add a case for covering preferring the directory of the manifest over the lockfile, and also interested in seeing if it might be possible to make the lockfile a bit smaller (though I'm less committed to that 😅)

It looks like it was `cmd/osv-scanner/fix/testdata/relax-npm/package.json`, except the version of `npm-registry-client` was `"^6.2.0"`

PR #2246, opened 2025-09-23
ci: introduce dedicated workflow for updating `osv-scalibr`

G-Rath:
Having renovate do this isn't currently working out as it doesn't support always updating a Go module to the latest digest so whenever a new version of `osv-scalibr` is published we have to manually update to the next pseudo version, so this introduces a new workflow similar to the one we use for updating the snapshots except for updating to the latest commit on `main` for `osv-scalibr`.

Currently I've just got this running on the same schedule as the snapshots workflow, though in theory we should be able to wire this up to `osv-scalibr` to have it actually trigger the workflow whenever a change is merged in 🤯

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -33,6 +33,11 @@ jobs:
           echo "updating osv-scalibr to $latest_commit"
           go get github.com/google/osv-scalibr@"$latest_commit"
           echo "latest_scalibr_commit=$latest_commit" >> "$GITHUB_ENV"
+          go mod tidy
+      - run: go test ./cmd/osv-scanner/ -run 'Test_run$'
+        env:
+          TEST_ACCEPTANCE: true
+          UPDATE_SNAPS: always
       - uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
         with:
           token: ${{ secrets.PR_TOKEN_BOT }}


.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          persist-credentials: false
+      - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+        with:
+          go-version: stable
+          check-latest: true
+      - run: |
+          latest_commit=$(git ls-remote https://github.com/google/osv-scalibr.git HEAD | cut -f1)
+          echo "updating osv-scalibr to $latest_commit"
+          go get github.com/google/osv-scalibr@"$latest_commit"
```
cuixq, line 34: run `go mod tidy` as well?

cuixq (commented):
There is scalibr's version in snapshot - we may need to update that if osv-scalibr releases a new version.

.github/workflows/dependencies.yml
G-Rath: I would have expected that to already be handled by `go get`? (not so much that it would tidy everything, but that it introduces mess in the first place...)

G-Rath:
> There is scalibr's version in snapshot - we may need to update that if osv-scalibr releases a new version.

have added a new `run` that should hopefully catch that, though I think we'll need to live with it for a bit to see if it does actually work without issue

.github/workflows/dependencies.yml
another-rex: I think go get doesn't remove the old version from the go.sum file, so the tests or lints will complain

G-Rath: ... of course it doesn't ...

another-rex:
The current test failure is caused by a new vuln, running the snapshot generator to fix it now. 

Though I would have thought the snapshot regeneration would have fixed this issue as well? 

Is it because it only run the test for `cmd/osv-scanner` and not `cmd/osv-scanner/...`?

another-rex:
It might actually be worth it to update the snapshots for everything because scalibr updates can cause behavior changes, and this way we can just compare it directly in the diffs, and if it's something we're fine with just merge the PR rather than having to manually update it ourselves.

G-Rath:
yeah for now I've it just update the specific snapshot since this is going to be run at the same cadence as the other workflow and won't know if the snapshots updated because of scalibr or because of changes to vulns.

At this point personally I'd prefer to just get this workflow going so we can know the "update scalibr" part of it is working as we want and then we can get into further changes

(I also wouldn't expect scalibr to result in any changes to our snapshots that wouldn't require a manual touch anyway, aside from the version string?)

PR #2254, opened 2025-10-01
test: use a different name for the default config file and ensure all vulns are ignored

G-Rath:
This should make it a lot easier to author and maintain cmd tests without disappointing tools like scorecard and dependabot, as it means going forward we can just drop a `osv-scanner.toml` configured to ignore everything into any `testdata` fixture that has actual vulns without it impacting our tests.

This is even more useful in the context of https://github.com/google/osv-scanner/pull/2216 which has us report ignored vulns that weren't found, which currently happens a lot in our test suite due to how artificial everything is 😅

Previously we'd discussed making this read from an env variable which I think could be useful for others, but this starts by just making our currently-constant "config file name" string an internally public variable so that we can change it as part of the setup of our tests

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -15,7 +15,7 @@ import (
 	"github.com/google/osv-scanner/v2/internal/imodels"
 )
 
-var OsvScannerConfigName = "osv-scanner.toml"
+var OSVScannerConfigName = "osv-scanner.toml"
 
 type Manager struct {
 	// Override to replace all other configs
@@ -198,7 +198,7 @@ func normalizeConfigLoadPath(target string) (string, error) {
 	} else {
 		containingFolder = target
 	}
-	configPath := filepath.Join(containingFolder, OsvScannerConfigName)
+	configPath := filepath.Join(containingFolder, OSVScannerConfigName)
 
 	return configPath, nil
 }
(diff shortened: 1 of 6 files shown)


internal/config/config.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/google/osv-scanner/v2/internal/imodels"
 )
 
-const osvScannerConfigName = "osv-scanner.toml"
+var OsvScannerConfigName = "osv-scanner.toml"
```
another-rex: `OSVScannerConfigName`

another-rex (approved):
Nice, I forgot we can just have a global variable that we set in tests.

internal/config/config.go
G-Rath: isn't it go convention to not do this? technically we can do it because the linter only enforces a built-in list, but it'll be inconsistent with other stuff like Id, Api, Json, etc

another-rex: https://google.github.io/styleguide/go/decisions#initialisms

All caps is the go convention, and I think we do that mostly in osv-scanner?

Sometimes we have inconsistencies (for Vulnerability.Id specifically) because proto3 generates it like that. But that shouldn't really affect most of osv-scanner (https://github.com/golang/protobuf/issues/555)

G-Rath: oh yeah whoops its the otherway around

PR #2256, opened 2025-10-01
feat: OSV-Scanner MCP server

another-rex:
Basic MVP of an osv-scanner MCP server fulfilling workflow 1. Directly vulnerability scanning of a project with prioritisation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/gkampitakis/go-snaps v0.5.15
 	github.com/go-git/go-git/v5 v5.16.3
 	github.com/google/go-cmp v0.7.0
-	github.com/google/osv-scalibr v0.3.7-0.20251021104859-79bb7c4ea658
+	github.com/google/osv-scalibr v0.3.7-0.20251023161426-90e9ac9cc1b3
 	github.com/ianlancetaylor/demangle v0.0.0-20250628045327-2d64ad6b7ec5
 	github.com/jedib0t/go-pretty/v6 v6.6.8
 	github.com/modelcontextprotocol/go-sdk v1.0.0
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -188,6 +188,25 @@ Total 0 packages affected by 0 known vulnerabilities (0 Critical, 0 High, 0 Medi
 
 ---
 
+[TestCommand/Go_project_with_an_overridden_go_version_and_licences - 1]
+Scanning dir ./testdata/go-project
+Scanned <rootdir>/testdata/go-project/go.mod file and found 1 package
+Total 0 packages affected by 0 known vulnerabilities (0 Critical, 0 High, 0 Medium, 0 Low, 0 Unknown) from 1 ecosystem.
+0 vulnerabilities can be fixed.
+
+
++--------------+-------------------------+
+| LICENSE      | NO. OF PACKAGE VERSIONS |
++--------------+-------------------------+
+| BSD-3-Clause |                       1 |
++--------------+-------------------------+
+
+---
+
+[TestCommand/Go_project_with_an_overridden_go_version_and_licences - 2]
+
+---
+
 [TestCommand/PURL_SBOM_case_sensitivity_(api) - 1]
 Scanning dir ./testdata/sbom-insecure/alpine.cdx.xml
 Scanned <rootdir>/testdata/sbom-insecure/alpine.cdx.xml file and found 15 packages
@@ -747,7 +766,7 @@ Scanned <rootdir>/testdata/sbom-insecure/only-unimportant.spdx.json file and fou
 Scanned <rootdir>/testdata/sbom-insecure/postgres-stretch.cdx.xml file and found 136 packages
 Scanned <rootdir>/testdata/sbom-insecure/with-duplicates.cdx.xml file and found 17 packages
 Filtered 9 local/unscannable package/s from the scan.
-Total 26 packages affected by 159 known vulnerabilities (20 Critical, 67 High, 48 Medium, 1 Low, 23 Unknown) from 4 ecosystems.
+Total 26 packages affected by 159 known vulnerabilities (20 Critical, 68 High, 48 Medium, 1 Low, 22 Unknown) from 4 ecosystems.
 8 vulnerabilities can be fixed.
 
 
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/google/osv-scalibr v0.3.4
 	github.com/ianlancetaylor/demangle v0.0.0-20250628045327-2d64ad6b7ec5
 	github.com/jedib0t/go-pretty/v6 v6.6.8
+	github.com/mark3labs/mcp-go v0.41.1
```
spencerschrock: Scorecard is looking into a MCP server as well, is there a reason you went with this library? We had been looking at the official SDK ( partly maintained by the Go team)?
https://github.com/modelcontextprotocol/go-sdk

I haven't personally looked at alternatives, so I'm just trying to avoid redoing any analysis that may have already been done.

another-rex: I haven't looked too much into the different mcp libraries, I mostly picked go-mcp as it was the most popular by stars, and was not aware that there is an official one. 

But what I found during implementation is that it really doesn't matter too much the mcp library being used, the actual part interfacing with the mcp library is quite minimal, so it should be very easy to switch between different libraries. I might switch to use go-sdk as it looks like an evolution of the go-mcp library that's more officially supported.

cmd/osv-scanner/mcp/command.go
```diff
@@ -0,0 +1,217 @@
...
+
+func handleCodeReview(_ context.Context, _ *mcp.GetPromptRequest) (*mcp.GetPromptResult, error) {
+	return &mcp.GetPromptResult{
+		Description: "Dependency vulnerability analysis",
+		Messages: []*mcp.PromptMessage{
+			{
+				Role: "assistant",
+				Content: &mcp.TextContent{
+					Text: `
+
```
cuixq: can we put the prompt in some const?

internal/output/mcp.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// PrintMCPReport prints a LLM friendly vulnerability report
+func PrintMCPReport(vulnResult *models.VulnerabilityResults, additionalInfo []string, outputWriter io.Writer) error {
+	stringRes := strings.Builder{}
+	stringRes.WriteString(`
+Output results are grouped into (Ecosystem -> Source file -> Packages -> Vulnerabilities),
+with a title for each section, and indentation to indicate that it belongs to the above section.
+When resolving these vulnerabilities, avoid manually updating individual packages, and use system tools.
+Use https://osv.dev/<VULN-ID> as the official record of the vulnerability.
```
cuixq:
```suggestion
Use https://osv.dev/vulnerability/<VULN-ID> as the official record of the vulnerability.
```

.github/workflows/checks.yml
```diff
@@ -16,10 +16,10 @@ name: Checks
 
 on:
   push:
-    branches: [main, v1]
+    branches: ["main", "v1", "mcp"]
```
cuixq, line 19: side node - do we still need v1 branch? maybe something to clean up for fixit week?

cmd/osv-scanner/mcp/command.go
```diff
@@ -0,0 +1,217 @@
...
+		vulnCacheMap[input.VulnID] = vuln
+	}
+
+	return &mcp.CallToolResult{}, vuln, nil
+}
+
+//go:embed configuration-instructions.md
+var configInstructions string
+
+func handleIgnoreVulnerability(ctx context.Context, _ *mcp.CallToolRequest, _ *IgnoreVulnerabilityInput) (*mcp.CallToolResult, any, error) {
```
cuixq: I am a bit confused about `IgnoreVulnerabilityInput` not being used?

another-rex: I can't seem to get it working with no input (either gemin-cli or the mcp framework will get confused and pass in a wrong type), so we need a placeholder input value. We don't actually need the value though. I'll make this clearer in a comment.

.github/workflows/checks.yml
another-rex: There is still significant traffic on v1 versions of osv-scanner, so folks are still using it, I think it's fine to keep around.

cmd/osv-scanner/mcp/command.go
```diff
@@ -0,0 +1,227 @@
...
+
+**Step 3: Prioritisation**
+
+Give advice on which vulnerabilities to prioritise fixing, and general advice on how to go about fixing
+them by updating. Don't try to automatically update for the user without input.
+`
+
+// vulnCacheMap is a cache of vulnerability details that have been retrieved from the OSV API during normal scanning.
+// This avoids unnecessary double queries to the osv.dev API.
+var vulnCacheMap = map[string]*osvschema.Vulnerability{}
```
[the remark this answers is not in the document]
michaelkedar: Could also use a `sync.Map`, I guess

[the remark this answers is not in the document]
another-rex: switched to use sync.Map

cmd/osv-scanner/mcp/command.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/google/osv-scanner/v2/pkg/osvscanner"
+	"github.com/jedib0t/go-pretty/v6/text"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+	"github.com/urfave/cli/v3"
+	"osv.dev/bindings/go/osvdev"
+)
+
+// scanDepsPrompt is the prompt that is sent to the AI model when the scan_deps prompt is requested.
+const scanDepsPrompt = `
```
michaelkedar: since you're already embedding `configuration-instructions.md`, is it worthwhile also putting this in its own file?

cmd/osv-scanner/mcp/command.go
```diff
@@ -0,0 +1,236 @@
...
+
+**Step 3: Prioritisation**
+
+Give advice on which vulnerabilities to prioritise fixing, and general advice on how to go about fixing
+them by updating. Don't try to automatically update for the user without input.
+`
+
+// vulnCacheMap is a cache of vulnerability details that have been retrieved from the OSV API during normal scanning.
+// This avoids unnecessary double queries to the osv.dev API.
+var vulnCacheMap = sync.Map{}
```
michaelkedar, line 30: nit: maybe comment that this is a map[string]*osvschema.Vulnerability

pkg/osvscanner/stats.go
```diff
@@ -8,13 +8,13 @@ import (
 	"github.com/google/osv-scanner/v2/internal/output"
 )
 
-type FileOpenedPrinter struct {
+type fileOpenedPrinter struct {
```
michaelkedar, line 11: technically a breaking change, but it's probably fine

cmd/osv-scanner/mcp/command.go
cuixq: +1

cmd/osv-scanner/mcp/command.go
```diff
@@ -0,0 +1,236 @@
...
+		vulnCacheMap.Store(input.VulnID, vuln)
+	}
+
+	return &mcp.CallToolResult{}, vuln, nil
+}
+
+// ignoreVulnerabilityInput is a placeholder to enable the tool call,
+// as it seems like go-sdk mcp does not support a tool call with no arguments.
+type ignoreVulnerabilityInput struct {
+	Verbose bool `json:"verbose" jsonschema:"ignore this parameter"`
```
cuixq, line 189: add a comment about why some field is needed here.

cmd/osv-scanner/mcp/command.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+//go:embed configuration-instructions.md
+var configInstructions string
+
+func handleIgnoreVulnerability(_ context.Context, _ *mcp.CallToolRequest, _ *ignoreVulnerabilityInput) (*mcp.CallToolResult, any, error) {
```
cuixq, line 197: also add a comment to `handleIgnoreVulnerability` explaining how this works relying on the instructions but not the input.

cuixq (approved):

PR #2267, opened 2025-10-07
test: introduce a `locks-many-with-insecure` collection of testdata

G-Rath:
Currently a lot of our tests use specific files from the `testdata/locks-many` set for stuff that doesn't actually need vulnerabilities meaning they'll trigger the warning about ignored vulns that weren't found that is being introduced in #2216 since there are two vulns in that set.

To make things cleaner I've copied the existing testdata set into a new `locks-many-with-insecure`, removed the vulns from the original testdata set, and updated any tests that do actually want vulnerabilities to use the new set


G-Rath:
For #2216 this is the before and after

Before:

https://github.com/google/osv-scanner/pull/2216/commits/f8443e7e30f709d74e125ba78a7740867281ff41

After:

https://github.com/google/osv-scanner/pull/2216/commits/3da7dedba730a3e9da70c447b271e8fbd647b488

PR #2268, opened 2025-10-07
ci: ensure filenames don't have undesired characters

G-Rath:
This should ensure that files are valid across OSs - I've gone with an inclusive pattern which purposely does not have some characters that are technically legal but that I don't think there's a good reason to use them in a filename.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -42,6 +42,16 @@ jobs:
         with:
           persist-credentials: false
       - run: scripts/report_uncleaned_snapshots.py
+  filenames:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - run: |
+          find . -mindepth 1 ! -regex '.*/[#@A-Za-z0-9._-]*' \
+            | xargs -I '{}' bash -c \
+              "echo '{}' && echo '::error file={}::This filename contains undesired characters' && false"
   format:
     permissions:
       contents: read # to fetch code (actions/checkout)
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -49,9 +49,9 @@ jobs:
         with:
           persist-credentials: false
       - run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -42,6 +42,16 @@ jobs:
...
           persist-credentials: false
       - run: scripts/report_uncleaned_snapshots.py
+  filenames:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - run: |
+          find . -mindepth 1 ! -regex '.*/[#@A-Za-z0-9._-]*' \
```
another-rex: Does the initial `.*/` mean the folder names can have invalid characters?

G-Rath: nope:

```
osv-scalibr on  ci/check-filenames [?] via 🐹 v1.24.4 via  v20.19.0
❯ find . -mindepth 1 ! -regex '.*/[#@A-Za-z0-9._-]*' | xargs -I '{}' bash -c "echo '{}' && echo '::error file={}::This filename contains undesired characters' && false"
./bad()
::error file=./bad()::This filename contains undesired characters
./bad()/bad:file.json
::error file=./bad()/bad:file.json::This filename contains undesired characters
./bad:dir
::error file=./bad:dir::This filename contains undesired characters
./.idea/shelf/semantic__use_`convertToBigInt`_instead_of_regexp
::error file=./.idea/shelf/semantic__use_`convertToBigInt`_instead_of_regexp::This filename contains undesired characters
./.idea/shelf/Uncommitted_changes_before_Checkout_at_2025-03-06_8_18_AM_[Changes]
::error file=./.idea/shelf/Uncommitted_changes_before_Checkout_at_2025-03-06_8_18_AM_[Changes]::This filename contains undesired characters
./bad:dir2
::error file=./bad:dir2::This filename contains undesired characters
```

another-rex (approved):
Nice, LGTM

G-Rath:
I have just discovered a bug though, so get ready to re-review 😅

another-rex (approved):
shell scripts has some magical characters...

PR #2279, opened 2025-10-14
test: hide the cli help flag to avoid panic when running parallel

G-Rath:
Unfortunately the upstream package does not feel there [is a reason to remove their global help flag variable hack](https://github.com/urfave/cli/issues/2176#issuecomment-3193884742) that can cause a panic when running in parallel, so this has us hide the help flag for all commands when running the tests so that the code path is skipped entirely


another-rex (approved):
Nice!

PR #2284, opened 2025-10-17
feat: Update scalibr to the latest version

another-rex:
Update osv-scalibr to the latest version. This required a few changes:

- Remove baseimagematch and use the base image enricher in osv-scalibr
- Add enrichers to the preset.go
- Custom clients for enrichers to set user agent. Create a insight client helper function for v3alpha
- 

I'm not sure what the snapshot update is for, and how we are getting new severity when we weren't before though.

Enrichers might be making network requests during initialisation (I need to look into how grpc initialisation works). This will be a problem if --offline is set to true.

### DIFF
--- a/internal/scalibrplugin/presets.go
+++ b/internal/scalibrplugin/presets.go
@@ -148,7 +148,7 @@ var enricherPresets = map[string]enricherlist.InitMap{
 }
 
 func baseImageEnricher() enricher.Enricher {
-	// The grpc client **does not** make any requests are start in an IDLE state until
+	// The grpc client **does not** make any requests. It starts in an IDLE state until
 	// the first function call is made. This means we can safely initialize the client even in offline mode,
 	// and the enricher plugin will be filtered out in offline mode.
 	insightsClient, err := datasource.NewInsightsAlphaClient(depsdev.DepsdevAPI, "osv-scanner_scan/"+version.OSVVersion)
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -561,8 +561,7 @@ func SetLogger(handler slog.Handler) {
 	cmdlogger.GlobalHandler = handler
 }
 
-// inventoryIsEmpty is a custom inventory is empty check because the default check
-// returns false if image metadata is filled. We want to still return empty if there is image metadata
+// inventoryIsEmpty ignores image metadata when checking if an inventory is empty
 func inventoryIsEmpty(i inventory.Inventory) bool {
 	if len(i.Packages) != 0 {
 		return false


internal/datasource/insightsalpha.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"crypto/x509"
+	"fmt"
+
+	pb "deps.dev/api/v3alpha"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+func NewInsightsAlphaClient(addr string, userAgent string) (pb.InsightsClient, error) {
```
cuixq, line 13: can you add a comment for this client?

internal/scalibrplugin/presets.go
```diff
@@ -148,16 +148,19 @@ var enricherPresets = map[string]enricherlist.InitMap{
 }
 
 func baseImageEnricher() enricher.Enricher {
+	// The grpc client **does not** make any requests are start in an IDLE state until
```
G-Rath: umm this reads weirdly - it looks like you couldn't decide between "the client does not make any request until ..." and "the client starts in an IDLE state until ...", and so mashed them together 😅

internal/scalibrplugin/resolve.go
G-Rath, line 1: I don't see why this change was needed - my guess would be maybe we've now got a thing whose name also happens to match a preset, but I don't think that is the case?

pkg/osvscanner/osvscanner.go
```diff
@@ -564,3 +560,22 @@ func overrideGoVersion(scanResults *results.ScanResults) {
 func SetLogger(handler slog.Handler) {
 	cmdlogger.GlobalHandler = handler
 }
+
+// inventoryIsEmpty is a custom inventory is empty check because the default check
+// returns false if image metadata is filled. We want to still return empty if there is image metadata
+func inventoryIsEmpty(i inventory.Inventory) bool {
```
G-Rath: I feel like a phrasing like this would be better:


```suggestion
// inventoryIsEmpty ignores image metadata when checking if an inventory is empty
func inventoryIsEmpty(i inventory.Inventory) bool {
```

I'm also just thinking about what if more fields are added - I guess the only way to have the compiler catch that for us would be to explicitly build a new inventory without field names and with metadata explicitly empty 🤔 

not sure if that is worth it, so will let you decide 🙂

G-Rath (approved):

internal/scalibrplugin/resolve.go
another-rex: The issue is I wanted base image enricher to be enabled by default, so I called the preset `artifact`. Now that the same preset name is repeated in the enricher list and extractor list continue stops working.

pkg/osvscanner/osvscanner.go
another-rex: I don't think that's worth it for now, especially since we don't add top level inventory fields that often.

PR #2294, opened 2025-10-22
feat: replace `requirementsenhanceable` extractor with transitive enricher

### LINKED ISSUE #2289: Use `transitivedependency` enricher for transitive extraction for requirements.txt
Currently, OSV-Scanner depends on [`requirementsnet` extractor](https://github.com/google/osv-scalibr/tree/main/extractor/filesystem/language/python/requirementsnet) for transitive extraction for `requirements.txt`. 

However, OSV-Scalibr is migrating to use [`transitivedependency` enricher](https://github.com/google/osv-scalibr/tree/main/enricher/transitivedependency/requirements) for this purpose.

We probably should migrate [`requirementsenhancable`](https://github.com/google/osv-scanner/tree/main/internal/scalibrextract/language/python/requirementsenhancable) to use the enricher as well.

G-Rath:
On the surface this was pretty much a drop-in replacement which is really exciting, but there are a couple of quirks that we may or may not want to deal with before landing this.

The first one is very obvious which is that our "found packages" count is very different - it's technically correct, but I think ideally it would be good if we could express that we found x more packages via the enricher.

The second one is more interesting and I suspect one we'll probably ignore for the time being, but right now `scalibr` implicitly adds extractors that are required by enrichers as part of doing the scan which is at odds with our (experimental) abilities to enable and disable plugins - for now I've made it so that the enricher is only enabled if the `requirements` extractor is enabled, since we don't have that many enrichers with plugin requirements right now.

Resolves #2289

### DIFF
--- a/pkg/osvscanner/scan.go
+++ b/pkg/osvscanner/scan.go
@@ -114,6 +114,8 @@ func scan(accessors ExternalAccessors, actions ScannerActions) (*imodels.ScanRes
 		actions,
 	)
 
+	// technically having one detector enabled would also be sufficient, but we're
+	// not mentioning them to avoid confusion since they're still in their infancy
 	if countNotEnrichers(plugins) == 0 {
 		return nil, errors.New("at least one extractor must be enabled")
 	}
--- a/pkg/osvscanner/osvscanner.go
+++ b/pkg/osvscanner/osvscanner.go
@@ -290,6 +290,8 @@ func DoContainerScan(actions ScannerActions) (models.VulnerabilityResults, error
 		actions,
 	)
 
+	// technically having one detector enabled would also be sufficient, but we're
+	// not mentioning them to avoid confusion since they're still in their infancy
 	if countNotEnrichers(plugins) == 0 {
 		return models.VulnerabilityResults{}, errors.New("at least one extractor must be enabled")
 	}


pkg/osvscanner/scan.go
```diff
@@ -101,7 +125,7 @@ func scan(accessors ExternalAccessors, actions ScannerActions) (*imodels.ScanRes
 		actions,
 	)
 
-	if len(plugins) == 0 {
+	if countNotEnrichers(plugins) == 0 {
```
cuixq, line 119: if we want to make sure at least one extractor is enabled, shall we count the number of extractors?

cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
```diff
@@ -1329,15 +1329,15 @@ unsupported output format "unknown" - must be one of: table, html, vertical, jso
...
-Scanned <rootdir>/testdata/locks-requirements/my-requirements.txt file and found 6 packages
-Scanned <rootdir>/testdata/locks-requirements/requirements-dev.txt file and found 8 packages
-Scanned <rootdir>/testdata/locks-requirements/requirements-transitive.txt file and found 14 packages
-Scanned <rootdir>/testdata/locks-requirements/requirements.prod.txt file and found 3 packages
-Scanned <rootdir>/testdata/locks-requirements/requirements.txt file and found 13 packages
+Scanned <rootdir>/testdata/locks-requirements/my-requirements.txt file and found 1 package
+Scanned <rootdir>/testdata/locks-requirements/requirements-dev.txt file and found 1 package
+Scanned <rootdir>/testdata/locks-requirements/requirements-transitive.txt file and found 4 packages
+Scanned <rootdir>/testdata/locks-requirements/requirements.prod.txt file and found 1 package
+Scanned <rootdir>/testdata/locks-requirements/requirements.txt file and found 3 packages
```
cuixq, line 1378: hmmm it will be nice that we have some logging to indicate the enircher has been run even if we don't have the number of extra packages extracted.

G-Rath: Do you have any thoughts on how that should look and how to do it? scalibr currently doesn't look to provide an `AfterEnricher` hook like it does for extractors and detectors, and enrichers look to be a lot more conditional so I'm not sure if we can reliable report on enrichers

pkg/osvscanner/scan.go
G-Rath: this is just preserving our current logic which technically doesn't account for detectors, and I think that's ok for now because the long-run goal is to be running detectors, it's just right now we don't have enough stable ones to advertise them or justify making us less "extractor-first" 🤷

cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
cuixq: I am not familiar with how post-run hooks run but I do see value adding an `AfterEnricher` hook - whether the stats is reliable sounds another problem.

Or alternative, we can add more logging to this transitive dependency enricher to provide more information. I will take a look at this later.

pkg/osvscanner/scan.go
cuixq: do you mean we are not currently running detectors and we would like to run them in the future?

also can you add a comment for this? the inconsistency between the function name and logging confuses me.

G-Rath: Currently we support running detectors via `--experimental-plugins` but we don't have any enabled by default, and the scanner was built with a focus around extractor-type plugins and vulnerabilities-type findings even though now we have stuff like detectors-type plugins and secrets-type findings, which is why we've got oddities like this messaging...

My understanding is that we're mostly ignoring this for now because there's still a lot of discussion going on how these sorts of things should be supported.

It's probably worth checking what @another-rex thinks too

cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
another-rex: AfterEnricher stats hook is likely the best way to do this, we should avoid logging directly in the transitive extractor.

pkg/osvscanner/scan.go
another-rex: Happy to leave it like this for now. Though probably a good idea to add a quick comment about why the error message mentions extractors if we are also counting detectors.

Eventually we'll want to support at least 1 enabled of: filesystem extractors, standalone extractors, detectors. 

Then in addition you can add Annotators and Enrichers.

pkg/osvscanner/scan.go
```diff
@@ -69,6 +75,11 @@ func getPlugins(defaultPlugins []string, accessors ExternalAccessors, actions Sc
 
 	plugins := scalibrplugin.Resolve(actions.PluginsEnabled, actions.PluginsDisabled)
 
+	// todo: use Enricher.RequiredPlugins to check this generically
+	if accessors.DependencyClients[osvschema.EcosystemPyPI] != nil && isRequirementsExtractorEnabled(plugins) {
```
another-rex: Hmm... We are now doing this in two different ways, and we should probably standardise on a specific way here.

Here we are depending on the accessors system, where we initialise the correct clients based on the input flags. 

However, for base images, I tried just initialising the base image enricher no matter what, and let it be filtered out by Capabilities in osv-scalibr. Since no actual network requests are made during the initiation, it should appear to be identical to the end user.

I can kind of see some pros and cons of both approaches?

Using accessors:
- **Pro:** Lazy initiazation, we don't have to do as much work if disabled by the flags.
- More control in osv-scanner about what is enabled/disabled
- **Con:** We'll be replicating the logic of capabilities to filter out plugins
- It's hard to figure out how to initialize custom enrichers if the accessors are disabled by the flag. 

Any opinions/suggestions on which way to go here?

cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
cuixq: considering `AfterEnricher` may not happen immediately, I think we can add some logging in the enricher.

I don't see why we should avoid logging in transitive extractor for extra information? we already have the logging about failed resolution.

pkg/osvscanner/scan.go
cuixq: as per offline discussion, considering we are moving towards using enrichers for job that requires network access, I am leaning to getting rid of the accessors.

G-Rath: yeah I think that's the right direction if "no actual network requests are made during the initiation" holds (+ "initialization isn't expensive")

G-Rath: Just to be clear though, that change is relating to the `--no-resolve` flag, where we either remove it entirely or change the implication to be "we don't add enrichers for resolving transitive dependencies"

pkg/osvscanner/scan.go
G-Rath: have added a comment about this

G-Rath:
@cuixq fwiw it seems like the enricher might be flakier than the existing extractor? it doesn't seem to be too common and might be a third-party thing, but afaik we've never had any flakey issues with the existing extractor.

See [this run for an example](https://github.com/google/osv-scanner/actions/runs/18760128680/job/53522494903):


```
--- FAIL: TestCommand (0.00s)
    --- FAIL: TestCommand/requirements.txt_can_have_all_kinds_of_names (6.94s)
        command_test.go:331: 
            - Snapshot - 2
            + Received + 2
            
            @@ -6,8 +6,8 @@
            
              Scanned <rootdir>/testdata/locks-requirements/requirements.txt file and found 3 packages
              Scanned <rootdir>/testdata/locks-requirements/the_requirements_for_test.txt file and found 1 package
              Scanned <rootdir>/testdata/locks-requirements/unresolvable-requirements.txt file and found 3 packages
            - failed resolution: no file can be used for parsing requirements for package flask-cors version 1.0
            - failed to parse metadata for file Flask-Cors-1.0.tar.gz: sdist: dependencies in setup.py, not in PKG-INFO
            + failed resolution: resolution impossible:
            + no candidates at all for: pytz ">=2011k"
              Total 15 packages affected by 53 known vulnerabilities (2 Critical, 20 High, 28 Medium, 0 Low, 3 Unknown) from 1 ecosystem.
              53 vulnerabilities can be fixed.
              ↵
            
            at __snapshots__\command_test.snap:1349
```

cuixq:
this error seems not like a flaky error: both resolution failure should be deterministic if the artifacts don't change - not sure if related to the artifacts from registry. and it only failed for windows :thinking: 

maybe don't worry too much about it now - see if this is going to happen again.

G-Rath:
yeah at this point I think we should just keep an eye on it, but fwiw its [not just a Windows thing](https://github.com/google/osv-scanner/actions/runs/18702900496/job/53335239290).

another-rex:
Huh reran this and it failed again immediately.

another-rex:
@cuixq Is there some retry logic that's missing in the osv-scalibr impl?

cuixq:
I missed the comments from last week - I will take a look later.

cuixq:
I think this is probably due to the order of packages is not deterministic after grouping them into a map: https://github.com/google/osv-scalibr/blob/main/enricher/transitivedependency/requirements/requirements.go#L86

this should be easy to fix since we already record the index in the map values.

cuixq:
however I think pip's [backtracking](https://pip.pypa.io/en/stable/topics/dependency-resolution/#backtracking) dependency resolution may not be affected by the order of dependencies, but this is a thing to follow up with the resolver instead of scalibr. 

For now I will make the change to keep the order of the packages as they are extracted and see if this helps with the flaky test results.

cuixq:
@G-Rath sorry I merged in the wrong snapshot you may need to re-generate them and push again.

cuixq:
https://github.com/google/osv-scalibr/pull/1532 is merged so I hope the flaky test should be fixed.

We probably still want to wait until https://github.com/google/osv-scanner/pull/2328 is in so we can update osv-scalibr further.

cuixq (approved):

cuixq:
@G-Rath could you help to merge the main branch add see if https://github.com/google/osv-scalibr/pull/1532 fixes the flaky test?

cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
another-rex: Benefit of the hook is that it makes it easy for us to customise the logging output, which is harder to do with it being inside the enricher. Happy with adding something in the short term though

another-rex:
Rerunning again just to make sure, otherwise good to merge.

PR #2300, opened 2025-10-24
fix: use global logger in cmdlogger when set

### LINKED ISSUE #2081: Allow configuration of the library's logging

spencerschrock:
Previously this code would always use the default slog handler, even if a user called `osvscanner.SetLogger`.

Also set osv-scalibr logger to same global logger. A small adapter was needed since osv-scalibr has its own logging interface.

Fixes #2081

### DIFF
--- a/pkg/osvscanner/osvscanner_test.go
+++ b/pkg/osvscanner/osvscanner_test.go
@@ -9,7 +9,15 @@ import (
 )
 
 // TestDoScan_LogHandlerOverride tests that the SetLogger override works correctly
+//
+//nolint:paralleltest // No parallel test since slog.SetDefault sets global behavior
 func TestDoScan_LogHandlerOverride(t *testing.T) {
+	// Restore default slog behavior at the ned of the test
+	defaultHandler := slog.Default()
+	defer func() {
+		slog.SetDefault(defaultHandler)
+	}()
+
 	actions := osvscanner.ScannerActions{
 		DirectoryPaths: []string{"../../cmd/osv-scanner/testdata/locks-many/Gemfile.lock"},
 	}


internal/cmdlogger/handler.go
```diff
@@ -8,7 +8,10 @@ import (
 	"strings"
 )
 
-var GlobalHandler slog.Handler
+var (
+	GlobalHandler slog.Handler
+	GlobalLogger  *slog.Logger
+)
```
spencerschrock: also happy to do this here if you want to avoid the `cmp.Or` and `slog.Default` calls in fmt.go:
```
GlobalLogger = slog.Default
```

spencerschrock:
> Also set osv-scalibr logger to same global logger. A small adapter was needed since osv-scalibr has its own logging interface.

The adapter and the call to osv-scalibr/log.SetLogger could live in Scorecard if needed, it just seemed applicable for a library logger to also set its libraries loggers.

another-rex:
Hmm... I'm not sure why the previous version was not working for you. We need the custom handler to set the correct error code when printing an error message, then redirect to the new handler.

another-rex:
Ah I see we are not setting the slog.handler in DoScan()

another-rex:
Looked at my original fix again, I'm not sure why I thought that will work 🤦. Thanks for the fix! I updated it a bit to fulfill the original intent of the GlobalHandler override, and added a test to verify.

PR #2309, opened 2025-10-30
fix: exit with a non-zero code when showing help

### LINKED ISSUE #2303: pre-commit hook performs should performs a scan by default
When `osv-scanner` is running from a pre-commit hook without `args` it does not fail.

Steps to reproduce:

1. Create a `.pre-commit-config.yaml` with the following content:
```yaml
repos:
  - repo: https://github.com/google/osv-scanner/
    rev: "v2.2.3"
    hooks:
      - id: osv-scanner
```

2. Run
```bash
pre-commit run --all-files --verbose osv-scanner
```
Image

I understand the [documentation](https://google.github.io/osv-scanner/usage/#pre-commit-integration) clearly states it should be used with `args`,  but i would argue that, when that's not the case, the hook should fail. 

The same behavior is observed when `osv-scanner` is run from the command line. 0 (zero) is returned as an exit code instead of non zero value.

This behavior hides mistakes like incorrect indentation on the pre-commit hook configuration or the all too common misspelled variable name in a bash script.

G-Rath:
Note I've purposely not used a new error code for now because I can't imagine a use-case for wanting to identify this particular situation in a scripting context, which is where exit code values are relevant (we can always use a more specific code later if a use case is presented)

Resolves #2303

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -177,8 +177,6 @@ jobs:
           ) | .name' output.json | while read -r image; do
             echo "Testing image $image"
 
-            docker run $image -h
-
             exit_code=0
             docker run -v ${PWD}:/src $image -L /src/go.mod || exit_code=$?
 


another-rex:
The scripting case I'm thinking of is less identifying this specifically, but differentiating argument errors vs some catch all transient errors that might occur. Right now any error that we don't specifically account for is 127.

G-Rath:
But that's still a scripting context right, in which case I don't see this as being a failure that is really actionable by the user?

another-rex:
I think the difference is if it's something caused by a network error, it'll go away by itself hopefully, so you don't need to do anything about it. But with invalid args you have to do something.

Hmm I probably have to think through this a bit more, let's merge this as is for now.

another-rex (approved):

PR #2311, opened 2025-10-31
fix: improve logic for `--all-packages` flag 

### LINKED ISSUE #2157: `--all-packages` flag should also show unscannable filtered packages.

Ly-Joey:
Updated logic so that unscannable packages are still filtered for vulnerabilities and license matching, but they will be re-added to the final JSON result if the `--all-packages` flag is passed in.

Also updated docs to clarify the order of precedence between `--all-packages` flag and `PackageOverrides` actions in the config file.

Fixes #2157

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/gkampitakis/go-snaps v0.5.15
 	github.com/go-git/go-git/v5 v5.16.3
 	github.com/google/go-cmp v0.7.0
-	github.com/google/osv-scalibr v0.3.7-0.20251028103653-ba3117fbf7a7
+	github.com/google/osv-scalibr v0.3.7-0.20251029213949-57f9a8bf7775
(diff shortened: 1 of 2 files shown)


another-rex (approved):
Nice! LGTM

another-rex:
Test error seems to be from a snapshot update, just run tests with UPDATE_SNAPS=true and it should be fixed.

PR #2313, opened 2025-10-31
feat: Use OSV-Scalibr directly

another-rex:
No more single lockfile extraction logic, just use the osv-scalibr scan function.

To keep the behaviour similar to what we currently have (mostly to make testing for the correct behaviour easier, we can change it later).

Some notes:
- OSV-Scalibr does not deduplicate the files being walked, so to prevent files from being extracted multiple times, we exclude child directories if they are covered by another scan directory. Ideally this is done in osv-scalibr.
- We keep track of all the specific files that are passed in via -L or --sbom, so that we can make sure to error if they fail to extract. This required using the stats collector to see all files that the extractor has ran on, which breaks the custom extractor input. We'll likely refactor this soon, so just left the field there for now.

### DIFF
--- a/pkg/osvscanner/__snapshots__/osvscanner_test.snap
+++ b/pkg/osvscanner/__snapshots__/osvscanner_test.snap
@@ -15,27 +15,27 @@
             "ecosystem": "",
             "commit": "33dffa3909a67e1b5d22647128ab7eb6e53fd0c7"
           },
-          "vulnerabilities": [
+          "groups": [
             {
-              "modified": "2025-11-06T01:21:21Z",
-              "published": "2025-10-31T07:15:38Z",
-              "schema_version": "1.7.3",
-              "id": "CVE-2025-63675",
+              "ids": [
+                "CVE-2025-63675"
+              ],
               "aliases": [
+                "CVE-2025-63675",
                 "GHSA-97w9-v595-3h5q"
               ],
-              "details": "cryptidy through 1.2.4 allows code execution via untrusted data because pickle.loads is used. This occurs in aes_decrypt_message in symmetric_encryption.py.",
-              "severity": [
-                {
-                  "type": "CVSS_V3",
-                  "score": "CVSS:3.1/AV:L/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:L"
-                }
-              ],
+              "max_severity": "6.9"
+            }
+          ],
+          "vulnerabilities": [
+            {
               "affected": [
                 {
+                  "database_specific": {
+                    "source": "https://storage.googleapis.com/cve-osv-conversion/osv-output/CVE-2025-63675.json"
+                  },
                   "ranges": [
                     {
-                      "type": "GIT",
                       "events": [
                         {
                           "introduced": "0"
@@ -44,7 +44,8 @@
                           "last_affected": "33dffa3909a67e1b5d22647128ab7eb6e53fd0c7"
                         }
                       ],
-                      "repo": "https://github.com/netinvent/cryptidy"
+                      "repo": "https://github.com/netinvent/cryptidy",
+                      "type": "GIT"
                     }
                   ],
                   "versions": [
@@ -57,12 +58,16 @@
                     "v1.2.2",
                     "v1.2.3",
                     "v1.2.4"
-                  ],
-                  "database_specific": {
-                    "source": "https://storage.googleapis.com/cve-osv-conversion/osv-output/CVE-2025-63675.json"
-                  }
+                  ]
                 }
               ],
+              "aliases": [
+                "GHSA-97w9-v595-3h5q"
+              ],
+              "details": "cryptidy through 1.2.4 allows code execution via untrusted data because pickle.loads is used. This occurs in aes_decrypt_message in symmetric_encryption.py.",
+              "id": "CVE-2025-63675",
+              "modified": "2025-11-06T01:21:21.629057Z",
+              "published": "2025-10-31T07:15:38Z",
               "references": [
                 {
                   "type": "PACKAGE",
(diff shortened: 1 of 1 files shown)


internal/utility/types/cast.go
```diff
@@ -0,0 +1,11 @@
+// Package types provides type conversion utility functions
+package types
+
+func MustCastSlice[OUT, IN any](a []IN) []OUT {
```
cuixq: is this used anywhere?

pkg/osvscanner/osvscanner.go
```diff
@@ -77,6 +77,7 @@ type ExperimentalScannerActions struct {
 	PluginsDisabled   []string
 	PluginsNoDefaults bool
 
+	// Currently unused.
```
cuixq, line 80: add a TODO?

pkg/osvscanner/scan.go
```diff
@@ -260,6 +290,68 @@ func scan(accessors ExternalAccessors, actions ScannerActions) (*imodels.ScanRes
...
+		if isDescendent(existing, absPath, recursive) {
+			return absPath, nil
+		}
+	}
+	rootMap[root] = append(rootMap[root], absPath)
+
+	return absPath, nil
+}
+
+func isDescendent(potentialParent, path string, recursive bool) bool {
```
cuixq, line 324: can you add a comment for this function?

pkg/osvscanner/scan.go
```diff
@@ -202,30 +194,68 @@ func scan(accessors ExternalAccessors, actions ScannerActions) (*imodels.ScanRes
...
 			SkipDirGlob:           nil,
 			UseGitignore:          !actions.NoIgnore,
-			Stats:                 statsCollector,
+			Stats:                 &statsCollector,
 			ReadSymlinks:          false,
 			MaxInodes:             0,
 			StoreAbsolutePath:     true,
 			PrintDurationAnalysis: false,
 			ErrorOnFSErrors:       false,
+			ExtractorOverride: func(api filesystem.FileAPI) []filesystem.Extractor {
```
cuixq, line 212: is this used somewhere?

internal/utility/types/cast.go
another-rex: No this is not, it was originally used in an earlier iteration, but I kept it in as it'll be useful just generally, and I don't want to write it again in the future.

michaelkedar: Should add a comment to the function, especially to explain the behaviour (panics?)

>  I don't want to write it again in the future.

It's only like 4 lines of code 🙂

cuixq: If we check-in some code because it's useful in the future, we may end up a codebase with a lot code that is not necessarily needed. since this code sits in its own file, you can still keep it without checking it in.

pkg/osvscanner/internal/scanners/lockfile.go
```diff
@@ -76,55 +75,26 @@ var lockfileExtractorMapping = map[string][]string{
...
 }
 
-// ScanSingleFile is similar to ScanSingleFileWithMapping, just without supporting the <lockfileformat>:/path/to/lockfile prefix identifier
-func ScanSingleFile(path string, extractorsToUse []plugin.Plugin) ([]*extractor.Package, error) {
-	invs, err := scalibrextract.ExtractWithExtractors(context.Background(), path, extractorsToUse)
-	if err != nil {
-		return nil, err
+func ParseLockfilePath(scanArg string) (string, string) {
+	if (runtime.GOOS == "windows" && filepath.IsAbs(scanArg)) || !strings.Contains(scanArg, ":") {
+		scanArg = ":" + scanArg
```
michaelkedar: nit: I feel like you might as well just `return "", scanArg` here

and in general, this function could use `strings.Cut(scanArg, ":")` instead of `strings.Contains` and `strings.SplitN`

pkg/osvscanner/scan.go
```diff
@@ -110,75 +98,79 @@ func scan(accessors ExternalAccessors, actions ScannerActions) (*imodels.ScanRes
...
+	// Build list of paths for each root
+	// On linux this would return a map with just one entry of /
+	rootMap := map[string][]string{}
+
+	// Also build a map of specific plugin overrides that the user specify
+	// map[path]parseAs
+	overrideMap := map[string]filesystem.Extractor{}
+	// List of specific paths the user passes in so that we can check that they all get processed.
+	//nolint:prealloc // Does not matter in this case
+	var specificPaths []string
```
michaelkedar: Wouldn't this just be
```go
	specificPaths := make([]string, 0, len(actions.LockfilePaths) + len(actions.SBOMPaths))
```

I feel like we should avoid using `nolint` if possible (or, disable the lint check altogether)

pkg/osvscanner/osvscanner.go
another-rex: Good idea, done

pkg/osvscanner/scan.go
another-rex: Added a description.

pkg/osvscanner/scan.go
another-rex: In Scalibr? yes

pkg/osvscanner/stats.go
```diff
@@ -10,11 +10,14 @@ import (
...
 	stats.NoopCollector
+
+	filesExtracted map[string]struct{}
 }
 
 var _ stats.Collector = &fileOpenedPrinter{}
 
-func (c fileOpenedPrinter) AfterExtractorRun(_ string, extractorstats *stats.AfterExtractorStats) {
+func (c *fileOpenedPrinter) AfterExtractorRun(_ string, extractorstats *stats.AfterExtractorStats) {
+	c.filesExtracted[extractorstats.Path] = struct{}{}
```
michaelkedar: I think we should either make a `newFileOpenedPrinter()` function that `make`s this map, or add here a 
```go
if c.filesExtracted == nil {
  c.filesExtracted = make(map[string]struct{})
}
```
so we don't have to rely on the creator of this struct initializing this map.

pkg/osvscanner/scan.go
another-rex: yeah, good point, we probably should just get rid of this lint entirely. I'll leave it to a separate PR and fix this.

pkg/osvscanner/stats.go
another-rex: Done, just did the if option for now.

pkg/osvscanner/internal/scanners/lockfile.go
another-rex: Done.

pkg/osvscanner/stats.go
michaelkedar: (nit: I think using `make` is more idiomatic than setting to the empty map)

internal/utility/types/cast.go
another-rex: Ok, removed now.

pkg/osvscanner/scan.go
```diff
@@ -260,6 +289,71 @@ func scan(accessors ExternalAccessors, actions ScannerActions) (*imodels.ScanRes
...
+		return true
+	}
+
+	if depths == 1 {
+		// Direct child of existing dir, skip
+		return true
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
another-rex: This is just incorrect, scalibr does not scan recursively when not in recursive mode.

PR #2318, opened 2025-11-03
chore: update universe and snapshots

cuixq:
This PR updates `universe.yaml` files as well as the relevant snapshots so that diff will be minimized if new formats of universe are generated.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/google/osv-scanner/v2
 
-go 1.24.6
+go 1.24.9
 
 require (
 	deps.dev/api/v3 v3.0.0-20250917073939-6ff3dd7d2eea


internal/remediation/testdata/override-workaround/universe.yaml
```diff
@@ -1,79 +1,1734 @@
+# Automatically generated by generate_mock_resolution_universe on 04 Nov 25 08:23 AEDT. DO NOT EDIT.
```
michaelkedar: I'm pretty sure I had manually created this file

cuixq: I reverted this file. PTAL.

michaelkedar (approved):

PR #2320, opened 2025-11-04
test: rewrite json replacement logic to better handle arrays and existence

G-Rath:
Currently, attempting to replace properties within arrays on objects that don't always have the same structure all the way down can result in undesired behaviours such as the property being added to all objects within arrays on the path, or only half of the potential matches being replaced.

While the problem is a bit awkward to describe, the solution is not: now we fully unroll a replacement path, and apply the replacement logic on each resulting path.

i.e.with the object `{ "arr": [1, 2, 3] }`, we unroll the path `arr.#` to `[]string{"arr.0", "arr.1", "arr.2"}`.

In addition we now also check that the property actually exists before doing the replacement to avoid adding replacements.

### DIFF
--- a/cmd/osv-scanner/internal/testcmd/jsonreplace.go
+++ b/cmd/osv-scanner/internal/testcmd/jsonreplace.go
@@ -143,6 +143,7 @@ func replaceJSONInput(t *testing.T, jsonInput string, path string, replacer func
 			continue
 		}
 
+		// optimistically replace the element, since we know at this point it does exist
 		json, err = sjson.SetOptions(json, pathElem, replacer(res), &sjson.Options{Optimistic: true})
 		if err != nil {
 			t.Fatalf("failed to set element")


cmd/osv-scanner/internal/testcmd/jsonreplace.go
```diff
@@ -91,62 +91,63 @@ var (
 	}
 )
 
-// replaceJSONInput takes a gjson path and replaces all elements the path matches with the output of matcher
-func replaceJSONInput(t *testing.T, jsonInput, path string, matcher func(toReplace gjson.Result) any) string {
+func expandArrayPaths(t *testing.T, jsonInput string, path string) []string {
```
cuixq, line 94: can you add some comment to this function?

G-Rath: are you meaning a comment on the function itself? I've already done my best to add comments throughout the function to explain its logic, and I didn't think a function-level comment was needed given the clear name, inner comments, and that the previous function didn't have one.

Note that the git diff seems to have been a bit mucked up by me introducing this above the existing function, so that might be what has thrown you off

cuixq: I am confused about what this function is for - I vaguely have the idea that it is related to path replacement, but what is the expected input? e.g. why there is "#" in the path?

> i.e.with the object { "arr": [1, 2, 3] }, we unroll arr.# to []string{"arr.0", "arr.1", "arr.2"}.

this description is also a bit confusing to me: where does `[1, 2, 3]` come from? is `arr.#` the pattern in JSON? why they are replaced as  `"arr.0", "arr.1", "arr.2"` but not `"arr.1", "arr.2", "arr.3"`?

G-Rath: I wonder is this because of the visibility of the method, or lack thereof? I feel like you're assuming it'll be used on its own when in practice it's not meant to be - it gets used by `replaceJSONInput`, which states it takes a `gjson` path [whose syntax includes the `#`](https://github.com/tidwall/gjson/blob/master/SYNTAX.md#arrays).

You might find it easier to understand looking at the [osv.dev version of this pr](https://github.com/google/osv.dev/pull/4306) as that uses an internal package where this function is private.

> i.e.with the object { "arr": [1, 2, 3] }, we unroll arr.# to []string{"arr.0", "arr.1", "arr.2"}.

the `[1, 2, 3]` is a property on an arbitrary object - it was an example to showcase the unrolling; I have updated my description to include the word "path" to make it clearer that's what the `arr.#` is.

cmd/osv-scanner/internal/testcmd/jsonreplace.go
```diff
@@ -91,62 +91,63 @@ var (
...
+		paths = append(paths, expandArrayPaths(t, jsonInput, static)...)
 	}
 
-	return json
+	return paths
 }
 
-func buildSJSONPaths(t *testing.T, pathToBuild *[]string, path string, structure gjson.Result) {
+// replaceJSONInput takes a gjson path and replaces all elements the path matches with the output of matcher
+func replaceJSONInput(t *testing.T, jsonInput string, path string, matcher func(toReplace gjson.Result) any) string {
```
another-rex: nit: Rename matcher to replacer or something.

cmd/osv-scanner/internal/testcmd/jsonreplace.go
```diff
@@ -91,62 +91,63 @@ var (
...
+		res := gjson.Get(jsonInput, pathElem)
+
+		if !res.Exists() {
+			continue
 		}
-		for i2 := range int(structure.Int()) {
-			newPath := strings.Replace(path, "#", strconv.Itoa(i2), 1)
-			*pathToBuild = append(*pathToBuild, newPath)
+
+		json, err = sjson.SetOptions(json, pathElem, matcher(res), &sjson.Options{Optimistic: true})
```
another-rex: nit: Add a comment about Optimistic being true because we already know the path is there.

cmd/osv-scanner/internal/testcmd/jsonreplace_test.go
```diff
@@ -3,13 +3,411 @@ package testcmd
...
+	}
+	for i, tt := range tests {
+		paths := slices.AppendSeq(make([]string, 0, len(tt.outputs)), maps.Keys(tt.outputs))
+		slices.Sort(paths)
+
+		for _, path := range paths {
+			t.Run(fmt.Sprintf("%d-%s", i, path), func(t *testing.T) {
+				t.Parallel()
+
+				got := replaceJSONInput(&testing.T{}, tt.input, path, matcher)
```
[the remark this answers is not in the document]
G-Rath: fwiw I'm planning on later refactoring us to not take `t`, I think

PR #2321, opened 2025-11-05
feat: support license checks for RubyGems

G-Rath:
Per https://github.com/google/osv-scalibr/issues/1511 and https://blog.deps.dev/rubygems/ we can now do license scanning for RubyGems


G-Rath:
yup I can do that now, I just wasn't sure if you were doing that as I've had that happen a couple of times on scalibr 😅

Ly-Joey:
Thank you so much!

PR #2333, opened 2025-11-08
deduplicate sarif outputs for GitHub

### LINKED ISSUE #2331: SARIF output does not contain `partialFingerprints` property, resulting in duplicate records eg. in GitHub output
Is there a reason SARIF output does not contain the `partialFingerprints` property? I see many duplicates in [GitHub's security tab](https://github.com/brabster/terraform-bootstrap-gcp/security) which I think would be resolved by this property being populated based [on GitHub's docs](https://github.com/brabster/terraform-bootstrap-gcp/security).

I can't see any mention of this property in the repo history, so checking whether there's a reason we don't generated it and whether populating it would be a welcome feature to add.

I am using the latest version 2.2.4 but this has been an issue for at least the past few versions.

brabster:
Fixes #2331 

I did use Copilot coding agent to help me with this but I've tweaked and sense checked the output. I hope it's acceptable, let me know if there are any problems :pray:

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -287,7 +287,7 @@ Total 2 packages affected by 7 known vulnerabilities (3 Critical, 4 High, 0 Medi
       "results": [
         {
           "partialFingerprints": {
-            "primaryLocationLineHash": "3dce2968c56c311ebb0e674a0a47ebaba0469b25df88c0f2c1cc6c2c18890b0c"
+            "primaryLocationLineHash": "[line-hash]"
           },
           "attachments": [],
           "codeFlows": [],
@@ -2702,7 +2702,7 @@ Total 1 package affected by 2 known vulnerabilities (0 Critical, 2 High, 0 Mediu
       "results": [
         {
           "partialFingerprints": {
-            "primaryLocationLineHash": "b5eb7e0bae3990d8bac5a5bee032cc835ff0decf7a0f433e68e42045c6a5ae52"
+            "primaryLocationLineHash": "[line-hash]"
           },
           "attachments": [],
           "codeFlows": [],
@@ -2738,7 +2738,7 @@ Total 1 package affected by 2 known vulnerabilities (0 Critical, 2 High, 0 Mediu
         },
         {
           "partialFingerprints": {
-            "primaryLocationLineHash": "8edc9c18b3a72a7ad1a33dcda0dd7979d152bb3adafc22edecc9845c7ea69afc"
+            "primaryLocationLineHash": "[line-hash]"
           },
           "attachments": [],
           "codeFlows": [],
--- a/cmd/osv-scanner/scan/source/command_test.go
+++ b/cmd/osv-scanner/scan/source/command_test.go
@@ -165,11 +165,17 @@ func TestCommand(t *testing.T) {
 		{
 			Name: "Empty sarif output",
 			Args: []string{"", "source", "--format", "sarif", "./testdata/locks-many/composer.lock"},
+			ReplaceRules: []testutility.JSONReplaceRule{
+				testutility.ReplacePartialFingerprintHash,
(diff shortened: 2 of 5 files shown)


brabster:
Hmm not sure what to do about "Copilot" not having signed the CLA? I guess if this solution looks good I can go rewrite the history back to me

internal/output/sarif_fingerprint_test.go
```diff
@@ -0,0 +1,79 @@
+package output
+
+import (
+	"testing"
+
+	"github.com/google/osv-scanner/v2/pkg/models"
+)
+
+func TestCreateSARIFFingerprint(t *testing.T) {
```
G-Rath:
```suggestion
func Test_createSARIFFingerprint(t *testing.T) {
```

G-Rath: also I think this test could do with some rework - rather than have a table test with the same inputs, and then a second (sub?) test that tests a bunch of other packages, I think we should have a single collection of table tests, and within each test we can run the function twice to do the deterministic test

another-rex (approved):
Looks good to me, thanks for the contribution!

I think the current tests are mostly fine, as the second test is covering something completely separate from the first.  Though can you move the second test into a separate Test_ function?

another-rex:
To make the snapshot tests pass, you need to run with `UPDATE_SNAPS=true` env variable when running the tests.

brabster:
Thanks @another-rex!
- I did refactor those tests to use more of a matrix approach - I think it's easier to see what's going on and be sure that all combinations are tested. I expect it to be less brittle for you going forward too.
- I also changed the signature tests to depend on the invariance rather than the specific hash values - again, I hope that means it's easier to see what's being tested, that it's comprehensive, and fewer changes to make if edge cases emerge.
- I think the snapshots are good

I'm not sure what to do about Copilot's authorship and the CLA - if I need to rewrite the history and force push or something I can do that, let me know.

another-rex (approved):

another-rex:
Yes, you would have to rewrite copilots commit and force push here, thanks!

brabster:
@another-rex hopefully the ownership is sorted now, CLA check seems happy now

brabster:
OK I think that's snapshots updated and a couple of linter issues fixed.

brabster:
OS-specific tests are failing, I think that's because the fingerprints depend on file paths, which vary by OS (both location and OS-specific formatting). Not sure what to do for the best - is a fingerprint be the same if the path to the file is different? Does it matter that a partialFingerprint for the same package doesn't match when you compare Windows and Ubuntu, but does match when it's seen more than once in the same OS? I don't think so?

A relatively simple way forward: remove the file element, so you get a duplicate when the CVE and package match regardless of where the file is. I think that should be robust across OS and fairly confident it solves the problem I have (screenshot)
image.

Alternative - remove the comparison from the snapshot tests - probably trickier, may eliminate useful tests

Complex - try and deal with the cross-OS case - not sure this is really doable robustly, how would I compare CVE-1234-5678 occurring in `file://var/lib/dpkg/status` with some equivalent on Windows?

I'll go with the first option unless advised otherwise

brabster:
Sorry for the noise @another-rex that lints and tests OK for me. I've enabled the checks pipeline in my own repo but I get failures uploading coverage reports and I can't tell whether the tests are passing otherwise. Anyway, hopefully they are!

another-rex:
Hmm... I think a fingerprint should depend on the filepath (I believe you would want to have two entries show up if you had two files with the same package/version),

For the snapshot tests, we can just replace it with a placeholder variable so it doesn't match. I pushed up a change with this replacement (needed to move a few packages around), let's see if this works.

another-rex:
Seems to work first try! Let me know if you're happy with this. 

@G-Rath Can you take a quick look at my additional changes in case I missed anything obvious.

brabster:
Works for me @another-rex thanks!! Now I don't have to write a hacky dedupe script in my own project :D

Issue #2354, opened 2025-11-19
feat: Support the packagedeprecation plugin


another-rex:
Support the packagedeprecation plugin in our output format. 

This requires a few changes:
- [ ] Add a flag for marking deprecated packages as vulnerabilities/notices. 
- [ ] Update enricher list to include packagedeprecation enricher when flag is passed i
- [ ] Update JSON output model and the translation layer copy over the Deprecated flag.
- [ ] Finally update the output formats to display this info.

Ly-Joey:
Feature released in #2405!

PR #2370, opened 2025-11-25
feat: Add support for packagedeprecation plugin

Ly-Joey:
Adds `--experimental-flag-deprecated-packages` flag, which uses `packagedeprecation` enricher from scalibr.
The result is available in all output formats except spdx, as it doesn't include vulnerability information in the first place.

Implements: #2354

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/go-git/go-git/v5 v5.16.3
 	github.com/goccy/go-yaml v1.18.0
 	github.com/google/go-cmp v0.7.0
-	github.com/google/osv-scalibr v0.4.1-0.20251127160229-ffdf84675c08
+	github.com/google/osv-scalibr v0.4.1-0.20251201165229-410845eae22d
 	github.com/ianlancetaylor/demangle v0.0.0-20251114061303-68c556c8ce09
 	github.com/jedib0t/go-pretty/v6 v6.7.5
 	github.com/modelcontextprotocol/go-sdk v1.1.0
--- a/go.sum
+++ b/go.sum
@@ -256,8 +256,8 @@ github.com/google/go-cpy v0.0.0-20211218193943-a9c933c06932 h1:5/4TSDzpDnHQ8rKEE
 github.com/google/go-cpy v0.0.0-20211218193943-a9c933c06932/go.mod h1:cC6EdPbj/17GFCPDK39NRarlMI+kt+O60S12cNB5J9Y=
 github.com/google/jsonschema-go v0.3.0 h1:6AH2TxVNtk3IlvkkhjrtbUc4S8AvO0Xii0DxIygDg+Q=
 github.com/google/jsonschema-go v0.3.0/go.mod h1:r5quNTdLOYEz95Ru18zA0ydNbBuYoo9tgaYcxEYhJVE=
-github.com/google/osv-scalibr v0.4.1-0.20251127160229-ffdf84675c08 h1:+y59dnFH4fOLFUMid5cFJULGCYkUrSmY7cqz4XSNU5s=
-github.com/google/osv-scalibr v0.4.1-0.20251127160229-ffdf84675c08/go.mod h1:9Ze2W6nQmu1WX2s95ezOAVZhPDbcA6ZGuEHgFT/sQEU=
+github.com/google/osv-scalibr v0.4.1-0.20251201165229-410845eae22d h1:QxDG7zi37mwGaTJfWCtt1iLpSRMfy7ckc+o+5EkujMU=
+github.com/google/osv-scalibr v0.4.1-0.20251201165229-410845eae22d/go.mod h1:9Ze2W6nQmu1WX2s95ezOAVZhPDbcA6ZGuEHgFT/sQEU=
 github.com/google/pprof v0.0.0-20250403155104-27863c87afa6 h1:BHT72Gu3keYf3ZEu2J0b1vyeLSOYI8bm5wbJM/8yDe8=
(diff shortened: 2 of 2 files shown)


jess-lowe (approved):
LGTM

cuixq:
LGTM - can we have a test case to check if this package deprecation information gets propagated as expected?

cmd/osv-scanner/internal/helper/flags.go
```diff
@@ -175,6 +175,10 @@ func BuildCommonScanFlags(defaultExtractors []string) []cli.Flag {
 			Usage: "report on licenses based on an allowlist",
 			Value: &allowedLicencesFlag{},
 		},
+		&cli.BoolFlag{
+			Name:  "show-deprecated",
+			Usage: "report if package versions are deprecated",
```
cuixq, line 180: it may be helpful to state that this is done scalibr's plugin based on deps.dev API.

cmd/osv-scanner/internal/helper/flags.go
```diff
@@ -175,6 +175,10 @@ func BuildCommonScanFlags(defaultExtractors []string) []cli.Flag {
 			Usage: "report on licenses based on an allowlist",
 			Value: &allowedLicencesFlag{},
 		},
+		&cli.BoolFlag{
+			Name:  "show-deprecated",
```
cuixq: rename it to `report-package-deprecation`?

`show-deprecated` sounds to me that we are going to display the  deprecated items, also not clear what is the deprecated item.

`report-package-deprecation` is clear about that we are telling people whether a package is deprecated or not.

Ly-Joey: Thanks for the suggestion! We have finalised on the name `experimental-flag-deprecated-packages`

cmd/osv-scanner/internal/helper/flags.go
Ly-Joey: There is now a dedicated doc page for the flag calling out that the data is based on deps.dev API.

internal/output/__snapshots__/table_test.snap
```diff
@@ -273,6 +273,21 @@ Total 1 package affected by 1 known vulnerability (0 Critical, 0 High, 0 Medium,
...
 ---
 
+[TestPrintTableResults_LongTerminalWidth_WithMixedIssues/one_source_with_one_deprecated_package - 1]
+Total 0 packages affected by 0 known vulnerabilities (0 Critical, 0 High, 0 Medium, 0 Low, 0 Unknown) from 1 ecosystem.
+0 vulnerabilities can be fixed.
+
+
+Total 1 package deprecated.
+
+╭───────────┬────────────────┬─────────┬──────────────────╮
```
another-rex: Can you add a header here (with a wide cell)? Similar to how uncalled vulns has a header.

pkg/osvscanner/osvscanner.go
```diff
@@ -56,6 +57,7 @@ type ScannerActions struct {
 	CallAnalysisStates map[string]bool
 	ShowAllPackages    bool
 	ShowAllVulns       bool
+	ShowDeprecated     bool
```
another-rex: I think this should go in ExperimentalScannerActions

cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -1579,3 +1579,52 @@ func TestCommand_Filter(t *testing.T) {
 		})
 	}
 }
+
+func TestCommand_ShowDeprecated(t *testing.T) {
+	t.Parallel()
+
+	tests := []testcmd.Case{
```
another-rex, line 1586: Can you also add some tests for the standard table output here? Probably just one with vulns + yanked, and one without vulns + yanked.

cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -1579,3 +1579,52 @@ func TestCommand_Filter(t *testing.T) {
...
+				"", "source", "--format=json",
+				"--experimental-flag-deprecated-packages",
+				"./testdata/exp-plugins-pkgdeprecate/deprecated-vuln/Cargo.lock",
+			},
+			Exit: 1,
+		},
+		{
+			Name: "package_deprecated_npm",
+			Args: []string{
+				"", "source", "--format=json",
```
another-rex, line 1625: Also. please add some replace rules to keep the snapshots smaller:

https://github.com/google/osv-scanner/blob/40599a9b1a31fe3c5febe3a230adcbb90357636c/internal/testutility/jsonreplace.go#L20

Probably OnlyIDVulnsRule and GroupsAsArrayLen. See the image test you added for how to add the rules.

another-rex (commented):
Nice, looks great! Just a few minor comments

internal/output/__snapshots__/table_test.snap
Ly-Joey: Added "Deprecated packages" as table title

cmd/osv-scanner/scan/source/command_test.go
Ly-Joey: Replace rules added to all JSON test cases!

pkg/osvscanner/osvscanner.go
```diff
@@ -82,6 +83,9 @@ type ExperimentalScannerActions struct {
 	StatsCollector stats.Collector
 
 	HTTPClient *http.Client
+
+	// Report deprecated packages as findings
+	ShowDeprecated bool
```
cuixq: maybe rename it to `ReportDeprecatedPackages`?

Ly-Joey: Renamed it `FlagDeprecatedPackages` to align with the command flag.

PR #2371, opened 2025-11-25
chore: update `github.com/jedib0t/go-pretty/v6` to v6.7.5

cuixq:
Unblocks https://github.com/google/osv-scanner/pull/2365

https://github.com/jedib0t/go-pretty/releases/tag/v6.7.4 introduces minor changes in the table rendering result. Snapshots are updated to reflect the changes. They are only different in styles and this should not impact GitHub annotation output.


cuixq:
thanks for the feedback - I have updated the PR description.

PR #2380, opened 2025-12-01
tests: Pin python deps

another-rex:
Pin the python dependencies so it doesn't change and break our cassettes


G-Rath:
doesn't this defeat the point though, as we're wanting to test transient resolving?

This is why some of the other Python tests don't use cassettes

another-rex:
I don't think that's this test, this is for testing scanning artifacts in image scanning, so the transitive resolution is being done by pip, not by osv-scanner.

cuixq:
@G-Rath the test in change is for image scanning but not source scanning

G-Rath:
Sure, but it still tests our dealings with external package managers right? I'm not saying this is completely wrong, I'm just checking that has been discussed especially as there is an alternative action we could take (not use cassettes for this particular test) 🙂

G-Rath:
also I just realized the pull request title checker doesn't seem to be running - I think its not running for PRs from forks?

(`tests` is not a valid conventional commit type, it should be `test`)

another-rex:
> also I just realized the pull request title checker doesn't seem to be running - I think its not running for PRs from forks?
> 
> (`tests` is not a valid conventional commit type, it should be `test`)

Huh that is weird, I'm sure it was running before for my PRs which have always been forks.

G-Rath:
yeah and I just opened #2381 where it ran, but it doesn't look like it ran on #2234?

another-rex:
> Sure, but it still tests our dealings with external package managers right? I'm not saying this is completely wrong, I'm just checking that has been discussed especially as there is an alternative action we could take (not use cassettes for this particular test) 🙂

For  this case I think it's fine, I've been wanting to pin the dependencies here for a while, just haven't gotten around to it. The images that we are testing should not be changing over time.

PR #2382, opened 2025-12-01
test: Faster tests by being consistent with what counts as TEST_ACCEPTANCE

another-rex:
This makes the standard tests (`make test`) run under 5s by making the following changes:

- Split up TEST_ACCEPTANCE into two things, 
  - TEST_ACCEPTANCE is now purely for tests that require external dependencies (or extreme reqs like high memory)
  - go test -short is now for tests that run under 5s (which is the majority of the tests, including container scanning tests)

Moved all of the following tests to be skipped if -short is passed
- Offline database tests
- Java archive reachability tests (does dep resolution)
- Transitive dependency tests 

- Moved all docker dependency tests under TEST_ACCEPTANCE=true
- Some tests have both TEST_ACCEPTANCE=true and SkipIfShort, this allows for short tests while TEST_ACCEPTANCE=true
  - Example is OCI container image tests, which require docker to build, but runs very fast.

This also includes some additional minor changes to aid with debugging and testing.

- Print out the stdout and stderr if the error code is not expected. Right now, especially for the tests with JSON replace rules, it just fails without any detail.
- Also for JSON replace tests, print out the stdout and stderr messages when json fail to parse.
- Make sure to sort keys when writing cassettes to disk
- Made container scanning tests with custom plugin flags actually test stuff, and run much faster.

### DIFF
--- a/cmd/osv-scanner/fix/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/fix/__snapshots__/command_test.snap
@@ -11282,7 +11282,7 @@ manifest or lockfile is required
 
 ---
 
-[TestCommandOffline/fix_non_interactive_in_place_package_lock_json_with_offline_vulns - 1]
+[TestCommand_OfflineDatabase/fix_non_interactive_in_place_package_lock_json_with_offline_vulns - 1]
 Loaded npm local db from <tempdir>/osv-scanner/npm/all.zip
 Scanning <tempdir>/package-lock.json...
 Found 9 vulnerabilities matching the filter
@@ -11297,11 +11297,11 @@ Rewriting <tempdir>/package-lock.json...
 
 ---
 
-[TestCommandOffline/fix_non_interactive_in_place_package_lock_json_with_offline_vulns - 2]
+[TestCommand_OfflineDatabase/fix_non_interactive_in_place_package_lock_json_with_offline_vulns - 2]
 
 ---
 
-[TestCommandOffline/fix_non_interactive_in_place_package_lock_json_with_offline_vulns - 3]
+[TestCommand_OfflineDatabase/fix_non_interactive_in_place_package_lock_json_with_offline_vulns - 3]
 {
   "name": "osv-fix",
   "version": "1.0.0",
@@ -13027,7 +13027,7 @@ Rewriting <tempdir>/package-lock.json...
 
 ---
 
-[TestCommandOffline/fix_non_interactive_relax_package_json_with_offline_vulns - 1]
+[TestCommand_OfflineDatabase/fix_non_interactive_relax_package_json_with_offline_vulns - 1]
 Loaded npm local db from <tempdir>/osv-scanner/npm/all.zip
 Resolving <tempdir>/package.json...
 Found 6 vulnerabilities matching the filter
@@ -13040,11 +13040,11 @@ Rewriting <tempdir>/package.json...
 
 ---
 
-[TestCommandOffline/fix_non_interactive_relax_package_json_with_offline_vulns - 2]
+[TestCommand_OfflineDatabase/fix_non_interactive_relax_package_json_with_offline_vulns - 2]
 
 ---
 
-[TestCommandOffline/fix_non_interactive_relax_package_json_with_offline_vulns - 3]
+[TestCommand_OfflineDatabase/fix_non_interactive_relax_package_json_with_offline_vulns - 3]
 {
   "name": "osv-fix",
   "version": "1.0.0",


G-Rath:
So the idea with `TEST_ACCEPTANCE` was about tests that had exotic dependencies such as Docker or the Rust build chain, not (initially) about speed, and imo the test suite should aim to run as many of our tests as possible by default.

I think rather than change that, we should make proper use of [`testing.Short()`](https://pkg.go.dev/testing#Short), probably with a helper like `testutility.IsSlow`, so that you can then run the tests fast with `go test -short ./...`

cuixq:
My understanding is that `TEST_ACCEPTANCE` is for tests that require additional dependencies but not for tests that are slow to run. Shall we have a separate env variable for the slow tests?

another-rex:
Huh did not know about the `-short` flag, will look into that.  

This change actually makes TEST_ACCEPTANCE actually cut out tests that require external dependency (there's a few docker tests that were not behind the flag), 

I'll do another pass to make sure only tests that require external deps are behind the flag.

G-Rath:
>  (there's a few docker tests that were not behind the flag),

despite what I've said, I actually don't think we should consider Docker as exoitc since I think thats a very common dependency and it would mean _all_ of our image scanning tests get disabled by default.

I think if we use `-short`, we get the nice layers that we want:

* `TEST_ACCEPTANCE=false go test -short ./...` for running the suite _fast_ (aka without _any_ external dependencies, including Docker)
* `TEST_ACCEPTANCE=false go test ./...` for running the suite with some external dependencies that we consider common (e.g. Docker and HTTP)
* `TEST_ACCEPTANCE=true go test ./...` for running _everything_ including tests that require specific toolchains like Rust

another-rex:
There's a misunderstanding here, we have two sets of docker tests, one for with the --archive flag, and one without. The one without `--archive` actually just shells out for `docker save blahblah > temp.tar` and then essentially runs `osv-scanner scan image --archive temp.tar`. 

So leaving them out of test acceptance is no issue, since it's really just testing some very thin convenience code.

In this PR I actually enabled image scanning with `--archive` when TEST_ACCEPTANCE=false, since it runs really fast with vcrs now.

another-rex:
Ok, made a few more changes:

Split up TEST_ACCEPTANCE into two things,

TEST_ACCEPTANCE is now purely for tests that require external dependencies (or extreme reqs like high memory)

go test -short is now for tests that run under 5s (which is the majority of the tests, including container scanning tests)
Made container scanning tests with custom plugin flags actually test stuff, and run much faster.


PTAL @cuixq @G-Rath

cmd/osv-scanner/fix/command_test.go
```diff
@@ -139,6 +131,45 @@ func TestCommand(t *testing.T) {
 	}
 }
 
+func TestCommandOffline(t *testing.T) {
```
cuixq: can you rename it to explicitly indicate this is to test the fix subcommand with offline vulnerability database? "offline" may be confused with "offline scanning".

cmd/osv-scanner/internal/testcmd/vcr.go
```diff
@@ -102,12 +102,16 @@ func InsertCassette(t *testing.T) *http.Client {
 
 			delete(i.Request.Headers, "User-Agent")
 
-			i.Request.Body = string(pretty.Pretty([]byte(i.Request.Body)))
+			// Force copy of default options
```
cuixq: if the options are default, why need to force copy?

cmd/osv-scanner/scan/image/command_test.go
```diff
@@ -13,51 +13,61 @@ import (
...
 			},
-			Exit: 127,
+			Exit: 1,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.Name, func(t *testing.T) {
 			t.Parallel()
 
-			// Only test on linux, and mac/windows CI/CD does not come with docker preinstalled
```
cuixq, line 56: this is still true right? we should still skip docker-based tests if not Linux

cmd/osv-scanner/internal/testcmd/vcr.go
another-rex: I don't want to change the default options, as that's the default of `pretty` package. I want to edit the default and use the edited version. I'll update the comment.

cmd/osv-scanner/scan/image/command_test.go
another-rex: Nope, it was true because it was not using the --archive option, we are now using tars, which are built on linux and copied over to macos and windows CI

cmd/osv-scanner/fix/command_test.go
another-rex: hmm I think that's already part of the way we are writing tests. 

All our tests are TestCommand, located in each subcommand. We don't mention the command we are testing, because it's already encoded as part of the test path.

When failing it looks like this:

```
=== Failed
=== FAIL: cmd/osv-scanner/fix TestCommandOffline (0.00s)
```

Or with go test ./...

```
--- FAIL: TestCommandOffline (0.00s)
FAIL
...
FAIL    github.com/google/osv-scanner/v2/cmd/osv-scanner/fix    10.912s
```

It could be useful to rename these to make grepping lines easier, but that should be done in a separate PR.

cuixq: my first impression of `TestCommandOffline` is that we are testing running the command offline without network access but actually we are just running it with offline database.

something like `TestCommand_OfflineDatabase` will help I think.

cmd/osv-scanner/internal/testcmd/vcr.go
cuixq: Is the comment updated?

cmd/osv-scanner/fix/command_test.go
another-rex: Ah I see what you mean

cmd/osv-scanner/internal/testcmd/vcr.go
another-rex: It is now :)

PR #2393, opened 2025-12-04
test: format cassette bodies as pretty json

G-Rath:
This should make it a lot easier to review changes to the cassettes

### DIFF
--- a/cmd/osv-scanner/scan/image/testdata/cassettes/TestCommand_Docker.yaml
+++ b/cmd/osv-scanner/scan/image/testdata/cassettes/TestCommand_Docker.yaml
@@ -6,9 +6,118 @@ interactions:
       proto: HTTP/1.1
       proto_major: 1
       proto_minor: 1
-      content_length: 1232
+      content_length: 2036
       host: api.osv.dev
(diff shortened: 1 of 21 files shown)


cmd/osv-scanner/internal/testcmd/vcr.go
```diff
@@ -208,13 +203,10 @@ func matchBody(r *http.Request, i cassette.Request) bool {
...
-		body2, err := normalizeBody(buffer.Bytes())
-		if err != nil {
-			// This file will only run in tests, and this should be very unlikely
-			panic(err)
-		}
-
-		if !bytes.Equal(body2, []byte(i.Body)) {
+		if !bytes.Equal(
+			pretty.PrettyOptions(buffer.Bytes(), &pretty.Options{SortKeys: true}),
+			pretty.PrettyOptions([]byte(i.Body), &pretty.Options{SortKeys: true}),
```
another-rex, line 208: I think this should actually be a cmp compare of  the parsed output, rather than formatting both JSONs and doing a byte compare.

G-Rath: could be, but what do you think of landing this as-is? just cause of its size and I've got a few other tweaks I'm planning to make which I do at the same time as moving to `cmp` (namely, I'm pretty sure we don't need to be touching some of the headers)

another-rex: Sure

another-rex (approved):

cmd/osv-scanner/internal/testcmd/vcr.go
G-Rath: so actually, I'm not sure if that is possible because parsing the output requires us to know the structure right? and currently this is generic - it so happens we're using it only for the `bulkquery` endpoint, but its implemented with the idea that it can be used for anything...

another-rex: Pretty sure you can just cmp `interface{}` or `any`. go JSON just parses it into native go types, and puts it into `any`, which cmp then uses reflection to match it.

PR #2394, opened 2025-12-04
test: update license snapshot

G-Rath:

PR #2395, opened 2025-12-05
fix: License scanning to match new deps.dev names

another-rex:
They are also looking at implementing a translation layer to keep supporting the old version format (eta sometime next week) to make sure existing osv-scanner queries are not broken. 

Closes #2394

### DIFF
--- a/internal/clients/clientimpl/licensematcher/licensematcher.go
+++ b/internal/clients/clientimpl/licensematcher/licensematcher.go
@@ -97,6 +97,8 @@ func (matcher *DepsDevLicenseMatcher) makeVersionRequest(ctx context.Context, qu
 
 func versionQuery(system depsdevpb.System, name string, version string) *depsdevpb.GetVersionRequest {
 	if system == depsdevpb.System_GO {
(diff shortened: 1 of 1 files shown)


G-Rath (approved):

internal/clients/clientimpl/licensematcher/licensematcher.go
```diff
@@ -96,8 +96,12 @@ func (matcher *DepsDevLicenseMatcher) makeVersionRequest(ctx context.Context, qu
 }
 
 func versionQuery(system depsdevpb.System, name string, version string) *depsdevpb.GetVersionRequest {
-	if system == depsdevpb.System_GO && name != "stdlib" {
-		version = "v" + version
+	if system == depsdevpb.System_GO {
```
cuixq, line 99: can you add a comment explaining this special behaviour? also probably a TODO once deps.dev implements the translation layer.

another-rex: Yep I'll add a comment, this is the desired behavior, the translation layer is just for backwards compatibility

PR #2408, opened 2025-12-09
test: have vulnerable Drupal contrib module

G-Rath:
For some reason when I added this, I used a version that wasn't vulnerable defeating the point of having this in the first place 🤦

(right now there's no vulnerability, but that'll change once the Drupal advisory database is being ingested into production)


cmd/osv-scanner/scan/source/testdata/cassettes/TestCommand.yaml
```diff
@@ -1624,7 +1916,7 @@ interactions:
       status: 200 OK
       code: 200
       duration: 0s
-  - id: 15
+  - id: 16
```
another-rex: hmm this incremental id number is a bit annoying to have in diffs. With the custom marshaller and unmarshaller we can probably remove it with some effort. Something for a separate PR.

cmd/osv-scanner/scan/source/testdata/cassettes/TestCommand.yaml
```diff
@@ -1580,6 +1580,298 @@ interactions:
       code: 200
       duration: 0s
   - id: 14
+    request:
```
another-rex: Can you run this with TEST_VCR_MODE=0? I don't think this should be adding a new cassette, but replacing an existing one?

G-Rath: yup that's resulted in a much nicer diff

cmd/osv-scanner/scan/source/testdata/cassettes/TestCommand.yaml
G-Rath: pretty sure we can't because the actual marshaling is handled by `go-vcr` which is called on the cassette struct so even if we somehow "deleted" it, it'd end up with its default zero value

another-rex: There's a custom yaml marshalling function: https://github.com/dnaeon/go-vcr?tab=readme-ov-file#custom-yaml-marshaling-function

Though I'm not sure if there's a corresponding unmarshalling func.

PR #2409, opened 2025-12-10
fix(rust): disable unsafe cargo build by default to prevent ACE

GHOryy5:
This PR mitigates a Critical Arbitrary Code Execution (ACE) vulnerability in the Rust source analysis pipeline.

**The Vulnerability**
Currently, `rustBuildSource` executes `cargo build` on the target repository. In the Rust ecosystem, `cargo build` executes `build.rs` scripts and procedural macros. If `osv-scanner` is run on an untrusted malicious repository, that repository can execute arbitrary code on the scanner's host.

### DIFF
--- a/internal/sourceanalysis/rust.go
+++ b/internal/sourceanalysis/rust.go
@@ -267,6 +267,9 @@ func rustBuildSource(source models.SourceInfo) ([]string, error) {
 
 	resultBinaryPaths := []string{}
 	for _, de := range entries {
+		// We only want .d files, which is generated for each output binary from cargo
+		// These files contains a string to the full path of output binary/library file.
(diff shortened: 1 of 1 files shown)


another-rex:
Thanks for the contribution!

Rust call analysis is not enabled by default, and has to be manually enabled, so I'm not sure adding an extra env makes sense.

It is also made clear that it's building the project to do call analysis, so it's up to the user to not run malicious build scripts. 
https://google.github.io/osv-scanner/usage/scan-source#call-analysis-in-rust

Perhaps a change to the --help message would be more appropriate here?

E.g. update `--call-analysis` flag help text to describe how it should be called (with the language, the list of available languages, and perhaps a warning for languages that have automatic build scripts.)

PR #2413, opened 2025-12-12
feat: migrate guided remediation to osv-scalibr implementation

michaelkedar:
Finally unifying the guided remediation. It's mostly a 1:1 mapping of config/functions.

I believe this will incidentally mean `osv-scanner fix` will now support python requirements.txt files, since that's been implemented in scalibr.

This is actually a bit slower now. One reason was that I am no longer caching the vulnerability responses, but even when I experimented with hackily doing that, it still was slower, so I haven't bothered re-implementing the vulnerability cache.


cuixq:
@michaelkedar considering offline vuln enricher is done in https://github.com/google/osv-scalibr/pull/1608, is this unblocked now?

cmd/osv-scanner/fix/command.go
```diff
@@ -225,160 +207,131 @@ func Command(stdout, stderr io.Writer, _ *http.Client) *cli.Command {
...
-			Client: osvdev.OSVClient{
-				HTTPClient:  http.DefaultClient,
-				Config:      config,
-				BaseHostURL: osvdev.DefaultBaseURL,
-			},
-			InitialQueryTimeout: 5 * time.Minute,
-		}
+		osvdevCl := osvdev.DefaultClient()
+		osvdevCl.Config.UserAgent = userAgent
+		opts.VulnEnricher = scalibrosvdev.NewWithClient(osvdevCl, 5*time.Minute)
```
michaelkedar, line 299: what's the preferred way to set up these clients?

cmd/osv-scanner/fix/command.go
```diff
@@ -13,56 +13,38 @@ import (
...
-type strategy string
-
-const (
-	strategyInPlace  strategy = "in-place"
-	strategyRelax    strategy = "relax"
-	strategyOverride strategy = "override"
 )
 
-var strategies = []string{string(strategyInPlace), string(strategyRelax), string(strategyOverride)}
+var strategies = []string{string(strategy.StrategyInPlace), string(strategy.StrategyRelax), string(strategy.StrategyOverride)}
```
cuixq, line 39: we may want to rename these to `strategy.Relax` which is more recommended in Go I guess but this is not the scope of this PR.

cmd/osv-scanner/fix/command.go
another-rex: Ideally each enricher will setup their own client. I sent you the design doc for how I envision this to work going forward.

cuixq: we probably should migrate the enricher to use the global configuration like  the offline vuln matcher so we only need to call `osvdev.New()` but this may be out of scope of this PR neither.

cuixq (approved):
LGTM - just two comments about potential refactoring.

Issue #2411, opened 2025-12-11
OSV-Scanner 2.x is not able to scan Pub packages used in flutter application from SBOM


JochemLindemannSchunk:
### Description
After upgrading from OSV Scanner version 1.9.2 to 2.3.1, SBOMs created by syft for flutter applications can not be scanned anymore (Pub ecosystem).

### Steps to reproduce
Scan SBOM containing Pub packages with OSV Scanner 2.3.1

`osv-scanner_windows_amd64_2_3_1.exe scan source -L minimalSbom.cdx.xml`
`Starting filesystem walk for root: C:\
Scanned C:\osv-scanner\minimalSbom.cdx.xml file and found 1 package
End status: 0 dirs visited, 1 inodes visited, 1 Extract calls, 560.8µs elapsed, 560.8µs wall time
Filtered 1 local/unscannable package/s from the scan.
No issues found`

Scanner reports one unscannable package, no scan result.

### Expected behaviour 
Behaviour similar to OSV Scanner 1.9.2. Pub packages from SBOM should get scanned and results should be reported.

`osv-scanner_windows_amd64_1_9_2.exe --sbom=minimalSbom.cdx.xml`
`Scanned C:\osv-scanner\minimalSbom.cdx.xml as CycloneDX SBOM and found 1 package
╭─────────────────────────────────────┬──────┬───────────┬────────────────────────────┬─────────┬─────────────────────╮
│ OSV URL                             │ CVSS │ ECOSYSTEM │ PACKAGE                    │ VERSION │ SOURCE              │
├─────────────────────────────────────┼──────┼───────────┼────────────────────────────┼─────────┼─────────────────────┤
│ https://osv.dev/GHSA-3hpf-ff72-j67p │ 3.0  │ pub:      │ shared_preferences_android │ 2.3.3   │ minimalSbom.cdx.xml │
╰─────────────────────────────────────┴──────┴───────────┴────────────────────────────┴─────────┴─────────────────────╯`

OSV Scanner 1.9.2 is able to identify package and reports result from identical SBOM.

### Files

[minimalSbom.cdx.xml](https://github.com/user-attachments/files/24099094/minimalSbom.cdx.xml)

`<?xml version="1.0" encoding="UTF-8"?>
<bom xmlns="http://cyclonedx.org/schema/bom/1.6" serialNumber="urn:uuid:d6a734a4-605d-4924-ad01-811604f6538c" version="1">
  <metadata>
[...]

PR #2415, opened 2025-12-12
feat(purl): Add more PURL-to-ecosystem mappings

cuixq:
Fixes https://github.com/google/osv-scanner/issues/2411

This PR adds more mappings for the following PURL types to their corresponding OSV ecosystems:
 - composer
 - conan
 - cran
 - github
 - hackage
 - k8s
 - swift

The `purlEcosystems` map also has been sorted alphabetically by PURL type.

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -461,21 +461,26 @@ overriding license for package Packagist/drupal/tfa/2.0.0-alpha4 with 0BSD
 overriding license for package Packagist/league/flysystem/1.0.8 with 0BSD
 overriding license for package Packagist/sentry/sdk/2.0.4 with 0BSD
 overriding license for package Packagist/theseer/tokenizer/1.1.3 with 0BSD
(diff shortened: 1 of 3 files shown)


internal/utility/purl/purl_to_package.go
```diff
@@ -9,21 +9,28 @@ import (
...
-	"golang":   {"*": osvconstants.EcosystemGo},
-	"maven":    {"*": osvconstants.EcosystemMaven},
-	"nuget":    {"*": osvconstants.EcosystemNuGet},
-	"npm":      {"*": osvconstants.EcosystemNPM},
-	"composer": {"*": osvconstants.EcosystemPackagist},
-	"generic":  {"*": osvconstants.EcosystemOSSFuzz},
-	"pypi":     {"*": osvconstants.EcosystemPyPI},
-	"gem":      {"*": osvconstants.EcosystemRubyGems},
+	"gem":     {"*": osvconstants.EcosystemRubyGems},
+	"generic": {"*": osvconstants.EcosystemOSSFuzz},
```
another-rex: Use the GIT ecosystem here is probably better.

cuixq: we don't yet have a GIT ecosystem - I have commented this out and we can add it back once we have the ecosystem added.

PR #2429, opened 2025-12-22
test: update testcases

Ly-Joey:
Update testcases, related snapshots and cassettes

Test cases changed:
- `requirements.txt can have all kinds of names`: Added "--no-resolve" flag to skip python transitive dependency resolution, as it is already covered in another test case
- `Empty Ubuntu 20.04 image tar with no vulns shown`: Added new package to ignore to config.toml

### DIFF
--- a/cmd/osv-scanner/scan/image/command_test.go
+++ b/cmd/osv-scanner/scan/image/command_test.go
@@ -249,7 +249,8 @@ func TestCommand_OCIImage(t *testing.T) {
 		},
 		{
 			// This tests that unimportant vulns are hidden properly
-			// If the test is failing (reporting new important vulns), add it as ignore=true to the config.toml
+			// If the test is failing (reporting new important vulns), add the package that introduced the vuln as ignore=true to the config.toml
+			// The package with unimportant vulns is pcre3, so if a new vulnerability appears for that package, don't ignore the entire package, just ignore the important vulnerability specifically.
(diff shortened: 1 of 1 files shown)


cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -325,7 +325,7 @@ func TestCommand(t *testing.T) {
 		// a bunch of requirements.txt files with different names
```
another-rex, line 325:
```suggestion
		// a bunch of requirements.txt files with different names
		// --no-resolve is used as transitive resolution tests are in a separate section
```

cmd/osv-scanner/scan/image/command_test.go
```diff
@@ -248,6 +248,8 @@ func TestCommand_OCIImage(t *testing.T) {
 			Exit: 1,
 		},
 		{
+			// This tests that unimportant vulns are hidden properly
+			// If the test is failing (reporting new important vulns), add it as ignore=true to the config.toml
```
another-rex:
```suggestion
			// If the test is failing (reporting new important vulns), add the package that introduced the vuln as ignore=true to the config.toml
			// The package with unimportant vulns is pcre3, so if a new vulnerability appears for that package, don't ignore the entire package, just ignore the important vulnerability specifically.
```

Issue #2420, opened 2025-12-17
allow library users to specify a user agent


spencerschrock:
osv-scanner v1 allowed library users to specify a user agent via the `RequestUserAgent` variable.
https://pkg.go.dev/github.com/google/osv-scanner@v1.9.2/pkg/osv#RequestUserAgent

`RequestUserAgent` was deleted in #1470 and the user agent is now hardcoded as `osv-scanner_scan/<version>` 
https://github.com/google/osv-scanner/blob/0dd3348e0daeabc68a44527fa7791a4c26b9baa6/pkg/osvscanner/osvscanner.go#L145-L148

PR #2433, opened 2025-12-29
fix: allow specifying user agent in v2 osvscanner package

Ly-Joey:
Users can specify a user agent with `ExperimentalScannerActions.RequestUserAgent`. 

Resolves: #2420

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -18,6 +18,7 @@ Use OSV-Scanner to find existing vulnerabilities affecting your project's depend
 OSV-Scanner provides an officially supported frontend to the [OSV database](https://osv.dev/) that connects a project’s list of dependencies with the vulnerabilities that affect them.
 
 You can use OSV-scanner in two ways:
+
 1. **As a CLI tool:** Run directly in a terminal or CI/CD pipeline to scan projects. See the [Installation Guide](./installation.md) to get started.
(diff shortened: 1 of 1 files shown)


pkg/osvscanner/osvscanner.go
```diff
@@ -121,20 +121,28 @@ var ErrVulnerabilitiesFound = errors.New("vulnerabilities found")
 // TODO(v2): Actually use this error
 var ErrAPIFailed = errors.New("API query failed")
 
+// RequestUserAgent allows users to specify a user agent
+var RequestUserAgent = ""
```
another-rex: Instead of a global variable, can this be a experimental osv-scanner actions field?

another-rex: That way we can default to osv-scanner-api, and override them in the actual binary files: `cmd/osv-scanner/...`

Ly-Joey: Done! Default user agent for API users is now `osv-scanner-api` and the user agent from the scan command is `osv-scanner_scan/<version>`. 
Do we want more granular tracking (source vs image)?

another-rex: Sure, might as well. Our logs split on _, so maybe something like `..._scan-image` and `..._scan-source`?

another-rex (approved):
LGTM!

PR #2434, opened 2025-12-31
chore: Create a better makefile to simplify test running

another-rex:
~Create a justfile to simplify all the various commands we have.~ 

Turns out we really don't need a justfile, just make the makefile better.

Added:
- Help menu (default)
- test with short commands so that it's easy to refer to and invoke.


Ly-Joey (approved):
Awesome! :D

PR #2437, opened 2026-01-02
docs: update OSV scanner reusable workflow references to v2.3.1

catatsuy:
This PR updates docs/github-action.md to use OSV-Scanner Action v2.3.1 everywhere.
The document previously mixed old and new versions; now it is consistent.


catatsuy:
Thank you!

Issue #2217, opened 2025-09-04
osv-scanner in offline mode has massive memory leak


kodem-eran:
When scanning vulnerabilities given an SBOM (SPDX) in an offline mode using osv-scanner the memory usage increases drastically.
It reaches more than 15GB RAM (and then the server kills it).
In order for the bug to be drastic, it requires relatively big SBOMs, I used 10MB SBOM (see attached) 

## Command:
`$ OSV_SCANNER_LOCAL_DB_CACHE_DIRECTORY="./osv-data" osv-scanner scan -L ./output/gitlab/sbom.spdx.json --offline`

## Tested versions: 
* osv-scanner@v2.0.0 - osv-scanner@latest (v2.2.2)
This behavior does not occur in online mode or in older versions, such as v1.9.2. 
* Go version: 1.24.6 (amd64)
* Debian Linux version:5.10

## stdout:
```
Neither CPE nor PURL found for package: &{IsUnpackaged:false PackageName:main PackageSPDXIdentifier:Package-main-95578677-9f52-4db1-b76e-88241bbcf5b2 PackageVersion:0 PackageFileName: PackageSupplier:0xc00011e300 PackageOriginator:<nil> PackageDownloadLocation:NOASSERTION FilesAnalyzed:false IsFilesAnalyzedTagPresent:true PackageVerificationCode:<nil> PackageChecksums:[] PackageHomePage: PackageSourceInfo: PackageLicenseConcluded: PackageLicenseInfoFromFiles:[] PackageLicenseDeclared: PackageLicenseComments: PackageCopyrightText: PackageSummary: PackageDescription: PackageComment: PackageExternalReferences:[] PackageAttributionTexts:[] PrimaryPackagePurpose: ReleaseDate: BuiltDate: ValidUntilDate: Files:[] Annotations:[] hasFiles:[]}
Scanned ./output/gitlab/sbom.spdx.json file and found 5521 packages
Loaded npm local db from osv-data/osv-scanner/npm/all.zip
Loaded RubyGems local db from osv-data/osv-scanner/RubyGems/all.zip
Loaded Go local db from osv-data/osv-scanner/Go/all.zip
Loaded Ubuntu local db from osv-data/osv-scanner/Ubuntu/all.zip
Loaded crates.io local db from osv-data/osv-scanner/crates.io/all.zip
<the memory limit kills it, after using more than 10GB>
```

## Stack trace: 
```
[...]

PR #2445, opened 2026-01-07
perf(local): don't read the entire zip into memory

G-Rath:
I'm a little suspect of the results I've produced, but it seems like this reduces memory usage by a decent amount e.g. with the Ubuntu SBOM used for #2217 we go from 1gb to 200mb total.

Ultimately, I think the theory is sound and that are not really any downsides so we might as well change this.

Note that this does not have any impact on scanning speed, just memory usage.

### DIFF
--- a/internal/clients/clientimpl/localmatcher/zip.go
+++ b/internal/clients/clientimpl/localmatcher/zip.go
@@ -77,56 +77,49 @@ func fetchRemoteArchiveCRC32CHash(ctx context.Context, url string) (uint32, erro
 	return 0, errors.New("could not find crc32c= checksum")
 }
 
-func fetchLocalArchiveCRC32CHash(f *os.File) (uint32, int64, error) {
+func fetchLocalArchiveCRC32CHash(f *os.File) (uint32, error) {
 	h := crc32.New(crc32.MakeTable(crc32.Castagnoli))
-	size, err := io.Copy(h, f)
 
-	if err != nil {
-		return 0, 0, err
+	if _, err := io.Copy(h, f); err != nil {
+		return 0, err
 	}
 
-	return h.Sum32(), size, nil
+	return h.Sum32(), nil
 }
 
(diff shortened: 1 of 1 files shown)


internal/clients/clientimpl/localmatcher/zip.go
```diff
@@ -211,13 +230,15 @@ func (db *ZipDB) loadZipFile(zipFile *zip.File, names []string) {
 func (db *ZipDB) load(ctx context.Context, names []string) error {
 	db.Vulnerabilities = []*osvschema.Vulnerability{}
 
-	body, err := db.fetchZip(ctx)
+	f, size, err := db.fetchZip(ctx)
```
another-rex: Since f has size inside, I don't think we need to return size as a separate return variable, it seems to be adding quite a bit of code.

G-Rath: `f` doesn't have the size inside though, unless I've missed something? the only way I found to get the size at this point was by calling `f.Stat`, which can return an error: https://github.com/google/osv-scanner/pull/2445/commits/f59188eda81a216503259aba725956acf6bb8297

another-rex: Yep that's what I meant, just make the Stat call out here, and return error if it errors (which should be unlikely, and if it does we probably will have problems reading the file anyway.)

G-Rath: right so you're voting for reverting/dropping f59188eda81a216503259aba725956acf6bb8297 - I'm not super fussed, I just don't think it actually saves that much code especially since this is all internal 🤷

another-rex: Yep, I think it's just cleaner to read without having to return size, and less likely size will be out of sync with the file itself.

another-rex (approved):

PR #2449, opened 2026-01-08
feat(local): support including and excluding malware advisories

G-Rath:
This adds an experimental flag for controlling how malware advisories (`MAL-`) are loaded as these advisories are very easy to create and are very common in some ecosystems e.g. of the 214057 advisories in the NPM database, only 4410 are actual vulnerabilities - the remaining 209647 are MAL.

Skipping these advisories entirely can greatly reduce the time taken and memory used when doing offline scans


G-Rath:
I'm not sure this is that useful now that #2450 has been landed as the cost of our automatic method of doing this has been decreased a great amount...

G-Rath:
yeah this hopefully isn't needed anymore

PR #2452, opened 2026-01-08
feat: remove deprecated `sbom` flag

### LINKED ISSUE #1960: Remove dedicated SBOM flag in favor of `-L` and friends

G-Rath:
This flag is so 2025

Resolves #1960


Wbaker7702:
I don't care I can't fix it, it won't let me

PR #2458, opened 2026-01-10
feat: add --exclude flag to skip paths during scanning

### LINKED ISSUE #2324: Option to skip or ignore files and folder while scanning for Vulnerabilities
Ignore certain files and folders while scanning for vulnerabilities from a path or location of the project such as Test files and Documentations.

Enhancement looking forward: Option to skip or ignore one or more files and folder from the project while scanning for vulnerabilities.

Reference URL: https://github.com/Kong/kong

When we try to scan the above mentioned repository most of the vulnerabilities were detected in the documents and test files. due to which the repository failed.

Ankitsinghsisodya:
## Overview

**Issue:** https://github.com/google/osv-scanner/issues/2324

This PR adds a `--exclude` CLI flag that allows users to specify glob patterns to exclude files and directories from vulnerability scanning.

Fixes #2324

### Problem
Users scanning large repositories often want to exclude test files, documentation, and other non-production code from vulnerability scans. While `.gitignore` is respected by default and [osv-scanner.toml](cci:7://file:///Users/ankitsinghsisodya/osv-scanner/internal/remediation/testdata/santatracker/osv-scanner.toml:0:0-0:0) can be used, there was no quick CLI option to exclude paths.

### Solution
Added `--exclude` / `-e` flag that accepts glob patterns:

```bash
osv-scanner scan source -r --exclude="**/test/**" --exclude="**/docs/**" ./path

### CODE REVIEW

**Review by another-rex — COMMENTED** [review]
This is definitely something we want, though might need a bit more design before including it. It is really hard for us to justify removing a flag once it's added.

I want a way for people to specify regex patterns without having to add another separate flag. Maybe something like if it is wrapped `/<regex>/` like in javascript?

File: internal/cachedregexp/regex.go [hunk 1]
```diff
@@ -1,18 +1,39 @@
...
+// Package cachedregexp provides a cached version of regexp.MustCompile and regexp.Compile.
 package cachedregexp
 
 import (
 	"regexp"
 	"sync"
 )
 
+// Regexp is an alias for regexp.Regexp so other packages don't need to import regexp directly.
+type Regexp = regexp.Regexp
```
G-Rath: nit: this is a neat trick, but I think it'd be best to stick with just importing `regexp` directly since it's what we're already doing throughout the codebase

G-Rath (changes requested):
Appreciate your work on this! 

Could you please:
  * move the flag to be `--experimental-` (this is mainly because right now scalibr only supports skipping directories, not files which'll hopefully be able to change but in the meantime this will mean we can do breaking changes if we go another direction instead)
  * update the docs to reflect that this only applies to directories, not files
  * add a couple of [cmd tests](https://github.com/google/osv-scanner/blob/main/cmd/osv-scanner/scan/source/command_test.go)
  * update the syntax to use `:`, matching our [`--lockfile` flag](https://google.github.io/osv-scanner/usage/scan-source#specify-lockfiles), which'll also let us use `DirsToSkip` too
      * if the string is prefixed with `:` or nothing, it should be applied as a directory (`DirsToSkip`)
      * if the string is prefixed with `g:`, it should be applied as a glob (`SkipDirGlob`)
      * if the string is prefixed with `r:`, it should be applied as a regex (`SkipDirRegex`)

another-rex:
> if the string is prefixed with :, it should be applied as a directory (DirsToSkip)

to clarify, also if it is not prefixed, use DirsToSkip, the prefixing with `:` is just an escape hatch incase your file path itself has a `:` inside.

Ankitsinghsisodya:
@another-rex sir, I have done the changes suggested you.

internal/cachedregexp/regex.go
```diff
@@ -8,6 +8,9 @@ import (
 
 var cache sync.Map
 
+// Regexp is an alias for regexp.Regexp, allowing packages to use this type without importing regexp directly.
+type Regexp = regexp.Regexp
```
G-Rath: nit: this is a neat trick, but I think it'd be best to stick with just importing regexp directly since it's what we're already doing throughout the codebase

pkg/osvscanner/osvscanner.go
```diff
@@ -49,6 +49,7 @@ type ScannerActions struct {
 	Recursive          bool
 	IncludeGitRoot     bool
 	NoIgnore           bool
+	SkipDirPatterns    []string
```
G-Rath: this should go in our `ExperimentalScannerActions` struct - I think we should also stick with the "exclude" terminology because ideally we do want to make it be for both files and directories, and I think most people will be using it for directories anyway

pkg/osvscanner/scan.go
```diff
@@ -250,15 +263,25 @@ SBOMLoop:
 			capabilities.Network = plugin.NetworkOffline
 		}
 
+		// Prepare skip patterns for this scan
+		var dirsToSkip []string
+		var skipDirGlob glob.Glob
+		var skipDirRegex *cachedregexp.Regexp
+		if skipDirPatterns != nil {
```
G-Rath: I think (hope?) it should be possible to have `skipDirPatterns` initialized in a way that means we can just pass these directly through to the scan config, rather than needing these extra variables

pkg/osvscanner/scan.go
```diff
@@ -237,6 +240,16 @@ SBOMLoop:
 	testlogger.BeginDirScanMarker()
 	osCapability := determineOS()
 
+	// Parse skip directory patterns (supports exact names, glob, and regex)
+	var skipDirPatterns *SkipDirPatterns
+	if len(actions.SkipDirPatterns) > 0 {
```
G-Rath: since we're looping over a slice, it should be safe to always call `ParseSkipDirPatterns`, rather then need to check the length

pkg/osvscanner/exclude_test.go
G-Rath, line 1: can you please only use only alphanumeric characters and underscores in test descriptions (i.e. no spaces)? that way the test name matches what Go will use, making them easier to run and debug.

(this is a convention we've started doing for new tests, which is why most of our existing suite isn't like this yet)

pkg/osvscanner/exclude.go
G-Rath, line 1: since this is in our public package, can you make sure everything is private?

cmd/osv-scanner/mcp/integration_test.go
G-Rath, line 1: the changes here don't seem to be related to the rest of the PR - if that is indeed the case, could you revert them?

G-Rath (changes requested):

Ankitsinghsisodya:
I initially removed the type alias as you suggested, but CI failed with a depguard error:

```
import 'regexp' is not allowed from list 'regexp':
Use github.com/google/osv-scanner/v2/internal/cachedregexp instead (depguard)
```

The `exclude.go` file needs the `*regexp.Regexp` type for the struct field, but can't import `regexp` directly due to the lint rule. Adding the type alias back in `cachedregexp` allows other packages to use `*cachedregexp.Regexp` without violating depguard.

Should I look into adding an exception to the depguard config for this file, or is the type alias acceptable given this constraint?

G-Rath:
right yeah of course - I think for now let's go with adding the exception as that mirrors what we're doing elsewhere for this situation, though I think I'll revisit that after this as it'll probably be better to just re-export stuff from `cachedregexp` and/or try to use the `forbidigo` linter instead

(I had forgotten we pretty much don't use regexp outside of `cachedregexp.MustCompile` 😅)

cmd/osv-scanner/scan/source/command.go
```diff
@@ -123,6 +127,7 @@ func action(_ context.Context, cmd *cli.Command, stdout, stderr io.Writer, clien
 	scannerAction.SBOMPaths = cmd.StringSlice("sbom")
 	scannerAction.Recursive = cmd.Bool("recursive")
 	scannerAction.NoIgnore = cmd.Bool("no-ignore")
+	experimentalScannerActions.ExcludePatterns = cmd.StringSlice("experimental-skip-dir")
```
G-Rath: can we do this assignment up with the rest of the `experimentalScannerActions` stuff?

cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -159,6 +159,27 @@ func TestCommand(t *testing.T) {
 			Args: []string{"", "source", "--recursive", "--no-ignore", "./testdata/locks-gitignore"},
 			Exit: 0,
 		},
+		// experimental skip-dir flag tests
```
G-Rath: can you update these test names too please, and add a couple that have multiple instances of the flag?

since the end-to-end tests can be more expensive, and you've got focused fatests for this functionality elsewhere, the tests here are mainly about covering our cmd usage, such as passing in the flag multiple times

pkg/osvscanner/exclude_test.go
```diff
@@ -0,0 +1,194 @@
+package osvscanner
+
+import (
+	"testing"
+)
+
+func TestParseExcludeArg(t *testing.T) {
```
G-Rath:
```suggestion
func Test_parseExcludeArg(t *testing.T) {
```

pkg/osvscanner/exclude_test.go
```diff
@@ -0,0 +1,194 @@
...
+				t.Errorf("parseExcludeArg(%q) patternType = %q, want %q", tt.arg, patternType, tt.wantPatternType)
+			}
+			if pattern != tt.wantPattern {
+				t.Errorf("parseExcludeArg(%q) pattern = %q, want %q", tt.arg, pattern, tt.wantPattern)
+			}
+		})
+	}
+}
+
+func TestParseExcludePatterns(t *testing.T) {
```
G-Rath:
```suggestion
func Test_parseExcludePatterns(t *testing.T) {
```

pkg/osvscanner/exclude_test.go
```diff
@@ -0,0 +1,194 @@
...
+	t.Parallel()
+	tests := []struct {
```
G-Rath, line 49: could you double check if there's any of these we can eliminate? I think all of the `wantX` ones can be inferred from the others e.g. we always set `globTestPath` when `wantGlob` is true

G-Rath (dismissed):
looking good, we just also do want to change the cmd flag to use `exclude` too

cmd/osv-scanner/scan/source/command.go
```diff
@@ -57,6 +57,10 @@ func Command(stdout, stderr io.Writer, client *http.Client) *cli.Command {
 				Usage: "include scanning git root (non-submoduled) repositories",
 				Value: false,
 			},
+			&cli.StringSliceFlag{
+				Name:  "experimental-exclude",
+				Usage: "exclude paths during scanning; use g:pattern for glob, r:pattern for regex, or just dirname for exact match (can be repeated)",
```
another-rex:
```suggestion
				Usage: "exclude directory paths during scanning; use g:pattern for glob, r:pattern for regex, or just dirname for exact match (can be repeated)",
```

another-rex (approved):
Looks good!

pkg/osvscanner/exclude.go
```diff
@@ -0,0 +1,113 @@
...
+// Returns (patternType, pattern) where:
+//   - patternType is "" for exact match, "g" for glob, "r" for regex, or the unknown prefix
+//   - pattern is the actual pattern to use
+//
+// Unknown prefixes are returned as-is so the caller can provide appropriate error messages.
+func parseExcludeArg(arg string) (string, string) {
+	// Handle Windows absolute paths (e.g., C:\path)
+	if runtime.GOOS == "windows" && filepath.IsAbs(arg) {
+		return "", arg
+	}
```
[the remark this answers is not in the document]
another-rex: This is just incorrect, exact directory match is matching on paths I believe.

another-rex:
Hmm the requested interactions not found errors are a bit odd in the CI, because it looks like you have added the new interactions to the cassettes. @G-Rath Can you take a look at why that could be happening?

G-Rath:
I think they might just need to be regenerated, as they're still using the old `X-Test-Name` and it doesn't look like there are interactions for the `exclude_with_multiple_exact_directories` and `exclude_with_multiple_pattern_types` tests.

First I'd do is just revert all changes to `cmd/osv-scanner/scan/source/testdata/cassettes/TestCommand.yaml`, re-run the suite, and commit the results

another-rex:
Apologies for the delay, tests are mostly passing now, though there does seem to be a legitimate issue on Windows. 

Is it easy for you to test on Windows to debug this? Otherwise we can look into it when we have time available.

Ankitsinghsisodya:
@another-rex Sorry to say but I can only test on macos and Linux(ubuntu).

another-rex:
@G-Rath Can you take a look at the windows error and see if it can be resolved easily?

pkg/osvscanner/exclude.go
```diff
@@ -0,0 +1,113 @@
...
+
+// parseExcludeArg parses a single exclude argument.
+// Returns (patternType, pattern) where:
+//   - patternType is "" for exact match, "g" for glob, "r" for regex, or the unknown prefix
+//   - pattern is the actual pattern to use
+//
+// Unknown prefixes are returned as-is so the caller can provide appropriate error messages.
+func parseExcludeArg(arg string) (string, string) {
+	// Handle Windows absolute paths (e.g., C:\path)
+	if runtime.GOOS == "windows" && filepath.IsAbs(arg) {
```
G-Rath: @another-rex this is why Windows is unhappy, because `r:/` and `g:/` are absolute paths on Windows

Solutions that come to mind are:
  1. we assume a lowercase `r` and `g` are our prefixes
      * Windows drive letters are case-insensitive and usually represented in uppercase
  2. we have a double colon rule (i.e. have both `r::/` and `r:/` be valid)
      * yuck, and yet more of a bandaid
  3. we make it a requirement to prefix with `:` if you're doing absolute paths, at least on Windows
      * this'd be inconsistent with our logic elsewhere, though we could change that too..
  4. we change the symbol 
      * `:` feels like the most intuitive symbol to me though for us to be using 😞 
  5. we explicitly check for `r:` and `g:`, and say Windows users have to prefix with `:` if they want to pass in those driver letters in lowercase
  6. we do nothing (at least for now), and tell people their regexps need to start with something e.g. `.*`

Those last two seem like the most straightforward for the time being (given this is experimental and all)

another-rex: I think 5 is probably the best way to go here, and on windows just force uppercase to be used for drive names if they want to have drive r: or drive g: to be interpreted as drives.

another-rex (approved):

Issue #2403, opened 2025-12-08
Question: Git ecosystem support


Avgor46:
Hi!

Are there any plans to support Git as an ecosystem for scanning?

## Problem

Currently, it is impossible to obtain vulnerabilities from an SBOM containing the Git URL of the collected packages. I am aware of the [discussion](https://github.com/google/osv.dev/discussions/2040) and other [topics](https://github.com/google/osv-scanner/issues/1352) about the synthetic nature of GIT, but that seems irrelevant because we can now request osv.dev like this:
```bash
curl -d \
  '{"package": {"name": "https://github.com/curl/curl.git", "ecosystem": "GIT"}, "version": "8.5.0"}' \
  "https://api.osv.dev/v1/query"
```
And some progress in establishing Git as an ecosystem was made in this [PR](https://github.com/google/osv.dev/pull/3700).

## Original problem

Initially, I intended to scan Homebrew packages for vulnerabilities. However, Homebrew is not considered an ecosystem due to its lack of advisories and organization. However, starting with version 4.3.0, Homebrew packages are supplied with SPDX BOM. This BOM often contains the Git repository URL from which the package came. Therefore, supporting the Git ecosystem will allow us to scan some Homebrew packages for vulnerabilities, which is extremely important for scanning the MacOS system, where Homebrew is the main package manager (imho).

## Proposed solution

Based on [this](https://github.com/google/osv-scanner/issues/1352) and [this](https://github.com/google/osv.dev/pull/3953), I can suggest the following structure for the Git SBOM component which will be accepted for scanning:
``` json
  "components": [
    {
      "bom-ref": "...",
      "type": "..."
      "name": "github.com/curl/curl",
      "version": "3.4.0",
      "purl": "pkg:generic/github.com/curl/curl@8.5.0"
    }
  ]
```

If it is useful and makes sense to be implemented, I would like to implement it by myself.

another-rex:
Definitely open to getting SBOM git urls, Do you have some examples of these homebrew SBOMs?

retanoj:
Sorry to bother you, but I have a question: I want to use purl to indicate that the application depends on openssl@3.0.0 compiled from GitHub source code. How should I express this?

I use 'pkg:generic/openssl@3.0.0' in spdx-2.3 format sbom. But osv-scanner can not find vulnerbilities related to it .

Avgor46:
@another-rex, sure.

For example, for openssl package we have the following component in homebrew sbom:
```json
    {
[...]

PR #2460, opened 2026-01-12
fix: Update osv-scanner.json git queries

another-rex:
Add test after fixing the issue mentioned in https://github.com/google/osv-scanner/issues/2403#issuecomment-3721090878

The fix is done in osv-scalibr, and already updated in another PR.

### DIFF
--- a/cmd/osv-scanner/scan/source/testdata/locks-insecure/osv-scanner-custom-git-tag.json
+++ b/cmd/osv-scanner/scan/source/testdata/locks-insecure/osv-scanner-custom-git-tag.json
@@ -1,0 +1,16 @@
+{
+  "results": [
+    {
+      "source": {},
+      "packages": [
+        {
+          "package": {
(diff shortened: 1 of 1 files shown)


cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -934,6 +934,11 @@ func TestCommand_GithubActions(t *testing.T) {
 			Args: []string{"", "source", "-L", "osv-scanner:./testdata/locks-insecure/osv-scanner-flutter-deps.json"},
 			Exit: 1,
 		},
+		{
+			Name: "scanning osv-scanner custom format with git tag",
+			Args: []string{"", "source", "-L", "osv-scanner:./testdata/locks-insecure/osv-scanner-custom-git-tag.json"},
```
cuixq, line 939: I don't think this test file is checked in?

PR #2466, opened 2026-01-16
feat: replace `pomxmlenhanceable` extractor with transitive enricher

### LINKED ISSUE #2465: Use `transitivedependency` enricher for transitive extraction for Maven pom.xml
We should be able to use the enricher added in https://github.com/google/osv-scalibr/pull/1573 and deprecate `pomxmlenhanceable` extractor.

G-Rath:
~Blocked by #2461~
~Blocked by https://github.com/google/osv-scalibr/pull/1674~

Resolves #2465

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -430,7 +430,7 @@ Scanned <rootdir>/testdata/locks-many-with-insecure/alpine.cdx.xml file and foun
 Scanned <rootdir>/testdata/locks-many-with-insecure/composer.lock file and found 6 packages
 Scanned <rootdir>/testdata/locks-many-with-insecure/package-lock.json file and found 1 package
 Scanned <rootdir>/testdata/locks-many-with-insecure/yarn.lock file and found 1 package
-Scanned <rootdir>/testdata/maven-transitive/pom.xml file and found 3 packages
+Scanned <rootdir>/testdata/maven-transitive/pom.xml file and found 1 package
 Filtered 1 local/unscannable package/s from the scan.
 Package npm/has-flag/4.0.0 has been filtered out because: (no reason given)
 Package npm/wrappy/1.0.2 has been filtered out because: (no reason given)
@@ -5486,7 +5486,7 @@ Fetching response from: https://repo.maven.apache.org/maven2/org/apache/logging/
 Fetching response from: https://repo.maven.apache.org/maven2/org/apache/logging/log4j/log4j-web/2.14.1/log4j-web-2.14.1.pom
 Fetching response from: https://repo.maven.apache.org/maven2/org/apache/logging/log4j/log4j/2.14.1/log4j-2.14.1.pom
 Fetching response from: https://repo.maven.apache.org/maven2/org/apache/logging/logging-parent/3/logging-parent-3.pom
-Scanned <rootdir>/testdata/maven-transitive/registry.xml file and found 59 packages
+Scanned <rootdir>/testdata/maven-transitive/registry.xml file and found 2 packages
 Total 2 packages affected by 6 known vulnerabilities (2 Critical, 1 High, 3 Medium, 0 Low, 0 Unknown) from 1 ecosystem.
 6 vulnerabilities can be fixed.
 
@@ -5541,7 +5541,7 @@ Total 3 packages affected by 11 known vulnerabilities (1 Critical, 4 High, 6 Med
 ---
 
 [TestCommand_Transitive/scans_dependencies_from_multiple_registries - 1]
-Scanned <rootdir>/testdata/maven-transitive/registry.xml file and found 59 packages
+Scanned <rootdir>/testdata/maven-transitive/registry.xml file and found 2 packages
 Total 2 packages affected by 6 known vulnerabilities (2 Critical, 1 High, 3 Medium, 0 Low, 0 Unknown) from 1 ecosystem.
 6 vulnerabilities can be fixed.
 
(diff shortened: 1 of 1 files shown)


pkg/osvscanner/scan.go
```diff
@@ -93,15 +86,39 @@ func getPlugins(defaultPlugins []string, accessors ExternalAccessors, actions Sc
...
-	if !actions.TransitiveScanning.Disabled && isRequirementsExtractorEnabled(plugins) {
-		p, err := transitivedependencyrequirements.New(&cpb.PluginConfig{
-			UserAgent: actions.RequestUserAgent,
-		})
-		if err != nil {
-			log.Errorf("Failed to make transitivedependencyrequirements enricher: %v", err)
-		} else {
-			plugins = append(plugins, p)
+	if !actions.TransitiveScanning.Disabled {
+		// TODO: Use Enricher.RequiredPlugins to check this generically
```
G-Rath, line 90: I'll do this refactor in a follow-up, now that we've got a couple of plugins

cuixq:
I believe this is unblocked now!

G-Rath:
@cuixq @another-rex I might need a hand with this - if we focus on just e.g. `TestCommand_Transitive/scans_transitive_dependencies_by_specifying_pom.xml`, for some reason [this `FS.Open` call](https://github.com/google/osv-scalibr/blob/main/enricher/transitivedependency/pomxml/pomxml.go#L123) is erroring:

```
open /home/jones/workspace/projects-oss/osv-scanner/cmd/osv-scanner/scan/source/testdata/maven-transitive/abc.xml: invalid argument
```

I don't feel familiar enough with `FS` and the expectations in scalibr around absolute vs relative paths to properly dig into this, but note if locally I hackly change this to a relative path (literally by just hardcoding the path in my local `osv-scalibr` with a `replace` in `go.mod`) it seems to work - but I'm not sure if that is an actual lead, or just a fluke?

Does any of this sound familiar to you?

Note that CI here currently does not show the error because of #2515

cuixq:
it seems absolute path will be considered invalid since it starts with a slash according to https://github.com/golang/go/blob/master/src/io/fs/fs.go

is there a way that we can get the relative path from the path?

G-Rath:
I mean we have the root path (which is `/`) so we should be able to make it relative, I just would have thought this to be an issue elsewhere in scalibr?

G-Rath:
~I'm also not seeing any reference to absolute paths in that go file?~

> // Paths must not start or end with a slash: “/x” and “x/” are invalid.

[Got it](https://github.com/golang/go/blob/master/src/io/fs/fs.go#L18C1-L18C72)

cuixq:
Filesystem extractor probably does not have this problem since its ScanInput has the [relative path](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/filesystem.go#L72)

G-Rath:
~that struct is what is being used here~, I think its just we are stripping `/`s and its just not documented anywhere 🤷 

I'll do a PR fixing the enricher

G-Rath:
https://github.com/google/osv-scalibr/pull/1753

G-Rath:
This is also blocked by https://github.com/google/osv-scalibr/pull/1765

G-Rath:
ok I think this should be good to go once https://github.com/google/osv-scalibr/pull/1765 is landed

pkg/osvscanner/scan.go
G-Rath: https://github.com/google/osv-scanner/pull/2526

pkg/osvscanner/scan.go
```diff
@@ -267,7 +285,6 @@ SBOMLoop:
 			Stats:                 &statsCollector,
 			ReadSymlinks:          false,
 			MaxInodes:             0,
-			StoreAbsolutePath:     true,
```
G-Rath, line 270: I'm not sure if we need to do something for `ScanContainer`, and it doesn't look like we've got any tests that would cover pomxml transitive dependencies?

@cuixq it might be useful to [the "full java" `pom.xml`](https://github.com/google/osv-scanner/blob/main/cmd/osv-scanner/scan/image/testdata/java-fixture/app/pom.xml) to include a transitive dependency, if it doesn't already

cuixq: I think both direct dependencies `org.apache.hadoop:hadoop-client` and `org.apache.commons:commons-compress` have transitive dependencies - is that what you mean?

G-Rath: maybe 🙃 

while I can read and write Java, I've not used it in years and never really had to use Maven a lot so I don't know if what's in that test is using transitive dependencies enough to cover all the different ways those can interact with the transitive enricher

cuixq: I guess you mean we need a good pom.xml that we can test whether the transitive enricher works or not?

the linked pom.xml lists two direct dependencies which "implicitly" depend on the transitive dependencies.

we rely on the snapshot for the number of package found as well as the vulnerabilities affecting transitive dependencies to know if the transitive dependencies are scanned:
 - I do see the number of package found has changed but that is due to the offline pomxml is always used now and I remember we have a TODO about how we can log the number of packages found in transitive enricher when we work on this for python requirements.txt
 - I don't see any snapshot change regarding the vulnerabilities found from the pom.xml so this probably indicates there is no regression on transitive scanning for pom.xml.

G-Rath: if you think the transitive enricher is running for that container test, then it should be good enough because it would be erroring otherwise

cuixq: ah you mean the container test - I was confused with source scanning, it seems we are running the [java/archive](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/java/archive/archive.go) extractor on the artifact built based on that pom.xml so we probably don't need to worry about the transitive enricher for this test.

pkg/osvscanner/scan.go
cuixq: also note that we are planning some refactoring work on enabling transitive enrichers https://github.com/google/osv-scanner/issues/2537

PR #2485, opened 2026-01-26
refactor: remove `noCFG` wrapper for enrichers

G-Rath:
While the `baseimage` enricher technically does not use `PluginConfig` yet, we can still remove our `noCFG` wrapper 🎉 

I've assumed that we want to return errors rather than continue panicking, since that's something we can now do

### DIFF
--- a/internal/scalibrplugin/presets.go
+++ b/internal/scalibrplugin/presets.go
@@ -1,6 +1,8 @@
 package scalibrplugin
 
 import (
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


G-Rath:
This'll conflict with https://github.com/google/osv-scanner/pull/2484 just because of code position.

Also I've not attempted to migrate the enricher in `osv-scalibr` as it takes a `baseimage.Client` in its config which I don't know how that gets handled with protobuffer, but happy to have a crack at that if someone has some pointers

PR #2499, opened 2026-02-09
refactor: use `Inventory` field on `ScanResults` as much as possible

G-Rath:
This progresses us towards leveraging `inventory.Inventory` by having us source `GenericFindings` and `PackageVulns` from the scan inventory, rather than our own fields on `ScanResults` and `PackageScanResults` respectively

### DIFF
--- a/pkg/osvscanner/vulnerability_result.go
+++ b/pkg/osvscanner/vulnerability_result.go
@@ -73,7 +73,7 @@ func buildVulnerabilityResults(
 		configToUse := scanResults.ConfigManager.Get(p.Location())
 
 		haveCheckedIgnore := false
-		for _, vuln := range scanResults.PackageVulns {
+		for _, vuln := range scanResults.Inventory.PackageVulns {
 			if vuln.Package == psr.PackageInfo.Package {
 				if !haveCheckedIgnore {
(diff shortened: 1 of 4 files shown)


pkg/osvscanner/vulnerability_result.go
```diff
@@ -72,17 +72,27 @@ func buildVulnerabilityResults(
...
-		if len(psr.Vulnerabilities) > 0 {
-			if !configToUse.ShouldIgnorePackageVulnerabilities(p) {
-				includePackage = true
-				pkg.Vulnerabilities = append(pkg.Vulnerabilities, psr.Vulnerabilities...)
-				pkg.Groups = grouper.Group(grouper.ConvertVulnerabilityToIDAliases(pkg.Vulnerabilities))
-				for i, group := range pkg.Groups {
-					pkg.Groups[i].MaxSeverity = output.MaxSeverity(group, pkg)
+		haveCheckedIgnore := false
+		for _, vuln := range scanResults.PackageVulns {
+			if vuln.Package == psr.PackageInfo.Package {
```
G-Rath, line 77: ~note: pretty sure this isn't the right way to do `Package` equality 😅~

This should be fine as we expect to be pointing to the same object in memory

another-rex (approved):

PR #2506, opened 2026-02-10
fix: bump Go to v1.25.7

G-Rath:
Addresses GO-2026-4337

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -84,7 +84,7 @@ Alternatively, you can install this from source by running:
 go install github.com/google/osv-scanner/v2/cmd/osv-scanner@latest
 ```
 
-This requires Go 1.25.6+ to be installed.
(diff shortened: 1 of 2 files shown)


cuixq:
we may need to update Go versions in Dockerfiles as well

G-Rath:
@cuixq these two places were the only instances of the exact version I could find, so I think everything else is using major.minor versions

another-rex (approved):

cuixq:
In [Dockerfile](https://github.com/google/osv-scanner/blob/main/Dockerfile), we are using Go 1.25.6 base image and updating `go.mod` to 1.25.7 will fail the build.

G-Rath:
hmm yeah ok I missed a bunch somehow - I did think it was odd that I only had to update two things...

G-Rath:
I think I've got them all now #2507 😓

PR #2523, opened 2026-02-17
refactor: update linter to v2.8 and do a bunch more pre-allocating

G-Rath:
`prealloc` has had some changes that makes it smarter, though now there are a few weird looking allocates; I had considered suggesting we disable it, but it has still flagged a few things that seemed silly for us to be doing so 🤷

### DIFF
--- a/internal/resolution/resolve.go
+++ b/internal/resolution/resolve.go
@@ -50,7 +50,7 @@ type NodeError struct {
 }
 
 func (res *Result) Errors() []NodeError {
-	var errs []NodeError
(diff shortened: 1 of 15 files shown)


internal/resolution/resolve.go
```diff
@@ -50,7 +50,7 @@ type NodeError struct {
 }
 
 func (res *Result) Errors() []NodeError {
-	var errs []NodeError
+	var errs []NodeError //nolint:prealloc
```
G-Rath, line 53: it looks like this is a bug in `prealloc` which is fixed in the next version

PR #2526, opened 2026-02-17
refactor: use `RequiredPlugins` to decide if transitive enrichers should be enabled

G-Rath:
Relates to #2537

### DIFF
--- a/pkg/osvscanner/scan.go
+++ b/pkg/osvscanner/scan.go
@@ -19,8 +19,6 @@ import (
 	transitivedependencyrequirements "github.com/google/osv-scalibr/enricher/transitivedependency/requirements"
 	"github.com/google/osv-scalibr/extractor"
 	"github.com/google/osv-scalibr/extractor/filesystem"
-	"github.com/google/osv-scalibr/extractor/filesystem/language/java/pomxml"
-	"github.com/google/osv-scalibr/extractor/filesystem/language/python/requirements"
 	"github.com/google/osv-scalibr/extractor/filesystem/simplefileapi"
 	"github.com/google/osv-scalibr/fs"
 	"github.com/google/osv-scalibr/inventory"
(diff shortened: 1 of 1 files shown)


pkg/osvscanner/scan.go
```diff
@@ -87,8 +77,7 @@ func getPlugins(defaultPlugins []string, accessors ExternalAccessors, actions Sc
 	plugins := scalibrplugin.Resolve(actions.PluginsEnabled, actions.PluginsDisabled)
 
 	if !actions.TransitiveScanning.Disabled {
-		// TODO: Use Enricher.RequiredPlugins to check this generically
-		if isRequirementsExtractorEnabled(plugins) {
+		if areRequiredPluginsEnabled(plugins, transitivedependencyrequirements.Enricher{}) {
```
G-Rath, line 94: This is probably relevant for #2537: I'm building the struct manually rather than using `enricher.New` because the latter actually does some work initializing API clients and such (which in some cases might even lead to an API request?).

I'm not sure if we can avoid that if we move to using a preset with our current structure

another-rex: We should be able to use New(), no network requests are made there. (The http clients just create a struct, the deps.dep gRPC clients are lazily initiated, no connection is opened until the first request is made)

G-Rath: it seems a little odd, but the best I can think of to deal with this without too much change would be to have some kind of `IsForRequirementChecks` flag in `cpb.PluginConfig`, with which ideally plugins should do:

```
// New creates a new Enricher.
func New(cfg *cpb.PluginConfig) (enricher.Enricher, error) {
	if cfg.IsForRequirementChecks {
		return Enricher{}, nil
	}
	// ...
}
```

🤔

another-rex: I'm not quite following, why do we need a separate initialisation for IsRequirements check?

another-rex: I was thinking we setup all the plugins, then do a loop and remove plugins that don't have their requirements met. Described in https://github.com/google/osv-scanner/issues/2537

G-Rath: yeah I posted that a couple of minutes after you posted your comment saying that we were ok with just always calling `New`.

I've had a look at the code paths and it looks like at most what we end up doing is potentially loading in some TLS PEM store stuff, but most of it does look very lazy (i.e. assigning a `setupThisThing` function field to be called later), and with no API calls being made.

Personally I'm still not the biggest fan of how much work it appears that we would be doing, but I don't think there's an easy way around that because Go 🤷

cuixq: I think this works for now but longer term I prefer to do all the checks in `scalibrplugin.Resolve()`

cuixq (approved):

PR #2534, opened 2026-02-19
feat: add experimental support for updating config files

G-Rath:
This is an attempt to port the `--update-config-ignores` feature from `osv-detector`, which has been useful especially for onboarding new codebases and updating existing ones after patching has been done to prune out old entries.

Currently this does not preserve comments or specific formatting, but I'm hopeful those are things we can improve on future especially if this proves to be a popular flag


internal/config/__snapshots__/config_internal_test.snap
```diff
@@ -0,0 +1,97 @@
...
+ignore = false
+
+---
+
+[TestConfig_UpdateFile/ids_are_deduplicated - 1]
+[[IgnoredVulns]]
+id = "GHSA-123"
+
+[[IgnoredVulns]]
+id = "GHSA-456"
```
another-rex, line 81: Should we add a comment/reason saying this is automatically added by the new flag?

another-rex (commented):
Being able to automatically remove not found vulnerabilities from the config files seem quite useful. 

However I'm not sure how much we want to encourage people to automatically add all their vulns to the ignore file. At least we should separate the two functionalities, maybe by making UpdateConfigIgnores not a boolean flag

G-Rath:
> However I'm not sure how much we want to encourage people to automatically add all their vulns to the ignore file

I originally felt against this kind of thing too, similar to also having tools like this exit as an error if there were unused ignores, but I've found in practice at least in an agency company kind of context it can really help reduce the friction - I'm able to tell other developers who are less familiar with this space to just add "xyz flag" rather than needing to try explain how to write ignores (which they can still learn about because they see the diff + it helps keep the actual order and format consistent), and when onboarding apps from other vendors we can get the security pipelines setup faster by initially ignoring all existing vulns and then go back to triage.

With `osv-detector` prior to this flag I was having to run this `jq` query:

```
osv-detector --json . | jq -r  '[.results[].packages | map("- " + .vulnerabilities[].id)] | flatten | unique | sort | .[]'
```

which has no hope of preserving any other config content including other properties and comments (the latter of which are not being preserved "yet")

> At least we should separate the two functionalities, maybe by making UpdateConfigIgnores not a boolean flag

I really like this idea, because then we could also later introduce an "interactive" mode which could let you set reasons and "until" dates.

---

One side thing to consider with this is if we do the updating before or after our filtering - currently I have it being done before which means the tool always presents "no vulns found" and with a non-zero exit code; currently `osv-detector` does it the other way around, presenting everything as if it wasn't going to add ignores, and just adds a "ignored <count> vulns" message at the end of its output.

Since this is being added an experimental flag we don't need to decide right now, but something to think about - I can explore this more once we've gotten the rest of the implementation settled, since it should just be a case of moving a function call to see the difference 🤷

internal/config/__snapshots__/config_internal_test.snap
G-Rath: at this point I'm leaning to not doing that just for simplicity, and I'd rather add it later to see how we feel - I'll explore this once I'm done with the core implementation

G-Rath:
ok this still needs work - I need to have `ConfigMap` hold pointers so that ignores are properly removed...

G-Rath:
@another-rex @cuixq I think this one should be good to go 🤞

G-Rath:
This is ready to go but needs #3103 first

PR #2538, opened 2026-02-20
test: label snapshots

G-Rath:
(this is for showing off a new feature [I'm proposing for `go-snaps`](https://github.com/gkampitakis/go-snaps/pull/152))

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -1,5 +1,5 @@
 
-[TestCommand/.gitignored_files - 1]
+[TestCommand/.gitignored_files - 1 - stdout]
 Scanning dir ./testdata/locks-gitignore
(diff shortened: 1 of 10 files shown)


cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
```diff
@@ -5636,11 +5637,7 @@ Total 5 packages affected by 21 known vulnerabilities (1 Critical, 9 High, 10 Me
 
 ---
 
-[TestCommand_Transitive/uses_native_data_source_for_requirements.txt - 2]
-
----
-
-[TestCommand_Transitive/uses_native_data_source_for_requirements.txt - 2]
+[TestCommand_Transitive/uses_native_data_source_for_requirements.txt - 2 - stderr]
```
G-Rath: this actually looks like a general bug, this shouldn't be here twice

another-rex (approved):

PR #2543, opened 2026-02-22
ci: allow snapshots workflow to record new cassette interactions

G-Rath:
This should hopefully make it easier to land our snapshot updates when an external service change results in "new" interactions, while not changing existing interactions


G-Rath:
Ideally it would be good to land this ASAP even though CI is failing, as if it works like I hope it'll result in #2535 passing CI

Ly-Joey (approved):

cuixq (approved):

another-rex (approved):

PR #2544, opened 2026-02-22
test: use a fixed length name for temp directories

G-Rath:
Most of the time `os.MkdirTemp` creates directories with the same number of characters, but its not required to and will occasionally has one less character than what is common, which can result in the table output being slightly different.

This is compounded by us replacing the temp directory with a fixed string `<tempdir>` marker, as it ends up looking like the table marks are just randomly changing for no reason 🤦

### DIFF
--- a/internal/testutility/utility.go
+++ b/internal/testutility/utility.go
@@ -3,7 +3,9 @@ package testutility
 import (
 	"flag"
 	"fmt"
+	"math/rand"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


another-rex (approved):
I think technically we're meant to check whether that random folder already exists and generate a new random name if it does to avoid conflicts.

But what are the chances?

G-Rath:
yeah that was my thinking - it should be easy to put in a loop for that if it ever happens, but it should already be pretty small and arguably by the same logic it's technically possible to generate directories with the same name that already exist on every loop, at which point 🤷 

so line has to be somewhere right

internal/testutility/utility.go
```diff
@@ -105,14 +107,23 @@ func ValueIfOnWindows(win, or string) string {
...
 }
 
+func fixedLengthTempDir(parent string) (string, error) {
+	n := rand.Int63n(1_000_000_000_000) //nolint:gosec // 10^12
+	suffix := fmt.Sprintf("%0*d", 12, n)
+
+	name := "osv-scanner-test-" + suffix
+	path := filepath.Join(parent, name)
+
+	return path, os.Mkdir(path, 0o700)
```
cuixq, line 117: do we need to check if the path already exists? though the probability is very low...

cuixq (approved):

internal/testutility/utility.go
cuixq: oops didn't load the latest comments when I sent out the review

G-Rath:
ok so turns out this wasn't needed because we're outputting absolute paths which will differ in length anyway due to different OS/path setups 🙃

PR #2552, opened 2026-02-26
feat: deprecate `--output` in favor of `--output-file` and `--output-files`

### LINKED ISSUE #2531: Consider renaming `--output` to `--output-file`
`--output` and `--format` are two very common names for "the output format of the cli", but we're using `--output` for writing to a file.

Consider this:

```
❯ osv-scanner --output vertical .
Scanning dir .
Starting filesystem walk for root: /
Scanned /home/jones/workspace/projects/repo/composer.lock file and found 245 packages
End status: 11 dirs visited, 31 inodes visited, 1 Extract calls, 3.745637ms elapsed, 3.745692ms wall time
```

This has just created a file called `vertical`, but gives no indication of that.

If we were to instead use `--output-path`, that'd let us catch when people provide `--output` when they intend `--format`, and as a bonus make it just generally clearer that the flag takes a path to a file

G-Rath:
Resolves #2531

### DIFF
--- a/cmd/osv-reporter/main.go
+++ b/cmd/osv-reporter/main.go
@@ -75,15 +75,25 @@ func run(args []string, stdout, stderr io.Writer) int {
 				Required:  true,
 			},
 			&cli.StringSliceFlag{
-				Name: "output",
-				Usage: "used to save files to various formats (--output=[format]:[path],[format]:[path]...).\n" +
+				Name: "output-files",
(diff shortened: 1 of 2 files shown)


cuixq (approved):

G-Rath:
@another-rex @cuixq just fyi it looks like the test for this might be flakey - I've got my eye on it and will dig into it more later this week

cuixq:
yes I also noticed this in https://github.com/google/osv-scanner/pull/2571. thanks for taking care of it!

PR #2571, opened 2026-03-02
feat: enable transitive scanning for requirements.txt using deps.dev API

cuixq:
This PR enables transitive dependency scanning for Python requirements.txt, utilizing the deps.dev API by default for resolution.
   - Supported the configuration toggle between deps.dev and native data sources for Python transitive resolution. 
   - Added a new test case to scan transitive dependencies in requirements.txt with deps.dev API by default.

~~The difference of scanning results in the snapshot is due to deps.dev API which is under investigation. I may not merge this PR until the issue is fixed but happy to seek review first.~~

I will also cleanup the test cases (naming and ordering) for transitive scanning in a follow up PR.


cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -1300,9 +1300,14 @@ func TestCommand_Transitive(t *testing.T) {
 			Args: []string{"", "source", "--data-source=native", "-L", "pom.xml:./testdata/maven-transitive/registry.xml"},
 			Exit: 1,
 		},
+		{
+			Name: "scans transitive dependencies in requirements.txt with deps.dev API by default",
```
G-Rath, line 1304:
```suggestion
			Name: "scans_transitive_dependencies_in_requirements.txt_with_deps.dev_API_by_default",
```

another-rex: let's keep it with spaces for now and do a pass through to replace all the names with _

cuixq: I will make this change in the follow up PR with other cleanup changes.

G-Rath: just to be clear I wasn't meaning this as a blocking change, I've just been switching tests over as I've touched this and introduced them so was just point that opportunity out here too 🙂

cuixq:
it seems the deps.dev API issue is fixed as I can see the same results for deps.dev API and the native registry, so once the merge conflicts are resolved, this PR should be ready to go.

PR #2574, opened 2026-03-03
fix: ensure "send everything to stderr" is set before any logging calls

G-Rath:
So it turns out we've got a race condition with logging that happens in flag actions since the action for the `--format` flag might not be run first - we've not hit this previously because the only reason we tend to do logging in our flag actions is for deprecated flags and the only one of those we had before #2552 was `--sbom` whose tests were always using the `table` format.

#2552 however deprecated `--output` which is used for tests relating to the HTML format which does trigger everything else going to stderr, and hence unearthed this race condition.

Thankfully it seems `urfave` has a `Validator` function which gets called as part of parsing flags before any `Action` is called - this does result in different messaging including for some reason the error message being repeated twice, so for now I've just used the `Validator` to handle setting "send everything to stderr" even though that's really not what a validator should be used for...

I've also updated the HTML output tests to snapshot both stdout and stderr as I didn't see a reason not to be doing that, and if we were it'd made the issue a lot more obvious...

### DIFF
--- a/cmd/osv-scanner/internal/helper/flags.go
+++ b/cmd/osv-scanner/internal/helper/flags.go
@@ -62,20 +62,21 @@ func BuildCommonScanFlags(defaultExtractors []string) []cli.Flag {
 			Aliases: []string{"f"},
 			Usage:   "sets the output format; value can be: " + strings.Join(reporter.Format(), ", "),
 			Value:   "table",
-			Validator: func(s string) error {
+			Action: func(_ context.Context, _ *cli.Command, s string) error {
 				if slices.Contains(reporter.Format(), s) {
-					if s != "vertical" && s != "table" && s != "markdown" {
-						// technically this is a side effect, which shouldn't be part of "validation"
-						// but we need to ensure this is set before any logging happens, which could
-						// be in the "action" of other flags (such as for deprecations), but should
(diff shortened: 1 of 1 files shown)


another-rex:
I'm not 100% following, so just to double check, this is not that one test is sending everything to stderr causing others to send to stderr. 

Just that in certain formats we have do send everything to stderr, but that might be ran at the same time as the format check? I thought they would happen in serial...

G-Rath:
@another-rex yup correct:

image

I too would have expected the flags to be `Action`'d in order, but maybe they use a map somewhere.. but I'm a little hesitant to report this upstream as a bug given their position on parallel exception.

(I have got a note to try and dig into further on the package side, but for now this should unblock us)

G-Rath:
hmm I think I might change my fix to be just doing the `cmdlogger.SendEverythingToStderr()` part in the `Validator`, and keeping the original `Action` for handling actually validating the flag value, and then look into what we want to do w/ `Validator` vs `Action` more generally as its own issue.

Even though that is bad from a responsibilities POV, it'll mean we can avoid dealing with the different output we're getting from `Validator`, and in particular the fact that we're having the "invalid value" message being printed multiple times for some reason...

PR #2575, opened 2026-03-04
feat: introduce dedicated `transitive` preset and more generally account for "required plugins" during resolving

### LINKED ISSUE #2537: Fix transitive plugin resolution
Related: #2327 

Right now in: https://github.com/google/osv-scanner/blob/f15560714dd622b745463ad50e3b314e2f5f11bf/pkg/osvscanner/scan.go#L93-L103

it forcibly enables the transitive plugins.

I think we need to rework this logic a bit, right now I think a viable solution is:

1. Put all the transitive enrichers into a preset
2. If transitive scanning is not disabled, add that preset to the list of enabled plugins (or default to having it in the list of enabled, and disable the preset of transitive scanning is disabled.)
3. All plugin in resolve should check its requirements, and remove itself if from the list of enabled plugins if it does not have the prereqs.  (This is the TODO comment basically)

G-Rath:
Resolves #2537

### DIFF
--- a/internal/scalibrplugin/resolve.go
+++ b/internal/scalibrplugin/resolve.go
@@ -101,10 +101,10 @@ func Resolve(enabledPlugins []string, disabledPlugins []string, cfg *cpb.PluginC
 		}
 	}
 
-	return withoutPluginsMissingRequiredPlugins(plugins, asSlice)
+	return filterPluginsMissingRequiredPlugins(plugins, asSlice)
 }
 
-func withoutPluginsMissingRequiredPlugins(pluginStatues map[string]bool, loaded []plugin.Plugin) []plugin.Plugin {
+func filterPluginsMissingRequiredPlugins(pluginStatues map[string]bool, loaded []plugin.Plugin) []plugin.Plugin {
 	plugins := make([]plugin.Plugin, 0, len(loaded))
 
 	for _, plug := range loaded {


internal/scalibrplugin/resolve.go
```diff
@@ -88,12 +90,36 @@ func Resolve(enabledPlugins []string, disabledPlugins []string, cfg *cpb.PluginC
...
 
 			asSlice = append(asSlice, plug)
 		}
 	}
 
-	return asSlice
+	return withoutPluginsMissingRequiredPlugins(plugins, asSlice)
+}
+
+func withoutPluginsMissingRequiredPlugins(pluginStatues map[string]bool, loaded []plugin.Plugin) []plugin.Plugin {
```
cuixq: the name is a bit confusing to me - probably rename it to `filterPluginsMissingRequiredPlugins` or simply ``filterPlugins` with some comment saying it filters out the ones missing required plugins?

internal/scalibrplugin/resolve_test.go
```diff
@@ -350,6 +354,118 @@ func TestResolve_Detectors(t *testing.T) {
...
+		},
+		{
+			name: "one_enricher_with_its_required_plugin_disabled",
+			args: args{
+				enabled:  []string{transitivedependencyrequirements.Name},
+				disabled: []string{requirements.Name},
+			},
+			want: []string{},
+		},
+		//
```
cuixq, line 385: delete this line?

G-Rath: this kind of line has been used throughout this file to break up blocks of tests, otherwise I find they get very hard to read through especially when having IDEs fold them - in this case the first "block" is focused on one specific enricher, while the second "block" is using the new `transitive` preset

internal/scalibrplugin/resolve.go
G-Rath: yeah this isn't a function name I'm proud of, but I couldn't come up with a concise name 😅

internal/scalibrplugin/__snapshots__/resolve_test.snap
```diff
@@ -23,6 +97,32 @@ weakcredentials/filebrowser
...
+rust/cargoauditable
+vex/os-duplicate/apk
+vex/os-duplicate/dpkg
+---
+
+[TestResolve_Enrichers_Presets/licenses - 1]
+license/depsdev
+---
+
+[TestResolve_Enrichers_Presets/transitive - 1]
```
G-Rath, line 117: note this is empty because none of the required plugins are enabled - ideally I'd like to have them here, but for now I've just added the "all presets" snapshot test since we just have the two plugins that require another plugin 🤷

internal/scalibrplugin/resolve_test.go
cuixq: ah I see - maybe add an comment to indicate this is a "separator" between the two types of tests.

PR #2577, opened 2026-03-04
ci: upgrade `golangci-lint` to v2.9 and use a versions file

G-Rath:
This version supports Go v1.26, and the version file should make it a bit easier to update things 🤞


G-Rath:
This should unblock #2576 and #2566

another-rex (approved):

PR #2585, opened 2026-03-04
test: have `mcp` integration tests use `osv-scanner-test.toml`

G-Rath:
This matches what we're doing in our other tests, and means that tools like scorecard won't ding us for the vulnerability in this test data


Ly-Joey (approved):

cuixq:
why we need this change?

G-Rath:
@cuixq same reason we have all the other `osv-scanner.toml` files in our testdatas - I've updated the description to reflect that

PR #2586, opened 2026-03-04
test: actually use a valid path in json replace rules

G-Rath:


G-Rath:
ugh nope I got myself confused - these are working fine 🤦

PR #2588, opened 2026-03-04
feat: Move away from github-pages gem and add ruby 4 support for docs

hopkincame:
This PR reworks the doc page deployment by removing the github pages gem and deploying using a github actions workflow. This was done to enable support for ruby 4, which is unsupported by github pages. As a result of the docs being deployed by github actions rather than pages, the deployment process for docs has changed, this PR also has an update to the contribution docs to reflect these changes. The github action workflow is set to deploy docs when there is a push to the docs branch.

This PR also adds some plugins for Jekyll. These were ones that github pages added automatically, now that we have removed github pages, thee plugins are included here instead.

Once this PR is merged, the "Pages" settings will need to be updated. The build and deployment source will need to be changed to github actions.

### DIFF
--- a/cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/image/__snapshots__/command_test.snap
@@ -440,7 +440,7 @@ Scanning local image tarball "./testdata/test-ubuntu.tar"
 
 
 Container Scanning Result (Ubuntu 22.04.5 LTS) (Based on "ubuntu" image):
-Total 22 packages affected by 46 known vulnerabilities (3 Critical, 14 High, 24 Medium, 3 Low, 2 Unknown) from 1 ecosystem.
+Total 22 packages affected by 46 known vulnerabilities (2 Critical, 15 High, 24 Medium, 3 Low, 2 Unknown) from 1 ecosystem.
 24 vulnerabilities can be fixed.
 
 
@@ -489,7 +489,7 @@ Scanning local image tarball "./testdata/test-ubuntu.tar"
 
 
 Container Scanning Result (Ubuntu 22.04.5 LTS) (Based on "ubuntu" image):
-Total 22 packages affected by 46 known vulnerabilities (3 Critical, 14 High, 24 Medium, 3 Low, 2 Unknown) from 1 ecosystem.
+Total 22 packages affected by 46 known vulnerabilities (2 Critical, 15 High, 24 Medium, 3 Low, 2 Unknown) from 1 ecosystem.
 24 vulnerabilities can be fixed.
 
 
@@ -557,7 +557,7 @@ Scanning local image tarball "./testdata/test-ubuntu-with-packages.tar"
 
 
 Container Scanning Result (Ubuntu 22.04.5 LTS) (Based on "ubuntu" image):
-Total 22 packages affected by 46 known vulnerabilities (3 Critical, 14 High, 24 Medium, 3 Low, 2 Unknown) from 1 ecosystem.
+Total 22 packages affected by 46 known vulnerabilities (2 Critical, 15 High, 24 Medium, 3 Low, 2 Unknown) from 1 ecosystem.
 24 vulnerabilities can be fixed.
 
 
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -5777,8 +5777,8 @@ Scanning dir ./testdata/locks-requirements/requirements.txt
 Scanned <rootdir>/testdata/locks-requirements/requirements.txt file and found 3 packages
 Loaded PyPI local db from <tempdir>/osv-scanner/PyPI/all.zip
 
-Total 3 packages affected by 12 known vulnerabilities (1 Critical, 4 High, 6 Medium, 1 Low, 0 Unknown) from 1 ecosystem.
-12 vulnerabilities can be fixed.
+Total 3 packages affected by 13 known vulnerabilities (1 Critical, 4 High, 7 Medium, 1 Low, 0 Unknown) from 1 ecosystem.
+13 vulnerabilities can be fixed.
 
 +-------------------------------------+------+-----------+----------+---------+---------------+----------------------------------------------+
 | OSV URL                             | CVSS | ECOSYSTEM | PACKAGE  | VERSION | FIXED VERSION | SOURCE                                       |
@@ -5797,6 +5797,7 @@ Total 3 packages affected by 12 known vulnerabilities (1 Critical, 4 High, 6 Med
 | https://osv.dev/GHSA-j8r2-6x86-q33q |      |           |          |         |               |                                              |
 | https://osv.dev/GHSA-9hjg-9r4m-mvj7 | 5.3  | PyPI      | requests | 2.20.0  | 2.32.4        | testdata/locks-requirements/requirements.txt |
 | https://osv.dev/GHSA-9wx4-h78v-vm56 | 5.6  | PyPI      | requests | 2.20.0  | 2.32.0        | testdata/locks-requirements/requirements.txt |
+| https://osv.dev/GHSA-gc5v-m9x4-r6x2 | 4.4  | PyPI      | requests | 2.20.0  | 2.33.0        | testdata/locks-requirements/requirements.txt |
 +-------------------------------------+------+-----------+----------+---------+---------------+----------------------------------------------+
 
 ---


.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
+name: Deploy Jekyll site to Pages
+
+on:
+  push:
+    branches: [docs]
```
G-Rath, line 5: I don't think we actually need to have a dedicated docs branch right?

I think right now the configuration is that we're deploying from `main`, and that should be fine; alternatively we could have it be deployed _just_ on workflow dispatch (since the docs don't get changed with every PR), and/or as part of our release workflow (to help ensure the docs are up to date, in case we forget to publish a small update or something...)

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
```
G-Rath: can we set the workflow permissions to none, and then set the permissions individually for each job based on what they need

```suggestion
# Restrict jobs in this workflow to have no permissions by default; permissions
# should be granted per job as needed using a dedicated `permissions` block
permissions: {}
```

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
```
G-Rath, line 12: very nitty but typically we have concurrency before the `permissions` block

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./docs
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Ruby
+        # https://github.com/ruby/setup-ruby/releases/tag/v1.289.0
+        uses: ruby/setup-ruby@19a43a6a2428d455dbd1b85344698725179c9d8c
```
G-Rath:
```suggestion
        uses: ruby/setup-ruby@19a43a6a2428d455dbd1b85344698725179c9d8c # v1.289.0
```

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          ruby-version: '4.0.1' # Not needed with a .ruby-version file
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+          cache-version: 0 # Increment this number if you need to re-download cached gems
+      - name: Setup Pages
+        id: pages
+        uses: actions/configure-pages@v5
```
G-Rath:
```suggestion
        uses: actions/configure-pages@983d7736d9b0ae728b81ab479565c72886d7745b # v5.0.0
```

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Build with Jekyll
+        run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
+        env:
+          JEKYLL_ENV: production
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
```
G-Rath:
```suggestion
        uses: actions/upload-pages-artifact@7b1f4a764d45c48632c6b24a0339c27f5614fb0b # v4.0.0
```

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
```
G-Rath:
```suggestion
        uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e # v4.0.5

```

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+      run:
+        working-directory: ./docs
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Ruby
+        # https://github.com/ruby/setup-ruby/releases/tag/v1.289.0
+        uses: ruby/setup-ruby@19a43a6a2428d455dbd1b85344698725179c9d8c
+        with:
+          ruby-version: '4.0.1' # Not needed with a .ruby-version file
```
G-Rath:
```suggestion
          ruby-version: '4' # Not needed with a .ruby-version file
```

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+  build:
+    env: # $BUNDLE_GEMFILE must be set at the job level, so it is set for all steps
+      BUNDLE_GEMFILE: ${{ github.workspace }}/docs/Gemfile
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./docs
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
G-Rath:
```suggestion
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          persist-credentials: false
```

docs/images/github-page.png
G-Rath, line 1: could you crop this to remove everything above the "GitHub Pages" header

docs/Gemfile
```diff
@@ -9,15 +9,20 @@ source "https://rubygems.org"
...
-# gem "jekyll", "~> 4.3.2"
-# This is the default theme for new Jekyll sites. You may change this to anything you like.
-# If you want to use GitHub Pages, remove the "gem "jekyll"" above and
-# uncomment the line below. To upgrade, run `bundle update github-pages`.
-gem "github-pages", "~> 232", group: :jekyll_plugins
-# If you have any plugins, put them here!
-group :jekyll_plugins do
-  gem "jekyll-feed", "~> 0.15"
-end
+gem "jekyll", "~> 4.3.2"
```
G-Rath, line 12: we're not necessarily using all of these plugins ay, they're just `github-pages` was using right?

Happy to do that for now, though it'd be good if we could try and remove ones we're obviously not using - I know for sure we have no `coffeescript`, and I'd guess `gist` and `paginate` probably are not being used either

.github/workflows/docs-deploy.yml
hopkincame: This project has a dedicated docs branch (osv.dev doesn't have one though) and based on the deployment times, it appears to be what is currently used to deploy docs. I can't see what the settings for deploys are though so I can't 100% confirm and I can change it to main if that is desired, but let's see what Google says first

docs/docs.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Use an official Ruby runtime as a parent image.
-FROM ruby:3
+FROM ruby:4
```
another-rex: Can you pin this to a specific hash? Once it's pinned it will be auto updated by renovate.

.github/workflows/docs-deploy.yml
another-rex: We are intentionally deploying from docs, the idea is if we need to add any mid release docs changes to the currently released version we can, while we still are able to make new docs updates to the upcoming changes currently in main but not released.

docs/_config.yml
```diff
@@ -30,10 +30,21 @@ ga_tracking: G-60SWD6J0TM
...
   - jekyll-feed
+  - jekyll-coffeescript
+  - jekyll-gist
+  - jekyll-github-metadata
+  - jekyll-paginate
+  - jekyll-relative-links
+  - jekyll-optional-front-matter
+  - jekyll-readme-index
+  - jekyll-default-layout
+  - jekyll-titles-from-headings
```
another-rex: This seems to need a lot more plugins now, can you add a bit to the PR description to explain why this is necessary?

hopkincame: Done

G-Rath:
@hopkincame just checking you've seen [this comment](https://github.com/google/osv-scanner/pull/2588/changes#r2886823180)?

PR #2591, opened 2026-03-05
ci: Create zizmor.yml github action

another-rex:


G-Rath:
Why wouldn't we run this on PRs that change workflow files? There's nothing dynamic to this right?

G-Rath:
we also don't have a `master` branch in this repo...

PR #2592, opened 2026-03-05
test: rename and reorder transitive tests

cuixq:
This PR reordered transitive scanning tests for pom.xml and requirements.txt.


cuixq:
I have made this PR to only rename and reorder the transitive tests since the underscores have been added in https://github.com/google/osv-scanner/pull/2605. PTAL!

another-rex (approved):

PR #2593, opened 2026-03-05
docs: improve snapshot update workflow

cuixq:
This PR improves the snapshot update workflow to ensure all snapshots are updated.
   * Updated `make update-snapshots`: changed the default behavior to run with `SHORT=false`. This ensures that snapshots for long-running tests are also updated.
   * Improved `CONTRIBUTING.md`: Clarified the difference between `make test SNAPS=true` and `make update-snapshots`.


Makefile
```diff
@@ -9,37 +9,46 @@ SNAPS ?= false
...
+help:
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} \
 		/^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2 } \
 		/^## / { printf "  %-20s %s\n", "", substr($$0, 4) }' $(MAKEFILE_LIST)
 
 ## Prevents make from trying to interpret the targets as files
 .PHONY: build scanner lint lint-fix format clean local-docs test update-snapshots refresh-all help
 
-build: ## Build scanner
+## Build scanner
```
another-rex: These changes will break the help output I'm pretty sure

another-rex (changes requested):

Makefile
cuixq: I have reverted the comments. PTAL.

PR #2597, opened 2026-03-05
refactor: remove unused `MarshalJSON`

G-Rath:


G-Rath:
that's so weird, coverage for some reason reports this function as _completely_ unused, but we do have tests of different sizes that clearly invoke it 🤔

cuixq:
I think this is due to implicit calls of the function: `MarshalJSON` is required for `Marshaler` interface and it may be called in functions like `json.Marshal(data)`

G-Rath:
I don't think that should be the reason, as otherwise all other (un)marshal functions would be marked as unused right?

G-Rath:
I think this might be related to how Go does coverage across packages? either way, I'll close this since we clearly are using the function

PR #2599, opened 2026-03-05
fix: load package deprecation enricher with the same config as other plugins

G-Rath:


pkg/osvscanner/osvscanner.go
```diff
@@ -285,17 +283,6 @@ func DoContainerScan(actions ScannerActions) (models.VulnerabilityResults, error
 		plugins = append(plugins, java.NewDefault())
 	}
```
cuixq: is it intentional that we do not append the package deprecation plugin here? or there is some context that I am missing here?

Ly-Joey: Since package deprecation enricher is being added to getPlugins() in [scan.go](https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/scan.go#L48) it should be all good! :D

PR #2612, opened 2026-03-06
refactor(imodels): use a global for purl cache

G-Rath:
Now that we're using functions rather tha methods, we should be able to make the "purl cache" global which will unblock axing our internal `imodels.PackageInfo` struct in favor of using `*extractor.Package` directory (and thus, be able to use `inventory.Inventory` directly too)

### DIFF
--- a/internal/imodels/imodels.go
+++ b/internal/imodels/imodels.go
@@ -4,6 +4,7 @@ package imodels
 import (
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/google/osv-scalibr/converter"
 	"github.com/google/osv-scalibr/extractor"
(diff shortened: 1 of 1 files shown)


internal/imodels/imodels.go
```diff
@@ -29,20 +30,52 @@ var gitExtractors = map[string]struct{}{
 	gitrepo.Name: {},
 }
 
+// todo: SBOM special case, to be removed after PURL to ESI conversion within each extractor is complete
+var cache = sync.Map{} // map[*extractor.Package]*models.PackageInfo
```
G-Rath, line 34: I've used a very generic name here since its internal and couldn't come up with a nice-but-short name 😅

PR #2630, opened 2026-03-09
perf: update `lipgloss` to improve start up performance for Windows Terminal

G-Rath:
I noticed that it was taking ~4 seconds for `osv-scanner --help` to resolve, which looks to be due to bug in `lipgloss` when running on Windows Terminal that was fixed via https://github.com/charmbracelet/lipgloss/pull/619

This would have gotten fixed automatically by the next minor versions update, but I figured I might as well do a dedicated PR for it since I noticed it 🤷


Ly-Joey (approved):

G-Rath:
fwiw I also noticed via `strace` that we seem to always be looking through a bunch of files even for `--help` such as `wl-copy`, `xclip`, and `xsel` - I'm pretty sure these are coming from `lipgloss` too because they also go away if I remove the `fix` command entirely.

Not the end of the world, but would be nice if we could at least have them deferred until they're actually needed

G-Rath:
@michaelkedar up for one last PR before landing #2413? 😅 

(actually looks like we do still end up directly importing `lipgloss` after that 🤷)

PR #2637, opened 2026-03-11
feat: Add configuration option to disable scanning Go version from go.mod

another-rex:
This commit addresses the issue where Go version warnings based on `go.mod` are misleading because the `go` directive specifies the minimum required language version, not the actual toolchain version. We introduce a new `ScanGoModVersion` configuration setting, which is disabled by default, ensuring we avoid assumptions regarding the toolchain version from `go.mod` files.

Snapshots and tests have been updated accordingly to reflect this change in the default behavior.

---
*PR created automatically by Jules for task [5355023063191685691](https://jules.google.com/task/5355023063191685691) started by @another-rex*


G-Rath:
Rather than introduce a new config property, what about reusing the existing one with a value of `ignore` or something?

We can use a custom marshaller to handle the type difference

another-rex:
I think the issue is that I want it to be disabled by default, and opt in, which makes the override field a bit odd to use. 

I guess we can have a term like "scan", but it's not really Overriding anything if the default is the version is not extracted.

G-Rath:
Ah yeah fair, if it's disabled by default that makes more sense - we could have it kind of like the version properties in things like gh actions e.g.

```
GoVersion: 'go.mod'
GoVersion: '.go-version'

GoVersion: 1.20
```

But thats probably not ergonomic enough to be worth it especially since it's off by default now 🤷‍♂️

another-rex:
@jules Rerun the snapshots building to resolve the conflict (make rebuild-all)

another-rex:
@jules please rebase this change onto the current branch and resolve any merge issues

another-rex:
Delaying this to the next release

Ly-Joey (approved):

PR #2641, opened 2026-03-12
fix: correctly output packages from osv-scanner.json source in spdx format

### LINKED ISSUE #2192: Scanning `osv-scanner.json` format does not correctly output spdx
Scanning `osv-scanner.json` format does not correctly output with the `--format=spdx` flag. 

Currently no packages will appear, and after #2177 it will show up with the incorrect purl type.

Mananshah237:
Fixes #2192

Packages loaded from an `osv-scanner.json` results file have `Inventory == nil`
because `PackageInfo.Inventory` is not populated during JSON deserialization.

`PrintSPDXResults` was blindly appending nil Inventory values, causing the
SPDX converter to silently skip all such packages.

This fix constructs a synthetic `extractor.Package` using the available
PackageInfo fields (name, version, ecosystem) when Inventory is nil.

Maven packages require special handling because their names are stored as
`groupId:artifactId` in osv-scanner. The scalibr PURL converter requires
`javalockfile.Metadata` with `GroupID` and `ArtifactID` to generate the correct
`pkg:maven/groupId/artifactId@version` PURL.

This ensures packages from osv-scanner.json sources appear correctly
in the generated SPDX SBOM output.

### DIFF
--- a/internal/output/spdx.go
+++ b/internal/output/spdx.go
@@ -3,9 +3,15 @@ package output
 import (
 	"encoding/json"
 	"io"
+	"strings"
 
 	scalibr "github.com/google/osv-scalibr"
 	"github.com/google/osv-scalibr/converter/spdx"
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/java/javalockfile"
+	"github.com/google/osv-scalibr/inventory/osvecosystem"
+	scalibr_purl "github.com/google/osv-scalibr/purl"
+	purlutil "github.com/google/osv-scanner/v2/internal/utility/purl"
 	"github.com/google/osv-scanner/v2/pkg/models"
 )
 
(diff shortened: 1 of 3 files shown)


internal/output/spdx.go
```diff
@@ -27,3 +39,38 @@ func PrintSPDXResults(vulnResult *models.VulnerabilityResults, outputWriter io.W
 
 	return encoder.Encode(doc)
 }
+
+// inventoryFromPackageInfo constructs a synthetic extractor.Package from a PackageInfo
+// for packages that don't have a populated Inventory field (e.g., loaded from osv-scanner.json).
+func inventoryFromPackageInfo(pkg models.PackageInfo) *extractor.Package {
```
another-rex: Have this function return an error as well, rather than just nil, to be handled in the caller.

another-rex (commented):
Nice, pretty neat fix!

Just need to fix the linting issues and one comment, otherwise LGTM.

another-rex (approved):

PR #2643, opened 2026-03-12
ci: allow dependencies workflow to record new cassette interactions

G-Rath:
I think it makes sense to do this for the same reason it made sense for the snapshots workflow, even though it is probably far less likely to be needed


G-Rath:
I don't know exactly why the cassette interactions have changed, but I think this would unblock https://github.com/google/osv-scanner/pull/2635

PR #2649, opened 2026-03-16
feat: allow unsafe plugins in scalibr capabilities

cuixq:
This PR updates `github.com/google/osv-scalibr` as well as enables `AllowUnsafePlugins` in `plugin.Capabilities` which is now required by the pomxml transitive dependency enricher.

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -5984,6 +5984,7 @@ No issues found
 Scanning dir <tempdir>/composer.lock
 Command "ssh -V": exit status 1
 Scanned <tempdir>/composer.lock file and found 1 package
+Warning: plugin transitivedependency/pomxml can be risky when run on untrusted artifacts. Please ensure you trust the source code and artifacts before proceeding.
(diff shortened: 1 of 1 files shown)


another-rex (commented):
Do we want a similar flag? It will be a breaking change though.

Should we log a warning if you are using transitive dependency scanning and rust call analysis for now.

cuixq:
as discussed offline, we probably delay having the new flag until next major version; for now, warning about the risky plugins in the logs.

pkg/osvscanner/scan.go
```diff
@@ -32,6 +32,19 @@ import (
...
 var ErrExtractorNotFound = errors.New("could not determine extractor suitable to this file")
 
+var unsafePlugins = []string{
+	"transitivedependency/pomxml",
+	"reachability/rust",
+}
+
+func logUnsafePlugins(plugins []plugin.Plugin) {
+	for _, plug := range plugins {
+		if slices.Contains(unsafePlugins, plug.Name()) {
```
another-rex: Rather than maintaining a list, can't we just check the plugin capabilities, and if it has need unsafe we can print the log?

cuixq: good point! done

PR #2650, opened 2026-03-16
ci: rename Dependencies workflow to Update osv-scalibr

cuixq:
It was not clear that this workflow is to update osv-scalibr specifically.


cuixq:
@another-rex @G-Rath how do you feel about renaming this workflow?

G-Rath:
I assume you mean it's not obvious when looking at the workflows in the actions tab?

When naming workflows and jobs I'm conscious that they get joined together when being displayed in relevant contexts (i.e. `checks / lint`), but in hindsight for these workflows it's probably better to name the workflow (and file, so that they are ordered together) something like `update-<thing>` and the job something like`run`

(By "they" I mean the scalibr, snapshots, and cassette workflows)

another-rex (approved):

ksealmalaysia41-lang (approved):

PR #2658, opened 2026-03-19
feat: support regex matching for package name overrides

### LINKED ISSUE #151: Exclude Packages by Regex Value from Scan

majiayu000:
Fixes #151

## Changes

Adds a `NameIsRegex` boolean field (`nameIsRegex` in TOML) to `PackageOverrideEntry`. When set to `true`, the `Name` field is treated as a regex pattern instead of an exact string match.

- Uses existing `cachedregexp.Compile()` for efficient cached compilation
- Anchors the regex with `^...$` to prevent partial matches
- Invalid regex patterns log a warning and do not match (no panic)
- Default behavior (`nameIsRegex = false`) is unchanged

Example config:
```toml
[[PackageOverrides]]
name = "internal-.*"
nameIsRegex = true
ignore = true
reason = "internal packages should not be sent to OSV API"
```

## Test Plan

Added `TestConfig_ShouldIgnorePackage_NameIsRegex` with 8 test cases:
- Regex match / no match
- Invalid regex handling
- Exact match preserved when `nameIsRegex = false`
- Empty name with regex flag
- Anchored matching (no partial match)
- Combined with ecosystem filter
- Regex match with ecosystem mismatch

```
go test ./internal/config/ -run TestConfig_ShouldIgnorePackage_NameIsRegex -v
```


internal/config/config.go
```diff
@@ -47,8 +49,20 @@ type PackageOverrideEntry struct {
 }
 
 func (e PackageOverrideEntry) matches(pkg *extractor.Package) bool {
-	if e.Name != "" && e.Name != imodels.Name(pkg) {
-		return false
+	if e.Name != "" {
+		if e.NameIsRegex {
+			re, err := cachedregexp.Compile("^" + e.Name + "$")
+			if err != nil {
+				cmdlogger.Warnf("warning: invalid regex %q in package override: %v\n", e.Name, err)
```
G-Rath: I'm pretty sure this'll generate a ton of warnings as it will happen for each package we extract

G-Rath (dismissed):
looks pretty good, though we should update the documentation to reflect the new property and add a CLI test for an invalid regex to showcase the output

majiayu000:
Thanks for the review! I've addressed the feedback:

1. **Moved regex validation to config load time** (`tryLoadConfig`) — invalid regex patterns are now caught once when the config is loaded, instead of warning per package in `matches()`. This means an invalid regex in a config file will produce a clean config loading error (same behavior as unknown keys).

2. **Updated documentation** — added `nameIsRegex` property with an example to `docs/configuration.md`.

3. **Added CLI test** — `config_file_with_invalid_regex` test case verifies the error output when a config has an invalid regex pattern.

internal/config/config.go
majiayu000: Done — moved regex validation to `tryLoadConfig()` so it's validated once at config load time. Invalid regex now produces a config loading error instead of a per-package warning. Added a unit test in `Test_tryLoadConfig` and a CLI snapshot test to verify the error output.

majiayu000: Done — moved regex validation to tryLoadConfig() so it's validated once at config load time. Invalid regex now produces a config loading error instead of a per-package warning. Added a unit test in Test_tryLoadConfig and a CLI snapshot test to verify the error output.

majiayu000: Good catch! This has been fixed — regex validation was moved to `tryLoadConfig()` so it happens once at config load time. An invalid regex now produces a config loading error (exit code 127) instead of per-package warnings. Added both a unit test in `Test_tryLoadConfig` and a CLI snapshot test (`config_file_with_invalid_regex`) to verify the error output.

majiayu000:
Hi @G-Rath, I've rebased onto the latest main and addressed the feedback:

1. **Regex validation moved to config load time** — invalid regex patterns are caught once in `tryLoadConfig()` instead of per-package warnings
2. **Documentation updated** — added `nameIsRegex` property with example to `docs/configuration.md`
3. **CLI test added** — `config_file_with_invalid_regex` test case verifies the error output
4. **Fixed errorlint** — `%v` → `%w` in the regex validation error

Could you please re-review when you get a chance? Thanks!

internal/config/config.go
majiayu000: Good catch! This has been fixed — regex validation was moved to `tryLoadConfig()` so it happens once at config load time. An invalid regex now produces a config loading error (exit code 127) instead of per-package warnings. Added both a unit test in `Test_tryLoadConfig` and a CLI snapshot test (`config_file_with_invalid_regex`) to verify the error output.

majiayu000: fixed — regex validation moved to config load time, docs updated with nameIsRegex in the field reference section too, and CLI snapshot test added for the invalid regex case.

another-rex (approved):
Sorry for the delay! I had another look and this looks good to me. Not sure about the name though, but I think it's good enough.

PR #2666, opened 2026-03-20
Add a docker variant of the pre-commit hook

pcastellazzi:
Updated .pre-commit-hooks.yaml to include a docker variant which avoids downloading and compiling when docker is available.

This change requires an updated to .pre-commit-hooks.yaml when a new release of the container image is made. Can we do that?


another-rex (approved):
Nice, LGTM

PR #2668, opened 2026-03-21
fix: prevent path traversal in ParentPOMPath via relativePath validation

johanr-8:
## Summary

This PR fixes a path traversal vulnerability in `ParentPOMPath` where 
a malicious `<relativePath>` value in a `pom.xml` file could cause 
`osv-scanner fix` to write generated content outside the project directory.

## Root Cause

`filepath.Join` resolves `../` sequences, allowing a `<relativePath>` 
like `../../../../tmp/malicious` to escape the project root with no 
validation.

## Fix

Added a containment check after path construction to ensure the resolved 
path stays within the project directory:
```go
if !strings.HasPrefix(path, filepath.Clean(projectRoot)+string(os.PathSeparator)) {
    return ""
}
```

## Testing

Verified that:
- Traversal paths like `../../../../tmp/malicious` are blocked
- Normal relative paths like `../pom.xml` continue to work
- Default behavior (empty relativePath) is preserved

Fixes security vulnerability reported via Google OSS VRP.

### DIFF
--- a/internal/utility/maven/maven.go
+++ b/internal/utility/maven/maven.go
@@ -127,15 +127,17 @@ func ParentPOMPath(currentPath, relativePath string) string {
 	if relativePath == "" {
 		relativePath = "../pom.xml"
 	}
-	projectRoot := filepath.Dir(currentPath)
-	path := filepath.Join(projectRoot, relativePath)
-	path = filepath.Clean(path)
 
-	// Prevent path traversal outside the project directory.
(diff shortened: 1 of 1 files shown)


cuixq:
@johanr-8 thank you for your contribution, however can you double confirm this is the change that you intend to make?
it seems the code does not align with what is said in the description.

also, we are migrating our `fix` command logic to [osv-scalibr](https://github.com/google/osv-scalibr/blob/main/guidedremediation/guidedremediation.go). we will soon deprecate the code in osv-scanner, so please make your contribution with osv-scalibr.

PR #2676, opened 2026-03-23
fix(maven): constrain local parent POM resolution to the project boundary

Omar-Khaleel:
This PR fixes an out-of-scope local parent POM resolution issue in Maven manifest parsing.

Previously, `ParentPOMPath` allowed `<relativePath>` values to resolve to locations outside the intended repository/project bounds.

To prevent this without breaking legitimate multi-module Maven setups (where the parent POM naturally sits at `../pom.xml`), this patch introduces `mavenutil.FindProjectRoot()`. This function climbs the directory tree looking for a `.git` repository. If found, it uses that as the highest allowable root. Otherwise, it falls back to the directory of the initial manifest. The computed `rootPath` is then passed to `ParentPOMPath` and checked via `mavenutil.IsWithinRoot` to ensure the resolved POM stays within the project boundary when present.

`IsWithinRoot` was tightened to use strict `filepath.Rel` component validation rather than simple string prefix matching.

Relevant unit tests were added to ensure:
- Sibling/child/parent POM paths resolve normally within the selected boundary.
- Unsafe out-of-scope traversals are blocked correctly.
- Production path tests utilizing `MavenReadWriter.Read` confirm out-of-scope files are successfully skipped.


cuixq:
@Omar-Khaleel thank you for your contribution.

We are migrating our Maven util package to [osv-scalibr](https://github.com/google/osv-scalibr/tree/main/internal/mavenutil), and we will soon deprecate the code in osv-scanner, please make your contribution with osv-scalibr.

Omar-Khaleel:
Thanks for the guidance. I’ll port the change to osv-scalibr.

PR #2680, opened 2026-03-24
chore: release v2.3.5

tobyhawker:
Changelog for v2.3.5: abandoning previous release and fixing workflow.


another-rex:
I think you need to update osv-scanner version and do a snapshot update run. No need to rebuild cassettes iirc.

PR #1730, opened 2025-03-23
feat: Add support for more .NET extrators

AlexLaroche:
This PR adds more lockfile extractors supported in OSV-Scalibr for .NET:
 - `packages.config`
 - `packages.lock.json`

PR #2715, opened 2026-04-11
Require explicit acknowledgement for risky Rust call analysis

sumitshahorg:
### Summary

Introduce a safety guard for Rust call analysis to prevent unintended execution of untrusted code during scanning, explicitly addressing a Remote Code Execution (RCE) risk.

### Problem

When `--call-analysis=rust` is enabled, the scanner invokes `cargo build`, which executes `build.rs` scripts. These scripts can contain arbitrary code and are executed automatically during the build process.

This creates a Remote Code Execution (RCE) risk where scanning an untrusted project can lead to arbitrary code execution on the host system without explicit user awareness.

### Solution

* Add a required flag: `--allow-risky-rust-call-analysis`
* Fail fast if Rust call analysis is requested without explicit acknowledgement
* Add documentation warning in README about execution of `build.rs`

### Behavior Change

Before:

```
--call-analysis=rust
→ cargo build executes automatically
→ build.rs may execute arbitrary code (RCE risk)
```

After:

```
--call-analysis=rust
→ fails unless --allow-risky-rust-call-analysis is provided
→ explicit user acknowledgement required
```

### Security Impact

* Prevents silent Remote Code Execution (RCE) via `build.rs`
* Eliminates implicit execution of untrusted code during scanning
* Introduces explicit user consent for risky operations
* Aligns with secure-by-default and least-astonishment principles

### Notes

This change does not alter functionality when explicitly enabled, but enforces a secure default by requiring user acknowledgement before executing potentially unsafe build steps.


sumitshahorg (commented):
Added tests validating the Rust call-analysis safety guard.

Ly-Joey:
Hi @sumitshahorg, thank you for your interest in contributing!

We have already documented the potential risks of running rust call analysis on untrusted project in our [documentation](https://google.github.io/osv-scanner/usage/scan-source#call-analysis-in-rust) and the `--call-analysis` flag [usage description](https://github.com/google/osv-scanner/blob/10db0bb14b144e00af505979d89c65754b3151af/cmd/osv-scanner/internal/helper/flags.go#L166).
Given rust call analysis is still an experimental feature, we currently don't see a need to add the warning to our `README.md`.

Additionally, your idea of requiring an explicit allow-flag is implemented in osv-scalibr [#1730](https://github.com/google/osv-scalibr/pull/1730). However, because adding the same flag to osv-scanner will be a breaking change, we will only be able to include it in the osv-scanner v3 release.

Thanks again for raising the concerns and PR! Please be assured that we are aware of this risk and are working towards stronger guardrails for risky plugins.

Issue #2675, opened 2026-03-23
Migrate to use OSV-Scalibr's data source clients


cuixq:
There are clients (packages) that we now have the code in OSV-Scalibr:
 - [data source](https://github.com/google/osv-scanner/tree/main/internal/datasource) 
 - [depsdev](https://github.com/google/osv-scanner/tree/main/internal/depsdev) 

We should migrate the current usage of these libraries to depend on osv-scalibr and then remove the code.

We may also need to migrate `update` command to use `func Update()` in [guidedremediation](https://github.com/google/osv-scalibr/blob/main/guidedremediation/guidedremediation.go#L185) to unblock code cleanup too.

PR #2719, opened 2026-04-13
refactor: use guidedremediation from osv-scalibr for version updates

cuixq:
https://github.com/google/osv-scanner/issues/2675

This PR refactors the update command to use the guidedremediation package from github.com/google/osv-scalibr to handle dependency updates.
   * Migrated to osv-scalibr: Replaced internal remediation/suggest and upgrade logic with guidedremediation.Update.
   * Updated Snapshots: Updated the test snapshots to reflect the slightly adjusted output messages and the behavior of the suggester which prefers stable package versions.


michaelkedar (approved):
nice!

PR #2726, opened 2026-04-16
chore(deps): update github.com/urfave/cli to 3.8.0

cuixq:
Unblocks https://github.com/google/osv-scanner/pull/2674

For `github.com/urfave/cli/v3` [version 3.8.0](https://github.com/urfave/cli/releases/tag/v3.8.0), it seems an empty flag is no longer treated as an error but instead an empty flag will be set, see https://github.com/urfave/cli/pull/2297.

Thus this PR updated the expected exit codes and error messages for test cases involving empty flags.


cmd/osv-scanner/scan/source/command_test.go
```diff
@@ -489,9 +489,9 @@ func TestCommand_ExplicitExtractors_WithDefaults(t *testing.T) {
 
 	tests := []testcmd.Case{
 		{
-			Name: "empty_plugins_flag_does_nothing",
+			Name: "empty_plugins_flag_does_default",
```
G-Rath, line 492: I don't think this is correct - the value being empty should not cause it to fallback to the default

what I would expect to happen is `plugins` ends up with an empty string, which we should consider an error

G-Rath:
relating to my inline comment, could you check if this changes what happens with `--config=`? if it means we're using the default instead of erroring (one way or another), then frankly I'd say the upstream is wrong in their behaviour because it feels dangerous to fallback to the default.

overall this change has some weird potential implications, so I think we should dig a bit deeper to confirm what the actual result is

another-rex:
I'm not sure that's what happens though? An empty plugins string with out `experimental-no-default-plugins` should just use the default plugins right?

cmd/osv-scanner/scan/source/command_test.go
another-rex: There's a separate test with --no-default that does what you describe, and stays that way.

another-rex:
Pushed similar changes to https://github.com/google/osv-scanner/pull/2674 directly.

PR #2669, opened 2026-03-22
  output: sanitize \r in gh-annotations to prevent GitHub Actions workflow command injection

djvirus9:
## Summary
                                                                                                          
  `--format=gh-annotations` sanitized `\n` → `%0A` to prevent annotation line breaks, but did not sanitize `\r`. GitHub Actions runners treat `\r` as a line boundary when parsing workflow commands, so a carriage return embedded in a package name or file path causes the annotation to be split — with anything after `\r` executed as a new workflow command.                                            
                                                                                                          
  ## Impact                                                                                               
                                                                                                          
  An attacker who introduces a dependency with `\r` in its package metadata (via a public registry or a crafted PR) can inject commands including  `::warning::`, `::add-mask::`, `::set-output::`, and `::stop-commands::` into any CI pipeline running `osv-scanner --format=gh-annotations`.                                     
                                                                                                          
  ## Reproduction                                                                                         
                                                                                                          
  ```bash                                                                                               
  MALICIOUS=$(printf 'scan\r::warning::INJECTED_BY_ATTACKER')
  mkdir -p "$MALICIOUS" && cp package-lock.json "$MALICIOUS/"                                             
  osv-scanner --format=gh-annotations "$MALICIOUS/package-lock.json" | cat -v                             
  # ^M visible in output — on GitHub Actions runner this fires ::warning::INJECTED_BY_ATTACKER            
                                                                                                          
  Live confirmation: https://github.com/djvirus9/osv-injection-poc/actions                                
                                                                                                        
  Fix

  Sanitize \r → %0D in renderedTable and artifactPath, mirroring
  the existing \n → %0A sanitization.

Issue #2749, opened 2026-04-27
Incomplete fix for #2669 / c2c9aae — raw `\r` in default/table/vertical output enables GitHub Actions workflow-command injection [via vrp team followup]


hits313:
## Summary

PR #2669 (commit c2c9aae) encoded `\r` as `%0D` for the `--format=gh-annotations` output path to fix GitHub Actions workflow-command injection. **The fix is incomplete:** three other code paths still emit raw `\r` bytes to stdout when a scanned source path contains `\r`.

Because the GitHub Actions runner treats `\r` as a line terminator while parsing `::command::value` sequences from stdout, an attacker who can introduce a directory name containing `\r::stop-commands::TOKEN\r` into a CI run can:

- Inject arbitrary workflow commands (`::error::`, `::warning::`, `::add-mask::`, `::stop-commands::`).
- Silently disable all subsequent command parsing via `::stop-commands::`, causing every later `::error::CVE-XXXX` annotation that osv-scanner itself emits to be dropped — the PR status check turns green and the maintainer merges believing the dependency change is clean.

This effectively defeats osv-scanner's value as a CI gate while the tool appears to run normally.

> **Severity note:** this primitive silently bypasses osv-scanner's CI gate — vulnerable-dependency PRs can be merged with a green check. The trigger does **not** require maintainer approval — Sink 1 fires on the scan invocation itself, so a first-time external PR contributor can exploit this on any public repository using `osv-scanner-action` as a required check.

Confirmed against **v2.3.5 (b4e672e, post-patch)** by `od -c` of stdout  raw `\r` bytes are present.

Coordinated disclosure via Google OSS VRP; triage requested public upstream coordination, hence this issue.

---

## Affected sinks (all post-c2c9aae)

All three paths interpolate a user-controlled source/directory path into a string that is written to stdout without `\r` / `\n` encoding.

### Sink 1 — `pkg/osvscanner/scan.go:153`

```go
cmdlogger.Infof("Scanning dir %s", path)   // `path` interpolated verbatim
```

This is emitted on every scan, **before any output format is selected** — so it fires regardless of whether the user passes `--format=gh-annotations` or not. `Info` level routes to stdout via `internal/cmdlogger/handler.go:79`:

```go
_, err := fmt.Fprint(c.writer(record.Level), record.Message+"\n")
//                                            ^ no \r encoding
```

### Sink 2 — `internal/output/vertical.go:158`

```go
[...]

PR #2750, opened 2026-04-27
fix: sanitize \r/\n in default/table/vertical output to prevent GitHub Actions workflow command injection

hits313:
# fix: sanitize `\r`/`\n` in default/table/vertical output to prevent GitHub Actions workflow command injection

Fixes #2669  and proposed in #2749

## Summary

PR #2669 (commit `c2c9aae`) URL-encoded `\r` as `%0D` for the `--format=gh-annotations` output path to fix GitHub Actions workflow command injection. The same class of bug still exists in **four** other code paths that also write user-controlled file paths / image references to `stdout`:

| # | Sink | User-controlled value |
|---|------|------------------------|
| 1 | `pkg/osvscanner/scan.go` (`Scanning dir <path>`) | `path` |
| 2 | `pkg/osvscanner/osvscanner.go` (`Scanning image <ref>`) | `actions.Image` |
| 3 | `pkg/osvscanner/stats.go` (`Scanned <path> file and found N packages`) | `systemPath` |
| 4 | `internal/output/vertical.go` (per-source header) | `result.Name` |
| 5 | `internal/output/table.go` (per-source table title) | `source.Name` |

Sinks 1 and 3 are emitted on **every scan**, regardless of `--format`, so they fire even for `--format=table`, `--format=vertical`, and the default output.

The GitHub Actions runner treats both `\r` and `\n` as line boundaries when scanning a step's stdout for `::command::value` sequences. A directory or package name containing `\r::stop-commands::TOK\r` silently disables all subsequent workflow-command parsing — every `::error::CVE-…` annotation `osv-scanner` itself emits afterwards is dropped, the PR check turns green, and the maintainer merges believing the dependency change is clean.

## Impact

Any contributor who can open a pull request against a repository that runs `osv-scanner` in CI can exploit this — no special permissions required. `git` does not sanitize `\r` from filenames, and `\r` is a valid filename byte on Linux, so a malicious PR can simply rename a directory.

A typical exploit:

1. Attacker opens a PR that renames a directory to `legit\r::stop-commands::TOK\r` and bumps a dependency to a version with a known CVE.
2. CI runs `osv-scanner`. Sinks 1 and 3 emit the `\r`-laden directory name to stdout regardless of `--format`.
3. The GHA runner parses `::stop-commands::TOK` as a live directive and stops processing all further workflow commands.
4. Every subsequent `::error::CVE-…` annotation `osv-scanner` emits is dropped. The vulnerability gate reports green.
5. Reviewer sees clean check, merges the vulnerable dependency.

The official `google/osv-scanner-action` is used as a required status check by hundreds of open-source projects.

This was originally reported via Google OSS VRP (issue tracker ID `505494979`); triage asked for the fix to be coordinated upstream publicly. Tracking issue: #2749.

## Reproduction (pre-patch, against `b4e672e`)

```bash
git clone --depth 50 https://github.com/google/osv-scanner.git /tmp/osv-src
cd /tmp/osv-src && go build -o /tmp/osv-bin ./cmd/osv-scanner/

DIR="$(printf '/tmp/poc/dir\r::stop-commands::T\r::error::INJECTED\r::warning::pwn')"
mkdir -p "$DIR"
cat > "$DIR/package-lock.json" <<'JSON'
{
  "name": "x", "version": "1.0.0", "lockfileVersion": 3, "requires": true,
  "packages": {
    "": { "name": "x", "version": "1.0.0", "dependencies": { "lodash": "4.17.20" } },
    "node_modules/lodash": {
      "version": "4.17.20",
      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.20.tgz",
      "integrity": "sha512-PlhdFcillOINfeV7Ni6oF1TAEayyZBoZ8bcshTHqOYJYlrqzRK5hagpagky5o4HfCzzd1TRkXPMFq6cKk9rGmA==",
      "license": "MIT"
    }
  }
}
JSON

/tmp/osv-bin --offline-vulnerabilities --download-offline-databases \
  "$DIR/package-lock.json" 2>/dev/null | tr '\r' '\n' | grep '^::'
```

**Pre-patch output:**

```
::stop-commands::T
::error::INJECTED
::warning::pwn
```

End-to-end PoC running the official binary inside a real GitHub Actions workflow:
https://github.com/hits313/osv-crlf-poc/actions/runs/24805983517

## Fix

Adds a small helper:

```go
// internal/output/sanitize.go
var workflowCommandReplacer = strings.NewReplacer("\r", "%0D", "\n", "%0A")

func SanitizeForWorkflowCommand(s string) string {
    return workflowCommandReplacer.Replace(s)
}
```

…and applies it at each of the five sinks listed above. The helper mirrors the encoding used by PR #2669 (`\r` → `%0D`, `\n` → `%0A`). A source path or package name containing `\r`/`\n` is never legitimate, so the helper is applied unconditionally rather than gating on `GITHUB_ACTIONS=true`.

I considered three alternatives and rejected each:

- **Sanitize at the `cmdlogger.Handler` writer** — would catch sinks 1–3 plus future log lines for free, but does *not* help sinks 4–5 (which write through `outputWriter` directly, not via `cmdlogger`), and would alter local-terminal output for legitimate multi-line error messages.
- **Sink-level inline `strings.ReplaceAll`** (matching #2669 exactly) — works, but duplicates the same two-call pattern at five sites and makes future regressions easy.
- **Gate on `GITHUB_ACTIONS=true`** — adds env-dependent behavior to output for no UX gain (no legitimate use of `\r`/`\n` in these strings exists outside CI either).

## Verification

Post-patch byte-level inspection of stdout for the same PoC:

```
$ /tmp/osv-bin-fixed --offline-vulnerabilities --download-offline-databases \
    "$DIR/package-lock.json" 2>/dev/null > /tmp/out-fixed.txt
$ tr -cd '\r' < /tmp/out-fixed.txt | wc -c
0
$ tr '\r' '\n' < /tmp/out-fixed.txt | grep '^::'
(no output — every injected command is now %0D-prefixed and inert)
$ head -c 200 /tmp/out-fixed.txt
Scanning dir /tmp/poc-fix/dir%0D::stop-commands::T%0D::error::INJECTED%0D::warning::pwn/package-lock.json
Starting filesystem walk for root: /
Scanned /tmp/poc-fix/dir%0D::stop-commands::T%0D::error::INJE...
```

Zero raw `\r` bytes in stdout, six `%0D` encodings (three from `Scanning dir`, three from `Scanned <path> file`).

Test results:

```
$ go test ./internal/output/... ./internal/cmdlogger/...
ok  github.com/google/osv-scanner/v2/internal/output   0.088s
ok  github.com/google/osv-scanner/v2/internal/cmdlogger 0.002s

$ go vet ./...
(clean)

$ go build ./...
(clean)
```

## Files changed

```
 internal/output/sanitize.go      | 29 +++++++++++++++++  (new)
 internal/output/sanitize_test.go | 69 ++++++++++++++++++ (new)
 internal/output/table.go         |  4 ++-
 internal/output/vertical.go      |  5 ++-
 pkg/osvscanner/osvscanner.go     |  4 ++-
 pkg/osvscanner/scan.go           |  8 ++++-
 pkg/osvscanner/stats.go          |  5 ++-
 7 files changed, 119 insertions(+), 5 deletions(-)
```

## Notes for reviewers

- The exported `SanitizeForWorkflowCommand` lives in `internal/output` because the existing `--format=gh-annotations` sanitizer also lives there; future consolidation of `internal/output/githubannotation.go`'s inline `ReplaceAll` calls onto this helper is a natural follow-up but kept out of scope to keep the security fix small.
- `cmd/osv-scanner/mcp/stats.go` has a structurally similar log line, but its output flows into MCP tool responses (LLM-context), not GHA stdout. Different threat model, intentionally out of scope.
- Happy to add snapshot tests against `--format=table` / `--format=vertical` if reviewers prefer end-to-end coverage in addition to the unit tests on the helper.

## Original report

Originally reported via Google OSS VRP, issue tracker ID `505494979`. Triage requested public coordination; this PR is the upstream fix.


internal/output/sanitize.go
```diff
@@ -0,0 +1,29 @@
...
+
+// SanitizeForWorkflowCommand returns s with carriage-return and newline bytes
+// replaced by their URL-encoded forms (%0D, %0A) so that an attacker-controlled
+// value cannot break out of its line and inject GitHub Actions workflow
+// commands such as ::error::, ::warning::, ::add-mask::, or ::stop-commands::.
+//
+// A source path or package name containing \r or \n is never legitimate, so
+// this is applied unconditionally regardless of whether the caller is running
+// under GitHub Actions.
+func SanitizeForWorkflowCommand(s string) string {
```
G-Rath, line 15: ~this doesn't need to be public as it's an implementation detail and~ shouldn't be tested directly

G-Rath (changes requested):
I think it would be good if you could rewrite the comments to avoid referencing PRs as those can be seen via the git history - instead we should just talk about the "here and now"

hits313:
Thanks for the review — addressed in 2e08a81:

- Rewrote the sink comments to describe behavior in the here-and-now (no PR/commit references; that history lives in git).
- Unexported the helper to `sanitizeForWorkflowCommand` and removed the direct unit test.
- Added a sibling private helper in `pkg/osvscanner` so the cross-package call sites don't need an exported API.

Behavior is unchanged; verified the exploit input still produces `%0D`-encoded output with zero raw `\r` bytes. Happy to add sink-level tests if you'd prefer behavioral coverage on the affected paths.

CLA signing in progress — will follow up with `@googlebot I signed it!` so the check re-runs.

G-Rath (dismissed):
sorry I missed that this was being used outside the `output` package, so it is fine to export the helper for that - but you should be adding tests for each output format.

They use [helpers](https://github.com/google/osv-scanner/blob/main/internal/output/helpers_test.go) so you should generally just need to add a couple of test cases, and then all output formats will be tested

hits313:
Thanks @G-Rath — pushed 708b8cc with the helper re-exported, format-level coverage via the shared fixture, and a couple more sinks (vertical totals line, table Source column, HTML template) that the new test surfaced. Verified end-to-end on a real GHA runner here: https://github.com/hits313/osv-crlf-poc/actions/runs/25016802187  zero raw `\r`, zero injectable commands.

Could you or someone on the team take another look when you get a chance? Happy to iterate quickly on anything else you'd like changed.

internal/output/html/package_view_template.gohtml
```diff
@@ -4,7 +4,7 @@
   <div class="ecosystem-sources-container">
     {{ range .Sources }}
     <div class="source-container">
-      <h3 class="source-heading">Source: <span class="source-path">{{ .Name }}</span></h3>
+      <h3 class="source-heading">Source: <span class="source-path">{{ .Name | sanitizeWorkflowCmd }}</span></h3>
```
another-rex: Is this change strictly necessary? Unless I'm misrememgbering I don't think the html output is shown in the github actions terminal?

another-rex (approved):
Thanks for identifying this. I think the current implmenetation after the fixes looks good, though I have a question.

internal/output/html/package_view_template.gohtml
hits313: Good catch @another-rex . HTML output goes to a file, not the GHA workflow command parser, so \r/\n in the source path can't escalate from there. Reverting this sink. The remaining sinks all touch stdout in CI runs, where the primitive applies.

hits313:
@another-rex  I've reverted that sink (removed the sanitizeWorkflowCmd pipe from the template and the funcmap entry) and squashed the branch into a single clean    commit. The five stdout sinks remain sanitized. thanks

another-rex (approved):

another-rex (approved):

another-rex:
Thank you for the clear report and PR! Released now in https://github.com/google/osv-scanner/releases/tag/v2.3.6

G-Rath:
For any future readers: we've discussed this internally and decided this kind of issue is not a concern for the scanner as

PR #2751, opened 2026-04-28
Sanitize GitHub Actions line terminators in human output

### LINKED ISSUE #2749: Incomplete fix for #2669 / c2c9aae — raw `\r` in default/table/vertical output enables GitHub Actions workflow-command injection [via vrp team followup]

saisathwikpeddireddy:
## Summary
- add a shared helper for escaping line terminators that GitHub Actions treats as workflow command boundaries
- sanitize cmdlogger messages before writing to stdout/stderr
- sanitize source names in table and vertical human output paths
- reuse the shared helper in GitHub annotation output and add regression coverage

Fixes #2749

## Testing
- GOCACHE=/tmp/go-build-small GOPATH=/tmp/go go test ./internal/utility/terminal ./internal/cmdlogger
- Attempted focused ./internal/output regression tests, but the local machine ran out of disk space while downloading/unzipping the package dependency graph.

### DIFF
--- a/internal/cmdlogger/handler.go
+++ b/internal/cmdlogger/handler.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"log/slog"
 	"strings"
+
+	"github.com/google/osv-scanner/v2/internal/utility/terminal"
 )
 
 var (
@@ -76,7 +78,7 @@ func (c *Handler) Handle(ctx context.Context, record slog.Record) error {
 		return c.overrideHandler.Handle(ctx, record)
 	}
 
-	_, err := fmt.Fprint(c.writer(record.Level), record.Message+"\n")
+	_, err := fmt.Fprint(c.writer(record.Level), terminal.EscapeGitHubActionsCommandChars(record.Message)+"\n")
 
 	return err
 }
--- a/internal/cmdlogger/handler_test.go
+++ b/internal/cmdlogger/handler_test.go
@@ -1,0 +1,38 @@
+package cmdlogger
+
+import (
+	"bytes"
+	"context"
+	"log/slog"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestHandlerEscapesGitHubActionsCommandChars(t *testing.T) {
+	t.Parallel()
+
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	handler := New(stdout, stderr)
+
+	err := handler.Handle(context.Background(), slog.NewRecord(
+		time.Time{},
+		slog.LevelInfo,
+		"Scanning dir safe\r::warning::pwn\nnext",
+		0,
+	))
+	if err != nil {
+		t.Fatalf("Handle() error = %v", err)
+	}
+
+	got := stdout.String()
+	if strings.ContainsAny(got, "\r") {
+		t.Fatalf("log output contains raw carriage return: %q", got)
+	}
+
+	want := "Scanning dir safe%0D::warning::pwn%0Anext\n"
+	if got != want {
+		t.Fatalf("log output = %q, want %q", got, want)
+	}
+}
--- a/internal/output/githubannotation.go
+++ b/internal/output/githubannotation.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/google/osv-scanner/v2/internal/utility/terminal"
 	"github.com/google/osv-scanner/v2/pkg/models"
 	"github.com/jedib0t/go-pretty/v6/table"
 )
@@ -82,8 +83,7 @@ func PrintGHAnnotationReport(vulnResult *models.VulnerabilityResults, outputWrit
 		// Sanitize artifactPath to prevent GitHub Actions workflow command injection.
 		// \r and \n in the file= parameter can terminate the annotation early and inject
 		// arbitrary workflow commands (e.g. ::warning::, ::add-mask::) into the runner output.
-		artifactPath = strings.ReplaceAll(artifactPath, "\r", "%0D")
-		artifactPath = strings.ReplaceAll(artifactPath, "\n", "%0A")
+		artifactPath = terminal.EscapeGitHubActionsCommandChars(artifactPath)
 
 		remediationTable, hasVulnTable := createSourceRemediationTable(source, groupedFixedVersions)
 		if hasVulnTable {
@@ -92,7 +92,7 @@ func PrintGHAnnotationReport(vulnResult *models.VulnerabilityResults, outputWrit
 			// so we URL encode the new line character
 			renderedTable = strings.ReplaceAll(renderedTable, "\n", "%0A")
 			// Sanitize \r to prevent workflow command injection via carriage return in package names
-			renderedTable = strings.ReplaceAll(renderedTable, "\r", "%0D")
+			renderedTable = terminal.EscapeGitHubActionsCommandChars(renderedTable)
 
 			// Prepend the table with a new line to look nicer in the output
 			fmt.Fprintf(outputWriter, "::error file=%s::%s%s", artifactPath, artifactPath, "%0A"+renderedTable)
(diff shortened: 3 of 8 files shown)


another-rex:
Duplicate of #2750

PR #2761, opened 2026-05-03
fix(filter): remove guard that silently drops unignored vulns when all groups are ignored

### LINKED ISSUE #2756: filterPackageVulns: orphan vulnerabilities silently dropped when all groups are ignored
## Summary

## Root Cause

```go
        }
    }
}
```

## Reproduction (pseudo-code)

```
Input:

After group filtering:

Vulnerability loop:

```

## Fix

```go
    }
}
```

---

nileshpatil6:
Fixes #2756

## Problem

`filterPackageVulns` in `pkg/osvscanner/filter.go` has a guard:

```go
        }
    }
}
```

The comment treats an assumption as an invariant. When all groups happen to be ignored, `len(newGroups) == 0` causes the vulnerability loop to be skipped entirely. Any vulnerability whose ID is **not** in `ignoredVulns` (i.e. not referenced by any group's `Aliases`) gets silently dropped instead of being reported.

Concrete case:

```

After group filtering:
  newGroups    = []
  ignoredVulns = {"CVE-2024-001": {}}

len(newGroups) > 0 → false → loop skipped

Result:   Vulnerabilities = []              (wrong)
Expected: Vulnerabilities = [CVE-2024-999] (CVE-2024-999 was never ignored)
```

This is a real security concern: a vulnerability can be silently hidden if an unrelated vulnerability in the same package happens to be the only group member and is ignored.

## Fix

Remove the guard. The `ignoredVulns` map already handles all filtering correctly on its own -- it is populated with every alias of every ignored group, so the inner check `if _, filtered := ignoredVulns[vuln.GetId()]; !filtered` correctly retains only vulnerabilities that were not part of any ignored group.

```go
    }
}
```

### DIFF
--- a/pkg/osvscanner/filter.go
+++ b/pkg/osvscanner/filter.go
@@ -42,6 +42,16 @@ func filterUnscannablePackages(scanResults *results.ScanResults, actions Scanner
 			}
 
 			continue
+		// Short commit hashes (< 40 hex chars) are rejected by the OSV API with
+		// "Invalid hash". Skip them with a warning rather than aborting the scan.
+		case imodels.Commit(psr) != "" && len(imodels.Commit(psr)) < 40:
+			cmdlogger.Warnf("Skipping %s: short commit hash %q cannot be queried; OSV API requires a full 40-character SHA.", imodels.Name(psr), imodels.Commit(psr))
+
+			if actions.ShowAllPackages {
+				filteredPsr = append(filteredPsr, psr)
+			}
+
+			continue
 		}
 
 		packageResults = append(packageResults, psr)
--- a/pkg/osvscanner/filter.go
+++ b/pkg/osvscanner/filter.go
@@ -172,11 +172,9 @@ func filterPackageVulns(pkgVulns models.PackageVulns, configToUse config.Config)
 	}
 
 	var newVulns []*osvschema.Vulnerability
-	if len(newGroups) > 0 { // If there are no groups left then there would be no vulnerabilities.
-		for _, vuln := range pkgVulns.Vulnerabilities {
-			if _, filtered := ignoredVulns[vuln.GetId()]; !filtered {
-				newVulns = append(newVulns, vuln)
-			}
+	for _, vuln := range pkgVulns.Vulnerabilities {
+		if _, filtered := ignoredVulns[vuln.GetId()]; !filtered {
+			newVulns = append(newVulns, vuln)
 		}
 	}
 
--- a/pkg/osvscanner/filter_internal_test.go
+++ b/pkg/osvscanner/filter_internal_test.go
@@ -7,7 +7,50 @@ import (
 	"github.com/google/osv-scanner/v2/internal/config"
 	"github.com/google/osv-scanner/v2/internal/testutility"
 	"github.com/google/osv-scanner/v2/pkg/models"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
 )
+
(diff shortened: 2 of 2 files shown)


G-Rath (changes requested):
You need to add tests for this change

nileshpatil6:
Thanks for the review @G-Rath! Added `Test_filterPackageVulns_orphanVulnKeptWhenAllGroupsIgnored` in `filter_internal_test.go` (commit `90720b80`) that covers the exact scenario described in the PR: CVE-2024-5678 is an orphan vuln (not in any group's aliases), CVE-2023-1234 is the only group member and is ignored -- the test verifies the orphan is still returned after filtering.

G-Rath:
Do you have an actual real world reproduction for this?

nileshpatil6:
Good question. The orphan scenario arises from a subtle mismatch between how `ignoredVulns` is built and what `Vulnerabilities` can contain.

`ignoredVulns` is populated from the **aliases of ignored groups only**. But `Vulnerabilities` contains **all raw vuln objects returned from the OSV API** for a package -- including any whose IDs were not aliased by any group.

Concrete real-world path:

1. OSV API returns vulns `[CVE-A, CVE-B]` for a package
2. The grouping logic creates one group: `G1 { Aliases: ["CVE-A"] }` -- `CVE-B` ends up with no group (e.g. because it's a newer advisory not yet deduplicated/aliased against `CVE-A`)
3. User ignores `CVE-A` in config
4. Result: `ignoredVulns = {"CVE-A"}`, `newGroups = []`
5. Guard fires: `len(newGroups) == 0` → vulnerability loop skipped entirely → `CVE-B` silently dropped

The comment in the original code -- `"If there are no groups left then there would be no vulnerabilities"` -- treats a grouping invariant as a guarantee, but the OSV API doesn't guarantee every vuln ID appears in at least one group's aliases. That invariant can break whenever a new advisory is published that aliases an existing CVE but the scanner receives it before the OSV dedup pipeline runs.

The fix is also just correct on its own terms: `ignoredVulns` already encodes exactly the set of vuln IDs that should be dropped. The outer `newGroups` guard adds nothing correct and actively loses data.

G-Rath:
No, do you have an actual real world example of this occuring? I suspect that this isn't possible, hence the comment in the code, and the discussion happening on the backing issue.

If we're to land this we'll want a CLI based test that showcases a real world big, not just a unit test

nileshpatil6:
After digging into the grouper I'm convinced this PR should be closed. Here's why:

`pkg.Groups` is always built from `pkg.Vulnerabilities` via:

```go
pkg.Groups = grouper.Group(grouper.ConvertVulnerabilityToIDAliases(pkg.Vulnerabilities))
```

And inside `grouper.Group`, every vulnerability starts as its own group:

```go
// Initially make every vulnerability its own group.
for i := range vulns {
    groups[i] = i
}
```

So every element of `Vulnerabilities` is **guaranteed** to appear in at least one group. There is no path in `buildVulnerabilityResults` that produces a `PackageVulns` where a vuln exists in `Vulnerabilities` but not in any group.

Consequence: when `len(newGroups) == 0` (all groups ignored), every vuln ID is necessarily in `ignoredVulns` (because it belongs to some ignored group), so the inner loop would produce `newVulns = []` regardless of the guard. **The guard is correct.**

The original issue was found by a static analysis tool, not a live reproduction, and the issue author themselves noted there's no concrete OSV record showing this state. Closing this -- sorry for the noise and thanks to @G-Rath and @Ly-Joey for pushing on the real-world reproduction question.

PR #2762, opened 2026-05-03
fix: skip packages with short commit hashes instead of aborting scan

### LINKED ISSUE #2683: bun.lock: git dependencies with short commit hashes cause "Invalid hash" API error
## Bug

Scanning a project with a `bun.lock` that contains git dependencies with short commit refs causes the scan to fail:

```
error when retrieving vulns: client error: status="400 Bad Request"
body={"code":3,"message":"Invalid hash."}
```

## Reproduction

1. Clone https://github.com/activepieces/activepieces
2. Run:
   ```bash
   osv-scanner scan ./ --format json -L bun.lock .
   ```
3. Observe the error:
   ```
   error when retrieving vulns: client error: status="400 Bad Request" body={"code":3,"message":"Invalid hash."}
   ```

The repo's `bun.lock` contains the following git dependency with a short hash:
```jsonc
"@dust-tt/client/@modelcontextprotocol/sdk": ["@modelcontextprotocol/sdk@github:dust-tt/typescript-sdk#bca26e4", ...]
```

## Version
```
osv-scanner --version
osv-scanner version: 2.3.3
osv-scalibr version: 0.4.2
commit: n/a
built at: n/a
```

nileshpatil6:
Fixes #2683

## Problem

`bun.lock` (and potentially other lockfiles) can record git dependencies using abbreviated commit refs -- e.g. `@modelcontextprotocol/sdk@github:dust-tt/typescript-sdk#bca26e4`. Bun stores only the short 7-char hash, not the full 40-char SHA.

When `filterUnscannablePackages` passes these packages through (the `imodels.Commit(psr) != ""` branch), the OSV bulk-query API rejects them with:

```
error when retrieving vulns: client error: status="400 Bad Request"
body={"code":3,"message":"Invalid hash."}
```

Because the error propagates and `res` is `nil`, the entire scan aborts -- no vulnerability results are returned for any package in the lockfile.

## Fix

Add a case to the second `switch` in `filterUnscannablePackages` that detects commit strings shorter than 40 characters, emits a per-package `Warnf` so the user knows which packages were skipped, and `continue`s past them. All other packages in the same scan are unaffected and the scan completes normally.

A full SHA1 is exactly 40 hex chars; a full SHA256 is 64. Anything shorter is an abbreviated ref the OSV API cannot resolve.

## Testing

Verified by reading the code path: packages with a short commit hash now fall into the new case and are excluded before `MatchVulnerabilities` is called, so the API never sees them. The fix is consistent with how Maven packages with `name == "unknown"` are already handled.


G-Rath (dismissed):
You need to add tests for this

another-rex (approved):

G-Rath (commented):
I think we should have an e2e cli test using e.g. a trimmed version of the original bun lock to give a real world example of why we're doing this why we're doing this

jess-lowe (approved):

another-rex (approved):

PR #2783, opened 2026-05-07
fix: keep transitive scanning with `--offline-vulnerabilities`

### LINKED ISSUE #2775: `--offline-vulnerabilities` also disables transitive scanning
I think `--offline-vulnerabilities` is currently doing more than the docs describe.

`docs/offline-mode.md` says:

> To use offline mode for just the vulnerability database, but allow other features to possibly make network requests (e.g. transitive dependency scanning), you can use the `--offline-vulnerabilities` flag instead.

My reading of this is that `--offline-vulnerabilities` should only switch vulnerability matching to the local DB. If a user wants to disable all network access, they should use `--offline`, which already sets both `--offline-vulnerabilities` and `--no-resolve`.

But in current `main`, both scan paths do this when `actions.CompareOffline` is set:

```go
if actions.CompareOffline {
    capabilities.Network = plugin.NetworkOffline
}
```

That happens in:

- pkg/osvscanner/scan.go
- pkg/osvscanner/osvscanner.go

Because Scalibr then filters plugins by capabilities, this drops network-backed plugins, including the transitive dependency enrichers. So `--offline-vulnerabilities` ends up disabling transitive scanning even when `--no-resolve` was not passed.

This seems to break the use case from #1339:

> Use a local vuln db, but still do transitive scanning in environments with access to the maven registry but not osv.dev.

I think the fix is to separate the two concepts in ScannerActions:

- one flag/action for “use the local vulnerability DB”
- another for “disable network-backed features”

Then CompareOffline can stay focused on selecting the vulnerability matcher, while Scalibr’s NetworkOffline capability is only set for the full `--offline` mode.

Happy to open a PR if this direction sounds right.

gotgolem:
## Overview

Fixes #2775.

`--offline-vulnerabilities` is documented as only switching vulnerability matching to the local DB, but it was also causing Scalibr to filter out network-backed plugins.
That meant transitive dependency scanning was disabled even when `--no-resolve` was not passed.

This PR keeps that behavior scoped to the full `--offline` flag instead.

## Details

I added a separate `ScannerActions.PluginNetworkDisabled` field and wire it from `--offline`.

`CompareOffline` still controls the local vulnerability DB path. The source and container scan paths now use `PluginNetworkDisabled` when setting Scalibr's network capability.

License behavior under `--offline-vulnerabilities` is intentionally left unchanged in this PR.

## Testing

- `go test ./pkg/osvscanner ./cmd/osv-scanner/internal/helper`
- `go test ./cmd/osv-scanner/scan/source -run TestCommand_Transitive -count=1`

Also ran:

- `./scripts/run_tests.sh`
  - Fails locally on unrelated snapshot drift from external data changes.

## Checklist

- [ ] I have run the unit tests using `./scripts/run_tests.sh`.


pkg/osvscanner/scan_test.go
```diff
@@ -3,8 +3,57 @@ package osvscanner
 import (
 	"path/filepath"
 	"testing"
+
+	"github.com/google/osv-scalibr/plugin"
 )
 
+func Test_networkCapability(t *testing.T) {
```
another-rex, line 10: Not too sure if this test is that useful, since it's just testing 1 if statement essentially.

another-rex (approved):
LGTM, thanks!

cuixq (approved):

pkg/osvscanner/scan_test.go
gotgolem: Thanks, fair point.

I added it as a regression guard for the CompareOffline / PluginNetworkDisabled split, but the transitive scan test covers the user-visible behavior. I can drop this one if you prefer :)

another-rex: Hmm... Let's keep it for now since it runs pretty fast, if we find it causing friction later we can remove it.

gotgolem: Sounds good, thanks for the review!

PR #2790, opened 2026-05-09
feat: add RPM scan plugin wiring

### LINKED ISSUE #254: Add RPM/Red Hat ecosystem support

qkal:
## Summary

Adds OSV-Scalibr's RPM package database extractor to OSV-Scanner's default plugin wiring so RPM databases can be scanned alongside the existing APK and DPKG OS package sources.

This also wires the RPM duplicate-package annotator for artifact scans and adds explicit lockfile parse aliases for common RPM database filenames: `rpmdb`, `rpmdb.sqlite`, `Packages`, and `Packages.db`.

Fixes #254.

## Testing

- `go test ./internal/scalibrplugin ./pkg/osvscanner/...`


another-rex (commented):
Thank you, can you also add a Dockerfile as a test for a redhat image so that we know it's extracting and matching correctly??

qkal:
@another-rex thanks for feedback.
I added a pinned Red Hat UBI Dockerfile fixture (test-redhat.Dockerfile) and wired it into the image scan acceptance test so it verifies RPM extraction from /var/lib/rpm/rpmdb.sqlite and Red Hat ecosystem matching.

The RPM extractor is Linux-only in SCALIBR, so the Red Hat image test is skipped on Windows and verified on Linux.

Validation run:
- `go test ./cmd/osv-scanner/scan/image ./pkg/osvscanner/internal/scanners ./internal/scalibrplugin -count=1`
- Linux acceptance subtest: `TEST_ACCEPTANCE=true TEST_VCR_MODE=ReplayOnly go test -v ./cmd/osv-scanner/scan/image -run "^TestCommand_OCIImage$/^Scanning_Red_Hat_image$" -count=1`

PR #2802, opened 2026-05-14
fix: preserve non-ignored vulns when all groups are filtered out

### LINKED ISSUE #2756: filterPackageVulns: orphan vulnerabilities silently dropped when all groups are ignored

SanskaarUndale21:
## Summary

Fixes #2756.

`filterPackageVulns` in `pkg/osvscanner/filter.go` had a logic guard:

```go
    for _, vuln := range pkgVulns.Vulnerabilities { ... }
}
```

This assumption does not hold. A vulnerability can exist in the `Vulnerabilities` slice without being referenced by any group's `Aliases`. When the ignore config caused all groups to be filtered out, the entire vulnerability loop was skipped -- silently dropping any vulnerability that was not explicitly ignored.

**Example scenario that was broken:**

- Groups: `[G1{Aliases: ["CVE-2024-001"]}]`
- Vulnerabilities: `[{ID: "CVE-2024-001"}, {ID: "CVE-2024-999"}]`
- Ignore config: ignore `"CVE-2024-001"`
- Before fix: both vulns dropped (CVE-2024-999 never ignored, but silently lost)
- After fix: CVE-2024-999 correctly preserved

## Changes

- `pkg/osvscanner/filter.go`: removed the `if len(newGroups) > 0` guard; the `ignoredVulns` map already handles filtering correctly on its own
- `pkg/osvscanner/filter_internal_test.go`: added `Test_filterPackageVulns_orphanVulnsPreserved` that directly exercises the bug scenario

## Test plan

- [x] New unit test `Test_filterPackageVulns_orphanVulnsPreserved` covers the exact failure case
- [x] All existing `Test_filterResults` snapshot tests still pass

### DIFF
--- a/pkg/osvscanner/filter.go
+++ b/pkg/osvscanner/filter.go
@@ -182,11 +182,9 @@ func filterPackageVulns(pkgVulns models.PackageVulns, configToUse config.Config)
 	}
 
 	var newVulns []*osvschema.Vulnerability
-	if len(newGroups) > 0 { // If there are no groups left then there would be no vulnerabilities.
-		for _, vuln := range pkgVulns.Vulnerabilities {
-			if _, filtered := ignoredVulns[vuln.GetId()]; !filtered {
-				newVulns = append(newVulns, vuln)
-			}
+	for _, vuln := range pkgVulns.Vulnerabilities {
+		if _, filtered := ignoredVulns[vuln.GetId()]; !filtered {
+			newVulns = append(newVulns, vuln)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


G-Rath:
Per our contributing guide, please dont open pull requests without having first discussed them on an issue

SanskaarUndale21:
Apologies i will discuss it first then

PR #2808, opened 2026-05-16
fix(osvmatcher): dedupe bulk queries

### LINKED ISSUE #2654: bulkquery requests don't deduplicate packages
https://github.com/google/osv-scanner/blob/30ab0fe125354737650556d412f094b8840ffd84/cmd/osv-scanner/scan/source/testdata/cassettes/TestCommand.yaml#L62-L122

This cassette recording shows we're requesting vuln info for the same packages multiple times in the same request, which is silly.

This is probably related to (iirc) we use the index to map results from the bulk query so if we were to just filter out duplicates before doing the querying we'd end up with a bunch of "off by 1" index type bugs and whatnot

rohan-patnaik:
## Overview

This fixes duplicate package queries being sent in the same OSV bulk request.

Fixes #2654

## Details

When the scan input contains the same package more than once, `OSVMatcher` currently sends that same package/version query multiple times to OSV. That does not change the final result, but it does add unnecessary API work and makes larger scans do more work than needed.

This PR deduplicates equivalent OSV queries before calling `BatchQueryPaging`, then expands the hydrated results back to the original package order. That keeps the scanner output behavior the same while reducing duplicate requests.

I also updated the cached matcher query collection to use the same query key, so it does not rely on pointer identity when collecting missing package queries.

## Testing

- Added a regression test that passes duplicate package inputs and verifies only unique queries are sent while all original result slots are preserved.
- Ran `go test ./internal/clients/clientimpl/osvmatcher`.
- Ran `go test ./internal/clients/clientimpl/...`.

## Checklist

- [x] I have run the unit tests using `go test ./internal/clients/clientimpl/...`.


rohan-patnaik:
Rebased this on current main and refreshed the affected VCR cassettes/snapshots after resolving the cassette conflicts.\n\nFocused checks run locally:\n- `go test ./internal/clients/clientimpl/osvmatcher -run 'TestOSVMatcher_MatchVulnerabilitiesDeduplicatesBulkQueries|TestOSVMatcher_MatchVulnerabilities' -count=1`\n- `TEST_VCR_MODE=ReplayOnly go test ./cmd/osv-scanner/scan/source -run '^TestCommand$' -count=1`

another-rex (approved):
Thanks, LGTM. Going to fix the merge issues and we should be good to go.

rohan-patnaik:
Pushed a small follow-up commit to fix the `nlreturn` lint failure in `osvmatcher.go`.

Local checks run on macOS arm64 with Go 1.26.3:
- `go test ./internal/clients/clientimpl/osvmatcher -count=1`
- `./scripts/run_lints.sh`
- `TEST_ACCEPTANCE=true go test ./cmd/osv-scanner/scan/source -run 'TestCommand/(fix_non-interactive_json_override_pom.xml|fix_non-interactive_override_pom.xml|fix_non-interactive_in-place_package-lock.json|fix_non-interactive_json_in-place_package-lock.json|fix_non-interactive_json_relax_package.json|fix_non-interactive_relax_package.json)$' -count=1`
- `TEST_ACCEPTANCE=true go test ./cmd/osv-scanner/update -run 'TestCommand/update_pom_with_in_place_changes_using_deps_dev_data_source$' -count=1`

The exact subtests that failed in the macOS CI log pass locally for me. The CI failure still looks like a VCR/deps.dev cassette mismatch after the cassette refresh, e.g. `rpc error: code = Unavailable desc = service unavailable` followed by `VCR: requested interaction not found` in `cmd/osv-scanner/scan/source`, plus the update snapshot mismatch. If this needs maintainer-side cassette regeneration, please let me know.

another-rex (approved):

PR #2816, opened 2026-05-18
test: fail fast on missing VCR cassettes by converting errors to 404

another-rex:
If a VCR cassette or a specific interaction is missing, the OSV client would previously retry the request 4 times with exponential backoff, causing significant delays in test execution (over 20 seconds).

This change wraps the VCR client transport in tests to intercept the "requested interaction not found" error (using `errors.Is` with `cassette.ErrInteractionNotFound` for robust matching) and convert it to a "404 Not Found" response. Since the OSV client does not retry on 404 status codes, this allows tests to fail immediately (~0.02s) when a cassette or interaction is missing while still preserving retry behavior for genuine transient network errors.

This approach avoids modifying the core `osvscanner` library API or adding new CLI flags.

- Implemented `vcrErrorWrappingTransport` in `internal/testcmd/vcr.go`
- Wrapped the VCR client transport in `InsertCassette`
- Used `errors.Is` with `cassette.ErrInteractionNotFound` for robust error matching

### DIFF
--- a/cmd/osv-scanner/internal/testcmd/vcr.go
+++ b/cmd/osv-scanner/internal/testcmd/vcr.go
@@ -3,6 +3,7 @@ package testcmd
 import (
 	"bytes"
 	"cmp"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -173,7 +174,10 @@ func InsertCassette(t *testing.T) *http.Client {
 		sortCassetteInteractions(t, path)
 	})
 
-	return r.GetDefaultClient()
+	client := r.GetDefaultClient()
+	client.Transport = &vcrErrorWrappingTransport{wrapper: client.Transport}
+
+	return client
 }
 
 // sortCassetteInteractions reorders the interactions in the given cassette, based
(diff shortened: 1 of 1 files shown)


internal/clients/clientimpl/osvmatcher/osvmatcher.go
```diff
@@ -45,13 +45,22 @@ type OSVMatcher struct {
...
 	if httpClient == nil {
 		httpClient = http.DefaultClient
 	}
 
 	config := osvdev.DefaultConfig()
 	config.UserAgent = userAgent
+	if maxNetworkQueryAttempts > 0 {
+		// osvdev.Config.MaxRetryAttempts actually controls the maximum number of attempts,
+		// where 1 means 1 attempt (0 retries).
+		config.MaxRetryAttempts = maxNetworkQueryAttempts
```
cuixq: I am a bit confused about how this config will be propagated through? is the config defined in osv-scalibr?

another-rex: It's not needed for scalibr, we have not switched to use the osv-scalibr osvdev enrichers yet.

another-rex: But I think I'll rework this a bit, I'm not sure we want to have this as a config in general.

cmd/osv-scanner/internal/testcmd/vcr.go
```diff
@@ -255,3 +258,23 @@ func matchBody(r *http.Request, i cassette.Request) bool {
...
 	return true
 }
+
+type vcrErrorWrappingTransport struct {
+	wrapper http.RoundTripper
+}
+
+func (t *vcrErrorWrappingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	resp, err := t.wrapper.RoundTrip(req)
+	if err != nil && strings.Contains(err.Error(), "requested interaction not found") {
```
cuixq: use `errors.Is` to check the error?

cuixq: the error is defined here: https://github.com/dnaeon/go-vcr/blob/v4/pkg/cassette/cassette.go#L50

another-rex: Good idea, done.

cuixq (approved):

Issue #2785, opened 2026-05-07
experimental-mcp: SSE server exposes unauthenticated scan_vulnerable_dependencies tool with no path restrictions


Tobsboy2:
## Summary

The `experimental-mcp` SSE server starts with no authentication and accepts 
arbitrary filesystem paths from any caller via the `scan_vulnerable_dependencies` 
tool. The `fileOpenedLogger` returns absolute paths of every file scalibr opens 
back to the MCP client verbatim.

## Affected code

- `cmd/osv-scanner/mcp/command.go:121` — `DirectoryPaths: input.Paths` 
  (no path validation or allowlist)
- `cmd/osv-scanner/mcp/stats.go:29` — `filepath.Join(Root, Path)` 
  returned to caller in tool response
- `cmd/osv-scanner/mcp/command.go:91` — SSE server starts with no 
  auth middleware

## Suggested fixes

1. Add token authentication to the SSE server
2. Restrict `input.Paths` to a declared workspace root using `filepath.Rel`
3. Exclude `fileOpenedLogger` output from the MCP tool response — 
   it was likely intended for local debug output only

## Notes

Reported to Google VRP — closed as below severity threshold with 
permission to disclose publicly.

PR #2830, opened 2026-05-21
fix: harden experimental MCP server access

maxts0gt:
## Summary
- restrict experimental MCP scan paths to configured workspace roots, including symlink-aware path resolution
- require an SSE bearer token when listening on non-loopback addresses, with optional `OSV_SCANNER_MCP_SSE_TOKEN` support
- redact workspace roots from MCP scan output and wire `ignore_glob_patterns` into scanner exclude patterns

Refs #2785. Thanks @Tobsboy2 for the public report and suggested hardening direction.

## Tests
- `go test ./cmd/osv-scanner/mcp ./pkg/osvscanner`
- `UPDATE_SNAPS=true go test ./cmd/osv-scanner/mcp`

### DIFF
--- a/cmd/osv-scanner/mcp/__snapshots__/integration_test.snap
+++ b/cmd/osv-scanner/mcp/__snapshots__/integration_test.snap
@@ -17,12 +17,12 @@ Total 1 package affected by 1 known vulnerability (0 Critical, 0 High, 1 Medium,
 
 Go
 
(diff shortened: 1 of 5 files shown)


maxts0gt:
@googlebot I signed it!

Tobsboy2:
Thanks for the fix and the detailed hardening work here, especially the workspace-root restrictions and SSE authentication changes. I appreciate the follow-up and the maintainer review.

PR #2857, opened 2026-06-02
chore: tag built Docker images with major version

### LINKED ISSUE #2803: Tag osv-scanner Docker images with major version
Right now, osv-scanner release Docker images are tagged as `latest`, `stable` and `v2.x.x`. The [documentation](https://google.github.io/osv-scanner/usage/#running-in-a-docker-container) suggests using the `latest` tag.

Using the `latest` tag in workflows can be a problem as at some point this will become a v3 release with potentially breaking changes. Using a specific version tag requires constant maintenance to stay on the current osv-scanner release.

In addition to the existing tags, I would like osv-scanner Docker images to be tagged with their major version. For example, `v2`. This would allow consumers to easily use the latest current release version while avoiding major version breaking changes.

Khuzaimx:
Currently, release Docker images for `osv-scanner` are only tagged with their full semver version (`v2.x.x`), `latest`, and `stable`. 

This PR adds major version tags (e.g. `v2`) to both:
- The multi-architecture `osv-scanner` image manifest.
- The single-architecture `osv-scanner-action` GitHub Action image.

This allows consumers to pin their Docker-based workflows and Actions to a major release version (like `:v2`) to safely receive patch and minor updates without the risk of breaking changes from future major releases.

### Changes
*   Added `ghcr.io/google/osv-scanner-action:v{{ .Major }}` template under the action builds in `.goreleaser.yml`.
*   Added `ghcr.io/google/osv-scanner:v{{ .Major }}` manifest template in `.goreleaser.yml` to assemble the amd64 and arm64 builds.

Fixes: #2803

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -113,6 +113,7 @@ dockers:
   # GitHub Action
   - image_templates:
       - "ghcr.io/google/osv-scanner-action:{{ .Tag }}"
+      - "ghcr.io/google/osv-scanner-action:v{{ .Major }}"
     dockerfile: goreleaser-action.dockerfile
     use: buildx
     extra_files:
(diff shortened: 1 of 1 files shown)


another-rex (approved):
Good idea!

PR #2858, opened 2026-06-02
feat: add GitLab Security Report output format

AlexLaroche:
## Summary

Adds a new `gitlab` output format that emits a [GitLab Security Report](https://docs.gitlab.com/ee/development/integrations/secure.html), so osv-scanner results can be consumed directly by GitLab's vulnerability/dependency dashboards.

```
osv-scanner scan --format gitlab ...
```

- Produces a **dependency-scanning** report by default, or a **container-scanning** report when the scan has image metadata (auto-detected).
- Maps OSV vulnerabilities to GitLab's schema: identifiers (CVE, CWE, GHSA, USN, RHSA, ELSA, OSVDB, HackerOne, GLAM, MAL, …), severity (derived from the most recent available CVSS vector), CVSS vectors, and reference links.
- Report output is validated against GitLab's official dependency- and container-scanning report schemas.

## Notes on robustness

The report builder defends against malformed advisory data:

- `ParseIdentifierID` guards the `USN`/`HACKERONE` cases (which read the part after the `-`) so a hyphen-less id/alias can't cause an index-out-of-range panic.
- Severity falls back to `Unknown` rather than being silently dropped when a rating can't be mapped to a defined level (e.g. `NONE` from a 0.0 CVSS score).
- Only CVSS severity types are emitted into `cvss_vectors`, so non-CVSS scores (e.g. Ubuntu) aren't written as invalid CVSS vectors.

## Testing

- Unit tests for identifier parsing, severity mapping, and version handling.
- Snapshot tests of `PrintGitLabResults` output.
- Schema-conformance tests are **hermetic**: GitLab's dependency/container scanning schemas are vendored under `testdata/schemas` and loaded from disk (no network at test time), and the validated reports are produced by the real `PrintGitLabResults` code path rather than hand-maintained fixtures, so the output can't drift from the schema.

### DIFF
--- a/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/scan/source/__snapshots__/command_test.snap
@@ -1864,7 +1864,7 @@ Total 1 package affected by 1 known vulnerability (0 Critical, 1 High, 0 Medium,
 ---
 
 [TestCommand/output_format:_unsupported - 2]
-unsupported output format "unknown" - must be one of: table, html, vertical, json, markdown, sarif, gh-annotations, cyclonedx-1-4, cyclonedx-1-5, cyclonedx-1-6, cyclonedx-1-7, spdx-2-3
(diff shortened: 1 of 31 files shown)


AlexLaroche:
@another-rex @cuixq Could you take a look when you have time? Please.

G-Rath:
Hey sorry for the delay - this sounds interesting but could you please open an issue first so we can discuss if this is something we want to take on?

PR #2882, opened 2026-06-12
fix: skip advisory affected entries with unrecognized ecosystems

### LINKED ISSUE #2867: IsAffected panics on an advisory with an unrecognized ecosystem, crashing the whole scan

Lougarou:
## Overview

`vulns.IsAffected` and `vulns.AffectsEcosystem` panic when an OSV advisory's `affected[].package.ecosystem` is a value that `osvecosystem.MustParse` does not recognize. Both run on the match path for every advisory, and the per-ecosystem local-db zips store the full advisory record (every `affected[]` entry, not just the matching one), so a single multi-ecosystem advisory that lists an ecosystem newer than the build's pinned `osv-schema` aborts the entire offline scan.

Fixes #2867.

## Details

`osvecosystem.MustParse` at `internal/utility/vulns/vulnerability.go:113` (`AffectsEcosystem`) and `:160` (`IsAffected`) panics with `Failed MustParse: base ecosystem does not exist in osvschema: "…"` on any unrecognized ecosystem string.

This is reachable without an attacker: osv.dev's ecosystem set grows over time while each osv-scanner release pins a vendored `osv-schema`, so once a new ecosystem appears in a multi-ecosystem advisory, older builds still pinned in CI panic on it. It is also deterministically triggerable today with a local advisory database.

This is the ecosystem-name analog of #2837, which fixed the *version*-parser panic in `rangeContainsVersion`; the ecosystem-name `MustParse` sites were not touched there.

The fix adds a small `parseAffectedEcosystem` helper that uses the non-panicking `osvecosystem.Parse` and skips the `affected[]` entry on error (logging at debug level), mirroring the existing "skip, don't panic on unsupported ecosystems" handling already used for version parsing. Recognized entries are matched exactly as before; the default path is unaffected.

## Testing

- Added `TestOSV_IsAffected_UnrecognizedEcosystemDoesNotPanic`: a multi-ecosystem advisory whose first entry uses an unrecognized ecosystem and whose second is a recognized npm range — asserts no panic, that the unrecognized entry is skipped, and that the recognized sibling still matches. The test fails (panics) without the fix.
- Extended `TestOSV_AffectsEcosystem` with an unrecognized + recognized ecosystem case.
- `go test ./internal/utility/vulns/` passes.
- Ran the pinned golangci-lint (`v2.11.4`): 0 issues.

## Checklist


internal/utility/vulns/vulnerability.go
```diff
@@ -108,9 +108,30 @@ func rangeAffectsVersion(vulnID string, a []*osvschema.Range, pkg *extractor.Pac
...
 }
 
+// parseAffectedEcosystem parses an advisory affected[] entry's ecosystem
+// without panicking. The vendored osv-schema ecosystem set is pinned per
+// release while osv.dev's set keeps growing, so a multi-ecosystem advisory can
+// carry an ecosystem this build doesn't recognize. Skip that entry instead of
+// crashing the whole scan, mirroring the "skip, don't panic on unsupported
+// ecosystems" handling used for version parsing in rangeContainsVersion.
+func parseAffectedEcosystem(affected *osvschema.Affected) (osvecosystem.Parsed, bool) {
+	ecosystem := affected.GetPackage().GetEcosystem()
```
another-rex, line 118: I think this also needs a GetPackage() == nil check, and don't create a warning.

We already do it in IsAffected, but not in AffectsEcosystem

another-rex (approved):

internal/utility/vulns/vulnerability.go
Lougarou: Done in f0fc7c4 added the silent nil-package guard to AffectsEcosystem (matching IsAffected), plus a test case.

another-rex:
Thanks, sorry for the delay, but the logic has been moved to osv-scalibr: https://github.com/google/osv-scalibr/blob/main/enricher/vulnmatch/osvlocal/osvlocal.go

Can you move the PR there?

Lougarou:
Moved to osv-scalibr: google/osv-scalibr#2429 (same fix ported onto `enricher/vulnmatch/osvlocal/internal/vulns/vulnerability.go`, with the regression test).

PR #2885, opened 2026-06-12
feat: include package purls in JSON output

### LINKED ISSUE #2874: JSON output doesn't include purls for scanned packages
The `--format=json` output describes each scanned package with just name, version and ecosystem:

```json
"package": {
  "name": "lodash",
  "version": "4.18.1",
  "ecosystem": "npm"
}
```

When I need a purl I have to rebuild it by hand from name, ecosystem and version, which is tedious and easy to get wrong. The advisories in the same output already use purls (`affected[].package.purl`).

So I'd add an optional field:

```json
"package": {
  "name": "lodash",
  "version": "4.18.1",
  "ecosystem": "npm",
  "purl": "pkg:npm/lodash@4.18.1"
}
```

With `omitempty`, packages without a purl (git commits, for example) just keep their current shape, like the recently added `deprecated` field.

Happy to open a PR. It's one new field on `models.PackageInfo`.

gotgolem:
## Overview

Fixes #2874.

JSON output currently includes each scanned package's name, version, and ecosystem, but not its package URL.

This PR includes the package URL when scalibr provides one.

## Details

I added an optional `purl` field to `PackageInfo` and populate it from scalibr's package URL in `buildVulnerabilityResults`.

The field is omitted (omitempty) when no PURL is available.

## Testing

- `go test ./pkg/osvscanner`
- `go test ./internal/output`
- `go test ./pkg/models`
- `./scripts/run_lints.sh`

## Checklist

- [ ] I have run the unit tests using `./scripts/run_tests.sh`.


another-rex (commented):
Apologies for the late review! 

Im happy with the change overall, though I think the test that's added is a bit too complicated for what we're actually doing. I would say try to simplify the test or just remove it entirely.

We're currently a bit busy recently, so might be a little while before we can get this merge in, but I will try to make sure this change is in the next osv-scanner release.

gotgolem:
@another-rex
Updated, thanks. I dropped the extra PURL-specific test since the existing snapshots already cover the JSON output.

another-rex (approved):

PR #2888, opened 2026-06-15
build(goreleaser): publish multi-arch (arm64) osv-scanner-action image

skialpine:
## Problem

`ghcr.io/google/osv-scanner-action` is published amd64-only. On `linux/arm64` GitHub-hosted runners — now GA for [public](https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/) and [private](https://github.blog/changelog/2026-01-29-arm64-standard-runners-are-now-available-in-private-repositories/) repositories — the `google/osv-scanner-action` action fails to start:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
exec /bin/bash: exec format error
```

This forces users onto x86 runners or a hand-rolled binary install just to run the dependency scan.

## Fix

Mirror the multi-arch setup already used for the `osv-scanner` scanner image, applied to the `osv-scanner-action` image:

- build the `osv-scanner-action` and `osv-reporter` binaries for `linux/arm64` in addition to `linux/amd64`;
- build an arm64 `osv-scanner-action` image (the existing amd64 image is suffixed `-amd64`, matching the scanner image convention);
- combine the two arches under `:{{ .Tag }}` and `:v{{ .Major }}` via `docker_manifests`, so the public tags become multi-arch manifests.

No Dockerfile change is needed — `goreleaser-action.dockerfile` copies the goreleaser-provided per-arch binary, and the base `golang`/`alpine` images are already multi-arch.

Net diff: **+32 / -3** in `.goreleaser.yml`.

## Testing

`goreleaser check` passes on the modified config. I don't have access to the tagged-release pipeline that publishes to GHCR, so the real multi-arch manifest is produced by your release job — happy to adjust if a maintainer wants to dry-run `goreleaser release --snapshot` against it first.


skialpine:
Would appreciate someone looking at this.

another-rex (approved):

another-rex:
Looks good!

PR #2892, opened 2026-06-16
fix(output): harden GitHub Actions annotation escaping

### LINKED ISSUE #2889: Harden gh-annotations workflow command escaping
## Overview

`gh-annotations` writes GitHub Actions workflow commands directly to stdout. The current implementation already encodes raw `\r` and `\n` bytes, which prevents direct line-break command injection. However, GitHub workflow command parsing has separate escaping rules for command data and command properties.

The `file=` annotation property currently does not escape `%`, `:`, and `,` according to the workflow command property grammar. Annotation message data also does not escape `%` before CR/LF encoding.

## Expected behavior

Values written into GitHub Actions workflow commands should follow the same split used by `actions/toolkit`:

- data: `%`, `\r`, `\n`
- properties: `%`, `\r`, `\n`, `:`, `,`

## Why this matters

Source paths and package metadata can be influenced by scanned repository contents. Applying the full workflow command escaping rules avoids malformed annotations and reduces the chance of command parsing confusion in GitHub Actions logs.

## Proposed fix

Add dedicated helpers for workflow command data/property escaping and use them only in the `gh-annotations` output path:

- property escaping for `file=...`
- data escaping for the annotation message and rendered tables

I have a small patch prepared with regression coverage for `%`, `,`, `:`, and `%0A`-style sequences.

Sayan2208:
## Overview

Harden the `gh-annotations` output format by applying GitHub Actions workflow command escaping rules separately for command properties and command data.

Fixes #2889

## Details

`PrintGHAnnotationReport` emits GitHub Actions workflow commands directly to stdout. The existing implementation encoded raw `\r` and `\n`, but GitHub Actions workflow command escaping also requires `%` escaping for command data, and `%`, `\r`, `\n`, `:`, and `,` escaping for command properties.

This change adds dedicated helpers for the two escaping contexts:

- command data escapes `%`, `\r`, and `\n`
- command properties escape `%`, `\r`, `\n`, `:`, and `,`

The change is scoped to GitHub annotation output:

- `file=` uses property escaping
- annotation message/table bodies use data escaping

Normal terminal table/vertical output remains unchanged.

## Testing

Added regression tests for workflow command escaping in `PrintGHAnnotationReport`:

- `TestPrintGHAnnotationReport_PercentSanitization`
- `TestPrintGHAnnotationReport_PropertyDelimiterSanitization`
- `TestPrintGHAnnotationReport_NewlineInMessageSanitization`
- existing CR sanitization coverage remains

Updated the existing GH annotation snapshot for the workflow command injection attempt path.

Ran locally:

```sh
./scripts/run_tests.sh -run "TestPrintGHAnnotationReport" -count=1
go vet ./internal/output/...
./scripts/run_lints.sh
git diff --check
```

All commands above passed locally.

## Checklist

- [x] I have signed the Contributor License Agreement.
- [x] I have run the relevant unit tests using `./scripts/run_tests.sh -run "TestPrintGHAnnotationReport" -count=1`.
- [x] I have made my commits and PR title follow the Conventional Commits specification.

### DIFF
--- a/internal/output/__snapshots__/githubannotation_test.snap
+++ b/internal/output/__snapshots__/githubannotation_test.snap
@@ -196,7 +196,7 @@
 ---
 
 [TestPrintGHAnnotationReport_WithVulnerabilities/one_source_with_workflow_command_injection_attempt_in_path - 1]
-::error file=dir%0D::stop-commands::T%0D::error::INJECTED%0A::warning::pwn/lockfile::dir%0D::stop-commands::T%0D::error::INJECTED%0A::warning::pwn/lockfile%0A+---------+-----------------------+------+-----------------+---------------+%0A| PACKAGE | VULNERABILITY ID      | CVSS | CURRENT VERSION | FIXED VERSION |%0A+---------+-----------------------+------+-----------------+---------------+%0A| mine1   | https://osv.dev/OSV-1 |      | 1.2.3           |               |%0A+---------+-----------------------+------+-----------------+---------------+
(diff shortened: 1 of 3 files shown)


G-Rath:
Please wait until the issue has been discussed with the maintainers and assigned to you before creating a pull request - we actually had an internal discussion about this yesterday and are considering a different direction.

Sayan2208:
Thanks for the clarification, and sorry for opening the PR before assignment.

I’ll wait for maintainer direction on #2889. If the approach here is useful or can be adapted to the direction you choose, I’m happy to rework it accordingly.

PR #2902, opened 2026-06-27
refactor: format with latest version of Prettier

G-Rath:


another-rex (approved):

G-Rath:
@another-rex fwiw I think this is probably something to just land w/o CI since the snapshots might need updating but nothing can get a green CI without this

(also the unit test jobs are stuck as GH is doing maintenance)

another-rex:
I agree, but I don't think the button exists on mobile app. If you can feel free to skip the checks.

G-Rath:
I don't have the permission to bypass checks 🙃

jess-lowe:
merged with bypass :)

G-Rath:
@jess-lowe fwiw I've messaged Rex about this, but I think the GH PAT used by the PR workflows has expired so they're all currently failing

jess-lowe:
RIP PAT. I'll look into it :) thanks for the heads up

PR #2904, opened 2026-06-28
Fix osv-scanner command to use custom UA

meterglost:
Apply https://github.com/google/osv-scalibr/pull/2269 patch on other commands as well


another-rex:
I think this is redundant now that we are using the new client factories? Please reopen if I'm mistaken.

PR #2907, opened 2026-06-28
fix: scope package VEX suppression to package

huynhtrungcip:
## Summary
- apply ComponentNotPresent package VEX signals to the annotated package instead of the entire source
- remove source-wide rendered-output suppression for PackageSource ExperimentalPES
- add a regression test for sibling packages sharing one source
- update the affected transitive scan cassette/snapshots so replay-only tests have the required interactions

## Tests
- `go test ./pkg/osvscanner ./internal/output -count=1`
- `TEST_VCR_MODE=1 go test ./cmd/osv-scanner/scan/source -run TestCommand_Transitive -count=1`

### DIFF
--- a/cmd/osv-scanner/fix/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/fix/__snapshots__/command_test.snap
@@ -5383,8 +5383,8 @@ UNFIXABLE-VULNS: 9
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.15.tgz",
       "integrity": "sha512-EwOCDEex4quD37XhqM3omwtMoJjr//isUZz1JopUNWms+4Z2ViyM/k1YIRePpoVNnQhENnxtFjLaxNHrT7xIUg==",
       "dependencies": {
-        "balanced-match": "^1.0.0",
-        "concat-map": "0.0.1"
+        "concat-map": "0.0.1",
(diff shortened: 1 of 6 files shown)


noqt (changes requested):
NOQT independently reviewed current head `697f8d7`. The rendered-output fix is headed in the right direction, and the new sibling-package regression passes, but package scope still leaks through the machine-readable JSON.

`packageVulnsGroup` still keeps one `annotations` slice per source. Each included package overwrites it at `pkg/osvscanner/vulnerability_result.go:169`, then it is serialized as `PackageSource.experimental_pes` at line 178. In the new test's shared lockfile, the signaled package is second, so raw JSON still places its `ComponentNotPresent` signal on the whole source even though the sibling vulnerability remains regular. `PrintJSONResults` emits that model unchanged, so a downstream consumer cannot tell which package owns the signal and may repeat the same source-wide suppression.

Please either serialize these signals with the matching package or stop emitting a package signal at source scope, and add a machine-JSON regression for two sibling packages.

The focused regression passes. `go vet` on the changed packages and `git diff --check` pass. The broader `pkg/osvscanner` test has one live-advisory snapshot drift on both this head and unchanged base `7a551f5`; `internal/output` passes on both.

G-Rath:
👋 appreciate the pull requests, but could you please open an issue first so we can discuss and make sure we're aligned on the goal :)

PR #2915, opened 2026-07-06
fix: send epoch-qualified version for RHEL-family RPM queries

Korulag:
## Summary

For RHEL-family RPM ecosystems, osv-scanner reports **already-fixed advisories as unfixed** on any package that carries an epoch (openssl, kernel, grub2, perl, tar, bind, libpng, …).

RHEL-family OSV records store fixed versions **with** the epoch (e.g. `1:3.2.2-7.el9_6`), but osv-scanner sends osv.dev the **epoch-less** version — scalibr keeps the epoch in `rpmmetadata.Metadata.Epoch`, separate from the version string, and it was never included in the query. osv.dev's RPM comparator treats a missing epoch as `0`, so an epoch-bearing installed package looks *older* than every fix and matches advisories it has already been patched against.

## Reproduction (api.osv.dev)

A fully-patched `openssl` reported as still vulnerable, purely due to the missing epoch:

| Ecosystem | query at patched version | epoch-less | epoch-qualified |
|---|---|---|---|
| Red Hat | `openssl 3.5.0-1.el9` | **111** | 10 |
| Rocky Linux 8 | `openssl 3.5.0-1.el8` | **11** | 0 |
| AlmaLinux 9 | `openssl-libs 3.5.4-1.el9` | **15** | 3 |

## Fix

Prepend the RPM epoch (when non-zero) to the query version in `pkgToQuery`, for ecosystems whose OSV records are known to encode it:

```go
var rhelFamilyEpochEcosystems = map[string]bool{
    "Red Hat": true, "AlmaLinux": true, "Rocky Linux": true,
}
```

Epoch-0 packages and non-RPM ecosystems (which carry any epoch inline) are unchanged.

## Why an allowlist rather than "all RPM ecosystems"

A blanket "always prepend the epoch" change would **regress** ecosystems whose feed stores epoch-less records even though their packages carry epochs. openEuler is one such case: its packages use epochs (e.g. `ImageMagick` epoch 1), but its OSV feed drops them (`fixed: "7.1.0.28-3.oe2203"`). Sending the epoch there flips real vulnerabilities to false negatives:

| openEuler `ImageMagick 7.1.0.28-3.oe2203` (a real fix point) | epoch-less | epoch-qualified |
|---|---|---|
| vulns returned | 10 | **0** |

So the correct query epoch-presence is per-ecosystem. The allowlist is restricted to ecosystems verified to encode epochs; leaving an ecosystem out is the safe direction (it keeps today's behavior). Entries can be added as other ecosystems are verified (Oracle Linux / CentOS / SUSE were not reachable on osv.dev to confirm).

## Testing

`TestQueryVersion` covers epoch prepending for Red Hat / AlmaLinux / Rocky Linux, the epoch-0 and non-RPM passthrough, and an explicit openEuler-exclusion guard so the regression above cannot silently return.


Korulag:
@another-rex @jess-lowe could you please take a look or ping someone who can?

another-rex (commented):
Good find! Please put this logic inside the imodels.Version() function instead of the current location though.

Korulag:
@another-rex moved to imodels as you've requested

another-rex (approved):

XavierMP14 (approved):

PR #2919, opened 2026-07-08
fix: add filter to show packages with license violations but no vulnerabilities

### LINKED ISSUE #1643: Add a "Show all packages" filter to the HTML output
https://github.com/google/osv-scanner/pull/1611 added license scanning results to the HTML output, but the current filter logic [hides all packages without vulnerabilities](https://github.com/google/osv-scanner/blob/e0206dedd17080021b768526f2a6fde635497c25/internal/output/html/script.js#L148). As a result, most packages with license violations are hidden and cannot be viewed.

To resolve this, we can add a filter to the HTML output to display packages with license violations even if they have no vulnerabilities (the filter logic determining which packages to hide needs to be updated).

summerpan688:
Fixes #1643

## Summary
Packages with license violations but no vulnerabilities were always
hidden in the HTML report, since the filtering logic only checked
`.vuln-tr` rows. This adds a "Show license violations without
vulnerabilities" filter checkbox (default: off) that reveals these
packages on request.
image

Also fixed two related issues surfaced during testing:
- Packages with zero vulnerabilities were mislabeled "Filtered out"
  (a label meant for packages whose vulns were filtered by call
  analysis) — now shows "No known vulnerabilities" instead.
image

- Clicking into these packages' details previously showed an empty
  vuln table — now shows a placeholder message instead.
image

## Design discussion
See discussion in the issue comment 
 "Show all packages“ vs. the more scoped "Show license violations without vulnerabilities"
（happy to adjust based on feedback）


summerpan688:
Looked into the failing Run unit tests checks — the failure is in TestDoScan/Test_curl_git_scanning, unrelated to the files changed in this PR (internal/output/html/*). The diff is a single modified timestamp on CVE-2025-63675 in the snapshot, which appears to just reflect the OSV DB updating that record between snapshot refreshes (saw the same pattern referenced in #2920). Confirmed locally that re-running with the latest main merged in still shows this as the only diff — happy to update the snapshot if preferred, just didn't want to bundle in an unrelated data-only diff without checking first.

summerpan688:
Hi @jess-lowe @another-rex , just checking in on this PR whenever you have a chance to take a look — no rush! Also wanted to flag that the CI test failures (Run unit tests) appear unrelated to this change; left a note with details above. Thanks so much! 
(Also, apologies for the gap in activity — my account was temporarily suspended and I just got access back)

internal/output/html/vuln_table_template.gohtml
```diff
@@ -17,3 +20,4 @@
     {{template "vuln_table_entry_template.gohtml" $args}}
   {{ end }}
 </table>
++{{ end }}
```
another-rex: hmm... this looks weird, why is there a + here.

internal/output/html/filter_template.gohtml
```diff
@@ -76,6 +76,12 @@
           <input type="checkbox" id="uncalled-type-checkbox" data-type-uncalled-count="{{ .VulnTypeSummary.Hidden }}">
           Uncalled/Unimportant ({{ .VulnTypeSummary.Hidden }})
         </label>
+        {{ if .LicenseSummary.Summary }}
```
another-rex: I think osv-scanner can have --license-violations without having the summary flag, `LicenseSummary.ShowViolations` is the bool here, we also need to show this.

internal/output/html/script.js
```diff
@@ -409,12 +412,30 @@ function resetTypeCheckbox() {
...
 document.addEventListener("DOMContentLoaded", () => {
   resetFilterText();
   showAndHideParentSections();
 
+  // Implement filter for "show license violations without vulnerabilities"
+  const licenseViolationCheckbox = document.getElementById(
+    "license-violation-type-checkbox"
+  );
+  if (licenseViolationCheckbox) {
+    licenseViolationCheckbox.addEventListener("change", () => {
```
another-rex: do we need to add a stopPropagation here?

Actually we might not need `applyFilers`, since it's already being called by the parent listener. Please have a look.

another-rex (commented):
Thanks! Sorry for the late review, been a bit busy recently!

summerpan688:
> Thanks! Sorry for the late review, been a bit busy recently!

Thanks for the thorough review! Pushed fixes for all three — replied inline on each.

internal/output/html/vuln_table_template.gohtml
summerpan688: Ah, that was a mistake and missing newline at end of file. Fixed — pushed in the latest commit.

internal/output/html/filter_template.gohtml
summerpan688: Good catch, thank you! You're right, updated to check LicenseSummary.ShowViolations instead of Summary so the checkbox shows up correctly even without the summary flag.

internal/output/html/script.js
summerpan688: You're right — this checkbox lives inside type-filter-option-container, which already has a change listener that calls applyFilters. Removed the separate listener and folded the license-violation checkbox's state into the existing one, so applyFilters only runs once now.

internal/output/html/filter_template.gohtml
another-rex: Not sure if we need both the summary and license violations here or just a license violation check...

another-rex (approved):

internal/output/html/filter_template.gohtml
summerpan688: I think just `ShowViolations` is enough here. `Summary` controls the separate license count summary, while `ShowViolations` is specifically set when an allowlist exists. Using both could also reintroduce the original issue where the checkbox disappears when violations exist but the summary isn't enabled.

summerpan688:
@another-rex hi! This pr should be ready to merge, any questions please let me know!

PR #2927, opened 2026-07-14
fix(security): sanitize package identity in workflow-parsed log sinks

adilburaksen:
## Problem

Several `Info`/`Warn` log sites emit a package **name / version / ecosystem** derived from **lockfile or SBOM contents** to stdout without applying `output.SanitizeForWorkflowCommand`. The `cmdlogger` handler writes the message verbatim to stdout (`internal/cmdlogger/handler.go`), and the GitHub Actions runner parses stdout for `::command::value` sequences.

On the `pull_request`-triggered `google/osv-scanner-action` path an attacker controls those values, and a package name may legally contain `\r`/`\n` (e.g. a `package-lock.json` `packages` path key, or a CycloneDX component). That lets an attacker break out of the log line and inject `::error::`, `::add-mask::`, `::stop-commands::`, `::notice::`, etc. into the runner's output.

Unlike the file-path sinks, this vector needs only crafted lockfile **contents** — no unusual filenames in the repository.

## Prior art in this repo

`pkg/osvscanner/scan.go:158` already applies the guard for the directory-name log line, and its comment states the exact threat model. This change extends the same guard to the remaining **package-identity** sinks that were not covered:

- `pkg/osvscanner/filter.go:48` — short-commit skip warning; reachable **with no config** via a git dependency whose commit hash is `< 40` chars
- `pkg/osvscanner/filter.go:104` — ignored-package notice
- `pkg/osvscanner/vulnerability_result.go:110,113` — license-override notices

## Reproduction (before this change)

`package-lock.json` git dependency with a short commit and a newline in the package name (no config needed):

```
Skipping ev
::error::INJECTED: short commit hash "abc1234" cannot be queried; ...
```

The `::error::` lands at the start of its own stdout line → parsed as a workflow command. After this change the bytes are URL-encoded (`ev%0A::error::INJECTED...`) and stay on one line.

## No behaviour change for benign metadata

`SanitizeForWorkflowCommand` is an identity function on strings that contain no `\r` or `\n`, so typical package name/version/ecosystem output is unchanged. `go test ./pkg/osvscanner/` passes with no snapshot changes.

### DIFF
--- a/pkg/osvscanner/filter.go
+++ b/pkg/osvscanner/filter.go
@@ -9,6 +9,7 @@ import (
 	"github.com/google/osv-scanner/v2/internal/config"
 	"github.com/google/osv-scanner/v2/internal/imodels"
 	"github.com/google/osv-scanner/v2/internal/imodels/results"
+	"github.com/google/osv-scanner/v2/internal/output"
 	"github.com/google/osv-scanner/v2/pkg/models"
 	"github.com/ossf/osv-schema/bindings/go/osvconstants"
 	"github.com/ossf/osv-schema/bindings/go/osvschema"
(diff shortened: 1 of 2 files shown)


G-Rath:

PR #2928, opened 2026-07-15
fix(security): sanitize extractor-emitted package identifiers in workflow-parsed log sinks

adilburaksen:
Follow-up to #2925 (cc @BiswajeetRay7, who requested a separate scoped PR for these sinks in https://github.com/google/osv-scanner/pull/2925#issuecomment-4984310843).

## Scope
#2925 sanitizes the **filesystem-walk path** sinks. This PR covers the complementary **extractor-emitted package-identity** sinks — package name/version/ecosystem read from lockfile or SBOM *contents* — which reach stdout unsanitized through the same `cmdlogger` → `Handler.Handle` path (`record.Message` is written verbatim; there is no global guard). A package name can legally contain `\r`/`\n` (e.g. a `package-lock.json` `packages` path key, or a CycloneDX component), so this is a strictly wider primitive than the path sinks: it needs only crafted lockfile contents, no unusual filenames.

## Sinks guarded (same `output.SanitizeForWorkflowCommand` as `pkg/osvscanner/scan.go:158`)
- `pkg/osvscanner/filter.go:48` — short-commit skip warning (`imodels.Name`). Reachable with **no config** via a git dependency whose commit hash is `< 40` chars.
- `pkg/osvscanner/filter.go:104` — ignored-package notice (`pkgString` = ecosystem/name/version, plus `reason`).
- `pkg/osvscanner/vulnerability_result.go:110,113` — license-override notices.

## Verification
- Before: a crafted `package-lock.json` emits `::error::…` at the start of its own stdout line (workflow-command injection). After: the bytes are URL-encoded (`…%0A::error::…`) and stay on one line.
- `go test ./pkg/osvscanner/` passes with no snapshot changes (`SanitizeForWorkflowCommand` is identity on strings without `\r`/`\n`).

Related VRP report: https://issuetracker.google.com/issues/533784193 (same threat model). Supersedes #2927 (which I opened before seeing the contributing guide; reopening as a scoped follow-up per @BiswajeetRay7's request). Happy to instead fold this into #2925 if the review team prefers a single commit.

### DIFF
--- a/pkg/osvscanner/filter.go
+++ b/pkg/osvscanner/filter.go
@@ -9,6 +9,7 @@ import (
 	"github.com/google/osv-scanner/v2/internal/config"
 	"github.com/google/osv-scanner/v2/internal/imodels"
 	"github.com/google/osv-scanner/v2/internal/imodels/results"
+	"github.com/google/osv-scanner/v2/internal/output"
 	"github.com/google/osv-scanner/v2/pkg/models"
 	"github.com/ossf/osv-schema/bindings/go/osvconstants"
 	"github.com/ossf/osv-schema/bindings/go/osvschema"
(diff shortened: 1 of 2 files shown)


G-Rath:
@adilburaksen we appreciate the pull request, but we've discussed this internally and decided this kind of issue is not a concern for the scanner as

PR #2934, opened 2026-07-21
feat: Fix all the snapshot problems for ever!*

another-rex:
> [!NOTE]
> I'll try to split this up

With the scalibr clients PR finally merged in, we now have access to everything we need to fully migrate to use scalibr end2end.

This PR does the following:

- Remove vulnerability and license matching from osv-scanner, and enable them as enrichers when running osv-scalibr
- All scanning and matching logic is in scalibr now, making it much easier to extend/onboard new functionality.
- Setup filtering as a plugin (temporary until we get a better solution, see design doc).
- Setup clients to allow cassette of enricher requests
- Create a grpc vcr library to allow cassetting of deps.dev requests
- Update how offline vulnerabilities are being tested so it's not downloading the full zips every time, making it much much faster to run.
- Full tests reduced from 60+ seconds or so on my machine to 7 seconds.
- Basically no more snapshot changes without code changes. *

~~Not quite true, as we still download live full vulns, which has a severity field that is likely to change.~~

This has been updated so that we replace the severities with a snapshotted severity map, solving this issue too.

Resolves #1567

PR #2935, opened 2026-07-22
refactor: migrate vulnerability and license matching end-to-end to scalibr

another-rex:
This PR migrates scanning, filtering, and matching in osv-scanner to use osv-scalibr end-to-end:

This is essentially the first part of #2934 split out into it's own PR.

- **Matching Offloaded to Scalibr Enrichers**: Removed legacy vulnerability and license matchers from osv-scanner, delegating them to osv-scalibr enrichers.
- **Filtering Annotator Plugin**: Implemented package filtering as a scalibr annotator plugin (osv-scanner/filter) to handle unscannable packages, container-specific package exclusions, and ignore rules.
- **Client Factories**: Introduced ClientFactories implementing scalibrconfig.ClientFactories to inject User-Agent headers and pool client connections.
  - This replaces usages in the fix and scan subcommands.
- **PURL Type Resolution**: Updated osvscannerjson extractor to map ecosystem names to valid PURL types (golang, gem, cargo, npm, etc.).
- **Model & Utility Cleanups**: Cleaned up imodels package and muffled asynchronous database loading logs in testlogger.

agy


cmd/osv-scanner/scan/image/testdata/stuff.py
michaelkedar, line 1: what is this file?

cmd/osv-scanner/internal/testcmd/vcr.go
```diff
@@ -199,7 +217,17 @@ func sortCassetteInteractions(t *testing.T, path string) {
...
+			return c
+		}
+		if c := cmp.Compare(a.Request.Method, b.Request.Method); c != 0 {
+			return c
+		}
+		if c := cmp.Compare(a.Request.URL, b.Request.URL); c != 0 {
+			return c
+		}
+
+		return cmp.Compare(a.Request.Body, b.Request.Body)
```
michaelkedar: `cmp.Or(...)` here?

internal/scalibr/client_factories.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/google/osv-scalibr/plugin/config"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+type closableGRPCConn interface {
+	grpc.ClientConnInterface
+	Close() error
+}
```
michaelkedar: nit: a more idiomatic name for this might be `grpcClientConnCloser` (or `grpcConnCloser` if that's too wordy for you)

internal/scalibr/client_factories.go
```diff
@@ -0,0 +1,127 @@
...
+		underlying: transport,
+		userAgent:  c.defaultUserAgent,
+	}
+
+	return &clientCopy
+}
+
+//nolint:nilnil // returning nil client is expected when not implemented
+func (c *ClientFactories) GoogleHTTPClient(_ context.Context, _ ...string) (*http.Client, error) {
+	return nil, nil
```
michaelkedar: Is this expected to be called by scalibr?
If not, should this panic or return an error (presumably if it doesn't return an error, it will panic anyway when something tries to use the nil client?)

internal/scalibrannotator/filter/filter.go
```diff
@@ -0,0 +1,192 @@
...
+
+// Requirements returns the requirements of this annotator.
+func (a *Annotator) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{}
+}
+
+// Annotate filters the package list inside the inventory.
+func (a *Annotator) Annotate(_ context.Context, _ *annotator.ScanInput, results *inventory.Inventory) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
```
michaelkedar: I'm not sure I understand the flow of scalibr enough to know what a mutex is necessary for an annotator (i.e. when would the same annotator be called in parallel)?

But also, should the mutex be covering this *whole* function, or would it be better to only hold the lock for the writes to `a`?

cmd/osv-scanner/scan/image/testdata/stuff.py
another-rex: Oops, I left that in when debugging the chainid issues with image scanning

internal/scalibr/client_factories.go
another-rex: I guess not. I have this returning an error now. Someone could probably configure osv-scanner to call osv-scalibr in a way that would try to get the GoogleHTTPClient, but it's probably fine to just error in those cases.

internal/scalibrannotator/filter/filter.go
another-rex: Yeah this is being overly cautious, I removed the mutex entirely.

Ly-Joey (approved):

michaelkedar (approved):

PR #2946, opened 2026-07-31
fix: omit -coverpkg=./... on macOS to avoid linker crashes and timeouts

another-rex:
Omit `-coverpkg=./...` when running tests on macOS in CI.

### Background & Issue
Running `go test ./... -coverpkg=./...` forces the Go linker (`cmd/link`) to compile and inject cross-package coverage symbols from every package across the entire module into every test binary.

On macOS runners (which have slower virtual disk I/O and tighter memory limits), this cross-package coverage instrumentation creates severe memory-mapping (`mmap`) thrashing and high CPU overhead, leading to:
1. Linker `SIGBUS` bus error crashes (`encoding/binary.littleEndian.Uint16` inside `cmd/link/internal/ld.(*deadcodePass).flood`) when memory-mapped object files in `GOCACHE` are truncated or corrupted under memory thrashing.
2. Intermittent CI test timeouts.

### Fix
In `scripts/run_tests.sh`, check if running on macOS (`RUNNER_OS=macOS` or `uname -s = Darwin`) when in CI, and run `go test ./... -coverprofile coverage.out` without `-coverpkg=./...`. Standard package coverage is still recorded to `coverage.out`.

agy

### DIFF
--- a/cmd/osv-scanner/fix/__snapshots__/command_test.snap
+++ b/cmd/osv-scanner/fix/__snapshots__/command_test.snap
@@ -13288,3 +13288,25 @@ UNFIXABLE-VULNS: 6
 }
 
 ---
+
+[TestCommand_OfflineDatabase/fix_non_interactive_relax_package_json_with_offline_vulns - 4]
+
+---
+
+[TestCommand_OfflineDatabase/fix_non_interactive_relax_package_json_with_offline_vulns - 5]
+{
+  "name": "osv-fix",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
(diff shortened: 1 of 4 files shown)


G-Rath (changes requested):
I would expect you need to adjust CI somehow, otherwise we'll be uploading different coverage structures to codecov, and that will likely be confusing

another-rex:
hmm it seems like it's not even fixing the original problem I'm running into with the local fix_non_interactive_relax_package_json_with_offline_vulns
being flaky and timing out on macos specifically. Might just shelve this one for now.

PR #2951, opened 2026-08-03
fix: localdb test deadlocks

another-rex:
While we had some separate localdb paths setup for the main set of osv-scanner offline tests, it's not applied where other places where localdb tests happen.

This is now done globally in addition to the existing local db path now.


cuixq (approved):

G-Rath (approved):

another-rex:
hmm the macos test is still taking a long time 🤔

PR #2955, opened 2026-08-05
fix: add gRPC transport keepalive parameters to prevent macOS test timeouts

another-rex:
### Background & Issue
osv-scanner's `fix` command initializes a gRPC client connection to `api.deps.dev:443` for dependency resolution

On macOS CI apparently streams without transport keepalives can stall, especially using the native network stack.

Trying this to see if it helps.

---

Long term this will be fixed by the grpc VCRs anyway.

### Fix
In `internal/scalibr/client_factories.go`, configure gRPC keepalive client parameters on `GRPCClientConn`:
```go
ourDialOpts := []grpc.DialOption{
	grpc.WithTransportCredentials(creds),
	grpc.WithKeepaliveParams(keepalive.ClientParameters{
		Time:                10 * time.Second,
		Timeout:             5 * time.Second,
		PermitWithoutStream: true,
	}),
}
```

With these parameters, gRPC will transmit HTTP/2 PING frames every 10 seconds to detect dead TCP connections and fail/reconnect within 15 seconds.

agy

### DIFF
--- a/internal/scalibr/client_factories.go
+++ b/internal/scalibr/client_factories.go
@@ -8,10 +8,12 @@ import (
 	"fmt"
 	"net/http"
 	"sync"
+	"time"
 
 	"github.com/google/osv-scalibr/plugin/config"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
 )
 
 type grpcClientConnCloser interface {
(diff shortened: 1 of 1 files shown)


jess-lowe (approved):

another-rex:
Hmm this worked, but is pretty close to 10 mins

PR #2964, opened 2026-08-05
fix: harden gh-annotations workflow command escaping

Rajkaran-122:
Add dedicated helpers for GitHub Actions workflow command escaping following the actions/toolkit specification:
- escapeGHCommandData() for message data: %, \r, \n
- escapeGHCommandProperty() for file= property: %, \r, \n, :, ,

Previously, only \r and \n were escaped in file paths, leaving %, :, and , unescaped in the file= property. This could cause malformed annotations or command parsing confusion. The annotation message data also did not escape % before CR/LF encoding.

This change applies the full escaping rules to prevent malformed annotations and reduce the chance of command parsing confusion in GitHub Actions logs, addressing issue #2889.

### DIFF
--- a/internal/output/__snapshots__/githubannotation_test.snap
+++ b/internal/output/__snapshots__/githubannotation_test.snap
@@ -196,7 +196,7 @@
 ---
 
 [TestPrintGHAnnotationReport_WithVulnerabilities/one_source_with_workflow_command_injection_attempt_in_path - 1]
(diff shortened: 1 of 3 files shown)


G-Rath:
We appreciate the pull request, but we've discussed this internally and decided this kind of issue is not a concern for the scanner as

PR #2967, opened 2026-08-05
Fix/gh annotations escaping

Rajkaran-122:
Implements the fix proposed in #2889 by hardening GitHub Actions workflow command escaping.

Changes
Added escapeGHCommandData() to escape %, \r, and \n in workflow command data.
Added escapeGHCommandProperty() to escape %, \r, \n, :, and , in workflow command properties.
Updated PrintGHAnnotationReport() to use the dedicated escaping helpers for both file= properties and annotation message data.
Replaced the previous inline escaping logic with the new helper functions.
Added TestPrintGHAnnotationReport_Escaping to verify correct escaping of special characters in both command properties and data.
Updated the relevant snapshot test to reflect the new escaping behavior.
Validation
✅ Existing tests pass.
✅ New regression test passes.
✅ Linting passes.
✅ Changes are minimal and focused solely on issue #2889.

### DIFF
--- a/internal/output/__snapshots__/githubannotation_test.snap
+++ b/internal/output/__snapshots__/githubannotation_test.snap
@@ -196,7 +196,7 @@
 ---
 
 [TestPrintGHAnnotationReport_WithVulnerabilities/one_source_with_workflow_command_injection_attempt_in_path - 1]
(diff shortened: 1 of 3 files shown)


G-Rath:
Duplicate of #2964

PR #2969, opened 2026-08-06
docs: fix tautology in README

Sup2point0:
"CLI interface" means "command line interface interface"


another-rex:
even though technically incorrect, I think it's mostly fine as it is, people reading know what it means.

PR #2973, opened 2026-08-09
fix(output): sanitize package name/version in table output to prevent workflow command injection

prasanna8585:
## Summary

The main vulnerability table and the filtered/hidden-vulnerability table in `table.go` append `pkg.Name` and the installed version directly to the output row, with no call to `SanitizeForWorkflowCommand` -- unlike other values in this exact file (source title, license violations, binary package names), which already go through it:

```
// line 187
outputRow = append(outputRow, pkg.Name, getInstalledVersionOrCommit(pkg), fixAvailable, totalCount)

// line 241
outputRow = append(outputRow, pkg.Name, eco.Name, getInstalledVersionOrCommit(pkg), totalCount, filteredReasons)
```

The table library is configured via `SetOutputMirror` to write **directly** to the output writer on `Render()`:

```
outputTable.SetOutputMirror(outputWriter)
```

with no intermediate capture-and-sanitize step -- unlike the gh-annotation path (`githubannotation.go`), which renders to a string, sanitizes it, then writes. A package name or version containing a raw `\r`/`\n` -- exactly the kind of value `osv-scanner` reads from a scanned, potentially untrusted manifest or lockfile -- reaches stdout unmodified.

**GitHub Actions scans the entire stdout of every workflow step for `::command::` sequences, regardless of output format**, so this is reachable via the default/plain table output, not only the `gh-annotation` format specifically.

This is the same vulnerability class already found and fixed multiple times in this file and its siblings -- `vertical.go`'s three call sites (`vertical_test.go` references this as *"not sanitized by the original fix in PR #2750"*), and the `gh-annotation` path. These two call sites in `table.go` were missed by that sweep.

## Fix

Wrap `pkg.Name` and `getInstalledVersionOrCommit(pkg)` in `SanitizeForWorkflowCommand` at both call sites, matching the existing pattern already used elsewhere in this exact file.

## Testing

- `TestPrintTableResults_CRSanitization` (new): a package name/version containing `\r::stop-commands::...` is confirmed to appear only in its `%0D`-encoded form in the rendered table output, following the same pattern as the existing `TestPrintVerticalResults_CRSanitization`.
- The second call site (the filtered/hidden-vulnerability table) receives the identical fix but isn't independently covered by a new test, since constructing a fixture that exercises the "hidden vuln" filtering path requires internal aggregation logic not reachable from this package's public test surface.

## Verification note

I wasn't able to run the full test suite in the environment I used to prepare this fix -- `go.mod` requires Go 1.26.5, which wasn't obtainable there. I verified the change with `gofmt` (syntax/formatting validity) and with a standalone reproduction using the real, verbatim `SanitizeForWorkflowCommand` function copied out of `sanitize.go`, confirming the vulnerable path leaks a raw `\r` byte and the fixed path doesn't. I'd appreciate CI (or a maintainer) confirming the new test passes as an independent check beyond that.

### DIFF
--- a/internal/output/table.go
+++ b/internal/output/table.go
@@ -184,7 +184,7 @@ func printSummaryResult(result Result, outputWriter io.Writer, terminalWidth int
 					}
 				}
 
-				outputRow = append(outputRow, pkg.Name, getInstalledVersionOrCommit(pkg), fixAvailable, totalCount)
+				outputRow = append(outputRow, SanitizeForWorkflowCommand(pkg.Name), SanitizeForWorkflowCommand(getInstalledVersionOrCommit(pkg)), fixAvailable, totalCount)
 
 				if isOSResult(source.Type) {
 					outputRow = append(outputRow, formatBinaryPackages(pkg.OSPackageNames))
@@ -238,7 +238,7 @@ func printSummaryResult(result Result, outputWriter io.Writer, terminalWidth int
 					outputRow := make(table.Row, 0, 5)
 					totalCount := pkg.VulnCount.AnalysisCount.Hidden
 					filteredReasons := getFilteredVulnReasons(pkg.HiddenVulns)
-					outputRow = append(outputRow, pkg.Name, eco.Name, getInstalledVersionOrCommit(pkg), totalCount, filteredReasons)
+					outputRow = append(outputRow, SanitizeForWorkflowCommand(pkg.Name), eco.Name, SanitizeForWorkflowCommand(getInstalledVersionOrCommit(pkg)), totalCount, filteredReasons)
 					outputTable.AppendRow(outputRow)
 				}
 			}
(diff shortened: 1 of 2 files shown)


prasanna8585:
Updated this PR to be comprehensive. The original commit only covered 2 of 6 call sites in `table.go` with the same missing-sanitization pattern. While double-checking this finding against the project's own prior VRP-driven fix history (#2749 / #2750), I re-read the whole file and found 4 more instances of the identical miss -- including two (`licenseViolationsTableBuilder`, `deprecatedPackagesTableBuilder`) where PR #2750 itself sanitized the `path` value on the exact same output row but missed `pkg.Package.Name`/`Version` right next to it.

Full list now covered:
- `printSummaryResult`'s main and filtered/hidden tables (the original commit)
- `tableBuilderInner`'s GIT-commit row and normal ecosystem row
- `licenseViolationsTableBuilder` and `deprecatedPackagesTableBuilder`
- `licenseSummaryTableBuilder`'s `license.Name` (defense-in-depth; this data flow wasn't traced as precisely as the others)

Regression tests added for the three distinct `PrintTableResults` code paths this touches (see updated PR description).

G-Rath:
We appreciate the pull request, but we've discussed this internally and decided this kind of issue is not a concern for the scanner as

PR #2977, opened 2026-08-10
fix(sourceanalysis): return an error instead of exiting when an rlib has no object file

### LINKED ISSUE #2976: extractRlibArchive exits the process via log.Fatalf on io.EOF, reporting an internal error as exit code 1 ("vulnerabilities or findings")
Following up on GHSA-2f6x-xr4j-f2h7 — @another-rex asked me to convert it into an issue, and having
looked at it again I think he was right that it isn't a vulnerability. Reframing it here as what I
now believe it actually is: an exit-code contract violation.

## What happens

`extractRlibArchive` calls `log.Fatalf` on any error from the archive iterator, including the
ordinary `io.EOF` that ends every ar archive:

https://github.com/google/osv-scanner/blob/5bc4ef461c3e57ae246d80a4c956a5245390bca8/internal/sourceanalysis/rust.go#L194-L199

```go
for {
    header, err := reader.Next()
    if err != nil {
        log.Fatalf("%v", err)   // os.Exit(1) — including on io.EOF
    }
```

The loop's only exit is finding a `/0` member or one ending in `.rcgu.o/`. An ar archive with
neither walks to end-of-archive, `ar.Reader.Next()` returns `io.EOF`, and the process exits.

No malformed or hostile input is needed — a well-formed ar file that simply isn't a Rust rlib is
enough. I reproduced it with `ar rc /tmp/plain.a /tmp/a.txt` and the loop above.

## Why I think this is worth fixing — the exit code is wrong

`docs/output.md` defines the contract:

| Exit code | Reason |
|---|---|
| `0` | Packages were found when scanning, but does not match any known vulnerabilities or findings |
| `1` | Packages were found when scanning, and there are vulnerabilities or findings |
| `127` | General Error |
| `129-255` | Reserved for non result related errors |

`log.Fatalf` exits **1**. So an internal failure is reported to the caller as *"there are
vulnerabilities or findings"* — a result-related code — when it should be `127`. A CI job can't
distinguish "this scan found problems" from "this scan died on an unrelated build artifact", and
the scan produces no output either way.

To be clear about what this is not: the process exits non-zero, so CI does go red. Nothing is
silently hidden. That's precisely why I no longer think it's a security issue — but the outcome is
still being misreported through the documented interface.

## The caller already handles this correctly

`rust.go:53-57` is written to survive exactly this:

```go
buf, err := extractRlibArchive(path)
if err != nil {
    cmdlogger.Errorf("failed to analyse '%s': %s", path, err)
    continue
}
```

`log.Fatalf` exits before that error can be returned, so the recovery never runs. One bad artifact
ends the whole scan instead of being skipped.

This is the only `log.Fatalf` in non-test library code in the repository.

## Suggested fix

Return the error and treat end-of-archive as a normal termination:

```go
header, err := reader.Next()
if errors.Is(err, io.EOF) {
    return bytes.Buffer{}, fmt.Errorf("no object file found in rlib archive '%s'", rlibPath)
}
if err != nil {
    return bytes.Buffer{}, fmt.Errorf("error reading rlib archive '%s': %w", rlibPath, err)
}
```

The existing `continue` then does what it was written to do: skip that artifact and carry on. The
exit code also stops being wrong, because the scan completes normally.

Reached via `--experimental-call-analysis` on a Rust project with a `.rlib` in its build output.

Happy to send a PR with a regression test if that's useful — @another-rex offered, so I'll open one
shortly and you can close it if you'd rather take a different approach.

Found against `5bc4ef461c3e57ae246d80a4c956a5245390bca8`. I used an AI assistant while
investigating; I read the code paths and ran the reproduction myself.

kobihikri:
## Overview

Fixes #2976.

`extractRlibArchive` calls `log.Fatalf` on any error from the archive iterator, including the
ordinary `io.EOF` that ends every ar archive. An ar archive with no `/0` member and no `.rcgu.o/`
member walks the loop to end-of-archive and terminates the whole `osv-scanner` process.

This came out of GHSA-2f6x-xr4j-f2h7. @another-rex judged it a regular bug rather than a
vulnerability and I agree with him — the process exits non-zero, so nothing is silently hidden.
Opening it here as the bug it is.

## Details

Two things were wrong, and both are fixed by returning the error:

**1. One artifact ended the entire scan.** The caller is already written to survive this:

```go
buf, err := extractRlibArchive(path)
if err != nil {
    cmdlogger.Errorf("failed to analyse '%s': %s", path, err)
    continue
}
```

`log.Fatalf` is `os.Exit(1)`, so it exits before that error can be returned and the `continue`
never runs.

**2. The exit code was wrong.** `docs/output.md` reserves `1` for *"packages were found when
scanning, and there are vulnerabilities or findings"* and `127` for *General Error*. `log.Fatalf`
exits `1`, so an internal failure was reported to CI through a result-related code, and no output
was produced. Letting the scan finish normally removes the ambiguity.

The change treats end-of-archive as "this is not an rlib we can analyse" and returns an error;
any other error is wrapped and returned. The `log` import goes with it — this was the only
`log.Fatalf` in non-test library code.

No malformed or hostile input is needed to hit this: a well-formed ar archive that simply is not a
Rust rlib is enough (`ar rc plain.a a.txt`).

## Testing

- **Added `Test_extractRlibArchive_noObjectFile`**, which builds a minimal GNU ar archive by hand
  (so the test needs no `ar` binary on the machine running it) and asserts an error is returned.
- **Verified the test actually catches the regression.** Against the unpatched `rust.go` it does
  not merely fail an assertion — the test binary is killed mid-test by `os.Exit(1)`:

  ```
  === RUN   Test_extractRlibArchive_noObjectFile
  2026/08/10 08:50:54 EOF
  FAIL    github.com/google/osv-scanner/v2/internal/sourceanalysis   0.005s
  ```

  With the fix: `--- PASS: Test_extractRlibArchive_noObjectFile (0.00s)`.
- The existing `Test_extractRlibArchive` cases (`simple.rlib`, `medium.rlib`) still pass; the new
  fixture is written to `t.TempDir()` so it is not picked up by that directory-driven test.
- `go test ./internal/sourceanalysis/...` — ok.
- `golangci-lint` at your pinned `v2.11.4` on the package — **0 issues**. `gofmt -l` clean, `go vet`
  clean.

All run in a `golang:1.26` container matching `go.mod`.

## Checklist

- [x] I have run the linter using `./scripts/run_lints.sh`. (golangci-lint `v2.11.4` on the changed package — 0 issues)
- [x] I have run the unit tests using `./scripts/run_tests.sh`. (`go test ./internal/sourceanalysis/...` — ok)

---

I used an AI assistant while investigating this; I read the code paths, wrote the test, and ran the
verification myself.


G-Rath:

G-Rath (dismissed):
The code change and test look fine, but could you scale back the comments?

Personally I'd say the test is a nice extra bit of coverage rather than guarding against a specific "fix" - the fact that this was a `log.Fatal` previously is discoverable from the git history, and us returning an error rather than exiting immediately is very standard behaviour so shouldn't need to be called out anywhere.

kobihikri:
Thanks — that's fair, the git history does carry it.

I've dropped the comments narrating the old `log.Fatal` and the one framing the test as guarding the fix, and trimmed the note above the hand-built archive. Code and test logic are unchanged.

Re-ran at your pinned golangci-lint v2.11.4 (0 issues), `gofmt` and `go vet` clean, and both `Test_extractRlibArchive` and `Test_extractRlibArchive_noObjectFile` pass in a `golang:1.26` container matching go.mod.

another-rex (approved):

PR #2984, opened 2026-08-11
chore: add "read contributing guide" to the PR checklist

G-Rath:


G-Rath:
@another-rex you cannot have the "run unit tests" as a required workflow _and_ also [not run it unless some types of files change](https://github.com/google/osv-scanner/blob/main/.github/workflows/tests.yml#L29-L33) 🙃

another-rex:
Yeah I've noticed that issue, not sure how to fix it yet. Might just have it automatically pass rather than not running.

another-rex (approved):

Issue #2968, opened 2026-08-06
osv-scanner panics on empty license field (index out of range) when --licenses is enabled


sl4x0:
### Summary

`osv-scanner scan --licenses <allowlist>` crashes with an unrecoverable panic (`index out of range`) when scanning packages whose license field is empty. An empty license string reaches `spdx.Satisfies("", allowlist)`, the tokenizer produces an empty token list, and `tokens.next()` dereferences `ts.tokens[0]` without a bounds check.

This is reachable from **valid parser output**, not malformed input: the APK extractor emits `Licenses: []string{record["L"]}` and, when an Alpine package record omits the `L:` field, that yields `[]string{""}`. The license enricher preserves it and it flows into the SPDX evaluator. With no `recover()` anywhere in the scan path, the panic aborts the whole scan.

### Affected code

`internal/spdx/satisfies.go:72-77`:
```go
// next returns the next token in the list of tokens, removing it from the list in the process
func (ts *tokens) next() string {
	token := ts.tokens[0]
	ts.tokens = ts.tokens[1:]

	return token
}
```
No bounds check exists before `ts.tokens[0]`, unlike `peek()` (`:63-69`) which does guard with `len(ts.tokens) == 0`.

Call chain: `Satisfies("", ...)` -> `parse` (`:157`) -> `parseOr` -> `parseAnd` -> `parseExpression` (`:216`) -> `nextAndIsNextNextValid` (`:92`) -> `next` (`:73`) -> panic.

### Reproduction

At HEAD `d87a911` (go 1.26+):

```bash
go test ./internal/spdx -run TestPocSatisfiesEmptyLicense -v
```

Observed output (`poc_satisfies_test.go` calling `spdx.Satisfies("", []string{"MIT"})`):
```
panic: runtime error: index out of range [0] with length 0
github.com/google/osv-scanner/v2/internal/spdx.(*tokens).next(...)
        C:/.../internal/spdx/satisfies.go:73
```

End-to-end: scan an Alpine repository/container image containing a package whose `lib/apk/db/installed` record omits the `L:` field with `osv-scanner scan --licenses MIT <target>`. The empty license reaches `spdx.Satisfies("")` and the process terminates.

### Impact

- Complete denial of the scan whenever `--licenses` is enabled and any scanned package lacks a license field.
- CI pipelines fail consistently until the package or flag is removed.
- Remaining packages in the scan are never evaluated.

### Root cause / suggested fix

[...]

PR #2995, opened 2026-08-15
fix(spdx): bound recursion depth when parsing license expressions

### LINKED ISSUE #2993: spdx.Satisfies: unbounded recursion causes uncatchable stack overflow on deeply nested license expressions
### Summary

`spdx.Satisfies` parses SPDX license expressions with an unbounded recursive descent. A license string with deeply nested brackets recurses until the goroutine stack overflows, which is a Go fatal error that `recover` cannot catch, so the whole scan process aborts.

License strings are untrusted input: they come from the metadata of packages in the scanned target. When `--licenses` is enabled, every package license is passed to `spdx.Satisfies` at `pkg/osvscanner/vulnerability_result.go:123`.

This is distinct from #2968. That issue is an empty-input `index out of range` in `tokens.next()`; its proposed fix guards `next()` against an empty token stream. That guard does not bound recursion depth, so it does not address this crash. The two are separate defects in the same parser, one shallow (zero tokens) and one deep (many tokens).

### Affected code

`internal/spdx/satisfies.go`. The recursion runs `parseOr -> parseAnd -> parseExpression`, and `parseExpression` calls back into `parseOr` on every `(` (`:215-237` at HEAD `567f3ea`):

```go
func parseExpression(tokens *tokens) (node, error) {
	next, err := tokens.nextAndIsNextNextValid()
	...
	if next == "(" {
		expr, err := parseOr(tokens)   // recurses; no depth limit
		...
	}
}
```

`allowed["("]` includes `"("`, so an arbitrarily long run of `(` is accepted and each one adds a stack frame. There is no bound on nesting depth anywhere in the descent.

### Reproduction

At HEAD `567f3ea`, go 1.26:

```go
package spdx_test

import (
	"strings"
	"testing"

	"github.com/google/osv-scanner/v2/internal/spdx"
	"github.com/google/osv-scanner/v2/pkg/models"
)

func TestReproDeepNest(t *testing.T) {
	license := models.License(strings.Repeat("(", 2_000_000))
	spdx.Satisfies(license, []string{"MIT"})
}
```

```
go test ./internal/spdx -run TestReproDeepNest
```

Observed:

```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
...
github.com/google/osv-scanner/v2/internal/spdx.(*tokens).isNextValid(...)
	internal/spdx/satisfies.go:99
github.com/google/osv-scanner/v2/internal/spdx.(*tokens).nextAndIsNextNextValid(...)
	internal/spdx/satisfies.go:94
github.com/google/osv-scanner/v2/internal/spdx.parseExpression(...)
	internal/spdx/satisfies.go:216
github.com/google/osv-scanner/v2/internal/spdx.parseAnd(...)
github.com/google/osv-scanner/v2/internal/spdx.parseOr(...)
github.com/google/osv-scanner/v2/internal/spdx.parseExpression(...)
	internal/spdx/satisfies.go:216
...
```

The closing brackets are not required to trigger it; an opening run alone overflows. On this machine the threshold sits between 1,000,000 nested brackets (survives) and 1,500,000 (overflows), so roughly a 3 MB license string. `fatal error: stack overflow` is not recoverable, so the process exits regardless of any `recover` on the scan path.

### Impact

- Denial of the scan when `--licenses` is enabled and any scanned package carries a sufficiently nested license expression.
- The trigger is package metadata from the scanned target, so a single crafted or malformed package aborts the run and no remaining packages are evaluated.
- Reachable through the same `--licenses` code path as #2968, so both crash the scanner from an untrusted license field, by different mechanisms.

Severity is bounded by the input size required (a multi-megabyte license string), so this is defense-in-depth for a parser that consumes untrusted input, rather than a trivially small trigger. It is still a reachable, unrecoverable crash from untrusted data in a security scanner, which is the reason to bound it.

### Suggested fix

Bound the nesting depth in the recursive descent and return a normal parse error past the limit, so a malformed or hostile license expression is rejected like any other invalid input instead of crashing the process. Real SPDX expressions nest only a few levels, so a generous ceiling (for example 1000) rejects nothing legitimate. A regression test drives a deeply nested expression and asserts an error rather than a crash.

CWE-674 (Uncontrolled Recursion).

I have this change and its test ready locally and can open a PR once assigned.

### Environment

- osv-scanner at HEAD `567f3ea`
- go 1.26.5

Amey-Thakur:
## Overview

Fixes #2993.

`spdx.Satisfies` parses SPDX license expressions with an unbounded recursive
descent. Because license strings come from the metadata of packages in the
scanned target, a crafted expression with deeply nested brackets recurses until
the goroutine stack overflows. That is a Go fatal error which `recover` cannot
catch, so the whole scan process aborts.

This is a single, self-contained change addressing the specific issue I filed
(#2993), which I intend to shepherd through review myself. It is distinct from
the empty-token index panic in #2968; the `next()` bounds check discussed there
does not bound recursion depth.

## Details

The recursion runs `parseOr -> parseAnd -> parseExpression`, and
`parseExpression` calls back into `parseOr` on every `(`, with no limit on
nesting depth. On this machine the crash threshold sits between 1,000,000
nested brackets (survives) and 1,500,000 (overflows), roughly a 3 MB license
string.

The fix threads a nesting-depth counter through the descent and returns an
ordinary parse error once it passes a generous ceiling:

- a `depth` field is added to the `tokens` struct, incremented when
  `parseExpression` recurses into a bracketed sub-expression, and
- when the depth exceeds `maxDepth` (1000), the parser returns a normal error
  rather than recursing further.

Real SPDX expressions nest only a few levels, so the ceiling rejects nothing
legitimate. An over-nested expression is now rejected the same way as any other
invalid input, instead of terminating the process. CWE-674 (Uncontrolled
Recursion).

## Testing

- Added `TestSatisfies_DeeplyNested` in `internal/spdx/satisfies_test.go`, which
  drives a deeply nested expression and asserts a parse error rather than a
  crash. It triggers a fatal stack overflow without this change and passes with
  it.
- Ran the full package suite: `go test ./internal/spdx/...` passes.
- `gofmt` and `go vet ./internal/spdx/` are clean.

## Checklist

- [x] I have run the linter (`gofmt`, `go vet`) on the changed package.
- [x] I have run the unit tests for the changed package and all tests pass.

If the maintainers would prefer to assign #2993 before review, I am happy to
wait; opening the pull request here to make the proposed fix concrete.

### DIFF
--- a/internal/spdx/satisfies.go
+++ b/internal/spdx/satisfies.go
@@ -55,8 +55,17 @@ func (n nodeLeaf) satisfiedBy(licenses []string) bool {
 
 var _ node = nodeLeaf{}
 
+// maxDepth bounds how deeply nested the parser will recurse into bracketed
+// sub-expressions. License expressions are supplied by scanned package
+// metadata, which is untrusted; without a limit a deeply nested expression
+// (e.g. many "(") recurses until the goroutine stack overflows, which is a
+// fatal error that recover cannot catch. Real SPDX expressions nest only a
+// handful of levels, so this ceiling is far above any legitimate input.
+const maxDepth = 1000
+
 type tokens struct {
 	tokens []string
+	depth  int
 }
 
 // peek returns the next token in the list of tokens, or otherwise an empty string
@@ -219,10 +228,16 @@ func parseExpression(tokens *tokens) (node, error) {
 	}
 
 	if next == "(" {
+		tokens.depth++
+		if tokens.depth > maxDepth {
+			return nil, fmt.Errorf("license expression nested too deeply (limit %d)", maxDepth)
+		}
+
 		expr, err := parseOr(tokens)
 		if err != nil {
 			return nil, err
 		}
+		tokens.depth--
 
 		if tokens.peek() != ")" {
 			return nil, errors.New("missing closing bracket")
--- a/internal/spdx/satisfies_test.go
+++ b/internal/spdx/satisfies_test.go
@@ -418,3 +418,28 @@ func TestSatisfies_Invalid(t *testing.T) {
 		})
 	}
 }
+
+func TestSatisfies_DeeplyNested(t *testing.T) {
+	t.Parallel()
+
+	// A license expression is untrusted input taken from scanned package
+	// metadata. Deeply nested brackets must be rejected with an error rather
+	// than recursing until the goroutine stack overflows, which is a fatal
+	// error that recover cannot catch. The input is built here rather than
+	// listed in the table above so the generated subtest name stays readable.
+	license := models.License(strings.Repeat("(", 2_000_000) + "MIT" + strings.Repeat(")", 2_000_000))
+
+	got, err := spdx.Satisfies(license, []string{"MIT"})
+
+	if got {
+		t.Errorf("Satisfies(deeply nested) = %v, want %v", got, false)
+	}
+
+	if err == nil {
+		t.Fatal("Satisfies(deeply nested) = nil error, want a nesting-limit error")
+	}
+
+	if !strings.Contains(err.Error(), "nested too deeply") {
+		t.Errorf("Satisfies(deeply nested) = %v, want a nesting-limit error", err)
+	}
+}


G-Rath:

PR #2996, opened 2026-08-16
fix(output): don't panic on an unparsable fixed version

### LINKED ISSUE #2936: getNextFixVersion panics on an advisory with an unparsable `fixed` version, crashing result output
### Summary

`internal/output/output_result.go` builds the scan result for every matched
package. For each vulnerability it calls `getNextFixVersion`, which parses the
`fixed` string of every affected range event with `semantic.MustParse`
(`output_result.go:590`). `MustParse` panics when the ecosystem's version parser
returns an error, so a single advisory whose `fixed` value is unparsable aborts
the entire scan **after** matching has already succeeded. Because every output
format (table, JSON, SARIF, HTML) routes through `BuildResults`, all of them
crash.

This is a distinct site from #2867 (match path, ecosystem-name `MustParse`) and
#2837 (match path, version parse). Neither touched the output/result-building
path.

### Affected code

`internal/output/output_result.go:590`:
```go
order, _ = semantic.MustParse(affectedEvent.GetFixed(), ecosystemPrefix).CompareStr(minFixVersion)
```

`semantic.Parse` returns `ErrUnsupportedEcosystem` for ecosystems with no
version parser, and `ErrInvalidVersion` for a malformed version in ecosystems
whose parser validates input (CRAN, Debian, Alpine, Ubuntu, Hackage, ...).
`MustParse` turns either error into a panic. The guard at `:584` does not
prevent this: `vp.CompareStr(GetFixed())` returns `(0, err)` on an unparsable
`fixed`, the error is discarded, so `order == 0`, the `order > 0` skip does not
fire, and execution reaches the panicking `MustParse`.

### Reproduction (public CLI, offline DB)

Reproduces end-to-end through the documented `--offline-vulnerabilities /
--local-db-path` feature — no internal calls. `mypkg@1.0.0` (CRAN) matches an
advisory by its `versions` list, and result building then parses the range's
malformed `fixed` value and panics. Real advisories routinely carry both a
`versions` list and `ranges`.

```bash
WORK="$(mktemp -d)"
mkdir -p "$WORK/target"
cat > "$WORK/target/renv.lock" <<'EOF'
{ "R": { "Version": "4.3.0" },
  "Packages": { "mypkg": { "Package": "mypkg", "Version": "1.0.0", "Repository": "CRAN" } } }
EOF

python3 - "$WORK" <<'PY'
import json,os,sys,zipfile
work=sys.argv[1]
adv={"id":"CB-CRAN-REPRO","affected":[
 {"package":{"ecosystem":"CRAN","name":"mypkg"},
  "versions":["1.0.0"],
  "ranges":[{"type":"ECOSYSTEM","events":[{"introduced":"0"},{"fixed":"notaversion"}]}]}]}
zp=os.path.join(work,"db","osv-scanner","CRAN","all.zip")
os.makedirs(os.path.dirname(zp),exist_ok=True)
zipfile.ZipFile(zp,"w",zipfile.ZIP_DEFLATED).writestr("CB-CRAN-REPRO.json",json.dumps(adv))
PY

go run ./cmd/osv-scanner scan source \
  --offline-vulnerabilities --local-db-path "$WORK/db" "$WORK/target"
```

### Actual behavior

```
panic: invalid version: 'notaversion' is not allowed
  semantic.MustParse            semantic/parse.go:36
  output.getNextFixVersion      internal/output/output_result.go:590
  output.updateVuln             internal/output/output_result.go:541
  output.processPackage         internal/output/output_result.go:462
  output.BuildResults           internal/output/output_result.go:202
  output.PrintTableResults      internal/output/table.go:30
```

### Expected behavior

An unparsable `fixed` version is skipped (that event simply cannot be offered as
a fix); the scan completes and reports the matched vulnerability with no
available fix.

### Why it is reachable without an attacker

The per-ecosystem DB zip stores the full advisory record, and the vendored
`osv-schema` / `semantic` parsers are pinned per release while osv.dev's data
evolves. Any advisory that carries a `fixed` string this build cannot parse —
future syntax, an ecosystem whose parser validates input, or a self-maintained
local DB — crashes result building on a package that otherwise matched. Broad
blast radius for a tool that runs in CI.

### Suggested fix

Use the non-panicking `semantic.Parse` and skip the event on error, mirroring
the "skip, don't panic on unsupported ecosystems/versions" approach already
taken in #2837/#2867. The default (parsable) path is unchanged.

I have a fix and reproducing tests ready and would be happy to open a PR once
this is assigned.

---

**Related, separate observation (not part of this fix):** `internal/output/table.go:334`
uses `osvecosystem.MustParse(eco.Name)` on the scanned package's ecosystem in the
table renderer. Reachability depends on whether an unrecognized ecosystem can
reach output (e.g. via SBOM input); worth a separate look but not covered here.

hamodywe:
Fixes #2936.

I asked to be assigned on the issue first per CONTRIBUTING, and I'm opening this so the change is concrete rather than described — happy for it to be closed if you'd rather not take it, or to adjust the approach.

## The bug

`getNextFixVersion` parses every range event's `fixed` value with `semantic.MustParse`, which panics when the ecosystem's parser rejects the value.

The guard above it does not prevent this:

```go
order, _ := vp.CompareStr(affectedEvent.GetFixed())
if affectedEvent.GetFixed() == "" || order > 0 {
    continue
}
order, _ = semantic.MustParse(affectedEvent.GetFixed(), ecosystemPrefix).CompareStr(minFixVersion)
```

`CompareStr` returns `(0, err)` for a version it cannot parse and the error is discarded, so `order` stays `0`, the `order > 0` skip does not fire, and execution reaches `MustParse`.

Advisories are third-party data, and result building runs *after* matching has already succeeded, so a single unparsable `fixed` value aborts an otherwise complete scan. Every output format goes through `BuildResults`, so table, JSON, SARIF and HTML are all affected.

## The change

Use `semantic.Parse` and skip the event when it fails: a version that cannot be compared cannot be offered as the next fix.

This mirrors what the same function already does one branch earlier — an unparsable *installed* version returns `N/A` rather than panicking — so the `fixed` value was the only place here that crashed instead of degrading. A valid fix listed alongside an unparsable one is still reported, which the second test covers.

It is also the same class the match path was hardened against in #2867 and #2837; neither touched the output path.

## Verification

- With this change reverted, the new test fails with the real crash: `panic: invalid version: 'notaversion' is not allowed`. With it applied, both new tests pass.
- The full `internal/output` package is green (`go test ./internal/output/`), and `go vet` is clean.

## Note on reproduction

As discussed on the issue, no currently published advisory triggers this — the reporter checked `fixed` values across the validating ecosystems and found none that fail on this path. I'd still suggest it is worth guarding: the input is third-party, the crash discards a completed scan, and the surrounding code already treats the neighbouring field defensively.

### DIFF
--- a/internal/output/output_result.go
+++ b/internal/output/output_result.go
@@ -581,13 +581,22 @@ func getNextFixVersion(allAffected []*osvschema.Affected, installedVersion strin
 		}
 		for _, affectedRange := range affected.GetRanges() {
 			for _, affectedEvent := range affectedRange.GetEvents() {
-				order, _ := vp.CompareStr(affectedEvent.GetFixed())
+				order, err := vp.CompareStr(affectedEvent.GetFixed())
 				// Skip if it's not a fix version event or the installed version is greater than the fix version.
-				if affectedEvent.GetFixed() == "" || order > 0 {
+				// A fixed version this ecosystem's parser rejects is skipped too: it cannot be
+				// compared, so it cannot be offered as the next fix. Advisories are third-party
+				// data, and using MustParse here aborted the whole scan after matching had already
+				// succeeded — the same reason the installed version above degrades instead of panicking.
+				if affectedEvent.GetFixed() == "" || err != nil || order > 0 {
 					continue
 				}
 
-				order, _ = semantic.MustParse(affectedEvent.GetFixed(), ecosystemPrefix).CompareStr(minFixVersion)
+				fixedVersion, err := semantic.Parse(affectedEvent.GetFixed(), ecosystemPrefix)
+				if err != nil {
+					continue
+				}
+
+				order, _ = fixedVersion.CompareStr(minFixVersion)
 				// Find the minimum fix version
 				if minFixVersion == UnfixedDescription || order < 0 {
 					minFixVersion = affectedEvent.GetFixed()
--- a/internal/output/output_result_internal_test.go
+++ b/internal/output/output_result_internal_test.go
@@ -1,0 +1,70 @@
+package output
+
+import (
+	"testing"
+
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+)
+
+// A `fixed` value the ecosystem's parser rejects must not abort the scan. Advisory
+// data is third-party input, and result building runs after matching has already
+// succeeded, so a panic here loses an otherwise complete scan.
+func TestGetNextFixVersion_UnparsableFixedVersion(t *testing.T) {
+	t.Parallel()
+
+	affected := []*osvschema.Affected{
+		{
+			Package: &osvschema.Package{Ecosystem: "CRAN", Name: "mypkg"},
+			Versions: []string{"1.0.0"},
+			Ranges: []*osvschema.Range{
+				{
+					Type: osvschema.Range_ECOSYSTEM,
+					Events: []*osvschema.Event{
+						{Introduced: "0"},
+						{Fixed: "notaversion"},
+					},
+				},
+			},
+		},
+	}
+
+	hasFix, fixVersion := getNextFixVersion(affected, "1.0.0", "mypkg", "CRAN")
+
+	if hasFix {
+		t.Errorf("getNextFixVersion() reported a fix from an unparsable version, got %q", fixVersion)
+	}
+	if fixVersion != UnfixedDescription {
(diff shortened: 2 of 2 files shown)


G-Rath:

PR #2997, opened 2026-08-16
feat(config): surface config suppressions and add --no-config-ignore

herdiyana256:
`Manager.Get` loads an `osv-scanner.toml` from the directory of each scanned manifest and applies its `IgnoredVulns` and `PackageOverrides` `ignore` directives to that directory's results. During a recursive scan this reaches directories the person running the scan did not author: a vendored dependency, a git submodule, an extracted archive, or a folder added by a pull request. A config placed in such a directory suppresses findings for it, so scanned code can carry a config that hides its own vulnerabilities. A single `[[PackageOverrides]]` with `nameIsRegex = true`, `name = ".*"`, `ignore = true` blanks the whole subtree without naming a vulnerability.

The suppression was only logged at info level, so under `--verbosity warn` (or a JSON/SARIF run) the scan printed "No issues found" and exited 0 with no trace, even though known-vulnerable packages had been dropped.

This raises every config-driven suppression to warning level, in both the vulnerability filter and the package annotator: the loaded config path, each suppressed vulnerability or package, and the filtered count. A suppressed result is now visible at the verbosity most CI runs use.

It also adds `--no-config-ignore`, which drops the vulnerability-suppressing directives (`IgnoredVulns` and the `ignore` flags on `PackageOverrides`) from auto-discovered configs while leaving license overrides in place. A config supplied explicitly with `--config` is trusted and unaffected. This gives a scan over code you do not fully control, such as CI on pull requests or vendored third-party trees, a way to refuse in-tree suppression outright. The docs gain a short trust note describing both.

`go test ./internal/config/... ./internal/scalibrannotator/... ./pkg/osvscanner/...` passes, including a new `config` test for the strict-mode helper. The log changes preserve message text, so existing stdout snapshots are unaffected at the default info verbosity.

### DIFF
--- a/cmd/osv-scanner/internal/helper/flags.go
+++ b/cmd/osv-scanner/internal/helper/flags.go
@@ -57,6 +57,11 @@ func BuildCommonScanFlags(defaultExtractors []string) []cli.Flag {
 			Usage:     "set/override config file",
 			TakesFile: true,
 		},
+		&cli.BoolFlag{
+			Name:  "no-config-ignore",
+			Usage: "ignore IgnoredVulns and PackageOverrides ignore directives from osv-scanner.toml files discovered next to scanned manifests (a config passed with --config is unaffected)",
+			Value: false,
+		},
 		&cli.StringFlag{
 			Name:    "format",
 			Aliases: []string{"f"},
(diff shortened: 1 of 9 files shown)


G-Rath:
Please read our contributor guide before opening pull requests

PR #3020, opened 2026-08-26
feat(output): add a "Fix available" filter to the HTML report

### LINKED ISSUE #3016: Feature request - Container scanning - HTML output: Add a filter to only show items with a fix available
Right now, there is no way to filter so I can only see items that have a fix available. So I have to scroll through the entire list and make sure I spot them visually, looking for "Fix available" or "partial fixes available".

Example of current output: 

Image

GitZohair:
## Overview

Fixes #3016

Adds a client-side "Fix available" filter to the HTML report, so users can show only vulnerabilities that have a fix available instead of scanning the fix column visually.

## Details

- `vuln_table_entry_template.gohtml` stamps `data-fixable="true|false"` on each vulnerability row, using the existing `VulnResult.IsFixable` field — no changes to result building are needed.
- A "Fix available (N)" checkbox joins the existing Filters panel, labelled with the already-computed `VulnCount.FixableCount.Fixed`. Like the other options in that panel, N counts regular vulnerabilities only.
- New `applyFixFilter()` in `script.js` runs inside `applyFilters()`, after the type and layer filters. It only ever adds `hide-block`, and `applyFilters()` starts from `showAllVulns()`, so it composes with the other filters regardless of order. Packages, sources and ecosystems left with no visible rows are collapsed by the existing `showAndHideParentSections()`, which is what makes the report scannable for the case in #3016.
- The checkbox sits inside `#type-filter-option-container`, so the existing `change` listener picks it up, and it participates in the existing reset paths (`resetFilterText()` / `resetTypeCheckbox()`) — typing in the search box clears it, exactly like the "Uncalled/Unimportant" toggle.
- `updateTypeFilterText()` now runs at the end of `applyFilters()` instead of at the start of `applyTypeFilter()` (its only caller). That lets the summary append "Fix available" *after* the "Default" shorthand has been applied rather than before it, and report the number of rows actually left visible. With the fix filter off, the summary text and count are computed exactly as they are today.

Filtering stays purely client-side, consistent with the existing type/layer filters; no new CLI flags or output schema changes.

## Testing

- Unit tests for `internal/output` (covering HTML rendering across ~30 result scenarios) pass locally via `./scripts/run_tests.sh`.
- `prettier --check` is clean for `script.js` (the `.gohtml` templates are covered by `.prettierignore`).
- Filter interactions were exercised against a DOM built from these templates, driving the real `script.js` with `change` / `click` / `keyup` events: the fix filter on its own, combined with the type filter, combined with the layer filter, with "Uncalled/Unimportant" enabled, on reports both with and without OS results, and through the search-box reset path. In each case the visible rows are exactly the fixable ones, the parent sections collapse correctly, and the summary count matches the number of rows shown.
- As a regression check, 13 interaction sequences that never touch the new checkbox produce DOM state identical to `main` — every row's classes, the shown/hidden state of every package, source and ecosystem container, and the filter summary text and count.

## Checklist

- [x] I have run the linter using `./scripts/run_lints.sh`. _(No Go source changes — this PR touches only a Go HTML template, the report's JS, and the filter template. `./scripts/run_formatters.sh` covers `script.js` and is clean.)_

### DIFF
--- a/internal/output/html/filter_template.gohtml
+++ b/internal/output/html/filter_template.gohtml
@@ -76,6 +76,10 @@
           <input type="checkbox" id="uncalled-type-checkbox" data-type-uncalled-count="{{ .VulnTypeSummary.Hidden }}">
           Uncalled/Unimportant ({{ .VulnTypeSummary.Hidden }})
         </label>
+        <label class="filter-option" for="fixable-type-checkbox">
+          <input type="checkbox" id="fixable-type-checkbox" data-fixable-count="{{ .VulnCount.FixableCount.Fixed }}">
+          Fix available ({{ .VulnCount.FixableCount.Fixed }})
+        </label>
       </div>
     </div>
   </div>
--- a/internal/output/html/filter_template.gohtml
+++ b/internal/output/html/filter_template.gohtml
@@ -76,8 +76,8 @@
           <input type="checkbox" id="uncalled-type-checkbox" data-type-uncalled-count="{{ .VulnTypeSummary.Hidden }}">
           Uncalled/Unimportant ({{ .VulnTypeSummary.Hidden }})
         </label>
-        <label class="filter-option" for="fixable-type-checkbox">
-          <input type="checkbox" id="fixable-type-checkbox" data-fixable-count="{{ .VulnCount.FixableCount.Fixed }}">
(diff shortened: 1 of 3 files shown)


G-Rath:

PR #3022, opened 2026-08-26
feat: add filter to show only vulnerabilities with a fix available

### LINKED ISSUE #3016: Feature request - Container scanning - HTML output: Add a filter to only show items with a fix available

amortizedX:
Closes #3016

Adds a checkbox to the HTML report that filters out vulnerabilities without a fix available. Went with a CSS-only approach using :has() per @G-Rath's suggestion, so no JS changes needed.

- vuln_table_entry_template.gohtml — tags rows with a `no-fix` class when IsFixable is false
- filter_template.gohtml — adds the checkbox
- style.css — hides `.no-fix` rows when the checkbox is checked

Tested locally against a project with known vulns, checkbox toggles the rows correctly.

### DIFF
--- a/internal/output/html/filter_template.gohtml
+++ b/internal/output/html/filter_template.gohtml
@@ -104,3 +104,5 @@
 
 
 </div>
+<input type="checkbox" id="filter-fix-available">
+<label for="filter-fix-available">Show only items with a fix available</label>
--- a/internal/output/html/style.css
+++ b/internal/output/html/style.css
@@ -684,3 +684,7 @@ div.title {
     transform: rotate(360deg);
   }
 }
+body:has(#filter-fix-available:checked).vuln-tr.no-fix
+{
+  display:none;
+}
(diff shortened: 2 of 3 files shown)


G-Rath (changes requested):
I've not had the time to dig into why, but running this locally doesn't seem to work.

I think it'd also be better to integrate this into our existing "filters" dropdown, and ideally omit the package entirely if there's no fixable vulns.

What we're wanting here should be very similar to the existing "Uncalled / Unimportant" filter, so you should be able to use that as a reference

another-rex:
@amortizedX quick ping! Have you had time to take another look at this?

amortizedX:
Hey @another-rex, sorry for the silence,I  got pulled away but I'm still on this. I've reworked the checkbox into the filters dropdown per G-Rath's feedback and it builds locally; I still need to verify it in-browser, fix prettier formatting, and sort out the failing unit tests before pushing. Should have an update within the next few days.

KoenDG:
It looks like there's an issue here with the unit tests. Amounts of CVEs have changed, as well as availability of fixes, changing from non to partially available.

This is probably something that a more senior coder should look at, to consider if changes can be made there, as that will impact more than just this PR.

PR #3023, opened 2026-08-27
fix(spdx): guard against panic on empty license expressions

### LINKED ISSUE #2968: osv-scanner panics on empty license field (index out of range) when --licenses is enabled

sl4x0:
## Overview

Fixes #2968

`osv-scanner scan --licenses <allowlist>` panics with `index out of range` when any scanned package has an empty license field. An empty license string reaches `spdx.Satisfies("", allowlist)`: the tokenizer produces zero tokens, and `tokens.next()` dereferences `ts.tokens[0]` without a bounds check, terminating the whole scan.

## Details

Two changes in `internal/spdx/satisfies.go`:

1. `nextAndIsNextNextValid` now returns an error when the token stream is exhausted instead of calling `next()` on an empty list, so an empty expression can never reach the index-out-of-range dereference. This is the root-cause guard (`peek()` already protected itself; `next()` did not).
2. `Satisfies` handles empty/whitespace-only expressions explicitly, returning `false, nil` so an empty license is treated as "not satisfied by the allowlist" and recorded as a violation (via `pkg/osvscanner/vulnerability_result.go`) rather than aborting the scan.

Regression tests:

- `internal/spdx/satisfies_test.go`: empty (`""`) and whitespace-only (`"   "`) license expressions added to `TestSatisfies`, asserting they evaluate to not-satisfied without an error (these panicked at HEAD).
- `pkg/osvscanner/vulnerability_result_internal_test.go`: `TestBuildVulnerabilityResultsWithEmptyLicense` drives a package with `Licenses: []string{""}` through `buildVulnerabilityResults` with a `MIT` allowlist and asserts the empty license is reported as a violation and the scan completes.

## Testing

- Reproduced the panic at HEAD with `go test ./internal/spdx -run TestSatisfies` before the fix.
- `go test ./internal/spdx/... ./pkg/osvscanner/...` passes after the fix (including the new regression tests).
- `go vet ./internal/spdx/... ./pkg/osvscanner/...` reports no issues.
- `go build ./...` succeeds.

## Checklist

- [ ] I have run the linter using `./scripts/run_lints.sh`.

### DIFF
--- a/internal/spdx/satisfies_test.go
+++ b/internal/spdx/satisfies_test.go
@@ -79,6 +79,23 @@ func TestSatisfies(t *testing.T) {
 				{"Apache-2.0"},
 			},
 		},
+		// empty expression licenses contain no tokens and can never satisfy an
+		// allowlist, but must not panic (regression for #2968)
+		{
+			license: "",
+			fail: [][]string{
+				{"MIT"},
+				{"MIT", "Apache-2.0"},
+				{},
+			},
+		},
+		{
+			license: "   ",
+			fail: [][]string{
+				{"MIT"},
+				{},
+			},
+		},
 		// AND expressions
 		{
 			license: "MIT AND Apache-2.0",
--- a/internal/spdx/satisfies_test.go
+++ b/internal/spdx/satisfies_test.go
@@ -80,7 +80,7 @@ func TestSatisfies(t *testing.T) {
 			},
 		},
 		// empty expression licenses contain no tokens and can never satisfy an
-		// allowlist, but must not panic (regression for #2968)
+		// allowlist.
 		{
 			license: "",
 			fail: [][]string{
--- a/pkg/osvscanner/vulnerability_result_internal_test.go
+++ b/pkg/osvscanner/vulnerability_result_internal_test.go
@@ -1,6 +1,7 @@
 package osvscanner
 
 import (
+	"slices"
 	"testing"
 
 	"github.com/google/osv-scalibr/extractor"
@@ -10,6 +11,7 @@ import (
 	"github.com/google/osv-scanner/v2/internal/config"
 	"github.com/google/osv-scanner/v2/internal/imodels/results"
 	"github.com/google/osv-scanner/v2/internal/testutility"
+	"github.com/google/osv-scanner/v2/pkg/models"
 	"github.com/ossf/osv-schema/bindings/go/osvschema"
 )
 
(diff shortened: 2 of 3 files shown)


G-Rath (changes requested):
can you tone the comments down a bit? we don't need to be describing the past behaviour because that's what git and commit messages are for

sl4x0:
Note on the checklist: the linter (\./scripts/run_lints.sh\ / golangci-lint v2.11.4) flags 19 files in the \osvscanner\ package as gofmt-unformatted on my Windows checkout, including files this PR does not touch (e.g. \pkg/osvscanner/stats.go\, \pkg/osvscanner/vulnerability_result.go\). Root cause is git's CRLF autocrlf conversion on Windows, not the diff. I verified the changed files (and an untouched control file) are gofmt-clean by running gofmt on LF-normalized copies, and CI on Linux will confirm.

internal/spdx/satisfies_test.go
```diff
@@ -79,6 +79,23 @@ func TestSatisfies(t *testing.T) {
 				{"Apache-2.0"},
 			},
 		},
+		// empty expression licenses contain no tokens and can never satisfy an
+		// allowlist, but must not panic (regression for #2968)
```
another-rex:
```suggestion
		// empty expression licenses contain no tokens and can never satisfy an
		// allowlist.
```

pkg/osvscanner/vulnerability_result_internal_test.go
```diff
@@ -175,3 +177,46 @@ func Test_assembleResult(t *testing.T) {
 		})
 	}
 }
+
+// TestBuildVulnerabilityResultsWithEmptyLicense ensures a package with an empty
+// license field does not panic when a license allowlist is configured, and is
+// instead recorded as a license violation (regression for #2968).
```
another-rex:
```suggestion
// TestBuildVulnerabilityResultsWithEmptyLicense ensures a package with an empty
// license field does not panic when a license allowlist is configured, and is
// instead recorded as a license violation.
```

another-rex (approved):

PR #3033, opened 2026-08-30
test: add output tests covering the license summary

### LINKED ISSUE #1840: Add cases to `output` tests to cover license summary
It turns out we're not covering logic in the `output` package related to the `LicenseSummary` field - we should expand our test fixtures for that package to include those cases.

I think we need at least two new sets of tests related to license summarizing since you can have `ShowViolations` either true or false - it might be possible though to reuse some of the existing test cases with a field or two changed, so we might want to introduce a new function(s) that return the test cases and accept some parameters

adityajha2005:
Resolves #1840

Adds test coverage for the license summary logic in the `output` package, which was previously untested.

As suggested in the issue, the existing license violation test cases are reused: `testOutputWithLicenseViolations` now delegates to a shared `testOutputWithLicenseIssues` function that takes the license config as a parameter. Two new helpers build on it:

- `testOutputWithLicenseSummary`: summary enabled with no allowlist, so `ShowViolations` is false
- `testOutputWithLicenseSummaryAndViolations`: summary enabled with an allowlist, so `ShowViolations` is true

When summarizing is enabled, each test case's `LicenseSummary` field is derived from its packages' licenses using the same counting and ordering as the scanner (descending count, UNKNOWN last).

New snapshot tests are added for the outputs where the license summary rendering lives: table (all three terminal widths), vertical, markdown table, and the `Result` struct built by `BuildResults` (via `PrintResults`), which covers the `ShowViolations` handling in `buildResult`.

This also fixes a small pre-existing slip the new tests surfaced: the license violation case list had two cases named "multiple_sources_with_a_mixed_count_of_packages,_some_license_violations", and the second one (the versions-and-commits variant, going by its data) was missing `ExperimentalAnalysisConfig`, so it silently rendered without any license config. It now has a distinct name and the config, and the corresponding snapshot keys are updated.

### DIFF
--- a/internal/output/__snapshots__/cyclonedx_test.snap
+++ b/internal/output/__snapshots__/cyclonedx_test.snap
@@ -137,80 +137,6 @@
 
 ---
 
-[TestPrintCycloneDXResults/CycloneDX14_WithLicenseViolations/multiple_sources_with_a_mixed_count_of_packages,_some_license_violations#01 - 1]
-{
-  "$schema": "http://cyclonedx.org/schema/bom-1.4.schema.json",
-  "bomFormat": "CycloneDX",
-  "specVersion": "1.4",
-  "version": 1,
-  "components": [
-    {
-      "bom-ref": "pkg:npm/mine1@1.2.3",
-      "type": "library",
-      "name": "mine1",
-      "version": "1.2.3",
-      "licenses": [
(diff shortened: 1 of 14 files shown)


G-Rath:
Please read our contributing guide before open pull requests

PR #3038, opened 2026-08-31
docs: clarify requirements range extraction

### LINKED ISSUE #3025: --no-resolve treats an unpinned lower-bound specifier (>=X.Y.Z) in requirements.txt as the exact version to check
## Title
`--no-resolve` treats an unpinned lower-bound specifier (`>=X.Y.Z`) in requirements.txt as the exact version to check

## Version
osv-scanner 2.5.1 (commit c84fa4568f2526d0333e9a914ea8a0a5f74ad68b)

## Summary
When scanning a `requirements.txt` with `--no-resolve`, a package declared with an open-ended lower-bound specifier (e.g. `PyJWT>=2.10.0`, no upper bound, no exact pin) is checked against the vulnerability database as if the installed version were *exactly* the floor value (`2.10.0`). This produces vulnerability findings for old, already-fixed versions even when the actually-installed version (per `pip freeze` / a lock file in the same project) is much newer and unaffected.

This is different from #3018 (deps.dev lexicographic version ordering during transitive resolution): this happens with `--no-resolve`, for a *direct* dependency, from parsing the manifest's own specifier — no transitive resolution or deps.dev lookup involved.

## Reproduction
```
mkdir /tmp/osv_repro && cd /tmp/osv_repro
printf 'PyJWT>=2.10.0\n' > requirements.txt
osv-scanner scan source --no-resolve --format json --offline-vulnerabilities --download-offline-databases .
```

Result: `"package": {"name": "pyjwt", "version": "2.10.0", ...}`, with 8 vulnerability groups reported (PYSEC-2025-183, PYSEC-2026-120/175/176/177/178/179/1814), all of which are fixed in PyJWT 2.13.0 — the version that actually gets installed by `pip install -r requirements.txt` today.

## Expected behavior
With `--no-resolve`, either:
- skip vulnerability matching entirely for specifiers without an exact pin (since the actual installed version is unknowable without resolution — the flag's own `--help` text says it "disable[s] transitive dependency resolution," which reasonably implies "don't guess versions," not "assume the floor"), or
- clearly document that unpinned lower-bound specifiers are matched against their floor value, so callers know this is a worst-case/reachability check rather than a check against what's actually installed.

## Impact
For any project using conventional `>=` version floors in `requirements.txt` (very common, e.g. framework/library minimums without upper caps) `--no-resolve` produces a large volume of findings against stale floor versions, unrelated to the real, currently-installed dependency tree. In our case this generated 25 findings across 5 packages for one manifest, all of which turned out to be already-fixed in the real installed versions (verified via a companion `pip freeze` lock file scanned directly, which returned 0 findings for the same packages).

## Workaround
Maintain a companion `pip freeze`-style lock file alongside the loosely-pinned `requirements.txt` and scan that instead (or in addition, filtering out the loose-manifest source when a lock file is present).

BenkiNew:
## Overview

Document how OSV-Scanner handles version ranges in Python `requirements.txt`
manifests.

Fixes #3025

## Details

- Clarify that `requirements.txt` range specifiers are extracted on a
  best-effort basis using the lowest allowed version.
- Explain that this extraction behavior is independent of `--no-resolve`,
  which controls transitive dependency resolution.
- Recommend scanning a resolved lockfile or a fully pinned `requirements.txt`
  generated by `pip freeze` when results should reflect selected environment
  versions.

This is a documentation-only change; it does not alter CLI or extractor
behavior.

## Testing

- `npx --yes prettier@latest --check docs/supported_languages_and_lockfiles.md`
- `./scripts/run_lints.sh` in the Go 1.26.5 container: `0 issues.`
- `make test` in the Go 1.26.5 container: 2008 tests passed, 17 skipped.
- Jekyll production build using `docs/docs.Dockerfile` and Podman.

## Checklist

- [x] I have run the unit tests using `make test` as documented in `CONTRIBUTING.md`.


G-Rath (approved):
looks like a good start to me - technically this applies to all manifest files, not just `requirements.txt`, but that's the only one we have enabled by default, and it'll be easy to refine this in future as needed

BenkiNew:
@googlebot rescan

docs/supported_languages_and_lockfiles.md
```diff
@@ -61,6 +61,22 @@ When scanning source code (`osv-scanner scan source ...`), OSV-Scanner automatic
...
+A `requirements.txt` file is a manifest and does not necessarily record the
+versions installed in an environment. When an entry uses a version range,
+OSV-Scanner's best-effort extractor selects the lowest version allowed by the
+specifier. For example, `PyJWT>=2.10.0` is extracted and checked as version
+`2.10.0`; this does not mean that `2.10.0` is the installed version.
+
+This extraction behavior is independent of `--no-resolve`. That flag controls
+transitive dependency resolution and does not change how direct dependencies
+are extracted from `requirements.txt`.
+
```
Ly-Joey: Since extraction behaviour is independent of `--no-resolve`, it's probably not necessary to call it out here?

docs/supported_languages_and_lockfiles.md
```diff
@@ -61,6 +61,22 @@ When scanning source code (`osv-scanner scan source ...`), OSV-Scanner automatic
...
+To scan the dependency versions selected for an environment, scan a resolved
+lockfile such as `poetry.lock` or a fully pinned `requirements.txt` generated
+by `pip freeze`.
+
```
Ly-Joey: Should we put this section upfront, since scanning lockfiles is the best practice?
And the "explaining manifest file scanning is best-effort basis" can follow.

docs/supported_languages_and_lockfiles.md
```diff
@@ -61,6 +61,22 @@ When scanning source code (`osv-scanner scan source ...`), OSV-Scanner automatic
...
 | Rust       | `Cargo.lock`                                                                                                                                           |
 
+### Python `requirements.txt` files
+
+A `requirements.txt` file is a manifest and does not necessarily record the
+versions installed in an environment. When an entry uses a version range,
+OSV-Scanner's best-effort extractor selects the lowest version allowed by the
+specifier. For example, `PyJWT>=2.10.0` is extracted and checked as version
+`2.10.0`; this does not mean that `2.10.0` is the installed version.
+
```
Ly-Joey: As @G-Rath said, this is applicable to all manifest files, so perhaps we can:
1. Generalise this section a bit (maybe turn it into a manifest vs lockfile explanation?)
2. Because extractor implementations and behaviours can vary across languages, instead of giving the concrete `PyJWT>=2.10.0` example, we can point readers to [osv-scalibr's language extractors](https://github.com/google/osv-scalibr/tree/main/extractor/filesystem/language) for the most up-to-date logic and test cases.

Ly-Joey (changes requested):
Thanks for making the PR!

docs/supported_languages_and_lockfiles.md
BenkiNew: Removed the `--no-resolve` paragraph, since it is unrelated to manifest extraction behavior.

docs/supported_languages_and_lockfiles.md
BenkiNew: Reordered the section so the resolved lockfile (or environment-generated fully pinned manifest) recommendation comes first, followed by the manifest caveat.

docs/supported_languages_and_lockfiles.md
BenkiNew: Generalized the section to “Lockfiles and manifests”, removed the Python-specific version example, and linked to OSV-Scalibr's language extractors and tests for current ecosystem-specific behavior.

Ly-Joey (approved):

PR #3059, opened 2026-09-07
Fall back to SBOM component version when PURL lacks one

tejas5038:
Use pkg.Version when a valid SBOM PURL omits its optional version and add a focused regression test.


G-Rath:

PR #3060, opened 2026-09-07
Skip unparsable fixed versions when building output

tejas5038:
Avoid panicking during result construction when advisory fixed versions cannot be parsed.


G-Rath:

PR #3071, opened 2026-09-10
fix(output): preserve UTF-8 when truncating text

### LINKED ISSUE #3056: Vertical output can emit invalid UTF-8 when truncating long Unicode text
## Summary

`internal/output/vertical.go`'s `truncate` helper counts runes, but its no-whitespace fallback slices the string using the rune limit as a byte offset.

For multibyte UTF-8 text with no whitespace, `limit` is a rune count while `str[:truncateAt]` requires a byte index. The slice can therefore split a rune and emit invalid UTF-8.

Current `main` inspected at `43026d7a28e5a8ad52c7b4ba3b62047fbf52ad31`.

## Minimal reproducer

A focused regression test using the existing helper should demonstrate it:

```go
s := strings.Repeat("界", 100)
got := truncate(s, 80)

if !utf8.ValidString(got) {
    t.Fatalf("truncate returned invalid UTF-8: %q", got)
}
```

At the 80-rune limit, the fallback slices at byte offset 80. Since `80 % 3 == 2`, that is inside a `界` rune rather than on a UTF-8 boundary.
## Reachability / impact

This helper is used in user-visible vertical output in at least two places:

- vulnerability summary/details via `describe(..., 80)`;
- container layer commands via `truncate(..., 100)`.

A long CJK (or other multibyte) string without spaces can therefore produce malformed terminal/log output. Downstream consumers that require valid UTF-8 may reject or replace the broken byte sequence.

## Expected behavior

Truncation should always preserve UTF-8 validity while keeping the existing word-boundary behavior for ordinary text.

## Suggested fix shape

Keep the change local to `truncate` and add focused tests covering:

- long multibyte text with no spaces;
- multibyte text with a whitespace boundary;
- existing ASCII behavior.

If this looks valid, please assign it to `@keeltrace` and I can submit a small fix with regression tests.

keeltrace:
## Overview

Preserve valid UTF-8 when vertical output truncates long multibyte text without a whitespace boundary.

`truncate` counts characters by rune, but its no-whitespace fallback used that rune count directly as a byte slice index. For multibyte text, that could split a UTF-8 sequence.

Fixes #3056

## Details

Use the byte offset of the current rune and its decoded width to truncate after the complete limit-th rune.

The existing behavior of preferring an earlier whitespace boundary is unchanged. Existing ASCII, short-string, and zero-limit behavior is also covered by the regression tests.

## Testing

- focused multibyte truncation tests with and without a whitespace boundary
- regression test verified against the previous implementation, which produced partial UTF-8
- ASCII, short-string, and zero-limit preservation cases
- `GOTOOLCHAIN=go1.27.0 go test ./internal/output -run '^TestTruncate$' -count=1`
- `GOTOOLCHAIN=go1.27.0 go test ./internal/output -count=1`
- `GOTOOLCHAIN=go1.27.0 ./scripts/run_lints.sh`
- `GOTOOLCHAIN=go1.27.0 make test` — 2026 passed, 17 expected skips

AI-assisted contribution; I am responsible for maintainer follow-up and any requested changes.

## Checklist

- [ ] I have signed the Contributor License Agreement
- [x] I have read the CONTRIBUTING guide and understand when to open a pull request
- [x] I have run the linter
- [x] I have run the unit tests
- [x] My commits and PR title follow Conventional Commits

### DIFF
--- a/internal/output/vertical.go
+++ b/internal/output/vertical.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/google/osv-scanner/v2/pkg/models"
 	"github.com/jedib0t/go-pretty/v6/text"
@@ -305,7 +306,12 @@ func truncate(str string, limit int) string {
 			// ideally we want to keep words whole when truncating,
 			// but if we can't find a space just truncate at the limit
 			if truncateAt == -1 {
-				truncateAt = limit
+				if limit <= 0 {
+					truncateAt = 0
+				} else {
+					_, size := utf8.DecodeRuneInString(str[i:])
+					truncateAt = i + size
+				}
 			}
 
 			return str[:truncateAt] + "..."
--- a/internal/output/vertical_internal_test.go
+++ b/internal/output/vertical_internal_test.go
@@ -1,0 +1,63 @@
+package output
+
+import (
+	"strings"
+	"testing"
+	"unicode/utf8"
+)
+
+func TestTruncate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name  string
+		input string
+		limit int
+		want  string
+	}{
+		{
+			name:  "multibyte without whitespace",
+			input: strings.Repeat("\u754c", 100),
+			limit: 80,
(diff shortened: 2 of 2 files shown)


another-rex (approved):

another-rex:
LGTM, thanks!

PR #3076, opened 2026-09-11
chore: Release osv-scanner v2.6.0

another-rex:
Changelog for osv-scanner 2.6.0 release.


G-Rath:
any chance we could get https://github.com/google/osv-scalibr/pull/2423 landed for this release? 🙏 

(it'd let me get rid of a workaround in one of our internal systems)

another-rex:
You can also use the disable plugin flag to disable the dotnet plugin too right?

G-Rath:
yeah that's one way (at the cost of not covering `*.deps.json` files, which we don't have any of 😅) - my implemented workaround is to just remove the file before running the scan 🤷

it's not a biggie if it gets delayed, it'd just be really nice having this + #3075 in this next release because those are the only two workarounds I had to do

Ly-Joey (approved):

PR #3081, opened 2026-09-12
grouper: keep all members of an alias group when groups merge

rootkiller6788:
Group() was splitting up vulnerabilities that are only linked through a chain of aliases, so the same set of findings could come back as different groups depending on the scan order.

The merge step only rewrote the group id of the vuln it was looking at, not the other members of the group being merged in, so those were left behind pointing at the old id. I swapped it for a regular union-find find() with path compression, so the whole group follows when two groups merge. The representative is still the lowest index, so results for everything already covered by TestGroup are unchanged.

A concrete case, using the order the vulnerabilities are actually handed to Group() (they come out of ConvertVulnerabilityToIDAliases sorted by MostUpstreamsOrder): CVE-2024-1111 aliases GHSA-cccc, GHSA-aaaa and GHSA-bbbb both alias CVE-2024-9999, and GHSA-cccc aliases both CVE-2024-1111 and GHSA-aaaa. All four are one group, but GHSA-bbbb comes back on its own. Reordering the same four into the group, and the group changes.

The regression test runs that set in a few different orders. I also compared the new Group() against a plain connected-components implementation over 20000 random alias graphs - they agree everywhere, while the old code disagrees within the first few.

This is reachable from every scan: pkg/osvscanner/vulnerability_result.go uses these groups for the output and for the per-group MaxSeverity, and internal/ci/vulnerability_result_diff.go uses them too. I scanned the existing fixtures and snapshots for two groups sharing an alias (the fingerprint of this bug) and found none, so nothing else should need updating.

### DIFF
--- a/internal/grouper/grouper.go
+++ b/internal/grouper/grouper.go
@@ -31,13 +31,32 @@ func Group(vulns []IDAliases) []models.GroupInfo {
 		groups[i] = i
 	}
 
+	// find returns the representative of the group holding i, following and
+	// shortening the chain as it goes.
+	find := func(i int) int {
+		for groups[i] != i {
+			groups[i] = groups[groups[i]]
+			i = groups[i]
+		}
+
+		return i
+	}
+
 	// Do a pair-wise (n^2) comparison and merge all intersecting vulns.
 	for i := range vulns {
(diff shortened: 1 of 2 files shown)


G-Rath:

PR #3097, opened 2026-09-18
perf: index vulnerabilities by package and group aliases with union-find

ajayk:
Removes quadratic work when building results and grouping aliases:

- `buildVulnerabilityResults` indexes vulnerabilities by package once instead of scanning all vulnerabilities per package.
- `grouper.Group` uses union-find over identifiers instead of a pairwise comparison. This also fixes transitive alias chains being split into separate groups when the joining link is discovered late (regression test added).
- `config.Manager.Get` memoizes the target path to config path resolution, avoiding a stat syscall per package.
- Max severity and fix version lookups no longer rescan the vulnerability list per group.

Benchmarks (`go test -bench`, Apple M-series):

| Benchmark | Before | After |
|---|---|---|
| Group/chained/1000 | 5.20 ms | 0.19 ms |
| Group/isolated/1000 | 2.63 ms | 0.28 ms |
| BuildVulnerabilityResults/1000pkgs_5vulns | 6.0 ms | 3.9 ms |
| BuildVulnerabilityResults/3000pkgs_5vulns | 38.9 ms | 7.3 ms |

Output is unchanged; all snapshot tests pass.


G-Rath:

